PR #871, opened 2013-08-20
makefile: require ARM toolchain to be installed under ./tools

solidgoldbomb:
The practice of searching for the toolchain in the user's
PATH when we can't find the toolchain under ./tools is
problematic.

If the user happens to have an old or incompatible ARM gcc
in their PATH and we change the toolchain version, we will
silently run the one in their PATH and possibly produce
a broken executable.

We now fail the build explicitly if we can't find the correct
toolchain version under ./tools.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,8 +254,9 @@ endif
 ifeq ($(shell [ -d "$(ARM_SDK_DIR)" ] && echo "exists"), exists)
   ARM_SDK_PREFIX := $(ARM_SDK_DIR)/bin/arm-none-eabi-
 else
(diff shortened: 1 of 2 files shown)


elafargue:
We should probably close this PR, right ? @peabody124 please confirm

mlyle:
Any objections to closing PR? We can open an issue so we don't lose the reference to the code/any remaining items.  Alternatively we should work to get this into a state that it can be merged.

mlyle:
Closing-- #1824 supersedes.

PR #933, opened 2013-09-03
Removes explicit dependency on SystemAlarms UAVObjID value

kubark42:
In https://github.com/TauLabs/TauLabs/pull/873, @guilhermito explained that enforcing that all plugins are compiled at the same time as the uavos library is a bad use of a shared library. "Using a shared library allows us (should allow us) to use different versions of the uavos library without failing the assertion. Imagine we only change the default value of an object field, we could recompile the uavos lib and only distribute that. All the plugins would then use it without any problems."

This PR removes the explicit dependency on the SystemAlarms UAVObjID value, as "if the methods used don't change (in terms of signature) I don't see a need for recompilation, that is how shared libs are supposed to work."

### DIFF
--- a/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
+++ b/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
@@ -89,8 +89,6 @@ bool SysAlarmsMessagingPlugin::initialize(const QStringList &arguments, QString 
  */
 void SysAlarmsMessagingPlugin::updateAlarms(UAVObject* systemAlarm)
 {
-    Q_ASSERT(systemAlarm->getObjID() == SystemAlarms::OBJID);
-
     foreach (UAVObjectField *field, systemAlarm->getFields()) {
         for (uint i = 0; i < field->getNumElements(); ++i) {
             QString element = field->getElementNames()[i];


kubark42:
Re-opening this PR because as per #1974 the Q_ASSERT was not necessary.

kubark42:
Any objections to merging this?

tracernz:
Have you verified that changing the UAVO doesn't silently break stuff? What about replacing the assert with a print to qWarning so at least it will show in red in the debug gadget?

kubark42:
@tracernz this was proposed in https://github.com/TauLabs/TauLabs/pull/1974, but the consensus is that sometime this summer the underlying root of the problem was solved.

tracernz:
Ah, that's right. In that case no objection. ;-)

peabody124:
I don't think we should remove that line. I agree it should not get executed, but checking assumptions of methods for things that should not happen is a reasonable thing to do. I'd prefer a well defined crash (if somehow this were to occur in the future) than something more cryptic.

I certainly won't fight merging this but don't think it is an improvement.

kubark42:
>  I agree it should not get executed, but checking assumptions of methods for things that should not happen is a reasonable thing to do. I'd prefer a well defined crash (if somehow this were to occur in the future) than something more cryptic.

I don't think a seg-fault/core-dump/hard crash is ever well-defined. Other modules are not aware of the segfault and the user isn't given time or opportunity to address the problem. The OS might survive, but unsaved states in the application will not.

This is the sole example in the code of checking that the UAVOs align, so the pattern is unique. If our goal is to sanity check UAVOs, it should happen elsewhere, right? I'm fine with not removing the check if we think this can still occur. But you agreed with us, https://github.com/TauLabs/TauLabs/pull/1974#issuecomment-149991004, that it could not. So I'm confused what you see as the ideal code state for this issue.

@mlyle, you solved the original bug, any thoughts on how this should look?

mluessi:
I think given that this should never happen (i.e., the UAVO ID should never mismatch) having an assert there is fine. If it gets triggered it indicates a bug somewhere else, which is what an assert is there for.

peabody124:
Yeah I agree with @kubark42 this should be consistent and anywhere we run into these issues with inconsistencies we should throw a Q_ASSERT to make sure we catch them further on. Agree with @mluessi this assert is fine. We good to close this PR?

kubark42:
Yes, but only because we need to reopen #1974. I was able to reproduce this build system error by accident. I didn't have time to look into the why, but I was very thankful that #1974 caught it instead of crashing.

IRL, @mluessi agreed that this kind of bug shouldn't lead to a crash with an obscure (i.e. `make clean`) resolution. I think we're on alignment that bugs should rarely lead to crashes if they can be handled more gracefully.

kubark42:
#2113 finally squished this. Closing forever.

PR #1291, opened 2014-09-17
Qt 5.2 related UI fixes

guilhermito:
Miscellaneous ui fixes.
Drops the manhattan style.
Drops the external stlyesheet files usage.
Cleans scattered stylesheets and palettes inside widgets.

The following are some of the issues this tries to address.


mlyle:
Any objections to closing PR? We can open an issue so we don't lose the reference to the code/any remaining items.  Alternatively we should get this ready for merge.

PR #1316, opened 2014-09-23
Make version dialog more user friendly

guilhermito:


mlyle:
Objections to closing PR?  We can open an issue so we don't lose the reference to the code/any remaining items.

Issue #947, opened 2013-09-10
Add support for one shot pwm output


lilvinz:
Like described here:

http://www.multiwii.com/forum/viewtopic.php?f=7&t=2729

http://www.multiwii.com/forum/download/file.php?id=1630&sid=0623051f2324245efeca82df359d0af8&mode=view

mlyle:
There's been a lot of oneshot work and it works great-- flying it on gemini often.  :)  Closed.

PR #1424, opened 2014-12-29
OneShot: inital commit

ernieift:
This adds support for the new OneShot125 servo mode. Normal ESCs use 1000..2000µs @ max. 400Hz for signal transfer. The new Ultra- and KISS-ESC use OneShot125 without a fixed rate. This implementation generates the wanted 125..250µs with 1500 steps based on an internal 12MHz clock. The OneShot is coupled with the PID update rate. The benefit is: There is no jitter between servo rate and sensor rate. The update rate can be up to 2kHz in OneShot according to the ESC description. I've tested it up to 3,9kHz without flying.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -526,7 +526,7 @@ static int16_t scaleChannel(float value, int16_t max, int16_t min, int16_t neutr
 static void setFailsafe(const ActuatorSettingsData * actuatorSettings, const MixerSettingsData * mixerSettings)
 {
 	/* grab only the parts that we are going to use */
-	int16_t Channel[ACTUATORCOMMAND_CHANNEL_NUMELEM];
+	uint16_t Channel[ACTUATORCOMMAND_CHANNEL_NUMELEM];
 	ActuatorCommandChannelGet(Channel);
 
 	const Mixer_t * mixers = (Mixer_t *)&mixerSettings->Mixer1Type; //pointer to array of mixers in UAVObjects
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -379,6 +379,9 @@ static void actuatorTask(void* parameters)
 		{
 			success &= set_channel(n, command.Channel[n], &actuatorSettings);
 		}
+#if defined(PIOS_INCLUDE_ONESHOT)
+		PIOS_Servo_OneShot_Update();
+#endif
 
 		if(!success) {
 			command.NumFailedUpdates++;
@@ -556,6 +559,9 @@ static void setFailsafe(const ActuatorSettingsData * actuatorSettings, const Mix
 	{
 		set_channel(n, Channel[n], actuatorSettings);
 	}
+#if defined(PIOS_INCLUDE_ONESHOT)
+	PIOS_Servo_OneShot_Update();
+#endif
 
 	// Update output object's parts that we changed
 	ActuatorCommandChannelSet(Channel);
@@ -686,6 +692,11 @@ static bool set_channel(uint8_t mixer_channel, uint16_t value, const ActuatorSet
 		case ACTUATORSETTINGS_CHANNELTYPE_PWM:
 			PIOS_Servo_Set(actuatorSettings->ChannelAddr[mixer_channel], value);
 			return true;
+#if defined(PIOS_INCLUDE_ONESHOT)
+		case ACTUATORSETTINGS_CHANNELTYPE_ONESHOT:
+			PIOS_Servo_OneShot_Set(actuatorSettings->ChannelAddr[mixer_channel], value);
+			return true;
+#endif
 #if defined(PIOS_INCLUDE_I2C_ESC)
 		case ACTUATORSETTINGS_CHANNELTYPE_MK:
 			return PIOS_SetMKSpeed(actuatorSettings->ChannelAddr[mixer_channel],value);
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -12,7 +12,7 @@
  *
  * @file       actuator.c
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2014
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2015
  * @brief      Actuator module. Drives the actuators (servos, motors etc).
  *
  * @see        The GNU Public License (GPL) Version 3
--- a/flight/PiOS/STM32F30x/pios_servo.c
+++ b/flight/PiOS/STM32F30x/pios_servo.c
@@ -194,3 +194,75 @@ void PIOS_Servo_Set(uint8_t servo, uint16_t position)
 			break;
 	}
 }
+
+#if defined(PIOS_INCLUDE_ONESHOT)
+#define OneShotFrequency 12000000
+
+/**
+* Set servo position for OneShot
+* \param[in] Servo Servo number (0-num_channels)
+* \param[in] Position Servo position in 1/12 microseconds based on OneShotFrequency
+*/
+void PIOS_Servo_OneShot_Set(uint8_t servo, uint16_t position)
+{
+	/* Make sure servo exists */
+	if (!servo_cfg || servo >= servo_cfg->num_channels) {
+		return;
+	}
+
+	const struct pios_tim_channel * chan = &servo_cfg->channels[servo];
+
+	/* stop the timer */
+	TIM_Cmd(chan->timer, DISABLE);
+
+	/* Update the position */
+	switch(chan->timer_chan) {
+		case TIM_Channel_1:
+			TIM_SetCompare1(chan->timer, position);
+			break;
+		case TIM_Channel_2:
+			TIM_SetCompare2(chan->timer, position);
+			break;
+		case TIM_Channel_3:
+			TIM_SetCompare3(chan->timer, position);
+			break;
+		case TIM_Channel_4:
+			TIM_SetCompare4(chan->timer, position);
+			break;
+	}
+}
+
+/**
+* Update the timer for OneShot
+*/
+void PIOS_Servo_OneShot_Update()
+{
+	if (!servo_cfg) {
+		return;
+	}
+
+	TIM_TimeBaseInitTypeDef TIM_TimeBaseStructure;
+	TIM_TimeBaseStructInit(&TIM_TimeBaseStructure);
+
+	for (uint8_t i = 0; i < servo_cfg->num_channels; i++) {
+		const struct pios_tim_channel * chan = &servo_cfg->channels[i];
+
+		/* Look for a disabled timer whick is probably used by OneShot */
+		if (!(chan->timer->CR1 & TIM_CR1_CEN)) {
+			/* Choose the correct prescaler value for the APB the timer is attached */
+			if (chan->timer==TIM6 || chan->timer==TIM7) {
+				// These timers cannot be used here.
+				continue;
+			} else if (chan->timer==TIM2 || chan->timer==TIM3 || chan->timer==TIM4) {
+				TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB1_CLOCK / OneShotFrequency) * 2 - 1;
+			} else {
+				TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB2_CLOCK / OneShotFrequency) - 1;
+			}
+
+			/* enable it again and reinitialize it */
+			TIM_Cmd(chan->timer, ENABLE);
+			TIM_TimeBaseInit(chan->timer, &TIM_TimeBaseStructure);
+		}
+	}
+}
+#endif
--- a/flight/PiOS/STM32F30x/pios_servo.c
+++ b/flight/PiOS/STM32F30x/pios_servo.c
@@ -8,7 +8,7 @@
  *
  * @file       pios_servo.c
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2012.
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
  * @brief      RC Servo routines (STM32 dependent)
  * @see        The GNU Public License (GPL) Version 3
  *
--- a/flight/PiOS/STM32F4xx/pios_servo.c
+++ b/flight/PiOS/STM32F4xx/pios_servo.c
@@ -194,3 +194,75 @@ void PIOS_Servo_Set(uint8_t servo, uint16_t position)
 			break;
 	}
 }
+
+#if defined(PIOS_INCLUDE_ONESHOT)
+#define OneShotFrequency 12000000
+
+/**
+* Set servo position for OneShot
+* \param[in] Servo Servo number (0-num_channels)
+* \param[in] Position Servo position in 1/12 microseconds based on OneShotFrequency
+*/
+void PIOS_Servo_OneShot_Set(uint8_t servo, uint16_t position)
+{
+	/* Make sure servo exists */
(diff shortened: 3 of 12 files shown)


ernieift:
I've done some outdoor tests with Quanton. I used 1000Hz MPU rate and set the PWM range 1500..3000 to produce 125..250µs signal range (multiplied by 12, because of the 12MHz time base). The copter was measured and set up by auto tune and the PIDs were high (higher than usually). It was very windy but the copter was still very good to control. The Flightmode Level was like nailed.

peabody124:
That is really good to hear. I have some KissESC in SeeingSpark so I'll try enabling that mode on Sparky in a bit.

One thing that would be good to improve - do you think the code that takes the UAVO settings to set the min and max period values could account for the prescalar? I think making sure it stays with physically real units for the user is probably important to avoiding confusion.

flight/Modules/Actuator/actuator.c
```diff
@@ -686,6 +692,11 @@ static bool set_channel(uint8_t mixer_channel, uint16_t value, const ActuatorSet
 		case ACTUATORSETTINGS_CHANNELTYPE_PWM:
 			PIOS_Servo_Set(actuatorSettings->ChannelAddr[mixer_channel], value);
 			return true;
+#if defined(PIOS_INCLUDE_ONESHOT)
```
peabody124: I think you need to move this conditional down a line. In the case oneshot is not included, we should probably do something like set an error to prevent arming. It might also make sense to set PIOS_Servo to 0 in this case to make sure the output is low.

ernieift: You mean, if PIOS_INCLUDE_ONESHOT isn't defined, the servo should be written to 0. There is a  fault handler at line 708. It generates SYSTEMALARMS_ALARM_ACTUATOR later.

peabody124: Oh, nevermind then!

peabody124:
Another thing I was just considering. We need to probably do a bit of work on the GCS side to protect the user from themselves. Basically if a TIM is assigned to one shot mode that also means that PWM input and PPM will not work currently.

We might be able to fix those input layers (by accounting for the prescalar when processing input) or we could block the user trying those combinations.

For now it might be ok to have a big warning in the wiki that shows up when enabling OneShot (since this is only through the UAVObject browser anyway)

ernieift:
Because of the units I have already considered. OneShot running in KISS with 8MHz. So the smallest unit is 125ns. Unfortunately, I can not adjust to 8MHz the prescaler. Otherwise, you could have simply leave the previous units so. See the I2C ESCs. Therefore, I have chosen as a compromise 12MHz and thus used an even higher resolution.
You could make it as I2C course. The unit would be fictitious and it is easy to PIOS layer x * 12MHz / 8MHz multiplied. The wasted later usable fine resolution. To really expect microsecond values of the variables of type uint16 must be changed to real.

ernieift:
Conflicts with TIM should not there be at Quanton. On Sparky, I haven't looked closely. Since I have only tested with HoTT.

sim-:
For the 125µs-250µs mode, what pulse length are you sending when idle? 83µs? I'm wondering because I could add support for this(*), but I want to avoid arming on leader pulses sent by the USB linker. Ideally, it should be longer than about 72µs.

Also, has anybody separated oneshot timer mode from shorter pulses? It seems like it would be useful to test both independently. On that note, what would be the longest loop time (slowest update frequency) that an ESC might see?

ernieift:
Internal (based on 12MHz) it uses 83ns steps. Oneshot uses 125ns steps. 125..250µs is the range like 1000..2000µs for original servos. There is no fixed interval frequency like on servos. The OneShot ESC  has a failsafe at 200ms without an update.
When you normally send 1000µs as idle, OneShot will send 125µs for that. It's just a factor. So you have to use 1500..3000 as range to get 125..250µs based on 12MHz.

sim-:
Ok, so the motor is expected to start at some point higher than 125µs so that 125µs will still allow arming?

200ms is a tad long for my case (https://github.com/sim-/tgy), since my usual timeout is 128ms. I suspect your maximum loop time won't be anything near that, though, right?

ernieift:
125µs is the lowest valid value for OneShot. The KISS spin at the last transferred value until a new value is read with a timeout of 200ms (according to Felix). 200ms are 5Hz rate. If we want to takeoff with such a low MPU rate, the UAV will probably fall out of the sky within seconds. Usually, we have MPU rates of about 300..666Hz (3..1,5ms). I've tested 2000Hz (0,5ms) on a Quanton. This uses 87% CPU but it's still ok. With 1000Hz (1ms) it uses 27% CPU on my setup. The highest accepted ESC rate is 3900Hz. This is far away from what we have now. There is a discussion about OneShot for BlHeli, so it's good to have it clean tested previously.
OneShot helps controlling overpowered acro copters. I've built a light weight 240g copter with MT1306 (3100kV) at 3S. It's like flying with 3S copter with 4S. This thing is rock stable and has high Kp.

sim-:
Does it not make sense to separate shorter pulses from synchronous PWM firing? Is there a reason to combine them? If it makes sense to do synchronous PWM at all, why not always do it?

The edge processing interrupt in my code takes <3 microseconds, so it should be safe to go way above 4kHz, but probably not useful at all. I set a timeout of 128ms because lower exposes some problems on really cruddy AM radios (like in my RC-10), and higher is long enough to cause safety problems with, for example, the i86 controller from HobbyKing, which disarms by sending one full-power pulse before disabling PWM (wonderful). It sounds like this is fine.

So, I'm still wondering what the default range is, then. If you just turn on this option, what is the running pulse length range by default?

ernieift:
I think it is better to have the choice. Normal (older) servos don't like 400Hz PWM frequency an work better on 50Hz. So you can run at selectable fixed frequency or coupled with MPU rate.
I think the KISS simply work on 8MHz time instead of 1MHz. They get the same resolution as before, but they run faster.
OneShot here uses 12MHz timebase. So you have to use 1500..3000 as range to get 125..250µs physically. The KISS will sample this at 8MHz and get the expected 1000..2000 range.
If you simply turn on the option the µs-value is now x*1/12µs.

fujin:
Hi team; just trying to play along at home here (on the brain board), trying to figure out what settings I'm looking to update in uavo. At a guess, I'm thinking something like this to match up with what @ernieift has said.

2015-01-21-182101_1920x1080_scrot

ernieift:
@fujin , you don't have to change the ChannelUpdateFreq. This value is ignored in OneShot. The MIN/NEUTRAL/MAX settings are still available in the output widget, but your values are ok. You have to set ChannelType to OneShot to enable the mode (look a little lower). Then the Update frequency is the MPU Rate. You find this value in the hardware widget on Quanton. It seems, that you can not change this value on brain yet. I think it's fixed to 500Hz. But this is ok.

fujin:
@ernieift thanks, I forgot to post here, but I looked into mpu rate etc and I see it is locked at 500hz. Thanks for the response! I did set the channel types to oneshot as well. I'm soldering everything up now.

fujin:
I got oneshot merged into a brain branch and BrainFPV talking to my KISS ESC's in oneshot with 1500..3000, per this thread. motors start at about 1580. too dark for a flight test / autotune. will post up in the morning!

ernieift:
My motors also start at 1580. I set NEUTRAL to 1600 then.

fujin:
https://www.youtube.com/watch?v=-fQBhNTxMno&feature=youtu.be

peabody124:
fixed #947

peabody124:
I did some tests with this too: http://buildandcrash.blogspot.com/2015/01/oneshot125-quantitative-testing.html

peabody124:
Also measuring the latency from gyro IRQ to the output shows that it is both synchronous and quick :)
screen shot 2015-02-01 at 11 25 27 am
screen shot 2015-02-01 at 11 24 48 am

peabody124:
BTW can you add this patch https://gist.github.com/062120b2805ddd6c8412 or enable it on Colibri yourself?

So the main issue I want to look at with this in the next few days is how we tag it to timers. Make sure nothing funky will happen and that we cleanly disable it for PWM inputs. My thinking is that the assignment of oneshot versus not should maybe happen only at startup.

flight/PiOS/STM32F30x/pios_servo.c
```diff
@@ -194,3 +194,75 @@ void PIOS_Servo_Set(uint8_t servo, uint16_t position)
...
+	TIM_TimeBaseStructInit(&TIM_TimeBaseStructure);
+
+	for (uint8_t i = 0; i < servo_cfg->num_channels; i++) {
+		const struct pios_tim_channel * chan = &servo_cfg->channels[i];
+
+		/* Look for a disabled timer whick is probably used by OneShot */
+		if (!(chan->timer->CR1 & TIM_CR1_CEN)) {
+			/* Choose the correct prescaler value for the APB the timer is attached */
+			if (chan->timer==TIM6 || chan->timer==TIM7) {
+				// These timers cannot be used here.
```
peabody124: why do they not work?

peabody124:
BTW can you update the original time notes to indicate the clock is 12mhz and not 8mhz.

peabody124:
Oh and if you have a chance to update the copyright on any changed files to 2015, that would be great.

flight/PiOS/STM32F30x/pios_servo.c
ernieift: That's what I also asked myself. It's the same like in PIOS_Servo_SetHz(). I think it has a historical reason. A deeper look in the datasheet may help here.
I looked a bit around and found:
Quanton and Colibri use TIM3/10/11/12, RevoMini uses TIM3/9/2/5 for output only.
On Sparky, TIM2 used for servo and pwm. This is worth a note in the wiki. It is not a problem with Spektrum or HoTT receivers.
It seems that TIM6/7 aren't used.

peabody124: Yeah agreed. I dug into the history and it comes from this PR: https://github.com/TauLabs/TauLabs/pull/982/files

Looking at the STM32F4 Reference manual it appears TIM6 and TIM7 are basic timers that do not support PWM output, so that is why :)

flight/PiOS/STM32F30x/pios_servo.c
```diff
@@ -194,3 +194,75 @@ void PIOS_Servo_Set(uint8_t servo, uint16_t position)
...
+{
+	/* Make sure servo exists */
+	if (!servo_cfg || servo >= servo_cfg->num_channels) {
+		return;
+	}
+
+	const struct pios_tim_channel * chan = &servo_cfg->channels[servo];
+
+	/* stop the timer */
+	TIM_Cmd(chan->timer, DISABLE);
```
peabody124: This makes me a bit uncomfortable. It is in a method that is called on one channel but as a result will influence other channels because it's changing the whole timer.

flight/PiOS/STM32F30x/pios_servo.c
```diff
@@ -194,3 +194,75 @@ void PIOS_Servo_Set(uint8_t servo, uint16_t position)
...
+	}
+
+	TIM_TimeBaseInitTypeDef TIM_TimeBaseStructure;
+	TIM_TimeBaseStructInit(&TIM_TimeBaseStructure);
+
+	for (uint8_t i = 0; i < servo_cfg->num_channels; i++) {
+		const struct pios_tim_channel * chan = &servo_cfg->channels[i];
+
+		/* Look for a disabled timer whick is probably used by OneShot */
+		if (!(chan->timer->CR1 & TIM_CR1_CEN)) {
```
peabody124: Then this is a bit off IMO. It's relying on that disabled command above to tag this as a OneShot  channel, but you could also look at the configuration structure to see this explicitly. That would be safer.

flight/PiOS/STM32F30x/pios_servo.c
fujin: Is there a way for this timer to only influence the timing of a particular channel, as opposed to influencing all?

peabody124:
@ernieift I'm tempted to merge this now since it works as advertised. It's an advanced feature that doesn't have a GUI, so it's ok that the user can misconfigure it (have other channels act up) if they go digging.

However, once merged I'd like to rework how it is configured to assign the OneShot settings to the output banks directly. This will make it clear that you have a set of channels on OneShot instead of individual channels.

flight/PiOS/STM32F30x/pios_servo.c
peabody124: No, it is a hardware issue. There is one physical timer (that can support up to four outputs) and the way it behaves in OneShot isn't compatible with traditional PWM output or PWM/PPM input. This isn't a real problem, but is something we should make clear on the GUI to prevent people trying to put one channel on OneShot and having another behave strangely.

ernieift: I still haven't found an easy way to generate a SYSTEMALARMS_ALARM_ACTUATOR for this configuration. It depends on the used flight controller.

flight/PiOS/STM32F30x/pios_servo.c
ernieift: I can not see the OneShot configuration at PIOS level. I thought about making a timer list at PIOS_Servo_OneShot_Set(). This will use more RAM, so i decided to take the timer itself.  All used timers are enabled in PIOS_Servo_Init().

flight/PiOS/STM32F30x/pios_servo.c
ernieift: You can still use normal Servos with 1000..2000µs in OneShot. Simply take 12000..24000µs/12 as range. This way you are able to bring a normal ESC in sync to the MPU rate. Take care not so set the MPU higher than 400Hz, i think.

ernieift:
@peabody124 The better way is to add a field in ActuatorSettings. I have avoided not to take up more space on smaller systems as needed. I want to test this addon on coptercontrol also. I know CF supports the coptercontrol and now sparky but they don't use the interrupt capability of the MPU, so they may get a sensor jitter. I'm interrested in a directly compare. There are some guys out there who switched to CF in order to get OneShot. Most of them are acro pilots. Let's see what they say after flying TL.

fujin:
Here's my "acro pilot research footage" from BrainFPV + oneshot125

- https://www.youtube.com/watch?v=-fQBhNTxMno
- https://www.youtube.com/watch?v=adBiUUJLyHA
- https://www.youtube.com/watch?v=lpL9PcL4h8w

I only fly rate: I previously flew Naze/cleanflight, then before that, baseflight. I support the code as-is with the eye for future refactoring/optimizations.

peabody124:
@ernieift I was planning to use the ActuatorSettings.ChannelUpdateFreq field and use ```(uint16_t) -1``` as a value that indicates OneShot (tempting to use an enum with a set of hardcoded update rates, but I'm not sure that we need to lose the generality.

Are you ok with merging this and having me make those modifications?

ernieift:
I've been thinking a little bit how we can integrate the sensor sync for normal PWM. What about setting the PWM frequency simply to zero to disable the fixed rate in PIOS_Servo_SetHz(). Then we can use the PIOS_Servo_OneShot_Set(12*value) function instead of PIOS_Servo_Set(value) in order to get the same µs-pulses.
This would avoid negative frequency values, since i don't like negative times and we do not need to change the  ChannelUpdateFreq from "uint16" to "int16".
So users need to set the frequency to zero and select OneShot on the used channel. These combinations could checked in actuator_update_rate_if_changed(). If something is wrong, it throws a config error. What do you think?

peabody124:
Yeah zero makes sense rather than (uint16_t) -1. I wasn't going to change to int16, I just didn't feel like looking up the max value.

I wouldn't do anything special (currently) to take normal times with sensor synced. The range of values will still allow 1-2ms pulses even after multiplying by 12, but I think we need to do some tests with ESCs to make sure they behave well. Also, a lot of the boards default to 500+ Hz for the sensors, which won't play nicely with doing that, so we'll need to add checks in the GCS.

You want to make these changes before merging, then?

ernieift:
I think we could merge this first. The PWM changes need more testing. Next week, i get a coptercontrol. So i can validate OneShot here too. It remained outside with the first commit. The reason was i hadn't got one for testing. So let's go ahead and let more users give it a try like it is.

PR #1436, opened 2015-01-04
DSM auto resolution detection

peabody124:
This further expands upon #1411. The original approach to DSM resolution autodetection
was done by Mathieu Rondonneau from the OpenPilot project. This approach was unsafe
because it could result in stack overflows (by infinite recursion for bad data) as well as
allows the resolution to change at any point during execution (e.g. while in air).

I have written a new approach to the resolution detection, based on the fact that it appears
all DSM packets start with channel orders of 1 then 5. This also only allows setting the resolution
one time and stays "unknown" until it gets a positive hit for either 10 or 11 bits.

This is combined with a set of unit tests ```make ut_dsm``` which uses dumps that I parsed from
some files I collected with my own satellite and some that are available online here:
http://wiki.paparazziuav.org/wiki/DSM#Raw_DSM_data_capture_files

- [x] flight tested on F1
- [x] flight tested on F3
- [x] flight tested on F4


peabody124:
Jenkins: test this please

peabody124:
@lilvinz could you post the contents of /tmp/hudson8529857193183861809.sh it is unclear why jenkins is failing on the dsm unit test since it works on my computer and linux VM

peabody124:
Ok, this has been now tested on all the platforms. The Jenkins error doesn't seem to indicate anything that is meaningful, so if we can't figure out why it is flagging the build as an error I'd suggest ignoring that and merging in another day with (hopefully) a few more user tests.

peabody124:
Code wise, I think this is good to merge. @lilvinz could you post "/tmp/hudson8529857193183861809.sh" so I can see why the build fails on the dsm test?

lilvinz:
https://gist.github.com/lilvinz/51e81c0cbe6505c96413

peabody124:
thanks @lilvinz -- looks like it fails all the tests, so different behavior than my computer and ubuntu vm. checking on irc with sounds like it fails on normal 64 bit ubuntu. Let me figure out where sizing matters and make sure that is fixed before merging - not that 64bit arm is something we are switching to soon.

peabody124:
Good to merge

Issue #1359, opened 2014-10-14
Remove feedforward code


peabody124:
This was from the days of old ESCs and terribly hard to tune anyway. This code is no longer needed in either the GCS or in the Mixer.

peabody124:
@elafargue btw if you are still interested in this, i'd say it would be great to strip out the GCS code sooner rather than later. that way if anyone is still using it the firmware will still support it and we can figure out why.

elafargue:
This issue is fixed on the GCS side. Shall we close it and open a separate issue to remove FF on the flight side ?

elafargue:
Closing and opening an issue for flight-side FF code removal.

PR #1437, opened 2015-01-06
Remove all Feed Forward configuration from Ground Control Station.

elafargue:
Relates to Issue #1359.

This removes Feed Forward from the "Vehicle" page of the config gadget, since use of Feed Forward is not considered as very useful anymore on recent ESCs and can be super tricky to tune.

No changes to flight side or UAVObjects, FF can still be configured manually if desired.


peabody124:
Awesome, I'll try and check this in the next few days. It seems like something we should get in before the pending release.

PR #1438, opened 2015-01-11
Update HISTORY.txt

peabody124:
This file had not been updated since the Tau Labs began.
Updated with notes from releases and pending release.


peabody124:
Let me know if there is anything that should be included that I missed. Let's hold this file until the beta testing on the release is done.

Issue #1439, opened 2015-01-11
GCS can't build on Ubuntu 12.04.4


thomaskausm:
Pleases see http://forum.taulabs.org/viewtopic.php?f=17&t=461&p=3840#p3751. 
The fix for the problem is on the way just by adding the required -lrt to the rawhid project.

PR #1441, opened 2015-01-12
Fix make problem "reference to `clock_gettime' collect2: ld returned 1 e...

thomaskausm:
on linux 32 and 64 bit by adding -lrt to libs list of the rawhid project file.
See also https://github.com/TauLabs/TauLabs/issues/1439.


peabody124:
@guilhermito since i know you have linux, can you confirm this doesn't mess up your installation? otherwise it passed jenkins so seems good to me.

PR #1443, opened 2015-01-14
GCS/Output: fix crash when slider range is zero

peabody124:
When the slider min and max values are identical then the
output neutral slider causes a crash when pointed at in
OSX 10.10

https://bugreports.qt.io/browse/QTBUG-43398

### DIFF
--- a/ground/gcs/src/plugins/config/outputchannelform.cpp
+++ b/ground/gcs/src/plugins/config/outputchannelform.cpp
@@ -3,6 +3,7 @@
  *
  * @file       outputchannelform.cpp
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2011.
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2014
  * @addtogroup GCSPlugins GCS Plugins
  * @{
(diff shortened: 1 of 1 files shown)


elafargue:
OK, one small issue: by changing the min/max slider values, you can end up in displays that don't make sense:
screen shot 2015-01-14 at 2 31 16 pm

-> The fix is easy, you can simply add a "setNeutral" call to reset the slider neutral value as well:

        ui.actuatorNeutral->setRange(ui.actuatorMin->value(), ui.actuatorMin->value()+1);
        setNeutral(ui.actuatorMin->value());
        ui.actuatorNeutral->setEnabled(false);

Can you add this call and let me know? Otherwise, 100% fine.

ernieift:
You can also use this:
```
ui.actuatorNeutral->setRange(ui.actuatorMin->value()-1, ui.actuatorMin->value()+1);
```
Then the slider is in the middle and disabled when MIN=MAX.

peabody124:
Updated with both suggestions.

ernieift:
Leaving this untouched:
```
setNeutral(ui.actuatorMin->value());
```
has the charm, that the slider jumps to the limit, when NEUTRAL is out of range.

Issue #1260, opened 2014-09-07
Pipxtreme broken in recent Next


carbogit:
I believe that support for Pipxtreme/Oplink is broken in recent versions. Settings are grey in configuration and object browser and cannot be saved.

mlyle:
Binding wizard makes this happy / I use pipxtreme all the time.

solidgoldbomb:
Closing this.  Should be fixed by #1451.  Feel free to re-open if it's not fixed in builds that include #1451.

PR #1451, opened 2015-01-22
RFM22b rework and binding wizard

peabody124:
This is a pretty major change to the RFM22b code that I have been working on
for a while within the Sparky2 code base. This splits out the RFM22b code from
the Sparky2 code and should get PipX modems working again with next.

 - backport of some of the work from Brian Webb done at OpenPilot
 - improved frequency hopping code
 - fixed some bugs (e.g. power setting was ignored and hardcoded to 100mw previously)
 - improved quality statistics and link monitoring
 - can monitor both directions of link from TauLink modem side
 - improved configuration (move configuration into hardware specific objects) as well as substantially simplifying it
 - implemented a binding wizard to facilitate configuration (see screenshot below). Access through Tools->RFM22B bind wizard
 - moved link monitoring from the config gadget to a separate widget (see screenshot below)
 - rework the receiver configuration to better fit our driver model. PIOS_Rcvr directly fetches channel values from RFM22b driver. RFM22bReceiver object is purely for monitoring.
 - modem side will automatically stream out serial port when port is set to telemetry and no USB is connected

Bind wizard configuration page:
screen shot 2015-01-22 at 12 58 41 pm

Link monitoring: 
screen shot 2015-01-22 at 1 02 09 pm


peabody124:
I've tested this quite heavily, but predominantly in the context of Sparky2. More testing with PipX->PipX and RevoMini would be very useful

peabody124:
@elafargue if you have a chance can you try this on your modems?

BrainFPV:
Is this planned to be included in the January release? I assume no, as it is quite a major change.

peabody124:
If someone was aggressive testing it on some hardware, then it's possible, but no - it's not a planned inclusion so probably not.

elafargue:
Quick feedback: PipXtreme now works again on this branch! But I cannot see the TauLink configuration tab in the Config gadget anymore? Did you remove it altogether ?

peabody124:
There also seems to be an issue where the System module stops running (Status LED stops blinking and the FlightTime doesn't go up) although the rest of the board isn't being sent. The RFM22BStatus object is also not sent from the FC side.

elafargue:
Managed to get PipX to PipX communication - my issue was a telemetry serial port default speed issue.  Our boards tend to use 57600 bps as default com speed, but the wizard defaults to 38400.

elafargue:
Suggestion for Wizard intro text:

This wizard will help you pair your flight controller modem and your ground modem. In the next two pages you will have to plug in first one and then the other.

Please make sure that both modems are not both plugged in at the same time, and that no flight controller modem other than the one you want to pair with is powered up.

If your primary goal is to receive telemetry, then it is best to plug in the flight controller modem first. If your goal is to use the modem for control (i.e. relaying PPM) then plug in the ground modem first.

If you want to bind multiple flight controllers to the same ground modem, the easiest way to do this is to make the ground modem the coordinator. Be aware that in that case, you should never have two flight controllers on at the same time.

peabody124:
Fixes #1260

peabody124:
jenkins: test this please

peabody124:
@BrainFPV @elafargue @solidgoldbomb @lilvinz anyone want to look at this code? I'm not sure many people have hardware to test this yet. I've tested it fairly extensively and it works well.

If no one wants to review it, can we merge it in the next few days?

peabody124:
Ok I'll merge this today if there are no objections

PR #1452, opened 2015-01-23
Stabilization: Horizon mode should saturate at ManualRate

peabody124:
Was accidentally using MaximumRate which only should apply to the
inner loop for Attitude mode.


peabody124:
Also added a fix for a typo

BrainFPV:
LGTM

Issue #1454, opened 2015-01-23
MWRate/RateMW naming consistency


fujin:
As discussed in IRC, the Stabilization tab refers to MWRate, however the Flight Switch Mode stabilization axis' selector refers to RateMW. I propose standardizing on one name for this terminology.

solidgoldbomb:
This appears to be fixed by #1455.

PR #1455, opened 2015-01-23
Flight: use MWRate consistency

peabody124:
The UI all uses the phrase 'MWRate' but the actual flight
mode was using RateMW so the options presented did not match
up. This fixes everything to use MWRate.


peabody124:
fixes #1454

fujin:
One minor thing we might want to get in here, I believe there is a reference in the MWRate panel that says to "set one of your flight modes to mwrate", or similar? when in fact, one must set the individual axes of a given Stabilization flight mode to MWRate, where applicable, as opposed to selecting a MWRate Flight Mode

peabody124:
Done

peabody124:
@BrainFPV could you check this code too?

BrainFPV:
LGTM

BrainFPV:
@peabody124 the OSD menu also had the MWRate/RateMW confusion, I blame you for that ;)

PR #1457, opened 2015-01-24
GCS: Implement update of "Temp/Baro" visualization within the GCS.

thomaskausm:
Temp/Baro: SYSTEMALARMS_ALARM_ERROR is set when the sensor valus are not readable
                    SYSTEMALARMS_ALARM_WARNING when the value read from the sensors a not a number

See also http://forum.taulabs.org/viewtopic.php?f=18&t=504&p=4124#p4124

Please review and give feedback.

Regards
Thomas

### DIFF
--- a/flight/Modules/Sensors/sensors.c
+++ b/flight/Modules/Sensors/sensors.c
@@ -221,18 +221,18 @@ static void SensorsTask(void *parameters)
 				// if we were in baro error state clear it.
 				// To avoid sending the clear signal all the time
 				// "in_baro_error_state" is used.
-				if (in_baro_alarm_state){
+				if (in_baro_alarm_state) {
 					AlarmsClear(SYSTEMALARMS_ALARM_TEMPBARO);
 					in_baro_alarm_state = false;
 				}
-
-			}
-			// Check that we got valid sensor datas
-			uint32_t dT_baro_datas = PIOS_DELAY_DiffuS(last_baro_update_time);
-			// if the last valid sensor datas older than 100 ms report an error
-			if (dT_baro_datas > MAX_TIME_BETWEEN_VALID_BARO_DATAS_MS) {
-				AlarmsSet(SYSTEMALARMS_ALARM_TEMPBARO, SYSTEMALARMS_ALARM_ERROR);
-				in_baro_alarm_state = true;
+			} else {
+				// Check that we got valid sensor datas
+				uint32_t dT_baro_datas = PIOS_DELAY_DiffuS(last_baro_update_time);
+				// if the last valid sensor datas older than 100 ms report an error
+				if (dT_baro_datas > MAX_TIME_BETWEEN_VALID_BARO_DATAS_MS) {
+					AlarmsSet(SYSTEMALARMS_ALARM_TEMPBARO, SYSTEMALARMS_ALARM_ERROR);
+					in_baro_alarm_state = true;
+				}
 			}
 
 		}
--- a/flight/Modules/Sensors/sensors.c
+++ b/flight/Modules/Sensors/sensors.c
@@ -170,7 +170,6 @@ static void SensorsTask(void *parameters)
 	lastSysTime = PIOS_Thread_Systime();
 	uint32_t good_runs = 1;
 	uint32_t last_baro_update_time = PIOS_DELAY_GetRaw();
-	bool in_baro_alarm_state = false;
 
 	while (1) {
 		if (good_runs == 0) {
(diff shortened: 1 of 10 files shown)


peabody124:
Congratulations on your first PR :). One problem with this approach is that we typically try and avoid working with UAVOs (including AlarmSet) from PIOS_Drivers. The changes in sensors.c look great. For the error you probably want to check for a timeout on the baro queue which should happen if the reads fail and nothing is pushed.

thomaskausm:
Sounds logical. I will follow your proposal and implement a timeout within the sensor.c.

thomaskausm:
As proposed I changed the code so that now the  "AlarmSet" gets called within the sensors.c after
a 100 ms timeout  and not within the  PIOS.

thomaskausm:
With the last change I removed the sensor read check completly .... now everything should be fine.
Sorry for that.
Thomas

thomaskausm:
Hello peabody124,
I changed the code in the way you have proposed. It would be great if you
could review it again.
Thanks and Regards
Thomas


2015-01-25 16:10 GMT+01:00 peabody124 <notifications@github.com>:

flight/Modules/Sensors/sensors.c
```diff
@@ -209,8 +212,18 @@ static void SensorsTask(void *parameters)
 		}
 
 		queue = PIOS_SENSORS_GetQueue(PIOS_SENSOR_BARO);
-		if (queue != NULL && PIOS_Queue_Receive(queue, &baro, 0) != false) {
-			update_baro(&baro);
+		if (queue != NULL) {
+			if (PIOS_Queue_Receive(queue, &baro, 0) != false) {
+				// we can use the timeval because it contains the current time stamp (PIOS_DELAY_GetRaw())
+				last_pios_sensor_update = timeval;
```
peabody124: this variable should be renamed. it is specifically the last baro update, not a general pios_sensor update. maybe ```last_baro_update_time```

peabody124:
Have you been able to test this? I don't see anything that clears that error if you start getting data from the baro again.

flight/Modules/Sensors/sensors.c
```diff
@@ -209,8 +212,18 @@ static void SensorsTask(void *parameters)
...
+				update_baro(&baro);
```
peabody124: Here you should probably add
```AlarmsClear(SYSTEMALARMS_ALARM_TEMPBARO);```
and wrap the code below in an else statement since there is no need to compare the time immediately after getting an update.

**Comment by thomaskausm** [comment]
I pressed the wrong  button ..... I totally agree to your comments and implement the suggestions in "80afe9f" and  "1edbd90".  I tested the change via a little change  in pios_ms5611.c 

static void PIOS_MS5611_Task(void *parameters)
{
..
..
...
		if (counter > 1000 && read_adc_result == 0) {
			PIOS_Queue_Send(dev->queue, &data, 0);
		}
		counter++;
                if (counter > 2000){
                	counter=0;
                }
	}

which shows now a nice blinking "Temp/Baro" sign. Initially I forgot the case that the sensor is only temporary not working completely.  

Thanks and Regards
Thomas

**Comment by peabody124** [comment]
Something got weird in the merges. Note how the change log shows lots of fixes that are unrelated now. Can you try typing this:

"git fetch origin
git rebase origin/next"

and see if it makes a clean history.  Sorry this is being so tricky.

**Comment by peabody124** [comment]
jenkins: test this please

File: flight/Modules/Sensors/sensors.c [hunk 3]
```diff
@@ -209,8 +213,28 @@ static void SensorsTask(void *parameters)
...
-			update_baro(&baro);
+		if (queue != NULL) {
+			if (PIOS_Queue_Receive(queue, &baro, 0) != false) {
+				// we can use the timeval because it contains the current time stamp (PIOS_DELAY_GetRaw())
+				last_baro_update_time = timeval;
+				update_baro(&baro);
+				// if we were in baro error state clear it.
+				// To avoid sending the clear signal all the time
+				// "in_baro_error_state" is used.
+				if (in_baro_alarm_state) {
```
peabody124: No need for this condition or tracking the state. Just call AlarmsClear directly. The library takes care of checking the previous condition https://github.com/TauLabs/TauLabs/blob/next/flight/Libraries/alarms.c#L74

flight/Modules/Sensors/sensors.c
```diff
@@ -209,8 +213,28 @@ static void SensorsTask(void *parameters)
...
+					AlarmsClear(SYSTEMALARMS_ALARM_TEMPBARO);
+					in_baro_alarm_state = false;
+				}
+			} else {
+				// Check that we got valid sensor datas
+				uint32_t dT_baro_datas = PIOS_DELAY_DiffuS(last_baro_update_time);
+				// if the last valid sensor datas older than 100 ms report an error
+				if (dT_baro_datas > MAX_TIME_BETWEEN_VALID_BARO_DATAS_MS) {
+					AlarmsSet(SYSTEMALARMS_ALARM_TEMPBARO, SYSTEMALARMS_ALARM_ERROR);
+					in_baro_alarm_state = true;
```
peabody124: Same here. Delete the in_baro_alarm_state

thomaskausm:
Hi peabody,
I change the code in the requested way. Thanks for checking.
Regards
Thomas

peabody124:
jenkins: test this please

PR #1458, opened 2015-01-25
Sparky ext mag

jihlein:
During bench testing I noticed a quick flash of the event system from green to yellow, then it remains green.  That might merit some investigation, however I can't find anything, at least anything obvious.

Note that I changed the I2C pin definitions for both the internal and external buses,  I made the pins open drain/pullup, rather than push pull/no pull.  I'm not sure how I2C worked with push pull outputs.  I also turned off the analog I2C filter, I've had instances where the HMC5883 did not like that filter for some reason.

Anyway, review this as it's my first adventure with this code base, make sure I did not do anything stupid.....

-JI


peabody124:
jenkins: add to whitelist
jenkins: test this please

peabody124:
I'll probably have to rework this patch for you. Your editor changed all the line endings to \r\n which is not our standard on the flight side.

jihlein:
Sorry for that, I'm just using textpad, I'll reconfigure it for next time.
Want me to re-work it and re-submit?

-JI

Date: Sun, 25 Jan 2015 06:17:08 -0800

peabody124:
Sure, that would be great. You can probably also use dos2unix to convert the files. I'd also recommend taking this opportunity to learn how to use ```git rebase``` to edit your previous commits and make the changes clean. Then you can use a ``git push <branch name> -f``` to overwrite the branch and update the PR.

Note: not required to do this, you can also open a new PR. It's just a good chance to learn.

peabody124:
Also while rebasing, could you maybe change the commit title. Normally we do something like

```
Topic: summary

Details go here
```

following the linux convention. So this could be "Sparky: support external mag"

jihlein:
Found a text editor that will let me save in the UNIX format with only LF at the end of the lines.  So it was easy to update the files.  I made a small improvement to the code too.
I'm new to GIT and the advanced features.  I researched rebase, and when I tried it, it would not allow me to rebase until I committed the changes.  After committing the changes, it said there was nothing to rebase.  Not sure where I went wrong.
Given that, would it be easier and cleaner at this point to reject the original pull request and just have me submit a new one?
Sorry for making this so difficult.......
-JI

Date: Mon, 26 Jan 2015 07:45:23 -0800

peabody124:
You aren't making it difficult, don't worry. Git is a hard tool to learn. Your work on the code is much appreciated. Absolutely feel free to create a new one. However, I'll suggest some commands in case you want to play :)

This makes a new branch too, so you don't risk you current code.

```
git checkout sparkyExtMag
```
This goes to your current branch
```git checkout -b sparkyExtMag_rebase```
This creates a new copy of it to be safe.
```git fetch```
Make sure you have recent code
```git rebase origin/next```
This applies your changes "on top" of next, so you only have those commits. It assumes our repository is "origin" and not your branch.

jihlein:
Closing Pull Request and issue new one with properly formatted files.

PR #1464, opened 2015-01-26
Sparky ext mag

jihlein:
Sparky: Adding support for external magnetometer

This changes allow use of an external HMC5883 magnetometer on Sparky's flexiport.

It has been bench tested, showing proper selection of the internal or external magnetometer, and proper
selection of the external magnetometer orientation.

jihlein@hotmail.com


peabody124:
I'll need to do some work on this. There are still a lot of whitespace changes being applied that shouldn't be.

peabody124:
Started working on it here #1481 -- you also had some changes to the I2C clocking I haven't ported yet. I wasn't sure why you were changing the behavior of the internal I2C bus.

jihlein:
Regarding I2C clocking, I've had issues with the analog filter. So I turned it off and recalculated the clock setup word.  Feel free to leave this out.
I really don't understand how the I2C bus was working with SCL and SDA defined as push-pull. All I2C masters I've worked with before were setup as open drain with pull-ups.  And if I recall correctly, that's how the I2C spec is written.
I think that covers my changes to the I2C setup. 
I'm on business travel the next few days, I can look for the references when I get home.

peabody124:
Ok I set the I2C bus to 50 KHz, but we can adjust it or the analog filter. I'm not sure we had to change the clock source though.

Regarding the push pull, I'm pretty certain when assign a pin to an alternative function it overrides the GPIO settings appropriately. You are correct that I2C requires pull up.

peabody124:
Closed this for #1481

jihlein:
The clock source change was to move from the internal oscillator to the system clock. It's a more stable clock source.

peabody124:
Ah cool. That's a reasonable thing to do, but we should probably do it in a separate PR and do it inside PIOS_I2C (as you suggested) and apply it to all targets.

PR #1477, opened 2015-02-02
Chibios

peabody124:
It seems reasonable to start working on a PR for ChiBios. This will need a lot of testing
and verification before merging.

 - [x] test on FlyingF3
 - [x] test on Sparky
 - [x] test on Quanton


peabody124:
This is not for merging yet. Test with caution.

@lilvinz please post any blockers or issues you can think of

peabody124:
rebased onto next

peabody124:
Sparky: heap remaining 7016 irq 512 cpu 35 (previously heap 6034 irq 512 cpu 62) behaves correctly

peabody124:
jenkins: test this please

peabody124:
FlyingF3 heap 13384 cpu 30 (previously heap 11674 irq 512 cpu 50)

peabody124:
Quanton heap 96436 irq 512 cpu 19 (previously heap 89828 irq 512 cpu 31)

peabody124:
Sparky2 heap 94296 irq 512 cpu 18 (previously heap 89752 irq 512 cpu 28)

peabody124:
Been flying this afternoon with Sparky2 running this. All seems to be working.

peabody124:
Merging tomorrow if there are no objections

elafargue:
What additional tests should we run before merging? This is a pretty major change, there must be a few borderline cases that can lead to issues...

peabody124:
I can't really come up with anything. The changes to abstract out the queue/thread/semaphore things have been in for a while and I've tested it a number of times. At this point I think rolling it out in next will give us the coverage we need in case there are other issues.

lilvinz:
IIRC this doesn't work for simulation yet. One option is sticking with FreeRTOS for simulation.

peabody124:
@lilvinz Correct. We can always make that change later.

elafargue:
Let's try to include Colibri in the targets before the merge ? I volunteer to test.

peabody124:
Ok I looked into this. I actually really think we should hold Colibri back. The reason is that this change requires an update to the bootloader for quanton and colibri. I'm fairly comfortable with that for Quanton (@lilvinz feel free to disagree here) because those users tend to be more savvy.

Gemini users are looking for a fairly plug-and-play solution. We'll need to coordinate with TBS if we make this change so they can update all their manuals accordingly. Once this is merged I'll create a PR for Gemini only so that @tbs-trappy can comment on it and decide what they want to do. We might also do something like strip some code (e.g. PicoC, FixedWingFollower) from Colibri so that remains with the configured flash space.

elafargue:
Looks good to me, I agree on the approach on Colibri, let's sync with TBS on that front.

PR #1478, opened 2015-02-02
Erase flash

peabody124:
Make it easy to compile an erase flash version of the firmware for targets. This was
already supported on Sparky.

Compile firmware with
```make fw_<board_name>```


peabody124:
@BrainFPV could you glance at this? Should be straightforward.

BrainFPV:
LGTM

PR #1482, opened 2015-02-06
Rework of the arming and disarming code

peabody124:
This fairly heavily reworks the arming code to make it cleaner and more
maintainable. On top of that, it introduces these new features

 - selectable arming and disarming times
 - a new mode that allows arming on a switch but only when throttle is low
 - arming by holding the sticks in the corners (similar to DJI). in this mode the same position arms and disarms


peabody124:
@BrainFPV @elafargue it would be great to have a few eyes on this to really verify the logic, since the arming/disarming code is a safety critical feature. I've tested this pretty extensively and can't find any issues.

cGiesen:
I like the new options!

I'n not sure, if it belong to arming. 
Very often my motors going on for a short time, after reset on quanton board

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -490,139 +478,205 @@ static void set_armed_if_changed(uint8_t new_arm) {
...
+	case MANUALCONTROLSETTINGS_ARMING_YAWRIGHT:
+		return lowThrottle && cmd->Yaw > ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_CORNERS:
+		return lowThrottle && (
+			(cmd->Yaw > ARMED_THRESHOLD || cmd->Yaw < -ARMED_THRESHOLD) &&
+			(cmd->Roll > ARMED_THRESHOLD || cmd->Roll < -ARMED_THRESHOLD) );
+	case MANUALCONTROLSETTINGS_ARMING_SWITCH:
+		return cmd->ArmSwitch == MANUALCONTROLCOMMAND_ARMSWITCH_ARMED;
+	case MANUALCONTROLSETTINGS_ARMING_SWITCHTHROTTLE:
+			return lowThrottle && cmd->ArmSwitch == MANUALCONTROLCOMMAND_ARMSWITCH_ARMED;
```
BrainFPV: clean indent

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -490,139 +478,205 @@ static void set_armed_if_changed(uint8_t new_arm) {
...
-				break;
+	case ARM_STATE_DISARMING:
+	{
+  		bool disarm = disarming_position(cmd, settings) && valid;
+		uint16_t disarm_time = (settings->DisarmTime == MANUALCONTROLSETTINGS_DISARMTIME_250) ? 250 : \
+			(settings->DisarmTime == MANUALCONTROLSETTINGS_DISARMTIME_500) ? 500 : \
+			(settings->DisarmTime == MANUALCONTROLSETTINGS_DISARMTIME_1000) ? 1000 : \
+			(settings->DisarmTime == MANUALCONTROLSETTINGS_DISARMTIME_2000) ? 2000 : 1000;
+  		if (disarm && timeDifferenceMs(armedDisarmStart, lastSysTime) > disarm_time) {
+			arm_state = ARM_STATE_DISARMING_STILL_HOLDING;
```
BrainFPV: If the user keeps holding the sticks while disarming, shouldn't we immediately disarm after the timeout? I see that it makes sense for arming, but why for disarming?

shared/uavobjectdefinition/manualcontrolsettings.xml
```diff
@@ -105,13 +105,15 @@
...
-				<option>Pitch Aft</option>
 				<option>Yaw Left</option>
 				<option>Yaw Right</option>
+				<option>Corners</option>
 				<option>Switch</option>
+				<option>Switch+Throttle</option>
 			</options>
 		</field>
+		<field name="ArmTime" units="ms" type="enum" elements="1" options="250,500,1000,2000" defaultvalue="1000"/>
+		<field name="DisarmTime" units="ms" type="enum" elements="1" options="250,500,1000,2000" defaultvalue="1000"/>
```
BrainFPV: I think we should have longer disarm timeouts. Someone might want to trun off throttle on a multirotor for a while, so it falls for a while and then throttle up again. It woud be bad if it disarms in this case. Before it was possible to set the timout in seconds, which wa better IMO.

flight/Modules/ManualControl/transmitter_control.c
peabody124: fixed

flight/Modules/ManualControl/transmitter_control.c
peabody124: the behavior as it goes to the next state is to immediately disarm (see line 670). however, it has to stay there so if they hold the sticks it doesn't enter the disarmed state and then immediately start arming again.

I've renamed ```ARM_STATE_DISARMING_STILL_HOLDING``` to ```ARM_STATE_DISARMED_STILL_HOLDING``` to make it clear this is already disarmed.

shared/uavobjectdefinition/manualcontrolsettings.xml
peabody124: the timeout is still there, this is a different thing. it is how long you have to hold in the arm/disarm position for it to become active. The logic for actually disarming based on no transmitter is here https://github.com/TauLabs/TauLabs/pull/1482/files#diff-4f4804c22880c52150730a11f3c4a5baR636

peabody124:
jenkins: test this please

peabody124:
@BrainFPV any other comments? is this good to merge...

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -490,139 +478,205 @@ static void set_armed_if_changed(uint8_t new_arm) {
...
+	case MANUALCONTROLSETTINGS_ARMING_ROLLRIGHT:
+		return lowThrottle && cmd->Roll > ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_YAWLEFT:
+		return lowThrottle && cmd->Yaw < -ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_YAWRIGHT:
+		return lowThrottle && cmd->Yaw > ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_CORNERS:
+		return lowThrottle && (
+			(cmd->Yaw > ARMED_THRESHOLD || cmd->Yaw < -ARMED_THRESHOLD) &&
+			(cmd->Roll > ARMED_THRESHOLD || cmd->Roll < -ARMED_THRESHOLD) );
```
BrainFPV: What about Pitch? I think we should also require pitch down, so it is the same as on DJI controllers.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -490,139 +478,205 @@ static void set_armed_if_changed(uint8_t new_arm) {
...
+	case MANUALCONTROLSETTINGS_ARMING_ROLLRIGHT:
+		return lowThrottle && cmd->Roll < -ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_YAWLEFT:
+		return lowThrottle && cmd->Yaw > ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_YAWRIGHT:
+		return lowThrottle && cmd->Yaw < -ARMED_THRESHOLD;
+	case MANUALCONTROLSETTINGS_ARMING_CORNERS:
+		return lowThrottle && (
+			(cmd->Yaw > ARMED_THRESHOLD || cmd->Yaw < -ARMED_THRESHOLD) &&
+			(cmd->Roll > ARMED_THRESHOLD || cmd->Roll < -ARMED_THRESHOLD) );
```
BrainFPV: same here

BrainFPV:
LGTM otherwise

tbs-trappy:
DJI-style arming works as advertised.

BrainFPV:
Nice :)

PR #1483, opened 2015-02-06
uploader: stop using a local variable for firmware CRC calculation

elafargue:
...and Malloc' the memory instead, so that there is no risk to blow the stack on devices with very large
firmwares (Happened on Mac on BrainFPV, MacOS 10.10.2).

Tested on MacOS X, with BrainFPV (BrainFPV fork of the GCS) and Quanton (TauLabs).

Needs to be thoroughly tested on other platforms and other flight controllers...


peabody124:
@elafargue can you actually do a ```commit --amend``` on this and then a git force push on your branch? Changing the title on github unfortunately doesn't clean up the commit message. I'll merge after that.

elafargue:
Done!

PR #1487, opened 2015-02-09
Trimming: allow trimming level in AH mode

peabody124:
Previously only a few modes used the vehicle trim,
but since now all of them do, it should be possible
to trim in altitude hold mode.


peabody124:
@elafargue can you glance at this? it should be an easy merge

PR #1489, opened 2015-02-11
Woraround OSX Qt bug for input sliders

BrainFPV:
Workaround Qt bug 43398. Like #1443 but for input sliders.

https://bugreports.qt.io/browse/QTBUG-43398


peabody124:
Tested and it works.

PR #1491, opened 2015-02-16
RFM22b Bind: update all settings in one shot

peabody124:
Updating the settings in multiple functions would
sometimes result in the last setting not getting
through reliably. This changes it to update all
the settings in one function.


peabody124:
I'll merge this if there are no objections. It's tested and just a small change.

peabody124:
Merged with good feedback from @elafargue

Kawobei:
Does the wizard give feedback if the save operations don't happen correctly?

PR #1493, opened 2015-02-21
Horizon: add separate exponential

peabody124:
Previously horizon used the exponential from the rate mode
for the rate component, but there was no ability to add
expo for the transition from attitude to rate. From experiments
with the transmitter, this is a better type of expo to use
in this mode.


elafargue:
Does this need a GCS-side configuration somewhere ?

peabody124:
I'm not sure - we could also just leave it in the browser as an advanced settings. I don't htink either this or the rate expo are exposed in the GCS though, so it would be reasonable to add to the advanced tuning settings.

tbs-trappy:
so much better now. flies amazing and it's a lot easier to control. flew 2 packs through it, going to fly some more to tune the settings. does the yaw expo affect the overall expo even if the yaw is set to axislock?

peabody124:
@elafargue does this code look good to go for you?

@tbs-trappy the rate yaw does have an effect in both regular rate and axis lock mode

peabody124:
I'll merge this if there are no comments

guilhermito:
Looks good

PR #1495, opened 2015-02-25
Autotune: reduce module stack size

ernieift:
This reduces the stack size from 2000 to 1500 in autotune module. It
frees some memory and autotune can be enabled on coptercontrol again.
Tested on CC3D and works.

### DIFF
--- a/flight/Modules/Autotune/autotune.c
+++ b/flight/Modules/Autotune/autotune.c
@@ -46,7 +46,7 @@
 #include "pios_thread.h"
 
 // Private constants
-#define STACK_SIZE_BYTES 2000
+#define STACK_SIZE_BYTES 1500
 #define TASK_PRIORITY PIOS_THREAD_PRIO_NORMAL
 
 #define AF_NUMX 13
--- a/flight/Modules/Autotune/autotune.c
+++ b/flight/Modules/Autotune/autotune.c
@@ -46,7 +46,7 @@
 #include "pios_thread.h"
 
 // Private constants
-#define STACK_SIZE_BYTES 1500
+#define STACK_SIZE_BYTES 1700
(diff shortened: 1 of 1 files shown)


peabody124:
So I tested autotune on an F4 using the original code. At the end it drops to 492 bytes (the last drop happens when you land and disarm). When testing this branch, I actually got a reboot.

You can see the extra stack remaining in ```TaskInfo.StackRemaining[Autotune]```. In principle you can
enable these diagnostics on CC3D by enabling the ```DIAG_TASKS``` flag in the Makefile. It costs a bit of stack to enable so can get you between a rock and a hard place for debugging.

I'm pretty sure the most we can safely drop autotune memory to is 1600.

ernieift:
When i disarmed in flight mode autotune, my CC rebooted once. This wasn't reproducable. After that, i switched to rate before i disarmed. Maybe UAVObjSave() uses sometimes more stack. If this is the problem, i can work around with 1600 bytes of stack. I'll try this tomorrow.
With 1500 i have 300 bytes memory free on CC. So there is space.

peabody124:
300 total bytes or for that particular module? It's possible the memory requirements are different on F1 and F4 ...

ernieift:
300 Bytes HeapRemaining in Systemstats.

ernieift:
I've done a little test without propellors. 1600 for Stack ist not enough and 1650 either. With 1700 ist works on CC. Then there are only 100 Bytes Heap left. I have to check a F4 target too.

peabody124:
Yeah for future reference there is total HeapRemaining but also per thread stack sizes...

peabody124:
Perfect. 196 bytes free in autotune stack on F4 after finished. No change for second run.

PR #1497, opened 2015-02-28
sparky: put pios_can_cfg in FLASH to save a few RAM bytes

nanodude:
On Sparky, pios_can_cfg is currently in RAM. As there appears to be no need to modify this, the modifier static const has been added to move this to FLASH and save a few precious RAM bytes.


nanodude:
@peabody124 is this OK with you?

peabody124:
jenkins: test this please

@nanodude looks good to me. I want to just test on hardware later.

Issue #1499, opened 2015-03-03
UAVOLightTelemetryBridge task runs even when disabled in config


taumcg:
Finally traced a recurring stack overflow to UAVOLightTelemetryBridge, which was running task even though it was disabled in config.  I have a PR to disable starting light telemetry when it is not configured, but I suspect it might need a larger stack when it is configured.

peabody124:
Good catch! https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/UAVOLighttelemetryBridge/UAVOLighttelemetryBridge.c#L102 it isn't doing the normal technique for how to decide whether to initialize.

elafargue:
Shall we close this one ?

mlyle:
Looks to be resolved-- closing.

PR #1500, opened 2015-03-03
Dont start UAVOLightTelemetryBridge unless requested

taumcg:
UAVOLightTelemetryBridge task was always starting regardless of whether selected in configuration.  Running on Sparky, this led to a stack overflow in the UAVOLightTelemetryBridge task even though it was not
being used.  This fix disables the UAVOLightTelemetryBridge task if it is not configured to run (no open port associated during init).  This fix does not attempt to solve the stack overflow when UAVOLightTelemetryBridge is intentionally enabled, and perhaps it just works in that case already.

### DIFF
--- a/flight/Modules/UAVOLighttelemetryBridge/UAVOLighttelemetryBridge.c
+++ b/flight/Modules/UAVOLighttelemetryBridge/UAVOLighttelemetryBridge.c
@@ -69,8 +69,8 @@
 // Private types
 
 // Private variables
+static bool module_enabled;
 static struct pios_thread *taskHandle;
-
 static uint32_t lighttelemetryPort;
 static uint8_t ltm_scheduler;
 static uint8_t ltm_slowrate;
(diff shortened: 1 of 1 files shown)


taumcg:
Is there some way to connect an issue to a PR?  Thx.

peabody124:
It would be good to add a flag to the modulesettings as well for consistency with the other optional modules.

peabody124:
jenkins: test this please

peabody124:
Fixes #1499

peabody124:
Done by the message I typed

taumcg:
Thanks, I'll have to try creating a link like that next time.  I thought maybe it was some sort of built-in github thing to link them.

taumcg:
Regarding the consistency in checking the modulesettings flag versus other UAVO modules, it seemed redundant to check the flags AFTER checking the port_id (implies PIOS_COM_Init succeeded), but if consistency is important then I'll be glad to add the extra test.  If port_id is valid but module flag is disabled then it would seem something else is wrong higher up, unless there are some valid hard-coded port_id values.  Still trying to learn how all the modules, tasks, and UAVOs are stitched together.

Thx.

peabody124:
Yeah, it is redundant but it is consistent with the way other modules are enabled which is probably a better user experience, ultimately. You can check that module flag first though, since if that is disabled then the port doesn't matter.

See how this flag is used: https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/CameraStab/camerastab.c#L104

taumcg:
Ok, revised to check both port and admin flag, using the similar "module_enabled" style.
Please note that I can only verify this works to disable the LightTelemetry module, but I don't have the hardware or knowledge to confirm that the module works as intended when intentionally enabled.
Thx.

peabody124:
jenkins: test this please
jenkins: add to whitelist

PR #1502, opened 2015-03-04
RevoMini: update RTC clock rate

peabody124:
It was running incorrectly which subsequently broke
the SBUS driver.

Thanks to @pug368 for lending me an SBUS TX/RX to
fix this.


peabody124:
Any objections to merging this? It's fairly straightforward. Not sure if anyone wants to test SBus on RevoMini.

PR #1506, opened 2015-03-08
EKF: 14 state INS with z-axis bias and mag heading mode

peabody124:
When engaging altitude hold using the INS it is common to see a
small glitch in the altitude. This arises because of untracked bias
in the accelerometer. This new INS fixes this issue by tracking
the accelerometer bias.

In addition it changes how the mag is dealt with so it only influences
yaw and cannot affect roll and pitch. This creates a better attitude
estimate in practice.

Finally it adds unit tests for the INS with regards to stability and
convergence rates.

Much of this work is written up here: http://buildandcrash.blogspot.com/2015/01/ekf-enhancements-and-unit-tests.html


peabody124:
Testing and code review extremely welcome! I feel quite good about this code.

flight/Libraries/insgps14state.c
```diff
@@ -0,0 +1,905 @@
...
+	//  Pnew = (I+F*T)*P*(I+F*T)' + T^2*G*Q*G' = scalar expansion from symbolic manipulator
+
+	T = dT;
+	Tsq = dT * dT;
+
+	for (i = 0; i < NUMX; i++)	// Create a copy of the upper triangular of P
+		for (j = i; j < NUMX; j++)
+			D[i][j] = P[i][j];
+
+	// Brute force calculation of the elements of P
```
BrainFPV: Where does this come from? Would be good to have some explanation.

flight/Libraries/insgps14state.c
```diff
@@ -0,0 +1,905 @@
...
+	Xdot[8] = (q3 * wx + q0 * wy - q1 * wz) / 2.0f;
+	Xdot[9] = (-q2 * wx + q1 * wy + q0 * wz) / 2.0f;
+
+	// best guess is that bias stays constant
+	Xdot[10] = Xdot[11] = Xdot[12] = 0;
+
+	// For accels to make sure things stay stable, assume bias always walks weakly
+	// towards zero for the horizontal axis. This prevents drifting around an
+	// unobservable manifold of possible attitudes and gyro biases. The z-axis
+	// we assume no drift becaues this is teh one we want to estimate most accurately.
```
BrainFPV: s/teh/the

shared/uavobjectdefinition/insstate.xml
```diff
@@ -1,8 +1,8 @@
 <xml>
     <object name="INSState" singleinstance="true" settings="false">
         <description>Contains the INS state estimate</description>
-        <field name="State" units="" type="float" elements="13"/>
-        <field name="Var" units="" type="float" elements="13"/>
+        <field name="State" units="" type="float" elements="16"/>
+        <field name="Var" units="" type="float" elements="16"/>
```
BrainFPV: I assume it is 16 so it could also work wih INS16 (which is not used).

BrainFPV:
I read through the code but didn't verify the math. Do you have the matlab scripts used to generate the update steps? It would help for double checking things.

I will test the next time I go flying.

shared/uavobjectdefinition/insstate.xml
peabody124: correct

flight/Libraries/insgps14state.c
peabody124: That's pretty old code from the 13 state INS, just generic matrix multiplication unrolled.

peabody124:
@BrainFPV I tried to write the python code so it was a bit more understandable and "mathy", I can dig out some of the matlab code that generated some of the pseudocode too but even that is fairly cryptic and the output requires a fair bit of massaging

peabody124:
One issue I'm seeing with this is that BaroAltitude is showing up in the SystemStats.ObjectManagerQueueID so somewhere that queue overflows (when not receiving GPS data).

peabody124:
@elafargue thsi should be good to test if you want to

mlyle:
While I hope for more work in this area.. INS14 is clearly better, the patch reads sane, and we have good flight test on it.

PR #1508, opened 2015-03-09
Battery monitoring fixes and improvements

BrainFPV:
Improvements to the battery monitoring module:

- Unchecking the voltage/current monitoring checkbox now actually disables voltage/current measurements. Before, it was necessary to set the ADC pin to `NONE` for this. Which is unintuitive and can lead to the problem that current monitoring remains enabled even when there is no sensor attached.

- Makes the flight time thresholds configurable

-  Shows consumed capacity and remaining flight time in battery monitoring config

- Moves ADC pin selection to the voltage / current config groups in the GCS

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -91,7 +91,6 @@ int32_t BatteryInitialize(void)
 	return 0;
 }
 MODULE_INITCALL(BatteryInitialize, BatteryStart)
-#define HAS_SENSOR(x) batterySettings.SensorType[x]==FLIGHTBATTERYSETTINGS_SENSORTYPE_ENABLED
 
 static bool battery_settings_updated;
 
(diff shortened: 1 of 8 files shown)


BrainFPV:
Please review, should be ready.

peabody124:
I'm wondering if the "SensorType" field in ```FlightBatterySettings``` provides any use beyond setting the Current/Voltage pins to none or not. Why have this redundancy?

BrainFPV:
Yes, it's completely redundant and could be removed. At first I wasn't even avware that it is there and changed the GCS code to set the pin to NONE when the checkbox was unchecked. Maybe this would be the best solution, i.e., still have the checkboxes but unchecking them sets the pin to NONE, checking them sets the pin to the default value. Would definitely be cleaner. Thoughts?

BrainFPV:
Also, the `Type` and `NbCells` fields are unused and could be removed. I assume the idea was to use them to determine the voltage thresholds but it was never implemented. It would make more sense to use this information in the GCS and set the thresholds accordingly (e.g. using a wizard). There is no need for the FC to know this information.

peabody124:
Yeah I think dropping all that redundancy would be great. If you want to do it for this PR that would be preferable but if not this improvement can also stand alone.

BrainFPV:
I will do it, pruning things now makes sense. Stay tuned.

fujin:
Looking forward to this!

lolorc:
Hi there,

I have a related use case, I'm using a Frsky FLVSS voltage monitor. It's connected to the S.PORT bus with the FC and the X4R. In my case the battery module is disabled so the s.port module doesn't send the cells/voltage s.port fields.

It works as expected, I get both voltage and TL telem on my TX.

I was thinking of implementing the reading of the s.port bus from the s.port module so TL can make use of the sensors on the s.port bus.

would that be ok to have the battery monitoring module to use the voltage from s.port instead of an ADC ?
would that be ok to have the voltage/current sources individually selectable in the gui ? (ADC/SPORT) ?

It would need another PR, I'm just wondering if that's a valid use case.

peabody124:
Yeah you could add a new source instead of ADC# to get the battery / voltage signal from. Then you'd need to add support in the SPort code. The last thing would be how do you transport the data. I'd say the SPORT driver should expose a PIOS_SPORT_GetBattery() and PIOS_SPORT_SetBattery() methods.

BrainFPV:
@peabody124 I removed the unused fields from settings (`SensorType` and `Type`). `NbCells` is used by the `UAVOFrSKYSensorHubBridge` so I left it there. I also added simplified the battery monitoring code and added some logic to the GCS so that unchecking the voltage/current checkbox sets the voltage/current pins to `NONE` (and the other way around).

BrainFPV:
Please review

peabody124:
Tested and works well. Good stuff!

BrainFPV:
Thanks :)

PR #1510, opened 2015-03-16
Alarms: set alarm when gyro bias is high

### LINKED ISSUE #1503: Warn about excessive gyro bias
When leveling (or maybe even in the system alarms) if the gyro bias is pretty extreme (e.g. > 10 deg/s) there is probably an issue with the sensor. It would be good for GCS to warn about this and potentially for us to add a system alarm to indicate it (only warning, probably).

peabody124:
A gyro bias of more than 10 deg/s is indicative of a potentially 
bad sensor. Throw a warning up to alert the user. This is a 
problem if after leveling and power cycling, the board does 
not hit level again.

fixes #1503


BrainFPV:
LGTM

PR #1511, opened 2015-03-16
No disarm in autonomous mode

peabody124:
The default behavior of applying the disarm timeout in all modes
is nice and conservative. However, it is easy to leave the
throttle low while doing navigation and then have it disarm
while flying. This provides an option to avoid that.

Once this has been well validated, eventually we should drop this
option and never disarm while in autonomous modes unless the
user gives the disarm command.

This fixes the disarm timeout for low throttle which was dropped
with #1482

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -633,8 +633,20 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
 	{
 		set_armed_if_changed(FLIGHTSTATUS_ARMED_ARMED);
 
-		// Check for arming timeout if transmitter invalid
-		if (!valid || cmd->Throttle < 0) {
+		// Determine whether to disarm when throttle is low
+		uint8_t flight_mode;
+		FlightStatusFlightModeGet(&flight_mode);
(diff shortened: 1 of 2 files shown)


flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -633,8 +633,19 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
...
 
-		// Check for arming timeout if transmitter invalid
-		if (!valid) {
+		// Determine whether to disarm when throttle is low
+		uint8_t flight_mode;
+		FlightStatusFlightModeGet(&flight_mode);
+		bool autonomous_mode = flight_mode == FLIGHTSTATUS_FLIGHTMODE_VELOCITYCONTROL ||
+		                       flight_mode == FLIGHTSTATUS_FLIGHTMODE_POSITIONHOLD ||
+		                       flight_mode == FLIGHTSTATUS_FLIGHTMODE_RETURNTOHOME ||
+		                       flight_mode == FLIGHTSTATUS_FLIGHTMODE_PATHPLANNER;
```
BrainFPV: what about `TABLETCONTROL`?

peabody124: added

peabody124:
@elafargue want to merge this?

BrainFPV:
LGTM

PR #1514, opened 2015-03-24
Chibios colibri

peabody124:
Now that #1477 has merged, it would be good to update Colibri to use ChibiOS as
part of the upgrading of all the targets to ChibiOS. This requires more flash space
than is currently allocated for the board.

I opted to remove some components not used by most Colibri users rather than
increase the bootloader, but am open to change.

### DIFF
--- a/flight/targets/colibri/fw/board.h
+++ b/flight/targets/colibri/fw/board.h
@@ -1,0 +1,69 @@
+/*
+    ChibiOS/RT - Copyright (C) 2006-20s13 Giovanni Di Sirio
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


peabody124:
@tbs-trappy can you comment on this? See this link https://github.com/peabody124/TauLabs/commit/175781ee48354f6c8515b29cd44386f70eeee97a to see the set of things I disabled.

tbs-trappy:
Looks good to me. No issues with disabling these. How does this translate
to the V2?

elafargue:
Removed most of the Fixed Wing path followers that make no sense for Gemini, but it looks like it preserves GPS navigation if you want to. @peabody124 you confirm ?

peabody124:
Yeah GPS navigation should still be doable. PicoC is gone which is a nice feature but probably not used by Colibri users

flight/targets/colibri/fw/board.h
```diff
@@ -0,0 +1,69 @@
...
+
+/*
+ * Setup for STMicroelectronics STM32F4-Discovery board.
+ */
+
+/*
+ * Board identifier.
+ */
+#define BOARD_QUANTEC_QUANTON
+#define BOARD_NAME                  "quanton flight control"
```
BrainFPV: Should be Colibri?

elafargue: Good point!

elafargue:
Will fly-test ASAP & report

elafargue:
Flight-tested: could not see a real difference in behavior of the airframe.

PR #1516, opened 2015-03-25
pios_thread: chibios requires 8 byte boundary for stacks

peabody124:
This was causing the GPS module to crash because 850 stack size.
The new function always computes a stack size that is aligned
to a multiple of 8 and has a size that is a multiple of 8 at
least as large as requested.


peabody124:
fixes #1516

lolorc:
I've just tried that on one of my quanton, been able to flash fw with bl84. thanks

elafargue:
Tested on Sparky - Battery stack space is a bit worrying (4 bytes) ?

peabody124:
@elafargue check #1519 regarding battery issue

Issue #1517, opened 2015-03-25
"FlightTime modul critical" prevents arming for 60s~80s with voltage sensor only


gompf:
Sparky1, chibios
If booted wo battery monitoring module red status led goes of with the "third beep from blheli", comes again for about 2s and then everything is fine. 
With Battery monitoring and only voltage sensing configures red led stays on and "FlightTime module critical" prevents arming. If I just let it sit there for 60s-80s red led goes off/module warning dissapears and I can arm&takeoff. 
I havent applied the stack boundary patch yet.
Modules Battery Monitoring and Altitude Hold; oneshot on 1,2,4; mpu666@53%cpu; 5368 bytes heap remaining

joergrohde:
Sparky1 taulabs_chibios_20150321_170854_290690f58f
With Battery Voltage only arming is possible after 70s
Modules Bat, Hott, OneShot125@1-4, mpu@666

elafargue:
Can you report stack remaining on the battery module? I noticed it can be super low (4 bytes only in my case), might be related?

gompf:
Within 3 boots I have seen 56,40 and 8byte (TaskInfo/MetaData/Stack remaining/Battery). It doesn't change when it's ready to arm. Arming with 8byte left was OK, after 70s. 
I found chibios branch now (didn't noticed the switch to next) and will test again with current version later.

joergrohde:
Stacksize 40, 16, 48Byte. Current pin is set to none, Current measurement is off. But live reading -1A and a Capacity left 146xxxxxxxxxxxxxxxxx mAH.

gompf:
Forgot to mention it, same config. I also assume a rollover or something in cap. calculation when current measurement isn't configured.

elafargue:
@joergrohde @gompf does the latest fix by peabody124 resolve your problem ?

gompf:
Yes, everything fine here. Tested in modul combination BatteryMonitoring/AltitudeHold/FrSkyHub. Thanks!

peabody124:
thanks

joergrohde:
Yes, Sorry only desk tested. Modul Batt, Hott on Quad & Batt, Hott, Autotune on Hexa. The weather here where I´m living in Germany is bad :(

PR #1519, opened 2015-03-26
Battery: make sure to load battery value

peabody124:
The module was using an uninitialized variable which
was causing the energy consumed to start at a random
value and the flight time to go to error state.

We could afford to only load this variable once before
the loop, but by doing it each loop it allows GCS to
push changes if desired. This is our more common
practice of making changes in place.

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -113,6 +113,8 @@ static void batteryTask(void * parameters)
 		FlightBatterySettingsData batterySettings;
 		float energyRemaining;
 
+		FlightBatteryStateGet(&flightBatteryData);
+
 		if (battery_settings_updated) {
 			battery_settings_updated = false;
 			FlightBatterySettingsGet(&batterySettings);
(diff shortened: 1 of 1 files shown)


peabody124:
@BrainFPV can you look at this since you most recently touched this module. The regression was not caused by #1508, however. It just seems like we were lucky that FreeRTOS was putting the stack space previously on all zero memory.

peabody124:
Fixes #1517

peabody124:
Note the commit history from second commit:

    Battery: increase stack size
    The previous commit fixed an issue triggering an inappropriate
    flight battery time warning. However, it also revealed the stack
    usage going extremely low for the Battery module in that case
    (6 bytes). That just adds insult to injury if the system is
    hitting limits, so this raises the stack size to be safe.

elafargue:
Getting 68 bytes free now on the Battery task stack, much better.

peabody124:
@elafargue Want to merge this then? Check your flight time is also valid.

BrainFPV:
LGTM, excuse the breakage..

peabody124:
@BrainFPV you didn't break it. the code had this bug and the value was simply uninitialized and was lucky to work with FreeRTOS

elafargue:
Tested and merging

PR #1520, opened 2015-03-26
AltitudeHold: make sure to get latest command when engaging

peabody124:
When switching to ChibiOS the order events were received changed
with regard to FlightStatus going to AH mode and then the update
for AltitudeHoldDesired. This meant the first pass was using all
zero for the AltitudeHoldDesired (or previous value) and was
causing the integral to hit the anti-windup saturation and go to
zero. The next update was correct but since the integral was
erased the aircraft would have a transient drop.

This can also be fixed by calling a continue in the FlightStatusHandle
processing block so the next loop catches the update (confirming
there is no event being dropped). However, this method is slightly
lower latency and it is reasonable to force a get of the AHD when
engaging it, anyway.


elafargue:
Merging now since this was leading to serious issues on AH. More feedback on flight tests appreciated.

PR #1523, opened 2015-03-28
Autotune: increase stack size usage to 1900 bytes

peabody124:


elafargue:
Autotune is stack-hungry...

PR #1527, opened 2015-03-29
Calibration: six issue that caused timeout error with RFM22b

peabody124:
There was no real error but because it takes a while to reset
the metadata when using RFM22b it was thinking it timed out and
also the progress bar kept moving when finished.


elafargue:
OK, patch is correct: the timeout timer is not there to catch metadata write timeouts (there is another timer for that). Merging.

PR #1532, opened 2015-03-31
Improve compile times by using ccache if available

mlyle:
These two changes should be safe (they auto-detect whether ccache is installed on the system) and are worth a build time speedup of 9-10x on my hardware.  They may not be as big a benefit for everyone (I'm on a little VM) but are likely to be worth more than 4x.  -mpl


peabody124:
jenkins: test this please
jenkins: add to whitelist

PR #1539, opened 2015-04-02
Conditionalize SIGFPE handling in posix simulator.

mlyle:
Add functionality to use getopt appropriately inside the posix simulator.
Floating point exceptions are now disabled by default.  Passing -f to the
simulator will give you exceptions.

This was necessary because navigation and IMU code throws a lot of SIGFPE's
and won't make it through initialization in the posix simulator (this is a
separate issue I hope to hunt).

Also plumb args into the OSX simulator even though they're presently unused.
Need a Mac user to confirm this still compiles.


peabody124:
Tested and it doesn't break OSX

PR #1542, opened 2015-04-02
Build: add FLIGHT_BUILD_CONF environment variable

mlyle:
It's a knob, with the values 'default' (which is the default), 'release',
and 'debug'.  In debug, DEBUG is set to YES in the flight firmwares.  In
debug and default, compilation caching is allowed (and ccache is smart
enough to cache default/debug separately because it includes command line
arguments in its hash).  In release, there is no compilation caching (and
jenkins/release build scripts should be updated to use this mode).

This parallels the GCS_BUILD_CONF environment variable which also disables
compilation caching for GCS build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,9 @@ include $(ROOT_DIR)/flight/targets/*/target-defs.mk
 # OpenPilot GCS build configuration (debug | release)
 GCS_BUILD_CONF ?= debug
 
+# And the flight build configuration (debug | default | release)
+export FLIGHT_BUILD_CONF ?= default
+
 ##############################
 #
 # Check that environmental variables are sane
@@ -98,6 +101,16 @@ ifdef GCS_BUILD_CONF
  ifneq ($(GCS_BUILD_CONF), release)
   ifneq ($(GCS_BUILD_CONF), debug)
    $(error Only debug or release are allowed for GCS_BUILD_CONF)
+  endif
+ endif
+endif
+
+ifdef FLIGHT_BUILD_CONF
+ ifneq ($(FLIGHT_BUILD_CONF), release)
+  ifneq ($(FLIGHT_BUILD_CONF), debug)
+   ifneq ($(FLIGHT_BUILD_CONF), default)
+    $(error Only debug or release are allowed for FLIGHT_BUILD_CONF)
+   endif
   endif
  endif
 endif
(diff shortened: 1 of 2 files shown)


mlyle:
This is per peabody124's irc comments (from stac).

Makefile
```diff
@@ -79,6 +79,20 @@ include $(ROOT_DIR)/flight/targets/*/target-defs.mk
...
 
+# And the flight build configuration (debug | default | release)
+export FLIGHT_BUILD_CONF ?= default
+
+ifeq ($(FLIGHT_BUILD_CONF), debug)
+export DEBUG:=YES
+else ifeq ($(FLIGHT_BUILD_CONF), release)
+export DEBUG:=NO
+else ifeq ($(FLIGHT_BUILD_CONF), default)
+export DEBUG:=NO
```
solidgoldbomb: This line will have the constant side effect of exporting ````DEBUG:=NO```` even for non-flight builds.  That might be very confusing if that variable ever becomes important for gcs, or androidgcs or uavo builds.

If the user happens to have ````FLIGHT_BUILD_CONF = debug```` in their environment, this code will constantly assert ````DEBUG:=YES```` even for non-flight builds which may be even more confusing.

It might make more sense to:
* move the setting of DEBUG out of this makefile and into the flight makefiles
* keep the validation of FLIGHT_BUILD_CONF values in this file but move it down under the "Check that environmental variables are sane" heading with the other validation code.

mlyle:
I agree this is broken based on your comments.  I will look if there's an elegant way to do this without making the change in many places.

mlyle:
OK, hopefully this is better?

mlyle:
Jenkins build failed because: 04:32:12 virtual memory exhausted: Cannot allocate memory

solidgoldbomb:
jenkins: test this please

solidgoldbomb:
@mlyle This is much better.  Thanks for making the changes.  I'm not thrilled about the DEBUG enable being in firmware-defs.mk since it mixes policy with the basic tool setup in firmware-defs but I can see why you would prefer this to replicating in each target makefile.

This version is OK to merge as soon as Jenkins gives it a clean bill of health.

I've poked @lilvinz to see if he can check on the memory exhaust error in Jenkins.

lilvinz:
jenkins: test this please

solidgoldbomb:
Thanks for fixing Jenkins @lilvinz.

PR #1543, opened 2015-04-04
VtolPathFollower: deadband, lookahead, and refactor

mlyle:
This adds lookahead to the path planner which should allow it to follow paths more
closely. In addition, it adds a deadband region - separately parameterized for path
legs and endpoint following. This should provide smoother control behavior along
paths and around endpoints.

It also refactors out some duplicated code.

### DIFF
--- a/flight/Modules/PathPlanner/pathplanner.c
+++ b/flight/Modules/PathPlanner/pathplanner.c
@@ -207,8 +207,11 @@ static void waypointsUpdated(UAVObjEvent * ev)
 		return;
 
 	WaypointActiveGet(&waypointActive);
-	if(active_waypoint != waypointActive.Index)
+	if(active_waypoint != waypointActive.Index) {
+		active_waypoint = waypointActive.Index;
+
 		activateWaypoint(waypointActive.Index);
+	}
 }
 
 /**
@@ -216,14 +219,14 @@ static void waypointsUpdated(UAVObjEvent * ev)
  */
 static void pathStatusUpdated(UAVObjEvent * ev)
 {
-	static PathStatusStatusOptions last_path_status;
-	PathStatusStatusOptions this_path_status;
-	PathStatusStatusGet(&this_path_status);
-	if(last_path_status == PATHSTATUS_STATUS_INPROGRESS &&
-	   this_path_status == PATHSTATUS_STATUS_COMPLETED) {
+	PathStatusData pathStatus;
+
+	PathStatusGet(&pathStatus);
+
+	if ((pathStatus.Status == PATHSTATUS_STATUS_COMPLETED) &&
+			(pathStatus.Waypoint == active_waypoint)) {
 		path_completed = true;
 	}
-	last_path_status = this_path_status;
 }
 
 /**
@@ -247,6 +250,7 @@ static void holdCurrentPosition()
 	pathDesired.StartingVelocity = 5; // This will be the max velocity it uses to try and hold
 	pathDesired.EndingVelocity = 5;
 	pathDesired.ModeParameters = 0;
+	pathDesired.Waypoint = -1;
 	PathDesiredSet(&pathDesired);
 }
 
@@ -293,7 +297,7 @@ static void advanceWaypoint()
 	// conditional logic desired here.
 	// Note: In the case of conditional logic it is the responsibilty of the implementer
 	// to ensure all possible paths are valid.
-	waypointActive.Index++;
+	waypointActive.Index = previous_waypoint+1;
 
 	if (waypointActive.Index >= UAVObjGetNumInstances(WaypointHandle())) {
 		holdLastPosition();	// This means last in path.
@@ -314,9 +318,7 @@ static void advanceWaypoint()
  */
 static void activateWaypoint(int idx)
 {
-	active_waypoint = idx;
-
-	if (idx >= UAVObjGetNumInstances(WaypointHandle())) {
+	if ((idx >= UAVObjGetNumInstances(WaypointHandle())) || (idx < 0)) {
 		// Attempting to access invalid waypoint.  Fall back to position hold at current location
 		AlarmsSet(SYSTEMALARMS_ALARM_PATHPLANNER, SYSTEMALARMS_ALARM_ERROR);
 		holdCurrentPosition();
@@ -327,6 +329,8 @@ static void activateWaypoint(int idx)
 	WaypointInstGet(idx, &waypoint);
 
 	PathDesiredData pathDesired;
+
+	pathDesired.Waypoint = idx;
 
 	pathDesired.End[PATHDESIRED_END_NORTH] = waypoint.Position[WAYPOINT_POSITION_NORTH];
 	pathDesired.End[PATHDESIRED_END_EAST] = waypoint.Position[WAYPOINT_POSITION_EAST];
@@ -350,7 +354,7 @@ static void activateWaypoint(int idx)
 			break;
 		case WAYPOINT_MODE_LAND:
 			pathDesired.Mode = PATHDESIRED_MODE_LAND;
-			break;
+		break;
 		default:
 			holdCurrentPosition();
 			AlarmsSet(SYSTEMALARMS_ALARM_PATHPLANNER, SYSTEMALARMS_ALARM_ERROR);
--- a/flight/Modules/PathPlanner/pathplanner.c
+++ b/flight/Modules/PathPlanner/pathplanner.c
@@ -264,7 +264,7 @@ static void holdLastPosition()
 
 	// Get the activated waypoint
 	WaypointData waypoint;
-	WaypointInstGet(idx, &waypoint);
+	WaypointInstGet(idx-1, &waypoint);
 
 	PositionActualData position;
 	PositionActualGet(&position);
--- a/flight/Modules/PathPlanner/pathplanner.c
+++ b/flight/Modules/PathPlanner/pathplanner.c
@@ -280,6 +280,7 @@ static void holdLastPosition()
 	pathDesired.StartingVelocity = 5; // This will be the max velocity it uses to try and hold
 	pathDesired.EndingVelocity = 5;
 	pathDesired.ModeParameters = 0;
+	pathDesired.Waypoint = -1;
 	PathDesiredSet(&pathDesired);
 }
 /**
--- a/flight/Modules/PathPlanner/pathplanner.c
+++ b/flight/Modules/PathPlanner/pathplanner.c
@@ -354,7 +354,7 @@ static void activateWaypoint(int idx)
 			break;
 		case WAYPOINT_MODE_LAND:
 			pathDesired.Mode = PATHDESIRED_MODE_LAND;
-		break;
+			break;
 		default:
 			holdCurrentPosition();
 			AlarmsSet(SYSTEMALARMS_ALARM_PATHPLANNER, SYSTEMALARMS_ALARM_ERROR);
--- a/flight/Modules/PathPlanner/pathplanner.c
+++ b/flight/Modules/PathPlanner/pathplanner.c
@@ -55,11 +55,10 @@ static struct pios_queue *queue;
 static PathPlannerSettingsData pathPlannerSettings;
 static WaypointActiveData waypointActive;
 static WaypointData waypoint;
-static bool path_status_updated;
+static bool path_completed;
 
 // Private functions
 static void advanceWaypoint();
-static void checkTerminationCondition();
 static void activateWaypoint(int idx);
 
 static void pathPlannerTask(void *parameters);
@@ -151,7 +150,9 @@ static void pathPlannerTask(void *parameters)
 
 	// Main thread loop
 	bool pathplanner_active = false;
-	path_status_updated = false;
+
+	pathStatusUpdated(NULL);
+	path_completed = false;
 
 	while (1)
 	{
@@ -183,8 +184,8 @@ static void pathPlannerTask(void *parameters)
 
 		/* This method determines if we have achieved the goal of the active */
 		/* waypoint */
-		if (path_status_updated)
-			checkTerminationCondition();
+		if (path_completed)
+			advanceWaypoint();
 
 		/* If advance waypoint takes a long time to calculate then it should */
 		/* be called from here when the active_waypoints does not equal the  */
@@ -215,21 +216,14 @@ static void waypointsUpdated(UAVObjEvent * ev)
  */
 static void pathStatusUpdated(UAVObjEvent * ev)
 {
-	path_status_updated = true;
-}
-
-/**
- * This method checks the current position against the active waypoint
- * to determine if it has been reached
- */
-static void checkTerminationCondition()
-{
-	PathStatusData pathStatus;
-	PathStatusGet(&pathStatus);
-	path_status_updated = false;
-
-	if (pathStatus.Status == PATHSTATUS_STATUS_COMPLETED)
-		advanceWaypoint();
+	static PathStatusStatusOptions last_path_status;
+	PathStatusStatusOptions this_path_status;
+	PathStatusStatusGet(&this_path_status);
+	if(last_path_status == PATHSTATUS_STATUS_INPROGRESS &&
+	   this_path_status == PATHSTATUS_STATUS_COMPLETED) {
+		path_completed = true;
+	}
+	last_path_status = this_path_status;
 }
 
 /**
@@ -302,13 +296,12 @@ static void advanceWaypoint()
 	waypointActive.Index++;
 
 	if (waypointActive.Index >= UAVObjGetNumInstances(WaypointHandle())) {
-		holdLastPosition();
+		holdLastPosition();	// This means last in path.
 	} else {
 		WaypointActiveSet(&waypointActive);
 	}
 
-	// Invalidate any pending path status updates
-	path_status_updated = false;
+	path_completed = false;
 }
 
 /**
@@ -389,7 +382,7 @@ static void activateWaypoint(int idx)
 	PathDesiredSet(&pathDesired);
 
 	// Invalidate any pending path status updates
-	path_status_updated = false;
+	path_completed = false;
 
 	AlarmsClear(SYSTEMALARMS_ALARM_PATHPLANNER);
 }
--- a/flight/Modules/VtolPathFollower/vtol_follower_control.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_control.c
@@ -225,19 +225,23 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
 	pathStatus.error = progress->error;
 	pathStatus.Waypoint = pathDesired->Waypoint;
 
-	// Determine if we hit criterion to finish this leg
+	// Figure out how low (high) we should be and the error
 	const float altitudeSetpoint = vtol_interpolate(progress->fractional_progress,
 	    pathDesired->Start[2], pathDesired->End[2]);
 	const float downError = altitudeSetpoint - positionActual.Down;
-	// TODO: make optional and make separate parametric error
-	const bool criterion_altitude = downError > -guidanceSettings.EndpointRadius;
 
 	// If leg is completed signal this
 	if (current_leg_completed || pathStatus.fractional_progress > 1.0f) {
+		const bool criterion_altitude =
+			downError > -guidanceSettings.WaypointAltitudeTol;
 
 		// Once we complete leg and hit altitude criterion signal this
-		// waypoint is done
-		if (criterion_altitude) {
+		// waypoint is done.  Or if we're not controlling throttle,
+		// ignore height for completion.
+
+		// Waypoint heights are thus treated as crossing restrictions-
+		// cross this point at or above...
+		if (criterion_altitude || (!guidanceSettings.ThrottleControl)) {
 			pathStatus.Status = PATHSTATUS_STATUS_COMPLETED;
 			PathStatusSet(&pathStatus);
 		} else {
--- a/flight/Modules/VtolPathFollower/vtol_follower_control.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_control.c
@@ -58,6 +58,44 @@ static VtolPathFollowerSettingsData guidanceSettings;
 static AltitudeHoldSettingsData altitudeHoldSettings;
 struct pid vtol_pids[VTOL_PID_NUM];
 
+static inline float vtol_interpolate(const float fraction, const float beginVal,
+	const float endVal) {
+	return beginVal + (endVal - beginVal) * bound_min_max(fraction, 0, 1);
+}
+
+static inline float vtol_magnitude(const float *v, int n) {
+	float sum=0;
+
+	for (int i=0; i<n; i++) {
+		sum += powf(v[i], 2);
+	}
+
+	return sqrtf(sum);
+}
+
+static inline void vtol_calculate_distances(const float *actual,
+	const float *desired, float *out, int normalize) {
+	out[0] = desired[0] - actual[0];
+	out[1] = desired[1] - actual[1];
+	out[2] = desired[2] - actual[2];
+
+	if (normalize) {
+		float mag=vtol_magnitude(out, 3);
+
+		out[0] /= mag;  out[1] /= mag;  out[2] /= mag;
+	}
+}
+
+static inline void vtol_limit_velocity(float *vels, float limit) {
+	float mag=vtol_magnitude(vels, 2);	// only horiz component
+	float scale = mag / limit;
+
+	if (scale > 1) {
+		vels[0] /= mag;
+		vels[1] /= mag;
+	}
+}
+
 /**
  * Compute desired velocity to follow the desired path from the current location.
  * @param[in] dT the time since last evaluation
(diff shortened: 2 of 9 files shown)


flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -58,6 +58,69 @@ static VtolPathFollowerSettingsData guidanceSettings;
...
+
+	for (int i=0; i<n; i++) {
+		sum += powf(v[i], 2);
+	}
+
+	return sqrtf(sum);
+}
+
+static inline void vtol_calculate_distances(const float *actual,
+	const float *desired, float *out, int normalize) {
```
peabody124: you can use bool instead of int here. we include "stdbool.h" in our code.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -87,107 +157,156 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
-	float downError;
-	float northCommand;
-	float eastCommand;
 	
-	const float cur_pos_ned[3] = { positionActual.North, positionActual.East, positionActual.Down };
+	VelocityActualData velocityActual;
+
+	VelocityActualGet(&velocityActual);
+
+	// XXX parameterize
```
peabody124: yup definitely parameterize. probably with default of 0 until we get practical testing experience.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -58,6 +58,69 @@ static VtolPathFollowerSettingsData guidanceSettings;
...
+
+/* "Real" deadbands are evil.  Control systems end up fighting the edge.
+ * You don't teach your integrator about emerging drift.  Discontinuities
+ * in your control inputs cause all kinds of neat stuff. */
+static inline float vtol_deadband(float in, float w, float b) {
+	/* So basically.. we want the function to be tangent to the
+	** linear sections-- have a slope of 1-- at -w and w.  In the
+	** middle we want a slope of b.   So the cube here does all the
+	** work b isn't doing. */
+	float m = cbrtf((1-b)/(3*powf(w,2)));
```
peabody124: following discussion on IRC b and w should probably come from some "effective deadband" setting in VtolPathFollowerSettings that is specified in meters and then can be precomputed/cached when settings change.

elafargue:
Won't have time to review until next week unfortunately....

mlyle:
Removed the review-only tag-- this is getting much closer to something I'd hope could be committed.

I'd really appreciate test feedback-- with VtolPathFollowerSettings.PositionFeedForward = 0.7

mlyle:
Tested.  Did a few PH flights of ~10 minutes each with different values of position feedforward parameter.

Values between 0.4 and 0.8 show obviously better PH performance than 0.0 on my 500-sizeish custom airframe, 4s, 950kv swinging 8 inch props at 1.2 kilo flying weight.  (I did not test higher than 0.8... so this isn't really "tuned",...  but now I suggest a value of 0.6 since it is in the middle of values I have proven "good").  At 0.6, I saw it stay in a 1 meter radius area for a couple minutes in a 5-12 mph variable wind.  (Over longer periods, the position wanders, but this is almost certainly GPS error).

But I have bad news, too.  I kept throttle control as I do not have the airframe foamed for AH yet.  I also was using this to excite the system-- dropping throttle to get attitude warble and throttling up.  One of the times I did this at posFF=0.0 the airframe "took off" at about 10mph.  I was only able to let it go in this direction for a second before I switched to attitude mode and recovered, so I don't know if this would have healed on its own.  I have seen similar things in the simulator when the system is deliberately excited with and without my codebase.

mlyle:
Ready for review.  I will re-read and flight test, too.

flight/Modules/PathPlanner/pathplanner.c
```diff
@@ -253,6 +250,7 @@ static void holdCurrentPosition()
 	pathDesired.StartingVelocity = 5; // This will be the max velocity it uses to try and hold
 	pathDesired.EndingVelocity = 5;
 	pathDesired.ModeParameters = 0;
+	pathDesired.Waypoint = -1;
```
peabody124: just to confirm my understanding, this is a sanity check and should not be relevant in normal usage.

regardless it should be duplicated in holdLastPosition

flight/Modules/PathPlanner/pathplanner.c
```diff
@@ -299,16 +297,15 @@ static void advanceWaypoint()
 	// conditional logic desired here.
 	// Note: In the case of conditional logic it is the responsibilty of the implementer
 	// to ensure all possible paths are valid.
-	waypointActive.Index++;
+	waypointActive.Index = previous_waypoint+1;
```
peabody124: why this change? no functional implication right?

flight/Modules/PathPlanner/pathplanner.c
```diff
@@ -357,7 +354,7 @@ static void activateWaypoint(int idx)
 			break;
 		case WAYPOINT_MODE_LAND:
 			pathDesired.Mode = PATHDESIRED_MODE_LAND;
-			break;
```
peabody124: just making changes to break indentation? :-p

flight/Modules/PathPlanner/pathplanner.c
mlyle: It could be important to avoid skipping waypoint 1 if you were previously on waypoint 1.  (Eliminates a minor race).  I agree with holdLastPosition-- willfix.

flight/Modules/PathPlanner/pathplanner.c
mlyle: No real functional implication.

flight/Modules/PathPlanner/pathplanner.c
mlyle: Whoops.  Maybe with all the merges I screwed this up.  Willfix.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -58,6 +58,131 @@ static VtolPathFollowerSettingsData guidanceSettings;
...
+/**
+ * Calculate pythagorean magnitude of a vector.
+ * @param[in] v pointer to a float array
+ * @param[in] n length of the amount to take the magnitude of
+ * @returns Root of sum of squares of the vector
+ *
+ * Note that sometimes we only take the magnitude of the first 2 elements
+ * of a 3-vector to get the horizontal component.
+ */
+static float vtol_magnitude(const float *v, int n) {
```
peabody124: this should probably be moved to a math library and just called vector magnitude. e.g. https://github.com/TauLabs/TauLabs/blob/next/flight/Libraries/math/coordinate_conversions.h#L70

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -58,6 +58,131 @@ static VtolPathFollowerSettingsData guidanceSettings;
...
+
+/**
+ * Clip a velocity 2-vector  while maintaining vector direction.
+ * @param[in,out] vels velocity to clip
+ * @param[in] limit desired limit magnitude.
+ *
+ * if mag(vels) > limit, vels=vels / mag(vels) * limit
+ */
+static void vtol_limit_velocity(float *vels, float limit) {
+	float mag=vtol_magnitude(vels, 2);	// only horiz component
```
peabody124: spaces around =

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -58,6 +58,131 @@ static VtolPathFollowerSettingsData guidanceSettings;
...
+
+/**
+ * Subtract two 3-vectors, and optionally normalize to return an error value.
+ * @param[in] actual the measured process value
+ * @param[in] desired the setpoint of the system
+ * @param[out] the resultant error vector desired-actual
+ * @param[in] normalize True if it's desired to normalize the output vector
+ */
+static void vtol_calculate_distances(const float *actual,
+	const float *desired, float *out, bool normalize) {
```
peabody124: On the flight code we try and stick to linux kernel convention. In this case the ```{```
should be on a new line for function definitions (but only for those)

https://www.kernel.org/doc/Documentation/CodingStyle

This is repeated throughout this function.

mlyle:
Marked not for merge pending more dev testing and dev review.

peabody124:
I made a PR against this to make one of the functions more readable. Other comments to come

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -58,6 +58,131 @@ static VtolPathFollowerSettingsData guidanceSettings;
...
+}
+
+/**
+ * Subtract two 3-vectors, and optionally normalize to return an error value.
+ * @param[in] actual the measured process value
+ * @param[in] desired the setpoint of the system
+ * @param[out] the resultant error vector desired-actual
+ * @param[in] normalize True if it's desired to normalize the output vector
+ */
+static void vtol_calculate_distances(const float *actual,
```
peabody124: this is a bit poorly named. it is a vector difference, not a difference (scalar)

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
 {
 	PositionActualData positionActual;
 	PositionActualGet(&positionActual);
+
+	VelocityActualData velocityActual;
+	VelocityActualGet(&velocityActual);
+
+	const float cur_pos_ned[3] = {
+		positionActual.North +
+		    velocityActual.North*guidanceSettings.PositionFeedforward,
```
peabody124: use spaces between operators

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+		positionActual.East +
+		    velocityActual.East*guidanceSettings.PositionFeedforward,
```
peabody124: ditto

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
-	const float cur[3] = { positionActual.North, positionActual.East, positionActual.Down };
-	
-	path_progress(pathDesired, cur, progress);
-	
+	// Interpolate desired velocity and altitude along the path
+	float groundspeed = vtol_interpolate(progress->fractional_progress,
+	    pathDesired->StartingVelocity, pathDesired->EndingVelocity);
+
+	if (progress->fractional_progress >= 1)
+		groundspeed = 0;
```
peabody124: I don’t like using this setting groundspeed to 0 as a hook to trigger ln 246, especially given this is essentially a known bug to do this. can you just use ```progress->fractional_progress >= 1)``` there? better yet since we have logic checks on whether fraction_progress is > / < 1 on line 217, 234, 242, 246 (implicit), could you just consolidate them into two if blocks? there is no need to compute these things when we simply ```return vtol_follower_control_simple(dT, pathDesired->End, 0, 1);```
see 16e2a8455e4b466cee7669e768ac9416d92c757ag in the PR against this branch on your repo

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
 	PathStatusGet(&pathStatus);
 	pathStatus.fractional_progress = progress->fractional_progress;
 	pathStatus.error = progress->error;
-	if (pathStatus.fractional_progress < 1)
+	pathStatus.Waypoint = pathDesired->Waypoint;
+
+	// Only if we are at the end, and above the end altitude, are we
+	// done.
+	if ((pathStatus.fractional_progress < 1) ||
+		(downError < -guidanceSettings.EndpointRadius)) {
```
peabody124: this needs to be both an option (whether to require a minimum altitude for path following) and a programmable threshold. possibly one thing. It also should be disabled when the user has throttle control set to false, or most users will simply never complete their path. also make sure to do it after merging my commit.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
-	if (progress->fractional_progress > 1)
-		groundspeed = 0;
-	
-	VelocityDesiredData velocityDesired;
-	velocityDesired.North = progress->path_direction[0] * groundspeed;
-	velocityDesired.East = progress->path_direction[1] * groundspeed;
-	
-	float error_speed = progress->error * 
+	if (groundspeed == 0) {
+		return vtol_follower_control_simple(dT, pathDesired->End, 0, 1);
```
peabody124: this is double settings the status is it not. also use boolean true/false instead of 1/0

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
-	float altitudeSetpoint = pathDesired->Start[2] + (pathDesired->End[2] - pathDesired->Start[2]) *
-	    bound_min_max(progress->fractional_progress, 0, 1);
+	commands_ned[1] = progress->path_direction[1] * groundspeed +
+	    progress->correction_direction[1] * error_speed;
 
-	float downError = altitudeSetpoint - positionActual.Down;
-	velocityDesired.Down = pid_apply_antiwindup(&vtol_pids[DOWN_POSITION], downError,
-		-guidanceSettings.VerticalVelMax, guidanceSettings.VerticalVelMax, dT);
+	/* Limit the total velocity based on the configured value. */
+	vtol_limit_velocity(commands_ned, guidanceSettings.HorizontalVelMax);
```
peabody124: good catch. apply the limit to the requested velocity seems more correct than applying it to the error component only.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
- * This method does not attempt any particular path, simply a straight line
- * approach. The calculated velocity to attempt is stored in @ref 
- * VelocityDesired. 
- *
- * Takes in @ref PositionActual and compares it to @ref PositionDesired 
- * and computes @ref VelocityDesired
- */
-int32_t vtol_follower_control_endpoint(const float dT, const float *hold_pos_ned)
-{
+static int32_t vtol_follower_control_simple(const float dT,
```
peabody124: you deleted all the doxygens! :( —also _simple isn’t a good name since it is more complicated than endpoint. possibly endpoint_core… i dunno

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
-	float northCommand;
-	float eastCommand;
 	
-	const float cur_pos_ned[3] = { positionActual.North, positionActual.East, positionActual.Down };
+	VelocityActualData velocityActual;
+
+	VelocityActualGet(&velocityActual);
+
+	/* Where would we be in ___ second at current rates? */	
+	const float cur_pos_ned[3] = {
```
peabody124: maybe consolidate into a vtol_path_lookahead with line 204. +doxygen :D

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+		positionActual.North +
+		    velocityActual.North*guidanceSettings.PositionFeedforward,
+		positionActual.East +
+		    velocityActual.East*guidanceSettings.PositionFeedforward,
+		positionActual.Down };
+
+	float errors_ned[3];
+
```
peabody124: too many empty lines for my aesthetic. now i’m just nitpicking

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+	vtol_calculate_distances(cur_pos_ned, hold_pos_ned, errors_ned, 0);
```
peabody124: for bools use true or false, just makes it a tad more readable

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+
+	float horiz_error_mag = vtol_magnitude(errors_ned, 2);
+	float scale_horiz_error_mag;
```
peabody124: just init to 0 and remove the else. less lines.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+
+	if (horiz_error_mag > 0.00001f) {
+		scale_horiz_error_mag = vtol_deadband(horiz_error_mag,
```
peabody124: can you explain what this math is doing (comments?). I think you can make this much cleaner if you make ```vtol_calculate_distances``` return the norm when normalizing (which it computed anyway). then you can do this

float horiz_error_mag = vtol_calculate_distances(cur_pos_ned, hold_pos_ned, errors_ned, false);
horiz_error_mag = vtol_deadband(horiz_error_mag, guidanceSettings.EndpointDeadbandWidth guidanceSettings.EndpointDeadbandCenterGain, vtol_end_m, vtol_end_r);
float damped_ne[2] = { errors_ned[0] * horiz_error_mag, errors_ned[1] * horiz_error_mag };

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
-	float distance2 = powf(northError, 2) + powf(eastError, 2);
-	if (distance2 < (guidanceSettings.EndpointRadius * guidanceSettings.EndpointRadius)) {
-		path_status = PATHSTATUS_STATUS_COMPLETED;
+	if (update_status) {
+		uint8_t path_status = PATHSTATUS_STATUS_INPROGRESS;
+
+		if (!landing) {
+			// Indicate whether we are in radius of this endpoint
+			// And at/above the altitude requested
+			if ((vtol_magnitude(errors_ned, 2) < guidanceSettings.EndpointRadius) &&
```
peabody124: again this needs to be both optional and settable. It also should be disabled when the user has throttle control set to false, or most users will simply never complete their path. that means it probably should be moved to a method like “vtol_control_altitude_criterion” since duplicated with line 243

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +197,184 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+ * This method does not attempt any particular path, simply a straight line
+ * approach. The calculated velocity to attempt is stored in @ref
+ * VelocityDesired.
+ *
+ * Takes in @ref PositionActual and compares it to @ref PositionDesired
+ * and computes @ref VelocityDesired
+ */
+int32_t vtol_follower_control_endpoint(const float dT, const float *hold_pos_ned)
+{
+	return vtol_follower_control_simple(dT, hold_pos_ned, 0, 1);
```
peabody124: true/false for 1/0

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -201,48 +387,7 @@ int32_t vtol_follower_control_endpoint(const float dT, const float *hold_pos_ned
 int32_t vtol_follower_control_land(const float dT, const float *hold_pos_ned,
```
peabody124: for now this is good restructuring. in the long run this will 99% be reverted for this method since landing really needs it’s own metric of “completed” that will allow us to spool down motors, etc.

peabody124:
vertical threshold settings. This definitely should (by default) not apply when throttle control is false. @elfargue do you think the default behavior when flying a path should wait for you to hit the target altitude or focus on the lat/lon behavior (this will be an option). I’m inclined to think most users will look at a map and expect it to progress once it reaches locations and worry less about altitude (since multi rotors should be doing this reasonably anyway).

peabody124:
@elafargue see comment above

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: No, I didn't delete all the doxygens.  Check again.

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: 99 probably overstates it :) either way this is good for now!

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: sorry i meant all in the sense of this block. ```vtol_follower_control_simple``` isn't documented. bad phrasing on my part

mlyle:
If the multirotor is doing a job capturing altitude, it's going to progress / not surprise people by waiting for a climb that it should have been doing smoothly over the leg.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Yes-- I saw the completed flag.  I still think it would make sense to combine this with endpoint following.

peabody124: You might be right. One of those let's cross it when we need to things

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Do you want vtol_calculate_vector_difference instead of calculate_distances ?  I don't see the word difference, I don't know what you mean.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: I have the PR.  The falling back to endpoint following is basically the same--- or perhaps more aggressive, than the known groundspeed = 0 bug (because it's going to not just try and stop, but pull back).

I like the way you latch STATUS_COMPLETED; it's effectively level triggered.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: I agree-- i thought all of these should go to a math lib and there are a whole lot of places i could replace code to use them elsewhere.  But maybe another pull?  Surely it's better here having once at the top than duplicated through the code below like before...

peabody124: Totally fine for another pass for math reorganization.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -72,123 +202,217 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
...
+		    velocityActual.East * guidanceSettings.PositionFeedforward,
+		positionActual.Down };
+
+	float errors_ned[3];
+
+	/* Calculate the difference between where we want to be and the
+	 * above position */
+	vtol_calculate_distances(cur_pos_ned, hold_pos_ned, errors_ned, 0);
+
+	float horiz_error_mag = vtol_magnitude(errors_ned, 2);
```
peabody124: You didn't like my nicer elegant way of writing this math I put in the previous comment?
https://github.com/TauLabs/TauLabs/pull/1543#discussion_r30106680

shared/uavobjectdefinition/vtolpathfollowersettings.xml
```diff
@@ -7,11 +7,20 @@
...
 		<field name="VelocityChangePrediction" units="" type="enum" elements="1" options="FALSE,TRUE" defaultvalue="FALSE"/>
 		<field name="MaxRollPitch" units="deg" type="float" elements="1" defaultvalue="20"/>
 
-		<!-- The distance within the waypoint (horizontally) that should be achieved to mark completed -->
-		<field name="EndpointRadius" units="m" type="uint8" elements="1" defaultvalue="2"/>
+		<!-- The distance within the endpoint (horizontally) that should be achieved to mark completed -->
+		<field name="EndpointRadius" units="m" type="float" elements="1" defaultvalue="2.0"/>
+
+		<!-- The distance that we can be below a waypoint and still mark it as completed.  Silently ignored if ThrottleControl == false -->
+		<field name="WaypointAltitudeTol" units="m" type="float" elements="1" defaultvalue="2.0"/>
```
peabody124: This default is probably reasonable. I can tell you one situation where this might cause issues: if there is z-accel bias then the velocity will be biased. You end up finding a balance point where the position error drives a velocity desired cancels out the velocity estimate and it hovers at the wrong altitude. 

Still we can always change value and this is reasonable default

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: where nicer is in the eye of the beholder...

mlyle:
<3 !

PR #1547, opened 2015-04-05
gcs: Sort objects in scope source pulldown

mlyle:
Mostly just an edit to get familiar with other parts of the codebase, but the
sheer number of data sources was annoying-- it'll be convenient for it to be
sorted.

### DIFF
--- a/ground/gcs/src/plugins/scope/scopegadgetoptionspage.cpp
+++ b/ground/gcs/src/plugins/scope/scopegadgetoptionspage.cpp
@@ -91,18 +91,34 @@ QWidget* ScopeGadgetOptionsPage::createPage(QWidget *parent)
     ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
     UAVObjectManager *objManager = pm->getObject<UAVObjectManager>();
     QVector< QVector<UAVDataObject*> > objList = objManager->getDataObjectsVector();
+    QStringList toAdd;
+
+    QString previous = NULL;
+
     foreach (QVector<UAVDataObject*> list, objList) {
         foreach (UAVDataObject* obj, list) {
+	    QString name = obj->getName();
+
             if (obj->isSingleInstance())
             {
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/scope/scopegadgetoptionspage.cpp
```diff
@@ -91,20 +91,36 @@ QWidget* ScopeGadgetOptionsPage::createPage(QWidget *parent)
...
         foreach (UAVDataObject* obj, list) {
+	    QString name = obj->getName();
+
             if (obj->isSingleInstance())
             {
-                options_page->cmbUAVObjects->addItem(obj->getName());
+		toAdd.append(name);
             }
-            else if(obj->getName() != options_page->cmbUAVObjects->itemText(options_page->cmbUAVObjects->count()-1))
+            else if(name != previous)
```
guilhermito: Not part of this PR I know but why are multi instance object limited to a single instance here?

mlyle:
I don't know much about multi instance objects, but all that's currently stored in the pulldown is the name.  So you'd have several of the object in a row, and the system would be unable to determine which you picked without further work.

lilvinz:
jenkins: test this please

ground/gcs/src/plugins/scope/scopegadgetoptionspage.cpp
```diff
@@ -91,20 +91,36 @@ QWidget* ScopeGadgetOptionsPage::createPage(QWidget *parent)
...
+
+            previous = name;
         }
     }
 
+    qSort(toAdd);
+
+    foreach (QString elem, toAdd) {
+	// Before we only added items to the spectrogram if they were not
+	// single instance; assuming that was in error
```
kubark42: FYI, we only added multi-instance items because each instance was one bin. So we should probably fix this so that the spectrogram only shows UAVO that it can plot, or teach it how to work with single-instance UAVOs (which would be pretty cool!)

Issue #1531, opened 2015-03-30
Safe boot doesn't seem to work on any F4 targets


peabody124:
I'm not sure why this is, but have noticed it with a variety of boards. Not sure if the non-volatile register is failing to be written, failing to be read, or getting overwritten.

PR #1550, opened 2015-04-06
Uploader: fix safe boot

peabody124:
There was an inconsistency between the bootloader code
and the uploader code. The bootloader swapped the bytes
of the safe word and compared to the constant 0x5AFE.
However, the uploader did not swap those bytes and thus
the safe_word was never valid to trigger a safe boot.

The cleanest thing to do would be remove the swap in the
bootloader, but for ease of the users not updating the
bootloader it swaps in the GCS.

### DIFF
--- a/ground/gcs/src/plugins/uploader/tl_dfu.cpp
+++ b/ground/gcs/src/plugins/uploader/tl_dfu.cpp
@@ -306,8 +306,11 @@ int DFUObject::JumpToApp(bool safeboot)
 {
     bl_messages message;
     message.flags_command = BL_MSG_JUMP_FW;
+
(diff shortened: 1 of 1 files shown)


peabody124:
It would be great if someone could test this. @guilhermito could you look at it in case you know why 0x5afe was flipped on the bootloader code.

Safe boot also doesn't appear to work on F1 targets (e.g. CopterControl) with or without this batch. I'll leave fixing (possibly) for later and it might require a new F1 bootloader.

peabody124:
fixes #1531

guilhermito:
This was done so it would maintain compatibility with the way the old boot loader handled words.

Issue #1526, opened 2015-03-29
Input wizard seems to disregard the center position


peabody124:
I think the input wizard might be takign the min+max average rather than the empirical center position. For transmitters that do not have well centered pots, this can create weird things.

lolorc:
I *might* have encountered this one.
If I have, it happened on an inverted sbus input, ChannelMin=1811, ChannelMax=172, ChannelNeutral was set to 1811
I did use input wizard not that long ago but I can't affirm ChannelNeutral was altered by the wizard, might as well been pebcak :)

elafargue:
Fixed with #1552

PR #1552, opened 2015-04-06
InputWizard: correctly identify stick center position

peabody124:
I introduced a bug when adjusting the throttle neutral
position and made all sticks neutral the middle of the
range which is not true for all transmitters. This could
create some annoying input calibrations.


peabody124:
This reverts 882a4ad7a486d6d4e3c306281130af49a7940abf and fixes #1526

I'm not sure why I created that patch. Probably cause DX7 has very nicely centered sticks compared to Taranis :)

PR #1558, opened 2015-04-14
OpenLRS support

peabody124:
This adds a PIOS_OpenLRS module that will bind to and process OpenLRS packets.
It has been tested with Sparky2. This PR does not enable it for any targets, but it would
be good to get in before the next release since it adds changes to ManualControlSettings
and ReceiverActivity (which are already changing for this release).


peabody124:
Video demonstration of this working: https://vimeo.com/124850842

elafargue:
Don't have a RevoMini to test, but otherwise no impact apart from the small changes to ReceiverActivity and ManualControlSettings. Merging.

PR #1560, opened 2015-04-18
Flight: Makefile updates to Colibri and Quanton

jihlein:
After merging the ChiBios updates, attempting to build the F4 targets (Colibri and Quanton) on Windows resulted in a status access violation crash from sh.exe.  The error was traced to an issue with the makefile syntax.  The Colibri and Quanton makefiles are updated to fix the syntax error in this pull request.

A make all_flight executed after the makefile updates successfully completed.  The resulting Quanton binary was loaded on a Quanton board and it executes correctly.  The Colibri binary was generated, but not tested.

Before merging, the makefile changes should be tested on OSX (already done by Peabody 124) and Linux platforms.


peabody124:
@jihlein i think you have another commit in here that is unrelated.

jihlein:
Yes, unfortunately I do.  I see how I messed this up, let me correct it.  Still wrestling with GIT.......

jihlein:
I made the makefile changes on my next, then made the pull request.  All was good at that point.  Then I made the DSM resolution changes on my next, and committed them.  That was my mistake.  

I should have made two branches, nextF4makefile and nextDSM, made the appropriate changes on those branches, and submitted the pull requests from those branches.

So, I backed out the DSM changes from my next, but I'm not sure if this corrects the existing pull request or not.

PR #1561, opened 2015-04-18
Flight: DSM 10 bit/11 bit auto detection

jihlein:
After the addition of the Spektrum DSM autodetection, the firmware failed to detect my satellite receivers.  I use a DSM2 Spektrum module in a JR PCM10 transmitter.

The issue is that for whatever reason, the Spektrum module outputs channels in a different order from a Spektrum transmitter,  This causes the current resolution detection to fail, and return DSM_UNKOWN.

The resolution detection routine was re-written to use the resolution setting embedded into the satellite data stream, rather than using channel order, which, as mentioned, appears to be different between the Spektrum module and Spektrum transmitters,

Since making this change, the firmware successfully detects both DSM2 and DSMX satellites (bound using 5 binding pulses to my Spektrum module) on Sparky, Quanton, and my Naze32Pro.

The PR needs to be checked over by users with other Spektrum TX configurations before being merged.  Note that the satellite must be bound as a master rx using the GCS.  Binding via a Spektrum receiver will result in a slave rx configuration which does not transmit the necessary resolution information.

Also note that at the moment, 4 binding pulses must be selected as there is a transient pulse being created that is interpreted as binding pulse.  The transient pulse plus the 4 generated pulses result in the 5 required pulses.  Correcting that will be the subject of another PR


peabody124:
jenkins: test this please

peabody124:
jenkins: add to whitelist

peabody124:
Sorry I haven't acted on this, yet. It will need some testing to make sure it is robust.

jihlein:
Of course.  It's hard to test without having all the different system permutations.  No rush, lets get it right.  I have it on a private branch right now so my builds will read my gear.  Keep in mind the satellite must be bound as a master or this definitely won't work.

Speaking of binding, I've not looked at the binding pulse issue with ChiBios yet, I'm curious if it's still there.

mlyle:
I vote for this change-- I need this for autodetect to work with my Spektrum DX6, OrangeRx satellite, sparky, and 4 ("5") binding pulses.  Could not get it to work reliably with old code (it worked one boot and then stopped).

peabody124:
It still passes unit tests. Code looks sensible to me. I put out a call for users to test it here: http://forum.taulabs.org/viewtopic.php?f=17&t=691 It will merge if there are no reports of problems.

jihlein:
Finally managed to test this on Sparky and Quanton.  Bound as master to my Spektrum module, RX is detected.  Bound as slave to my Spektrum module, RX is not detected.  This is as expected.

To finish testing, verification is needed that a slave that puts out channels in the "correct" order for what is expected in the slave resolution detection code is properly detected.

After that, I think this can merge.

peabody124:
Tested bound as a slave and as a master. Works in both cases.

Issue #1524, opened 2015-03-28
Make PWM/OneShot selection configurable in GCS


BrainFPV:
We should have an user friendly way to enable OneShot in the GCS. One option could be to add a selection to the output channel config as shown below. Thoughts?

![out_type](https://cloud.githubusercontent.com/assets/7647499/6882964/20a42e50-d56f-11e4-85af-fcddf754a06e.png)

joergrohde:
That would be nice. But also the true value. When you select OneShot the factor 12 has to be calculated in the GCS in the right way. So one has to make his settings 1000-2000 or 125-250, imho.

peabody124:
Agreed to both these points. More specifically, I think the settings should apply at the top the way update rate does since it influences an entire timer bank. This will avoid confusing situations where two channels on the same timer are assigned different tasks which will produce unexpected outcomes.

joergrohde:
You meen at this place?
![oneshot](https://cloud.githubusercontent.com/assets/4664234/6884554/16893350-d5f8-11e4-9f68-58db29702880.jpg)

BrainFPV:
I will look into adding it to the update rate selection. This makes me wonder why OneShot was added as an option to `ChannelType` and not `ChannelUpdateFreq`. As @peabody124 said, configuring some channels of a group as OneShot and others as PWM will produce unexpected outcomes.

How about changing `ChannelUpdateFreq` to be an enum and list OneShot there? This means it would be removed from `ChannelType`.

peabody124:
See the discussion in #1424 for rational. TL;DR: step 1 get it in and tested, step 2 refactor for user friendliness. I've been meaning to work on it (and actually consider this a release blocker, or at least that we don't announce OneShot).

@ernieieft might want to comment here.

The enum option is a bit difficult - for example of the brushed sparky I run at 8000 Hz output, so the freedom to freetype is nice. I thought a value of 0 or the max value might be appropriate for indicating OneShot. The GCS already maps that field to an enum combo box so from the GCS perspective it can end up in the same combo box.

It would also be nice to have a way of indicating x12 resolution while not using OneShot. For example again on BrushedSparky I'd prefer to run to even higher update rates with higher resolution, but not synchronously updating. In this case, overloading the Update Rate field might be different than the timer mode. We could have three timer modes:

 - x1 resolution, periodic
 - x12 resolution, periodic
 - x12 resolution, OneShot

[...]

PR #1566, opened 2015-04-30
HPWM: Expanding the Limits of the current OneShot implementation

ernieift:
OneShot-only mode is replaced by HPWM. OneShot is still available by setting the frequency value to zero. HPWM can be used to attach brushed motors and run them on higher frequency with sub microseconds resolution. This is also a BLHeli mode.


peabody124:
Ok. One problem. The ChannelType cannot be containing the "HWPWM" information. That needs to be in a separate field because we need that to be with the same granularity as the ChannelUpdateFrequency (i.e. associated with a timer bank). This still allows invalid/nonsenical things like having Ch1 PWM and Ch2 HPWM.

peabody124:
Tested and flew BrushedSparky2 using 12000 HPWM. Works great!

It's probably outside of the scope of this, but at this point I think we can drop the channel types "MK and ASTEC" and also the ChannelAddr field entirely.

ernieift:
There is no elegant way to solve this problem in the firmware. Only the GCS has the hardcoded information about which individual channel belongs to which bank. Within the firmware it is a cosmetic bug. The PWM channels will give out wrong information then. My Idea: In the Output widget we can trigger on channel type change and set the new type in all channels of this bank. Only when going to HPWM. If a channel was HPWM before and we want to go to PWM or similar like Info_LED we set it to PWM. The third possiblity is MK or ASTEC. These channel are left untouched by this procedure.
I would keep these channel types. Maybe we port the whole firmware in a completely different hardware without PWM timers in the future. Then we need to communicate with intelligent hardware or there is are newer or better ESC. It is better to have half of the work already done, as inventing all new.
The screen shot in #1524 with all channel types is great. I have often used Info_LED on tiny copters. Many users have never looked deeper in the settings and don't know this possiblity, i guess.

peabody124:
I'll try and build upon this over the weekend. I really think it should be possible to represent the information so that it accurately maps to the hardware and do it that way. Being able to configure the timers in such a way they do the wrong thing (e.g. set Ch0 to PWM and Ch1 to HPWM) isn't really a cosmetic bug IMO...

We configure the timers in the same method that responds to updates. Based on the timer downsampling the code that sets the output  period can determine if that particular output channel (which the config structure maps to a physical timer) is high res or normal res.

Let me give it a shot and see what you think. I'll also add the requisite GCS changes.

peabody124:
Take a long at this implementation of pios_servo:

https://gist.github.com/67d50c46b583ff99f7e1

The core method is PIOS_Servo_SetMode which configures both the sync/async information and the resolution. Everything is cached for each channel appropriately to make it easy to set the channel period in us blindly and have the period calculation sync update work.

ernieift:
I spent the whole forenoon with Qt. This isn't my play ground ;). Have a look at my solution. If i spent a little more time, i could also check if all channels of a bank are set to HPWM. So there won't a be conflict.

ernieift:
I've looked at the pios_servo source. I like the idea of expanding SetMode() with pwm_mode. There is one little problem in HPWM_Update(). You have to check if the timer is enabled, otherwise it will stretch the pulse length due multiple calls of TIM_SetCounter().

peabody124:
Ah, good point. I thought that was being used as an ad hoc method of detecting which channels were synchronous but it is more critical than that ;)

I'll adopt this to F3 and F1 as well ad the other related changes and push a full PR tonight.

ernieift:
I again pondered over PIOS_Servo_SetMode(). It is hardly possible to find out for actuator.c module which timer belongs to which channel. Accordingly, it can not provide the pwm_mode. Similarly, the function can not recognize itself in PIOS_servo.c the set PWM mode. If now the HAL driver passes even more information, you thwarted the functionality of HAL.
My suggestion. We unite PIOS_Servo_Set() and PIOS_Servo_HPWM_Set(). The PIOS_Server_SetHz() function receives an invisible boundary of "> 400Hz || == 0" to decide which base frequency to be adjusted. PIOS_Servo_HPWM_Update() decides on the basis of a stopped timer if the timer is set to zero and should be re-started.
This approach would eliminate the need for a separate HPWM mode. It also saves FLASH and RAM.
The channel types I'd like to leave in the GCS, as I have already Info/ArmingLED often used on small copters.

peabody124:
I agree. I didn't post it, but the method of Actuator.c I was using only calls HPWM if that is enabled and calls PWM set if that option is not compiled in. Since we have it enabled for all targets at this point, I'll actually remove the flag entirely and have the code always use it consistently.

The channel type field remains, but only has a "PWM" option and a TimerUpdateRate and TimerResolution field determine the resolution and sync properties.

ernieift:
Wait a minute. Right now, I'm testing a version that removes the channel type completely. If this works it is like you've described.

ernieift:
And here it is. Saves RAM and FLASH as promised.

fujin:
This is looking great. In need of any testers? I have the previous oneshot stuff still running on all of my boards.

jihlein:
Built this on one of my branches.  400 Hz PWM works fine,  50 Hz PWM buzzer works fine.  Tried 0 Hz for the oneshot mode.  Got the proper oneshot frequency (gyro update rate at 500 Hz), but I could not get the correct pulse widths.  With a min pulse of 1000, I got an 83.3 uSec pulse width.  Changed min pulse to 1500, but still got the 83.3 uSec pulse width.  On the high side, it wouldn't even let me enter 3000.  Didn't check the actual high pulse width.

Not sure if I built it correctly or??

ernieift:
I think you built it correctly. This is a rollback to physics. In OneShot mode, simply enter 125 for MIN and 250 for MAX. My KISS start at 131 which i entered as NEUTRAL.

jihlein:
Let me try it again. I was still entering 1500-3000 as the limits, not 125-250!  Thanks!

fujin:
2015-05-12-110550_1920x1080_scrot

jihlein:
Okay, got it working.  Couple of notes.  Here's my output window config.  Channel 1, 50 Hz pwm driving gimbal ptch servo, channels 3-6, oneShot125 driving motors (blHeli 13.2), channel 9, 50 Hz pwm buzzer function.

outputs

All functions work fine.  The box around the neutral setting is a bit small for 5 digit numbers.

Is this statement, #define HiresMinPeriod 500, correct?  It's compared against speeds[set], which from the comments is in Hz.  So the code is comparing a period (in Usec?) against Hz.  Should this be #define HiresMinFreq 2000?

ernieift:
@jihlein you are right. HiresMinPeriod wasn't physically  correct. I didn't want to use frequency again. Now it's called speed.
I heard about problems with BlHeli and OneShot due auto detect. Have you tried the brushed mode with higher frequencies on them?
Your Neutral values font is larger then mine. Have you tried a fresh configuration?

jihlein:
All my GCS build on Windows 7 have that font for the Neutral values.

I've not been following the BlHeli development too closely, other than to know that Rev 13.2 autodetects the one shot pulse widths.  Any specific test scenarios/setups you would like run?

ernieift:
I read about configs with higher MPU rates which BlHeli detected as brushed mode. There is a discussion at the BlHeli github about the problematics (https://github.com/bitdump/BLHeli/issues/55). 4kHz PWM may be the better choice for BlHeli ESCs.

peabody124:
Hey @ernieift . I finally had a chance to finish the code I started a few weeks ago. Can you take a look at #1615. It expands on your work to make the settings per timer explicitly. Also from talking to the devs at the retreat, setting the timer resolution explicitly is also preferrable. This also allows dropping that old
prescalar thing that was being passed around.

I hope you don't mind me taking this and extending this - I saw you continued working on it in parallel and I think I grabbed your changes that fit too (and kept the commit message).

ernieift:
I do not mind. Maybe we can do something together to make it perfect. Unfortunately, I had little time to chat during the retreat. I would continue here for tests but we merge #1615 finally.

peabody124:
this PR is superseded by #1615

PR #1567, opened 2015-05-01
Unify Mac and Posix simulator...

mlyle:
There's still one file of difference conditionalized at build time, which I may eventually fix/kill.  In the meantime it's clean enough.

I have smoke tested this on both OSX and Linux, but I will be doing some additional testing tomorrow-- shouldn't merge until I say that's OK.

Sending the pull request to start the code review process and to see how Jenkins likes this.

Yay, 15811 deletions...  I like writing code.... but I like deleting code more!


mlyle:
Confirmed working on meaningful simulation flights on Linux.  And also, I confirmed that the weird anomalous behavior I was seeing before-- aircraft jumping hundreds of miles away from home on startup, etc.. is gone.  OSX still to test.

mlyle:
Confirmed healthy on OS X, flew a return to home flight!  OK to go as far as I'm concerned.

mlyle:
When this is pulled, I will update wiki documentation to say you do make sim_posix_revolution on OSX.

Also the wiki documentation says you need to get gcc from brew, and I don't really think that's true-- it all seems to compile with the system compiler/clang.

ernieift:
Could you also change the makefile in order to compile the right OS version by default?

peabody124:
Compiled and minim tests performed on OSX. More testing to come.

Makefile
```diff
@@ -846,7 +845,7 @@ EF_BOARDS  := $(ALL_BOARDS)
 ifeq ($(UNAME), Linux)
 SIM_BOARDS := sim_posix_revolution
 else ifeq ($(UNAME), Darwin)
-SIM_BOARDS := sim_osx_revolution
+SIM_BOARDS := sim_posix_revolution
```
mlyle: I thought this would change to build the correct one.  Did I mess up?  will test a complete build later if able..

mlyle:
@ernieift Just checked-- make all does indeed build sim_posix_revolution on OSX.

mlyle:
We might also want to kill the win32 simulator.  It's not even close to working and I'm not sure anyone plans to put in the effort to get it going... It'd be one less combination to worry about with chibios.

peabody124:
Ok with me. Want to add to this PR?

mlyle:
Yup, added.  Jenkins likes it, and I confirmed it builds happy on OSX with the win32 stuff removed.

peabody124:
Did some more flying around in OSX. I'm happy with this.

mlyle:
Yay!  Thank you.

PR #1568, opened 2015-05-01
Ground - Fixes Dragons

guilhermito:


peabody124:
Perfectly in time for the retreat. Would not be ok without dragons!

BrainFPV:
I like dragons, why fix them?

PS: Can you explain what dragons are in the context of this PR? ;)

mlyle:
Nice!  I think this is the google maps stuff I've been whining about!

mlyle:
Unfortunately, it doesn't seem to fix my issues of missing tiles-- though different ones are missing.. Going to double check whether it's better on Linux next.  
maps

mlyle:
Huh.  Obviously way better on Linux-- there is still an anomaly at maximum zoom-out where we only get one quadrant, but zoomed in I can see my flying field, etc.

mlyle:
Happy on OSX after resizing window smaller.

It looks like there are separate bugs in the opmap widget, where on OSX at max resolution on 5k imac it fails to render properly.   Also, at smaller res, I occasionally see tiles go to the wrong place (e.g. Africa at the top of South America).  Regressions not linked to this commit, this commit definitely makes things better.  I vote for this being pulled.

PR #1570, opened 2015-05-02
python-api: some initial commits for review

mlyle:
I need to add a few convenience methods, documentation, and some example code... but this is shaping up.  It's now working again, now has a lock model, select/nonblocking, and generator goodness.

This is fine to actually get pulled-- it's more useful than what was there before-- but it is not complete.

You can use either the "old way" and look through the uavo_list objects, or use as an iterator to inspect telemetry objects as they come in.  Or both, actually.  And you can use single threaded or with a dedicated handler thread, in blocking and nonblocking variants.

### DIFF
--- a/python/minimal_connection.py
+++ b/python/minimal_connection.py
@@ -57,8 +57,8 @@ def main():
 
     telemetry.open_network()
 
-    while True:
-        telemetry.serviceConnection()
+    for obj in telemetry:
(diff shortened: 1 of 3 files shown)


mlyle:
closing.  Much bigger pull request coming soon, hopefully it's palatable.

PR #1571, opened 2015-05-05
Simulator: run on chibios instead of freertos.

mlyle:
Another day, another sim commit.  Tried to make this as clean as possible.  It seems to work well on both Linux and OSX; I've been using this for my development of the ground api for the past few days.


mlyle:
Ah, some of the earlier stuff needs fixing for other targets, I think.. looking...

mlyle:
Looks good.

peabody124:
I'm not sure if you want to include it in this patch, but I think we should drop the "revolution" target. This was essentially a development board and I don't think anyone is still using it. Since that is what we use for simulation, it would make sense to totally convert it to the simulation target, especially now simulators are unified. e.g. ```make fw_simulation```

flight/PiOS/Common/pios_semaphore.c
```diff
@@ -397,6 +397,6 @@ bool PIOS_Semaphore_Give_FromISR(struct pios_semaphore *sema, bool *woken)
 
 	return result;
 }
-#endif /* !defined(SIM_POSIX) && !defined(SIM_OSX) */
+#endif /* defined(PIOS_INCLUDE_FREERTOS) && (!defined(SIM_POSIX) && !defined(SIM_OSX)) */
```
peabody124: Can you remove all occurrences of ```SIM_OSX```

flight/PiOS/Common/pios_semaphore.c
```diff
@@ -341,7 +341,7 @@ bool PIOS_Semaphore_Give(struct pios_semaphore *sema)
 }
 
 /* Workaround for simulator version of FreeRTOS. */
-#if !defined(SIM_POSIX) && !defined(SIM_OSX)
+#if defined(PIOS_INCLUDE_FREERTOS) && (!defined(SIM_POSIX) && !defined(SIM_OSX))
```
peabody124: This change is not needed. It is already in a FreeRTOS only conditional.

flight/PiOS/inc/pios_semaphore.h.save
```diff
@@ -0,0 +1,80 @@
+/**
```
peabody124: this file was probably not intended

PR #1574, opened 2015-05-08
Input Config: fix glitch importing reversed channels

peabody124:
Importing a .uav file that has an input channel reversed causes
the neutral value to hit the limits, because the neutral is set
before both channel limits are set.

Because the addUAVObjectToWidgetRelation preserves the order things
are registered, simply connecting min/max first fixes this issue.
A better solution would be to allow making changes to the whole
object atomically, but that would be a pretty major overhaul.


BrainFPV:
LGTM

elafargue:
good catch

PR #1578, opened 2015-05-09
FIX: don't modify board rotation during calibration

### LINKED ISSUE #1572: BUG: Six point calibration can modify FC orientation relative to UAV
When performing the 6-point calibration, the GCS first fetches `AttitudeSetting.BoardRotation` and then sets all values to zero for the calibration. When the calibration is successful, the orignial values are written back to the FC. 

This is problematic, as the calibration can be interrupted, e.g. due to a USB disconnection when moving the UAV around during calibration, which will leave `AttitudeSetting.BoardRotation` set to all zero instead of the original values. When the calibration is attempted again without restarting the FC, the GCS will fetch the all-zero values and write them back to the FC after the calibration, which may be different from the values that the user originally entered. This can have bad consequences, e.g., if the FC is mounted facing backwards (Yaw=180), the UAV will perform a "flip of death" when spooling up.

The above also applies to the sensor scalings (`SensorSettings.AccelScale`, `SensorSettings.MagScale`). 

I see two potential ways to solve this:

1. Leave `AttitudeSetting.BoardRotation` unchanged and the GCS applies the inverse of the rotation/scaling matrix to the xyz sensor values during the calibration. 

2. Leave `AttitudeSetting.BoardRotation` unchanged and add a special mode to the Sensor module, which causes it not to apply scaling/rotation to the sensor values. This mode is only used for calibration and the Sensor module will remain in error state as long as it is enabled.

IMO, solution 2 may be better, as 1 needs us to replicate the calibration from the FC side in the GCS and if there is a discrepancy, the resulting coefficients will be wrong. Thoughts?

BrainFPV:
Fixes #1572

Board rotation is applied in GCS instead of resetting the rotation on the FC. Scales and bias are still reset, but this is much less critical.

Ready for review.

### DIFF
--- a/ground/gcs/src/plugins/config/calibration.cpp
+++ b/ground/gcs/src/plugins/config/calibration.cpp
@@ -1053,13 +1053,13 @@ bool Calibration::storeSixPointMeasurement(UAVObject * obj, int position)
         // Store the average accelerometer value in that position
         if (calibrateAccels) {
             // undo the board rotation that has been applied to the sensor values
-            double xyz[3] = {listMean(accel_accum_x), listMean(accel_accum_y), listMean(accel_accum_z)};
-            double xyz_body[3];
-            rotate_vector(boardRotationMatrix, xyz, xyz_body, true);
+            double accel_body[3] = {listMean(accel_accum_x), listMean(accel_accum_y), listMean(accel_accum_z)};
+            double accel_sensor[3];
+            rotate_vector(boardRotationMatrix, accel_body, accel_sensor, false);
 
-            accel_data_x[position] = xyz_body[0];
-            accel_data_y[position] = xyz_body[1];
-            accel_data_z[position] = xyz_body[2];
+            accel_data_x[position] = accel_sensor[0];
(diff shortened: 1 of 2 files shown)


BrainFPV:
Ready for review / merge. Tested 6-point with several random board rotations and got the same results.

BrainFPV:
ping @peabody124

peabody124:
I tested this three times and all three times it told me it failed to get a valid calibration.
screen shot 2015-05-22 at 10 34 37 am
screen shot 2015-05-22 at 10 35 09 am

BrainFPV:
@peabody124 very strange.. did the accelerometer calibration succeed?

BrainFPV:
Ping @peabody124 can you have another look? I just tested this again and it's working fine here..

peabody124:
Ok. Retested with a erased board and things are fine:
screen shot 2015-06-08 at 10 09 55 am
screen shot 2015-06-08 at 10 10 39 am

Nicely produces very similar values with level or rotated board.

PR #1585, opened 2015-05-10
PathPlanner Fixes

peabody124:
This primarily fixes issues with the behavior at the end of a flight
plan. It also prevents a race condition where we can potentially
run through multiple waypoints quickly.


mlyle:
Looks good to me.

elafargue:
All fixes found while flying during the Spring 2015 meetup - no quads were badly harmed during this test.

PR #1589, opened 2015-05-11
wizard: adds SUMD to the possible choices

ernieift:
Adds the SUMD protocol to the radio wizard on enabled targets.


BrainFPV:
LGTM

PR #1592, opened 2015-05-12
pathplanner: CRITICAL: prevent accessing uninit'd mem

mlyle:


peabody124:
Good catch!

peabody124:
Tested in simulator and it definitely fixed a regression.

Issue #1588, opened 2015-05-10
Initial altitude is not 0 at power up


elafargue:
When powering up (or arming) the altitude estimate should be set to zero
so that navigation is relative to launch altitude.

elafargue:
This is now fixed

PR #1595, opened 2015-05-12
Attitude: take barometric altitude on powerup as zero

peabody124:
Previously we tried to compare the home location altitude
against the current gps altitude and preserve that offset.
In practice, this is inconvenient and we should take the
powerup altitude as zero.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -1101,7 +1101,7 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 			getNED(&gpsData, NED);
 
 			// Initialize barometric offset to current GPS NED coordinate
-			baro_offset = -NED[2] - baroData.Altitude;
+			baro_offset = -baroData.Altitude;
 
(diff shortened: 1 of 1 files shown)


peabody124:
Fixes #1588

elafargue:
IMHO we should go one step further, and zero altitude at arming time. This is similar to what you do in a real plane at takeoff, and should be the safest/most reliable option.

Thoughts ?

peabody124:
I made a separate issue for that. We should make that option also hit the home location and it's a touch more work.

flight/Modules/Attitude/attitude.c
```diff
@@ -1101,7 +1101,7 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 			getNED(&gpsData, NED);
 
 			// Initialize barometric offset to current GPS NED coordinate
```
elafargue: Update comment too - Initialize baro offset to zero

mlyle:
This is sane for now as a height datum for now-- pressure altitude above home location.  We'll need to figure out how to handle the mean sea level (with baro corrections) and above ground level datums too.

PR #1598, opened 2015-05-12
PathPlanner: use a queue to respond immediately to FlightStatus

peabody124:
Because the PathPlanner runs at a fairly slow rate, we could miss
rapid toggles of the FlightMode switch. This would, for example,
allow the follower to briefly enter RTH mode and then resume the
current path segment. For longer switches a different behavior
would emerge (the path planner detects it and resets to path
beginning).


mlyle:
Doesn't fix everything-- but if it passes tests it is a good change on the way.

PR #1599, opened 2015-05-12
Simulation: make a standalone target

peabody124:
The simulation code really has nothing to do with any of the board targets
and was only grouped together for historical reasons. This makes a standalone
simulation target

```make simulation```

In a separate PR we can probably drop the revolution (full) hardware entirely.


peabody124:
@mlyle let me know what you think of this. One downside of keeping the old macro expansion code is that the simulation ends up in ```build/sim_simulation_posix```. It probably is worth dropping that expansion code, don't you think?

mlyle:
sure, I like shorter names, it can become sim_posix.  Once it builds it sounds good :)

mlyle:
(The reason this is failing is, I think, because jenkins is explicitly making sim_posix_revolution).

peabody124:
Grr jenkins

peabody124:
Closing this. Superseded by #1631

PR #1600, opened 2015-05-13
vtolpathfollower: fsm improvements

mlyle:
Removes the "return to home" state toggle landmine and makes the FSM code slightly cleaner.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -271,13 +271,10 @@ const static struct vtol_fsm_state *current_goal;
 //! The current state within the goal fsm
 static enum vtol_fsm_state_num curr_state;
 
-/**
- * Process any sequence of automatic state transitions
- */
-static void vtol_fsm_process_auto()
+static int vtol_fsm_enter_state(enum vtol_fsm_state_num state)
 {
-	while (current_goal[curr_state].next_state[FSM_EVENT_AUTO]) {
-		curr_state = current_goal[curr_state].next_state[FSM_EVENT_AUTO];
+	if (state != FSM_STATE_UNCHANGED) {
+		curr_state = state;
 
 		/* Call the entry function (if any) for the next state. */
 		if (current_goal[curr_state].entry_fn) {
@@ -285,7 +282,32 @@ static void vtol_fsm_process_auto()
 		}
 
 		configure_timeout(current_goal[curr_state].timeout);
+
+		return 1;
 	}
+
+	return 0;
+}
+
+static int vtol_fsm_process_event(enum vtol_fsm_event event)
+{
+	enum vtol_fsm_state_num next = current_goal[curr_state].next_state[event];
+
+	/* Special if condition to not have to explicitly define auto
+	 * (don't do fault transitions on auto) */
+	if ((event != FSM_EVENT_AUTO) || (next != FSM_STATE_FAULT)) {
+		return vtol_fsm_enter_state(next);
+	}
+
+	return 0;
+}
+
+/**
+ * Process any sequence of automatic state transitions
+ */
+static void vtol_fsm_process_auto()
+{
+	while (vtol_fsm_process_event(FSM_EVENT_AUTO));
 }
 
 /**
@@ -295,12 +317,8 @@ static void vtol_fsm_process_auto()
 static void vtol_fsm_fsm_init(const struct vtol_fsm_state *goal)
 {
 	current_goal = goal;
-	curr_state = FSM_STATE_INIT;
 
-	/* Call the entry function (if any) for the next state. */
-	if (current_goal[curr_state].entry_fn) {
-		current_goal[curr_state].entry_fn();
-	}
+	vtol_fsm_enter_state(FSM_STATE_INIT);
 
 	/* Process any AUTO transitions in the FSM */
 	vtol_fsm_process_auto();
@@ -315,32 +333,11 @@ static void vtol_fsm_fsm_init(const struct vtol_fsm_state *goal)
  */
 static void vtol_fsm_inject_event(enum vtol_fsm_event event)
 {
-	// No need for mutexes here since this is called in a single threaded manner
-
-	/*
-	 * The STATE_UNCHANGED indicates to ignore these events
-	 */
-	if (current_goal[curr_state].next_state[event] == FSM_STATE_UNCHANGED)
-		return;
-
-	/*
-	 * Move to the next state
-	 *
-	 * This is done prior to calling the new state's entry function to
-	 * guarantee that the entry function never depends on the previous
-	 * state.  This way, it cannot ever know what the previous state was.
-	 */
-	curr_state = current_goal[curr_state].next_state[event];
-
-	/* Call the entry function (if any) for the next state. */
-	if (current_goal[curr_state].entry_fn) {
-		current_goal[curr_state].entry_fn();
-	}
+	vtol_fsm_process_event(event);
 
 	/* Process any AUTO transitions in the FSM */
 	vtol_fsm_process_auto();
 }
-
 
 /**
  * vtol_fsm_static is called regularly and checks whether a timeout event has occurred
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -95,9 +95,10 @@ enum vtol_fsm_state {
 };
 
 //! Structure for the FSM
-struct vtol_fsm_transition {
+struct vtol_fsm_state {
 	void (*entry_fn)();       /*!< Called when entering a state (i.e. activating a state) */
 	int32_t (*static_fn)();   /*!< Called while in a state to update nav and check termination */
+	uint32_t timeout;	  /*!< Timeout in milliseconds. 0=no timeout */
 	enum vtol_fsm_state next_state[FSM_EVENT_NUM_EVENTS];
 };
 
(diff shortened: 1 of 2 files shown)


flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -80,7 +80,7 @@ enum vtol_fsm_event {
  * sets the nav mode and a default methods will farm it out to the appropriate
  * algorithm.
  */
-enum vtol_fsm_state {
+enum vtol_fsm_state_num {
```
peabody124: why this change? it is the state. not a number *confused*

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -95,10 +95,11 @@ enum vtol_fsm_state {
 };
 
 //! Structure for the FSM
-struct vtol_fsm_transition {
+struct vtol_fsm_state {
```
peabody124: yep that's a better name.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -114,6 +115,8 @@ enum vtol_nav_mode {
 	VTOL_NAV_IDLE    /*!< Nothing, no mode configured */
 };
 
+#define MILLI 1000
```
peabody124: Maybe call this S2MS or something?

peabody124: Meh. nevermind, not worth ti.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -204,6 +208,7 @@ const static struct vtol_fsm_transition fsm_land_home[FSM_STATE_NUM_STATES] = {
 	[FSM_STATE_PRE_RTH_RISE] = {
 		.entry_fn = go_enable_rise_here,
 		.static_fn = do_ph_climb,
+		.timeout = 10 * MILLI,	/* Not sure this is good */
```
peabody124: The thinking here was that if it is for some reason not getting up there, give up and start coming home. We can increase timeout.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -250,38 +267,58 @@ void configure_timeout(int32_t s)
...
-static enum vtol_fsm_state curr_state;
+static enum vtol_fsm_state_num curr_state;
 
-/**
- * Process any sequence of automatic state transitions
- */
-static void vtol_fsm_process_auto()
+static int vtol_fsm_enter_state(enum vtol_fsm_state_num state)
 {
-	while (current_goal[curr_state].next_state[FSM_EVENT_AUTO]) {
```
peabody124: Note to myself: review this logic change again.
Note to Mike: can you explain to me what this makes it do differently?

mlyle: It factors the state transition into a single function rather than the code being duplicated below (vtol_fsm_enter_state).   It also factors timeout configuration into the same place.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -342,10 +358,21 @@ static int32_t vtol_fsm_static()
 			return ret_val;
 	}
 
-	current_count++;
-
-	if ((timer_duration > 0) && ((current_count - set_time_count) * DT) > timer_duration) {
-		vtol_fsm_inject_event(FSM_EVENT_TIMEOUT);
+	/* If there's a timer... */
```
peabody124: This logic should be moved into a function that just checks the timer for expiration. Will also help for refactoring into library later, right?

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: honestly?  Because I named the structure vtol_fsm_state below (because it made more sense to me) and then had to change the name of this.  Glad to put back if it's particularly objectionable.

peabody124:
First pass review looks good. I need to review again a few segments more closely. Nice work.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: Will do later.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: Yah, and with the ability to be stuck at offsets probably some timeout is good.

We really need "not making [much] progress" alarms to do this well.  Maybe some of the alarm infrastructure i'm thinking about can be in the FSM (but there's probably other subsystems that want to use it, so maybe there's a clever way to factor it out).

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
peabody124: The old names were consistent with other state machines in the code:
https://github.com/TauLabs/TauLabs/blob/next/flight/PiOS/STM32F4xx/inc/pios_i2c_priv.h#L53
https://github.com/TauLabs/TauLabs/blob/next/flight/PiOS/STM32F4xx/pios_i2c.c#L80

I added the static_fn to the current state structure (hence making the name _transition) slightly less accurate. My preference would be to keep these names the same and consistent with other FSM's in our code (I think @solidgoldbomb took this convention/design pattern from the linux kernel).

Not a blocker though if you really want to change it :)

mlyle: Nope, being consistent is really important.

mlyle:
OK, review feedback incorporated.

peabody124:
Not relevant for this review but thought triggered by it:

Here: https://github.com/mlyle/TauLabs/blob/mpl-fsmtimers/flight/Modules/VtolPathFollower/vtol_follower_fsm.c#L171 the static reaction while initiailizing the do_path fsm is the ideal place to do any form of handshaking to the path planner

Here https://github.com/mlyle/TauLabs/blob/mpl-fsmtimers/flight/Modules/VtolPathFollower/vtol_follower_fsm.c#L179 would eventually be a better place to be indicating to the pathplanner a segment is complete.

Totally not relevant for review, but highlighting them.

mlyle:
New, nicer documented diff.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -233,16 +240,56 @@ const static struct vtol_fsm_transition fsm_land_home[FSM_STATE_NUM_STATES] = {
...
 {
-	set_time_count = current_count;
-	timer_duration = s;
+	if (ms) {
+		uint32_t now = PIOS_Thread_Systime();
+		vtol_fsm_timer_expiration = ms+now;
+
+		// If we wrap and get very unlucky, make sure we still
+		// have a timer 1ms later.
+		if (vtol_fsm_timer_expiration == 0) {
```
peabody124: Good catch! That one in 4294967296 times this failed would be very hard to reproduce but would essentially leave a quad hanging in RTH.

mlyle: It'd have to be flying 49 days first... then get unlucky :)

mlyle: Part of why I overkilled this timer mechanism is that .. I think that you're (no one else) going to end up using tau labs firmware for odd things that may have high uptimes.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
guilhermito: Or make it a macro

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -255,18 +302,55 @@ const static struct vtol_fsm_transition *current_goal;
...
+		curr_state = state;
 
 		/* Call the entry function (if any) for the next state. */
 		if (current_goal[curr_state].entry_fn) {
 			current_goal[curr_state].entry_fn();
 		}
+
+		/* 0 disables any pending timeout, otherwise it's set to the
+		 * value for this state */
+		vtol_fsm_timer_set(current_goal[curr_state].timeout);
```
peabody124: doesn't matter for now, but later this might need to move up a few lines. that way a state can actually control it's timeout from the entry_fn() (imagine endpoints with programmable dwell times)

mlyle: ah, my thought was if the entry_fn was time consuming and bound up the fsm thread, it should get a chance to do that before the timer gets set.  Willing to move up or even to do something fancy (set 0, then if still 0 set to prespecified timeout), just lmk.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -293,36 +373,18 @@ static void vtol_fsm_fsm_init(const struct vtol_fsm_transition *goal)
...
+ * afterwards.
  */
 static void vtol_fsm_inject_event(enum vtol_fsm_event event)
 {
-	// No need for mutexes here since this is called in a single threaded manner
-
-	/*
-	 * The STATE_UNCHANGED indicates to ignore these events
-	 */
-	if (current_goal[curr_state].next_state[event] == FSM_STATE_UNCHANGED)
```
peabody124: so a side effect of this change: an event that does not make a change will still process auto events. however, since any auto event for a state should have already happened that doesn't seem to matter.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
peabody124: it's fine for now since we don't have that functionality. just point it out for future.

mlyle:
Thank you!

PR #1601, opened 2015-05-13
python-api: further api work

mlyle:
The focus of this one was to factor the log handling code into the uavo talk subsystem, and to create a magical module that can be imported to run these scripts standalone instead of only using ipython.

This also separates logview.py into shell.py, dump.py, and logview.py for the different modes.  Note I am unable to test logview.py in my environment.


peabody124:
Note to self: verify ```logview.py -v``` still works.

mlyle:
Rebased to next to escape conflicts.

mlyle:
Oh, also note, it's just logview.py, shell.py, and dump.py for the 3 different modes.  no -v anymore

python/taulabs/telemetry.py
```diff
@@ -177,28 +199,62 @@ def serviceConnection(self, timeout=None):
...
+        self._do_io(0)
+
+        while (len(self.recv_buf) < 1) and self._do_io(finishTime):
+            pass
+
+        if len(self.recv_buf) < 1:
+            return None
 
-        self.__select(0)
+        ret=self.recv_buf
```
BrainFPV: style: space around `=`

python/taulabs/telemetry.py
```diff
@@ -177,28 +199,62 @@ def serviceConnection(self, timeout=None):
...
+        self.recv_buf=''
```
BrainFPV: same

python/shell.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/python -B
+
+from taulabs import telemetry,uavo
```
BrainFPV: style: space after `,`

BrainFPV:
Looks good to me

python/logview.py
```diff
@@ -1,180 +1,22 @@
...
-        else:
-            # Instantiate an ipython shell to interact with the log data.
-            import IPython
-            from IPython.frontend.terminal.embed import InteractiveShellEmbed
-            e = InteractiveShellEmbed(user_ns = user_ns, user_module = user_module)
-            e.enable_pylab(import_all = True)
-            e("Analyzing log file: %s" % src)
+    # Start the log viwer app
+    from PyQt4 import QtGui
+    from logviewer.gui import Window
```
peabody124: we need to ```import``` sys here

python/mag_calibration.py
```diff
@@ -1,3 +1,7 @@
+#!/usr/bin/python
```
peabody124: how would you feel about leaving mag_calibration in a calibration directory? I can see us having a number of calibration routines and it might be good to organize them. is that a problem for importing?

python/taulabs/readbyargs.py
```diff
@@ -0,0 +1,3 @@
+import telemetry
```
peabody124: is this method used anywhere or callable? what does this do?

python/taulabs/telemetry.py
```diff
@@ -9,35 +9,40 @@
...
             else:
-                uavo_defs.from_uavo_xml_path("../shared/uavobjectdefinition")
+                uavo_defs.from_uavo_xml_path("shared/uavobjectdefinition")
+
+        self.githash = githash
 
         self.uavo_defs = uavo_defs
         self.uavtalk_generator = uavtalk.processStream(uavo_defs,
-            useWallTime=useWallTime)
+            useWallTime=useWallTime, logTimestamps=weirdTimestamps)
```
peabody124: name could be more descriptive ;-)

peabody124:
small comments, but essentially good to merge once they are addressed

python/mag_calibration.py
mlyle: Just adds a line of code to set the path (on each importer).

python/taulabs/readbyargs.py
mlyle: So, when you import a python module, the code runs.

This creates a shorthand when you do from taulabs.readbyargs import uavo_list ... that it runs that method.  It's slight abuse of the module system but seemed like a convenient way to do this.

python/logview.py
mlyle: Good catch.

python/taulabs/readbyargs.py
mlyle: (the standalone programs other than minimal_connection do things this way)

mlyle:
Fixed up according to review feedback, and moved mag_cal script.  Check out the change to that-- if the new wordy boilerplate is OK I'm happy for commit.

(Also, we should be careful to cut and paste that boilerplate exactly, just in case i need to sed it with anything subsequently)

peabody124:
LGTM. @solidgoldbomb you want to review this at all? I think it should be good, though.

solidgoldbomb:
I'll look it over later today.

python/calibration/mag_calibration.py
```diff
@@ -97,4 +100,17 @@ def distortion(x,mag_x=mag_x,mag_y=mag_y,mag_z=mag_z,LH=LH,LV=LV):
...
 
-	return res, cor_x, cor_y, cor_z
\ No newline at end of file
+	return res, cor_x, cor_y, cor_z
+
+def main():
+	print mag_calibration(uavo_list.as_numpy_array(UAVO_Magnetometer), uavo_list.as_numpy_array(UAVO_Gyros))
+
+	# Wait for user to close window.
+        matplotlib.pyplot.show()
```
solidgoldbomb: indentation (should be tabs)

python/calibration/mag_calibration.py
```diff
@@ -97,4 +100,17 @@ def distortion(x,mag_x=mag_x,mag_y=mag_y,mag_z=mag_z,LH=LH,LV=LV):
...
+
+if __name__ == "__main__":
+	import sys, os
+	sys.path.insert(1, os.path.dirname(sys.path[0]))
+        from taulabs.readbyargs import uavo_list
```
solidgoldbomb: indentation (should be tabs)

python/calibration/mag_calibration.py
```diff
@@ -1,100 +1,116 @@
...
+def main():
+        print mag_calibration(uavo_list.as_numpy_array(UAVO_Magnetometer), uavo_list.as_numpy_array(UAVO_Gyros))
+
+        # Wait for user to close window.
+        matplotlib.pyplot.show()
+
+if __name__ == "__main__":
+        import sys, os
+        sys.path.insert(1, os.path.dirname(sys.path[0]))
+        from taulabs.readbyargs import uavo_list
```
solidgoldbomb: Eww, uavo_list is global and initialized as a side effect of importing this?  We need a better way to deal with this.

python/calibration/mag_calibration.py
```diff
@@ -1,100 +1,116 @@
...
-	matplotlib.pyplot.subplot(1,2,2)
-	matplotlib.pyplot.plot(sqrt(cor_x**2+cor_y**2+cor_z**2))
-
-	return res, cor_x, cor_y, cor_z
\ No newline at end of file
+        """ Calibrates the magnetometer data by fitting it to a sphere,
+        ideally when constantly turning to spread the data around that
+        sphere somewhat evenly (or at least in a horizontal plane)"""
+
+        import numpy
```
solidgoldbomb: is this still required?

python/calibration/mag_calibration.py
mlyle: So the existing stuff that ran in the ipython shell expect uavo_list to be global-- that's how it's been provided to them through the ipython instantiation.  This was an effort to come up with a minimal adapter to allow them to be run either way.

python/minimal_connection.py
```diff
@@ -62,7 +57,7 @@ def main():
 
     for obj in tStream:
         print obj
-        pass
+#        pass
```
solidgoldbomb: Remove debug code.

python/calibration/mag_calibration.py
mlyle: (Though in this case those lines could easily/safely be moved to main)

python/taulabs/telemetry.py
```diff
@@ -177,28 +199,62 @@ def serviceConnection(self, timeout=None):
...
+                gcsTimestamps=False,  *args, **kwargs)
 
-        self.__select(0)
+        self.recv_buf = ''
+        self.send_buf = ''
+
+        self.fd = fd
+
+    def _receive(self, finishTime):
+        """ Fetch available data from TCP socket """
```
solidgoldbomb: This wouldn't have to be strictly a TCP socket now that you've abstracted out the FDTelemetry class.

python/taulabs/telemetry.py
```diff
@@ -212,36 +268,139 @@ def __select(self, finishTime):
...
+                self.send_buf = self.send_buf[written:]
 
             didStuff = True
 
         return didStuff
 
-    def __receive(self, finishTime):
-        """ Fetch available data from TCP socket """
+    def _send(self, msg):
+        """ Send a string out the TCP socket """
```
solidgoldbomb: This wouldn't have to be strictly a TCP socket now that you've abstracted out the FDTelemetry class.

python/minimal_connection.py
mlyle: If you want i can just delete minimal_connection.py and keep it outside of tree... it does nothing useful-- it is purely a scaffolding to develop the live telemetry interface with (and that i end up playing with as development continues)

python/taulabs/telemetry.py
```diff
@@ -212,36 +268,139 @@ def __select(self, finishTime):
...
+        return buf
+
+    def _done(self):
+        return self.done
+
+def _normalize_path(path):
+    import os
+    return os.path.normpath(os.path.join(os.getcwd(), path))
+
+def GetUavoBasedOnArgs(desc=None):
```
solidgoldbomb: Could (optionally) pass in a local argv here (with default of sys.argv) to allow callers first crack at parsing sys.argv or even generating their own args to the underlying telemetry decoder.

python/calibration/mag_calibration.py
mlyle: Unfortunately yes, though some of hte later ones aren't.  The top lines match the conventions in how ipython is invoked, while the code in the function assumes that numpy stuff will be in the numpy namespace.  Both are used (numpy.array, minimize, etc).

I'd be glad to clean this up in a subsequent pull.

Importing * from numpy is kinda bad, because htere's a big polluted namespace in there.  Most numpy code i've seen imports numpy as np to balance the namespace issues vs. the wordiness.

solidgoldbomb: Agreed, we should be doing ```import numpy as np``` and fixing up all references to it.  Ok to defer to later PR of course.

python/taulabs/telemetry.py
mlyle: I thought about this.  argparse kinda assumes it's doing it all or doing none-- there's not an easy way if a parent called it to get the stuff that didn't match args (unless I missed it).   If they really care, they can instantiate FileTelemetry directly, I thought.

python/taulabs/telemetry.py
mlyle: Yup.  Next pass definitely has a new proper documentation pass with pydoc.

python/calibration/mag_calibration.py
solidgoldbomb: We should set a precedent that all of these helper scripts accept the ```uavo_list``` as an argument.  That way, the shell can still dump uavo_list into the namespace so it's easy to work with interactively but then require the helper script to be called like ```mag_calibration(uavo_list)```.  The ```main``` can then be responsible for constructing ```uavo_list``` and passing it as a parameter to the ```mag_calibration``` function.

solidgoldbomb:
I really like the way you've collapsed the log file processing and telemetry streams into the same abstract class.  Nice.

python/minimal_connection.py
solidgoldbomb: Up to you.  It's nice to have an example user of the module but even examples should have debug/dead code removed from them.

solidgoldbomb:
Finished reviewing.  I would prefer to have something better than the ```readbyargs.py``` import side-effect mechanism before merging this.  Are you willing to take that on in this PR @mlyle?

mlyle:
@solidgoldbomb I'm willing to do whatever, but anything I do will be wordier than that.  My objective with the readbyargs stuff was to get something that would be as close as possible to the original ipython behavior with a minimal set of cruft needed to be appended to the end.

Let me know what you'd prefer instead.

mlyle:
(if you want, i can just remove it and replace it with its contents everywhere it's imported).

solidgoldbomb:
My vote would be what I wrote here: https://github.com/TauLabs/TauLabs/pull/1601#discussion_r30545185

mlyle:
That's like what I had before.  I can unwind later and submit that way.

mlyle:
@solidgoldbomb -- from IRC-- I don't really think they're making things worse.  It's a choice of two evils, between replicating 3-4 lines of boilerplate code in (potentially) a lot of places or by abusing a module import to have some side effects.  I originally had it close to the other way, and if that's preferred I'm fine with putting it back.. just need to get up the courage to do so :P

mlyle:
Done.

mlyle:
Yay, thanks for the merge. :)

PR #1609, opened 2015-05-14
stab: Compute desired pitch/roll properly.

mlyle:
Changed in horizon to multiply by PitchMax/RollMax instead of keeping the
desired body angle in the range from -1 to 1 deg.

Changed in attitude, horizon, weakleveling to not limit output range based on
trim settings.


SystemsGuy:
Bit of a windy day for testing, but really seems to improve Horizon.  I got maybe 20 minutes flying in - nothing unexpected.  Horizon flys great, and no changes to MWRate or Rate.  I flew a bit of attitude as well, and it feels a bit different, but I rarely fly it in as much wind as I had today.

peabody124:
Ok. This fixes what was a bug and is more consistent with what we intended. Since SystemsGuy had success let's give it a shot.

PR #1615, opened 2015-05-15
OneShot: explicitly set the Sync mode and Resolution for Timers

peabody124:
The previous code had a few issues. We could not independently set the resolution
and the synchronization (to sensors) properties of channels. In addition, the settings
were per channel instead of per timer as the hardware specifies. This makes it consistent
with the update frequency properties of timers.

Support for OneShot configuration is also added to the setup wizard.


peabody124:
I have checked using the wizard to go back and forth between oneshot/normal mode, no issues. Also weird combinations of things seem to work:
screen shot 2015-05-15 at 11 42 23 am
screen shot 2015-05-15 at 11 06 02 am
screen shot 2015-05-15 at 11 05 09 am

peabody124:
Updated GCS
screen shot 2015-05-15 at 12 49 10 pm

fujin:
@peabody124 in the wizard SyncPWM is referred to as PWMSync too.

fujin:
I integrated and tested hpwm into my long-running brain dev branch (chibiosrt, ...), manually configured, reset, ran the wizard, logged all the consistent naming issues I saw. Wizard outputs are great: "oneshot125(TM)" perfectly usable! I took some screenshots on my workstation of the wizard. Will run it again with the latest consistency fixes and re-screenshot everything.

Flight footage here: https://www.youtube.com/watch?v=XFcqJ-U81FU

:+1: :100: from me.

flight/PiOS/STM32F10x/pios_servo.c
```diff
@@ -82,28 +87,36 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
 		TIM_Cmd(chan->timer, ENABLE);
 	}
 
-	/* Allocate memory */
-	output_timer_frequency_scaler = PIOS_malloc(servo_cfg->num_channels * sizeof(typeof(output_timer_frequency_scaler)));
-	// Check that memory was successfully allocated, and return if not
-	if (output_timer_frequency_scaler == NULL) {
+	output_channel_clk_rate = PIOS_malloc(servo_cfg->num_channels * sizeof(typeof(output_channel_clk_rate)));
```
ernieift: this occupies 40 bytes.

flight/PiOS/STM32F10x/pios_servo.c
```diff
@@ -82,28 +87,36 @@ int32_t PIOS_Servo_Init(const struct pios_servo_cfg * cfg)
...
+	if (output_channel_clk_rate == NULL) {
 		return -1;
 	}
-	memset(output_timer_frequency_scaler, 0, servo_cfg->num_channels * sizeof(typeof(output_timer_frequency_scaler)));
+	memset(output_channel_clk_rate, 0, servo_cfg->num_channels * sizeof(typeof(output_channel_clk_rate)));
+#if defined(PIOS_INCLUDE_HPWM)
+	/* Allocate memory for frequency table */
+	output_channel_frequency = PIOS_malloc(servo_cfg->num_channels * sizeof(typeof(output_channel_frequency)));
```
ernieift: and this 20 bytes.

flight/PiOS/STM32F10x/pios_servo.c
```diff
@@ -156,62 +173,84 @@ void PIOS_Servo_SetHz(const uint16_t * speeds, uint8_t banks)
...
+
+			// Configure this timer appropriately.
+			TIM_TimeBaseInit(chan->timer, &TIM_TimeBaseStructure);	
+
+			/* Configure frequency scaler for all channels that use the same timer */
+			for (uint8_t j=0; (j < servo_cfg->num_channels); j++) {
+				if (chan->timer == servo_cfg->channels[j].timer) {
+#if defined(PIOS_INCLUDE_HPWM)
+					/* save the frequency for these channels */
+					output_channel_frequency[j] = speeds[set];
```
ernieift: You save all these values for comparing. As predefined frequencies will use, we can represent the whole thing with the help of flags.

flight/PiOS/STM32F10x/pios_servo.c
```diff
@@ -239,44 +279,30 @@ void PIOS_Servo_OneShot_Set(uint8_t servo, uint16_t position)
...
-				if (PIOS_PERIPHERAL_APB1_CLOCK == PIOS_SYSCLK)
-					TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB1_CLOCK / OneShotFrequency) - 1;
-				else
-					TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB1_CLOCK*2 / OneShotFrequency) - 1;
-			}
-
+		/* Check for channels that are using synchronous output */
+		/* Look for a disabled timer using synchronous output */
+		if (!(chan->timer->CR1 & TIM_CR1_CEN) &&
+		     (output_channel_frequency[i] == 0)) {
```
ernieift: A double check. The timer is off, because the frequency was zero in PIOS_Servo_Set().

ernieift:
I like the elimination of the output_timer_frequency_scaler. But I would use the resolution enum and (update rate ==0) to generate a uint8 field. This could be used to switch in PIOS_Servo_Set(). The Compiler wouldn't generate a multiply calculation to get the position value. Because all frequencies are predefined, the compiler replaces the brackets by a simple factor.

SystemsGuy:
When using the Vehicle Setup Wizard, and select ESC type of "OneShot" the functionality of the "Output Calibration" window differs from "non-oneshot" mode in that the slider only has one step position - eg off or on.

Not sure if this is expected or not, but if so, the text should be updated?

fujin:
@peabody124 verified the SyncPWM fix went into the setupwiz.
2015-05-17-102505_1920x1080_scrot

fujin:
have been trying to help @SystemsGuy in IRC to no avail, cannot replicate on a fresh board.

fujin:
Took screenshots of the wizard process here: https://imgur.com/a/dyisw#0

SystemsGuy:
Update : it's the slider step values - if I click (as show on the video), it jumps from min to max.  If I drag the slider, it works as expected...

peabody124:
@ernieift i like the idea of storing hte enum natively, but we need the actual divisor to use. That is a function of both the timer configuration _and_ the desired timer resolution. I don't see how to get it from just the enum value without adding another layer of logic into the set method.

ernieift:
@peabody124 I'm in the next week on vacation. I hope that I'll be there somewhere in the net. In my most recent changes I have the output_timer_frequency_scaler[] taken it. If this is now free, it could carry the enums and flags. PIOS_Servo_Set() could handle it like this:
```

if (output_timer_config[servo] & OneShotFlag) {
 TIM_Cmd(chan->timer, DISABLE);
}

uint32_t compare;

switch (output_timer_config[servo] & 0xf) {
 case PWM_MODE_1US:
  compare = position;
  break;
 case PWM_MODE_83ns:
  compare = position * HiresFrequency/1000000;
  break;


// etc for other modes...
}
....

HiresFrequency/1000000 is solved by the compiler then. What do you think?

**Comment by ernieift** [comment]
or better:
```
if (output_timer_config[servo] & OneShotFlag) {
 TIM_Cmd(chan->timer, DISABLE);
}

switch (output_timer_config[servo] & 0xf) {
 case PWM_MODE_1US:
  // do nothing the position is already in µs.
  break;
  // recalculate the position value to hires base.
 case PWM_MODE_83ns:
  position = position * HiresFrequency/1000000;
  break;
// etc for other modes...
}

// and later
TIM_SetCompareX(chan->timer, position);
```

**Comment by peabody124** [comment]
Ok @ernieift I updated the way it looks up the conversion to similar to what you suggested. The case statement only returns the conversion from us to counts so in the case of 1us mode it has one more multiply than necessary. It's an integer multiply and I think worth it for code symmetry. Also, most people will probably start using high res anyway :)

**Comment by ernieift** [comment]
Looks good ;). I agree, most people use serial or ppm receivers, so there are less problems with timers.
And you still want to store all the speed values only for (speed==0) checks? As I wrote before, in PIOS_Servo_Update() there could be a problem when the frequency changes between Servo_Set() and Update(), the timer could be disabled in a worst case scenario.
I may be looking as a hack, but why not use a single bit in output_channel_resolution[] as indicator?

**Comment by peabody124** [comment]
@ernieift i made an enum to indicate the sync rather than store the full speed. I know I could save 10 bytes by using bitflags but properly labeled enums just make me so much more happy and make the code more maintainable (IMO) that I'll sacrifice those 10 ... 

Everything except CC3D has a fair bit of ram for the intended modules.

File: shared/uavobjectdefinition/actuatorsettings.xml [hunk 5]
```diff
@@ -1,12 +1,12 @@
...
     <object name="ActuatorSettings" singleinstance="true" settings="true">
         <description>Settings for the @ref ActuatorModule that controls the channel assignments for the mixer based on AircraftType</description>
-        <field name="ChannelUpdateFreq" units="Hz" type="uint16" elements="6" defaultvalue="50"/>
-        <field name="ChannelMax" units="us,us/12" type="uint16" elements="10" defaultvalue="2000"/>
-        <field name="ChannelNeutral" units="us,us/12" type="uint16" elements="10" defaultvalue="1000"/>
-        <field name="ChannelMin" units="us,us/12" type="uint16" elements="10" defaultvalue="1000"/>
-        <field name="ChannelType" units="" type="enum" elements="10" options="PWM,OneShot,MK,ASTEC4,PWM Alarm Buzzer,Arming led,Info led" defaultvalue="PWM"/>
-        <field name="ChannelAddr" units="" type="uint8" elements="10" defaultvalue="0,1,2,3,4,5,6,7,8,9"/>
+        <field name="TimerUpdateFreq" units="Hz" type="uint16" elements="6" defaultvalue="50"/>
+        <field name="TimerPwmResolution" units="" type="enum" elements="6" options="1us,80ns" defaultvalue="1us"/>
```
BrainFPV: Would it make sense to make 400Hz / 80ns the default? Most users use multirotors, where the 50Hz will give very poor performance.

peabody124: The setup wizard gives those defaults, so I'm not sure it's critical. The reason for the 50Hz issue was old analog servos could burn out. I'm not sure that is a major concern anymore, though. I'll defer to group opinion.

ernieift:
@peabody124 i can live with the 10 extra bytes. If there is a race condition between Servo_SetMode() and Servo_Update() we will see...
I'd prefer always enabling the disabled timer. There is nothing against this behaviour IMO.

flight/PiOS/STM32F10x/pios_servo.c
```diff
@@ -239,44 +293,30 @@ void PIOS_Servo_OneShot_Set(uint8_t servo, uint16_t position)
...
-				else
-					TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB1_CLOCK*2 / OneShotFrequency) - 1;
-			}
-
+		/* Check for channels that are using synchronous output */
+		/* Look for a disabled timer using synchronous output */
+		if (!(chan->timer->CR1 & TIM_CR1_CEN) &&
+		     (output_channel_mode[i] == SYNC_PWM_TRUE)) {
 			/* enable it again and reinitialize it */
 			TIM_Cmd(chan->timer, ENABLE);
```
peabody124: @ernieift when you say "enabling the timer" do you mean move this line outside the conditional? Before or after?

shared/uavobjectdefinition/actuatorsettings.xml
jihlein: I'd leave default at 50 Hz, that's the safe option.  I still have some tricopter's with 50 Hz analog servo's.  I'd rather have to change setup parameters than a burnt out servo FWIW........

peabody124:
@ernieift regarding races, the real race is between when we actually configure the timer and set either of the flags. The order of the flags being set is much less than that.

Since you've pointed it out I feel a need to address it. Perhaps we should disable the timer in set mode and set all compare registers to 0? I'm worried about unintended side effects though.

shared/uavobjectdefinition/actuatorsettings.xml
peabody124: Works for me. Burning less peoples stuff is better.

flight/PiOS/STM32F10x/pios_servo.c
```diff
@@ -239,44 +293,30 @@ void PIOS_Servo_OneShot_Set(uint8_t servo, uint16_t position)
...
-				if (PIOS_PERIPHERAL_APB1_CLOCK == PIOS_SYSCLK)
-					TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB1_CLOCK / OneShotFrequency) - 1;
-				else
-					TIM_TimeBaseStructure.TIM_Prescaler = (PIOS_PERIPHERAL_APB1_CLOCK*2 / OneShotFrequency) - 1;
-			}
-
+		/* Check for channels that are using synchronous output */
+		/* Look for a disabled timer using synchronous output */
+		if (!(chan->timer->CR1 & TIM_CR1_CEN) &&
+		     (output_channel_mode[i] == SYNC_PWM_TRUE)) {
```
ernieift: i mean this line.

peabody124: I thought we needed to verify it was disabled here to prevent pulse lengthening.

ernieift: The condition is only one time true.

ernieift: What i exactly mean is this: "(output_channel_mode[i] == SYNC_PWM_TRUE)".

peabody124: Ah ok. I'm ok with that.

peabody124:
Any more issues?

ernieift:
No more issues. Looks good to me. Just one possible problem crossed my mind. what happens when there will be a PWM_MODE_1MS? But we can care about this when it comes reality. For all finer resolutions it is good as it is.
Sadly i can't test this right now. Next week i'm back, so i suggest testing on all supported targets to be sure. I don't expect any problems.
I also would use 50Hz and 1MHz as default. Not all models are multirotors.

peabody124:
Yea really slow modes might require larger refactoring in the code. At least the configuration should be able to handle this fairly easily.

peabody124:
Flight tested on Sparky2 and scoped on Sparky, seems good to merge.

peabody124:
Scoped on CC3D and working (notice the oneshot channels running at 666 like the settings)
screen shot 2015-05-21 at 6 21 52 pm

peabody124:
@elafargue want to merge this?

SystemsGuy:
Gents - I still think the issue with the slider performance under windows should be fixed.  The "step size" when you click on the bar means that it jumps from min to max in one "click". If you "slide" it, it works as expected in small steps.

https://www.youtube.com/watch?v=adA896Jdchs

peabody124:
@SystemsGuy check most recent commit. This should fix your issue.

SystemsGuy:
Looks great @peabody124 - steps work properly!

PR #1616, opened 2015-05-16
gps-ubx: retrieve accuracy values from messages.

mlyle:
Should get plumbed down to INS13 variances eventually...

### DIFF
--- a/flight/Modules/GPS/NMEA.c
+++ b/flight/Modules/GPS/NMEA.c
@@ -765,6 +765,10 @@ static bool nmeaProcessGPGSA(GPSPositionData * GpsData, bool* gpsDataUpdated, ch
 	// next field: VDOP
 	GpsData->VDOP = NMEA_real_to_float(param[17]);
 
+	// NMEA doesn't have this.  Assume we have a nominally 5m accurate
+	// receiver and scale accordingly.
+	GpsData->Accuracy = GpsData->PDOP * 5.0f;
+
 	return true;
 }
 
--- a/flight/Modules/GPS/UBX.c
+++ b/flight/Modules/GPS/UBX.c
@@ -215,6 +215,7 @@ static void parse_ubx_nav_sol (const struct UBX_NAV_SOL *sol, GPSPositionData *G
 {
 	if (check_msgtracker(sol->iTOW, SOL_RECEIVED)) {
 		GpsPosition->Satellites = sol->numSV;
+		GpsPosition->Accuracy = sol->pAcc / 100.0;
 
 		if (sol->flags & STATUS_FLAGS_GPSFIX_OK) {
 			switch (sol->gpsFix) {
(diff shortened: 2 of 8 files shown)


shared/uavobjectdefinition/gpsposition.xml
```diff
@@ -9,6 +9,7 @@
         <field name="Heading" units="degrees" type="float" elements="1"/>
         <field name="Groundspeed" units="m/s" type="float" elements="1"/>
         <field name="Satellites" units="" type="int8" elements="1"/>
+        <field name="Accuracy" units="" type="float" elements="1"/>
```
peabody124: no units?

peabody124: looks like cm from the notes

flight/Modules/GPS/NMEA.c
```diff
@@ -28,6 +28,8 @@
  * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
  */
 
+#include <math.h>
```
peabody124: this was needed to do multiplication?

flight/Modules/GPS/UBX.c
```diff
@@ -251,6 +252,7 @@ static void parse_ubx_nav_velned (const struct UBX_NAV_VELNED *velned, GPSPositi
 			GpsVelocity.North	= (float)velned->velN/100.0f;
 			GpsVelocity.East	= (float)velned->velE/100.0f;
 			GpsVelocity.Down	= (float)velned->velD/100.0f;
+			GpsVelocity.Accuracy	= (float)velned->sAcc/100.0f;
```
peabody124: according to ```struct UBX_NAV_SOL``` this starts at cm/s accuracy. you are trying to convert to 100µm/s ?

peabody124:
This will be really useful - to see if this drops when we see issues, even if PDOP or Sat doesn't.

flight/Modules/GPS/NMEA.c
mlyle: Woops.  I wavered with whether I should set it to NAN, so I picked up that line.  Willfix.

flight/Modules/GPS/UBX.c
mlyle: cm/s * 1m / 100cm = m/s, just like the other terms.

mlyle:
Rewrote history based on comments.

elafargue:
Ideally, the accuracy info should also be added to the GPS Display gadget

flight/Modules/GPS/UBX.c
peabody124: don't tell anyone i did that ... :-p

peabody124:
Looks good. Hopefully we get some logs with this info for study. I'll merge the firmware changes and we can have a separate issue for the GCS side.

flight/Modules/GPS/UBX.c
mlyle: i want forgiveness/the same immunity next time i do ;)

PR #1622, opened 2015-05-18
Sparky2: update kicad schematic following fixes in converter

peabody124:
This fixes the power port symbols.
screen shot 2015-05-18 at 11 58 39 am


peabody124:
Thanks for the upstream fix in the converter https://github.com/thesourcerer8/altium2kicad/issues/4

PR #1623, opened 2015-05-18
ENH: Native FrSky PWM RSSI Support

BrainFPV:
FrSky receivers, e.g., model D4R-II, output a PWM RSSI frequency signal with a frequency of 110kHz, which is too high to be captured with normal PWM inputs in Tau Labs. This adds native support for such PWM signals by using two capture-compare channels. The disadvantage is that an entire timer is used, which limits the number of available PWM outputs.

I only have the Brain FC to test this, an example configuration for this target is given below.

[Example configuration for Brain](https://github.com/BrainFPV/TauLabs/blob/brain_frsky_rssi/flight/targets/brain/board-info/board_hw_defs.c#L1828)

### DIFF
--- a/flight/PiOS/STM32F4xx/inc/pios_frsky_rssi.h
+++ b/flight/PiOS/STM32F4xx/inc/pios_frsky_rssi.h
@@ -1,0 +1,46 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS driver for PWM RSSI input
+ * @brief Input RSSI PWM
+ * @{
+ *
+ * @file       pios_frsky_rssi.h
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2015.
+ * @see        The GNU Public License (GPL) Version 3
+ *
+ *****************************************************************************/
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ */
+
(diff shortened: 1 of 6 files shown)


fujin:
Yeah!!!!

flight/PiOS/STM32F4xx/inc/pios_frsky_rssi.h
```diff
@@ -0,0 +1,46 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS driver for PWM RSSI input
+ * @brief Input RSSI PWM
+ * @{
+ *
+ * @file       pios_frsky_rssi.h
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2015.
```
peabody124: Copyright just 2015, right?

This file should probably go in pios_common. I don't think there is a problem having the ```TIM_ICInitTypeDef``` in there is other pios driver include files have it there.

Also strictly to follow our normal conveniton this should be two files. a _priv.h variant that has the cfg/Init method and then the public API component that has the _Get() method.

flight/PiOS/STM32F4xx/pios_frsky_rssi.c
```diff
@@ -0,0 +1,125 @@
...
+ * @{
+ * @addtogroup PIOS driver for FrSky RSSI input
+ * @brief Driver for FrSky RSSI, which is a PWM signal with a frequency of
+ * about 110kHz. The driver uses two capture-compare channels of the timer
+ * to measure the PWM pulse width. In contrast to the normal PWM driver, this
+ * procedure does not require any interrupts.
+ * @{
+ *
+ * @file       pios_frsky_rssi.c
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2015.
```
peabody124: Ditto copyright

flight/PiOS/STM32F4xx/pios_frsky_rssi.c
```diff
@@ -0,0 +1,125 @@
...
+
+#if defined(PIOS_INCLUDE_FRSKY_RSSI)
+
+#include "pios.h"
+#include <pios_stm32.h>
+#include "stm32f4xx_tim.h"
+#include "pios_tim_priv.h"
+
+#if defined(PIOS_INCLUDE_FREERTOS)
+#include "FreeRTOS.h"
```
peabody124: Is this actually needed? I don't see any FreeRTOS primitives  below.

peabody124:
Awesome work!

guilhermito:
References to RssiPwmPeriod should be dropped from code and UAV object since it was an attempt to handle the high frequency PWM signals which this PR takes care of.

BrainFPV:
@peabody124 @guilhermito thanks for the review. Comments have been addressed.

PR #1625, opened 2015-05-19
GCS - Channel bank display algorithm fix.

jihlein:
While working on my Naze32Pro target, I discovered the channel bank definition 1<<2<<9 would display in the GCS as 1-9, rather than 1-2,9.

This pull request fixes the algorithm that creates the display string.  For my Naze32Pro target, I now get the desired 1-2,9 channel bank display on the GCS.  I checked Quanton and Sparky for correctness, and they display correctly also.

This change is purely cosmetic, dealing with the GCS display strings only.

### DIFF
--- a/ground/gcs/src/plugins/coreplugin/iboardtype.cpp
+++ b/ground/gcs/src/plugins/coreplugin/iboardtype.cpp
@@ -11,64 +11,45 @@ QStringList IBoardType::queryChannelBanks()
 {
     QStringList banksStringList = QStringList();
 
-    foreach(QVector<qint32> channelBank, channelBanks) 
-	{
-        quint8 sequential = false;
+    foreach(QVector<qint32> channelBank, channelBanks) {
+        bool sequential = false;
         QString banksString;
 
-		for (int i = 0; i < channelBank.size(); i++) 
-		{
-			if (i == 0)															 // First channel in bank
-			{
+		for (int i = 0; i < channelBank.size(); i++) {
+			if (i == 0)	{														 // First channel in bank
 				banksString.append(QString("%1").arg(channelBank[i]));
 				
-				if (channelBank.size() > 1)
-				{
-					if ((channelBank[i] + 1) == channelBank[i + 1])
-					{
-						banksString.append(QString("-"));							 // Next channel is sequential, add '-' to the string
+				if (channelBank.size() > 1)	{									 // If more than 1 channel in bank, add separator
+					if ((channelBank[i] + 1) == channelBank[i + 1])	{
+						banksString.append(QString("-"));						 // Next channel is sequential, add '-' to the string
 						sequential = true;
 					}
-					else
-					{
-						banksString.append(QString(","));							 // Next channel is not sequential, add ',' to the string
+					else {
+						banksString.append(QString(","));						 // Next channel is not sequential, add ',' to the string
 						sequential = false;
 					}
 				}
 			}
-					
-			///////////////////////////
-				
 			else if (i == channelBank.size() - 1)								 // Last channel in bank
 				banksString.append(QString("%1").arg(channelBank[i]));
-					
-			///////////////////////////
-				
-			else																 // Mid channel(s) in bank
-			{
-				if (sequential)
-				{
-					if ((channelBank[i] + 1) == channelBank[i + 1])				 // Still sequential channels, nothing to add to string
-					{
+			else {																 // Mid channel(s) in bank
+				if (sequential)	{
+					if ((channelBank[i] + 1) == channelBank[i + 1]) {			 // Still sequential channels, nothing to add to string
 						sequential = true;
 					}
-					else
-					{
+					else {
 						banksString.append(QString("%1,").arg(channelBank[i]));  // End of sequence, add channel and ',' to string
 						sequential = false;
 					}
 				}
-				else
-				{
+				else {
 					banksString.append(QString("%1").arg(channelBank[i]));
 
-					if ((channelBank[i] + 1) == channelBank[i + 1])
-					{
+					if ((channelBank[i] + 1) == channelBank[i + 1])	{
 						banksString.append(QString("-"));						 // Next channel is sequential, add '-' to the string
 						sequential = true;
 					}
-					else
-					{
+					else {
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
 {
     QStringList banksStringList = QStringList();
 
-    foreach(QVector<qint32> channelBank, channelBanks) {
-        qint32 tmpChannel = -1;
-        qint32 lastWrittenChannel = -1;
+    foreach(QVector<qint32> channelBank, channelBanks) 
+	{
```
guilhermito: "{" should be at the end of the last line as it was previously

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
+        quint8 sequential = false;
```
guilhermito: should be bool

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
-                    banksString.append(QString("-%1,%2").arg(tmpChannel).arg(channel));
-                } else {
-                    banksString.append(QString("-%1").arg(channel));
-                }
-                lastWrittenChannel = channel;
-            }
-            tmpChannel = channel;
-        }
+		for (int i = 0; i < channelBank.size(); i++) 
+		{
```
guilhermito: "{" moved above

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
+			if (i == 0)															 // First channel in bank
+			{
```
guilhermito: "{" moved above

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
peabody124: yea just to elaborate we try and mostly follow the linux kernel style convention. For function definition { start new lines. For everything else at the end of the line.

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
+				banksString.append(QString("%1").arg(channelBank[i]));
+				
+				if (channelBank.size() > 1)
+				{
```
guilhermito: "{" moved above will mute now, but all other "if" and "else" braces should be moved to the same line

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
+					}
+					else
+					{
+						banksString.append(QString(","));							 // Next channel is not sequential, add ',' to the string
+						sequential = false;
+					}
+				}
+			}
+					
+			///////////////////////////
```
guilhermito: remove comment and empty space

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
+			}
+		}
+		
+		// If there are no elements in the bank, print a hyphen
         if (banksString.isEmpty())
             banksString.append("-");
-
-        // Add string to list
+		
+		// Add string to list
```
guilhermito: Indentation seems of

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
```diff
@@ -11,41 +11,76 @@ QStringList IBoardType::queryChannelBanks()
...
-                else if (i < channelBank.size()-1) { // If this isn't the last element, add a comma
-                    banksString.append(QString("-%1,%2").arg(tmpChannel).arg(channel));
-                } else {
-                    banksString.append(QString("-%1").arg(channel));
-                }
-                lastWrittenChannel = channel;
-            }
-            tmpChannel = channel;
-        }
+		for (int i = 0; i < channelBank.size(); i++)
```
guilhermito: Prefer ++i to i++

ground/gcs/src/plugins/coreplugin/iboardtype.cpp
jihlein: Made the formatting changes.  I did not change the i++ to ++i, as that was the way it was originally and I did not change it.  If truly necessary, let me know and I'll change it.

jihlein: Did some research on i++ vs ++i.  Seems either will work in a for loop, and ++i is the preferred syntax.  I'll change this and add another commit tonight.

jihlein:
Made the formatting changes. I did not change the i++ to ++i, as that was the way it was originally. If truly necessary, let me know and I'll change it.

Update:

Did some research on i++ vs ++i. Seems either will work in a for loop, and ++i is the preferred syntax. I'll change this and add another commit tonight.

peabody124:
I didn't know there was a preferred way. Cool. Ok will wait for fix.

tracernz:
Likewise, I hadn't thought about that before. When you think about it, it makes sense though. When post-incrementing the compiler has to keep the original value and the post-incremented value, probably using an extra register? I imagine a modern compiler would optimise this though, at least in the case of a for loop.

mlyle:
Postincrement vs. preincrement will not make a difference on any compiler from the last couple decades, and not many before.  It's purely a question of style.

jihlein:
I prefer post increment, to me it reads easier.  

That said, I just changed it due to the code review request.

Ready to merge unless we want to change it back.  :)

Issue #1621, opened 2015-05-18
extended GPIO functionality


Trex4Git:
Hi, 
as i want to use a active buzzer in taulabs and the quanton already has an port for this, i thought about using it with a new GPIO function in picoC (that i want to write soon). For quanton this pin is already initialized in pios_board.c. 

But i wanted to implement it not only for me, but in a more general way.
In my first thoughts i wanted to do a "security check" (if the pin on this port is really configured as a GPIO output) in the picoC function. But from my understanding of the taulabs code structure such driver things should be better done in pios.
But i have seen that there are already structures for GPIO in board-info/pios_board.h that could be used to define GPIO pins, and then no further security check in the dynamic picoC usage would be needed.

So i think the best way would be using the GPIO functionality which is already in taulabs, with defining the GPIO Pins in pios_board.h for the desired target,  initializing the pins in pios_board.c, and also defining PIOS_INCLUDE_GPIO in pios_config.h to use the already implemented hardware abstraction for GPIO operations from pios_gpio.c.
I want to do a commit for the changes that are needed for the quanton, and a separate commit/pull request for the picoC GPIO code.

As this will be my first coding for taulabs, i wanted to be sure if it is the right way I, or if there are any problems with taulabs doing it like that.

Trex4Git:
As i have read that you  should split pull request into small requests with well defined functionality, i did some quick coding in separate branches. Here are the firsts test commits:
One for the PicoC changes: https://github.com/Trex4Git/TauLabs/commit/969b4f41c407a5828ec4f91ce44eed3864555e99
And one for the Quanton changes: https://github.com/Trex4Git/TauLabs/commit/b1c9fea4fc8f9839bda0b37a78c3419696094172

For testing i merged them together in a other branch, and it seems to compile. But as i just have setup the build environment, i have to do further testing if i really works.

But i would be nice if some expeienced taulabs developer might have a look at it, and can tell me if this could me merged in principal, or if there will be some problems.

Trex4Git:
As i made a pull request, this can be closed.

PR #1628, opened 2015-05-19
extended GPIO output functionality, PicoC and PIOS_GPIO

Trex4Git:
As already posted in this issue https://github.com/TauLabs/TauLabs/issues/1621 i want to make use of the Quanton buzzer output.
I have not seen any functions in taulabs to do this, so i started a implementation for general GPIO output in PicoC.

At first i thought about implementing the complete configuration, initialization,... in PicoC. But from my understanding this driver stuff should be better in PIOS.
So i implemented it with the PIOS GPIO functionality which is already there.
* defining PIOS_INCLUDE_GPIO in pios_config.h to use the already implemented hardware abstraction for GPIO operations from pios_gpio.c
* defining the GPIO Ports/Pins/.. in pios_board.h as arrays for the desired target (like it was already prepared for the coptercontrol target)
* initializing the pins in pios_board.c
	* for the buzzer pin of the quanton it is already done
	* when it should be done for other pins, or on other targets the PIOS_GPIO_Enable(uint8_t Pin) function can be used

The PicoC function uses then the already defined and initialized GPIO output pins with the PIOS_GPIO functions, and has no direct hardware access, so no additional security check should be necessary. The developer should have tested if it works, and the user cant make any mistakes.

A disadvantage of this implementation is, that the GPIO must be configured before compiling, and can not be changed dynamically by the user.
But as on most targets there is only an easy access to already used pins, and if a user wants to use one of them, there might be problems with the rest of the system. I don't know if it is worth to implement security checks for this, as it would be complicated to define what is allowed. 
Only on the flyingF3/4 targets are easy accessible pins, which are not used with taulabs, and therefore there cant be a conflict, when they are defined during runtime as GPIO output pins.

Maybe an additional implementation in PicoC for advanced users makes sense. There would be only basic security checks, and  the user could define all Pins as GPIO output and use them. But then the user is responsible for all errors that might occur.

Also i implemented no GPIO input functionality yet, as i have seen no interest in it. But if needed i can implement it similar to GPIO output, before it might get merged.

First test showed that it works, but if there are principal problems, or some additional functionality (dynamic gpio initialization, input) should be implemented some feedback would be nice before implementing more functionality makes sense.

Also if i better should split the commits in different pull requests or not. At first i wanted to do it, but  ernieift mentioned i should not be necessary.


Trex4Git:
Additional tests with different PicoC scripts on quanton, and it works.
No problems noticed so far.

For example:
```
#include "system.h"
TestValSet(1);
sync(0);

void beep3(unsigned int b1,unsigned int p1, unsigned int b2, unsigned int p2, unsigned int b3)
{
	short test = 0;

	test = GPIOWrite(0,1);
	sync(b1);
	test = GPIOWrite(0,0);
	sync(p1);
	
	test = GPIOWrite(0,1);
	sync(b2);
	test = GPIOWrite(0,0);
	sync(p2);

	test = GPIOWrite(0,1);
	sync(b3);
	test = GPIOWrite(0,0);
	TestValSet(1);
}


while(TestValGet() > 0)
{
	if(TestValGet() == 2)
	{
		beep3(400,150,150,150,150);
	}
	else if(TestValGet() == 3)
	{
		beep3(150,150,150,150,400);
	}
	else
	{ sync(1000); }
}
TestValSet(99);
```

peabody124:
jenkins: test this please

peabody124:
jenkins: add to whitelist

peabody124:
@ernieieft since this is touching on picoC could you take a look at it?

peabody124:
@ernieift  since this is touching on picoC could you take a look at it?

ernieift:
@Trex4Git i can test this next week. What i can say for now is, i'd remove the return value. Why? all called PIOS functions are from type void. The only useful information you can get here is wether you called it with the right parameters. You simply have to check if the pin number is ok. If the wanted subfunction is available? Who cares?
PicoC generates much overhead for variable definitions, so keep it low. I suggest making the function type void.

Trex4Git:
Thanks for your reviews.
As this is my first pull request on github ever, and my first coding for taulabs i am glad for your support.

I wasn't aware of the overhead, sorry. So i will make the function type void.

As it seems, that you can do a lot problematic stuff with PicoC if you don't know what you do, PicoC should better be used experienced users. To much unnecessary "security" code in PicoC itself would bloat just the code.

mlyle:
This looks awesome!

Have you considered letting you set the gpio hi-z (I know there's no corresponding read call yet, but this is part of what you'd need) and manage the pull-up from this call?

Trex4Git:
I have removed the return value and tested again, still works ;)

@mlyle: i am not sure what you mean?
a dynamic change of the pin mode from PicoC? Or a special configuration for input pins?

At the current implementation there is now dynamic change of pin configuration, everything has to be setup for each target in the code. For the input i thought about doing it the same way.
So if the pin should be input/output_pp/output_od and if there should be pull up or push up resistor activated, is configured during the initialization in the code.

I am open for changes, but in need to know what needs to be implemented.

ernieift:
Yeah, much cleaner. When you warp PIOS_GPIO_Enable() also, it's almost a complete GPIO functionality. Just clean up your the comments above.
Have you got a discoveryf4? So you could add some pins for this target.
For input functionality we have to extent the pios_gpio.c in a further step. As i said, i'll try this next week.

Trex4Git:
Implementing a call to PIOS_GPIO_Enable() in PicoC would be no problem, but wouldn't it be useless to init pin dynamically from PicoC, when in the current implementation the pins have to be in the confgured array, to use the functionality?

I have  discoveryf4 but not worked with that together with taulabs yet. But if someone could tell me which pins should be configured, it would be a easy task to implement it.

I am also already working on input functionality at the moment, will post a link to it soon. I am not sure if its a good idead to include it in this pull request, or open a nother one for it.
Also not shure if its worth to implement input functionality to PIOS , also for F1 (already some time gone since i worked with this µC) and F3 (never coded for this µC) as PicoC only supports F4 targets. But i think its good to keep PIOS functionality consistent across the different platforms.

Trex4Git:
Another target of interest for this might be also the Sparky 2, with its  buffered outputs that might also be used to  drive a buzzer or led.

Has someone a  good idea how to implement dynamic gpio definitions during runtime. For example to use a PWM Input/Output Pin? The PWM input pins can now be configured in different configurations as normal PWM Input, PPM Input, ADC In, PWM Output. But i think adding GPIO to this would be more complicated, than the stuff i am doing here.

Trex4Git:
For a more dynamic usage of GPIO Port/Pins, i think a bigger change to the PIOS_GPIO functionality is necessary. Instead of using the predefined Port/Pin in the arrays for read/write functions,  more general read/write functions are needed, that accept individual Port/Pin combinations as input.
In the first step i think it might be better to add additional functions to PIOS and not replace the old ones for backward compability. Also having both functions in PIOS would allow more secure access for a standard user with the old functions, and advanced access for experienced users.
 
This new functions would have less security. 
So i am thinking about adding functions that can check if the Port/Pin is configured as GPIO Output (write function) or as GPIO Input (read function).
But i have not seen a easy way to check if the Port/Pin in GPIO mode is already used by another module. So its up to the user to proof this.This check could be run in every read/write access, or maybe just once before using read/write functionality.
But as there are still some possible security issues i don't know if its worth to implement it, or just inform the user (in the wiki) to be responsible and really take care for possibles problems.
Together with enabling the init functionality you can have dynamic GPIO functionality in PicoC. 

I don't know if such a "dirty" implementation would be ok, and worth to do it.

flight/Modules/PicoC/picoc_library.c
```diff
@@ -620,6 +620,31 @@ void SystemI2CWrite(struct ParseState *Parser, struct Value *ReturnValue, struct
 }
 #endif
 
+#ifdef PIOS_INCLUDE_GPIO
+/* void int GPIOWrite(unsigned int, unsigned int): Writes/sets a general purpose output pin */
```
ernieift: i meant these comments. GPIOWrite isn't int type any more...

ernieift:
@Trex4Git, GPIO should be implemented for really free pins only. The Sparky2 led drivers can be easily programmed via accessory channels from picoC.
Did you tested your buzzer code with an active 5V buzzer? I guess, that is exactly the function the output was designed for. There are many usages like triggering a camera shutter at a waypoint etc. 
But GPIO_IN needs a little more testing. The battery input on Quanton may work in high impedance configuration and you have to desolder R2 or bridge R3 to make it functional. Not the best way for the most. All pwm inputs can already be read with picoC. So we can lower this priority. As you have already pointed out, it is more complicated and can cause conflicts with other modules/hardware. So why not add the battery pin to the output list?

Trex4Git:
Yes, tested my buzzer code with a 5v active buzzer on my PDB, because i also thought that was the function it was designed for.

As the battery input on Quanton has a voltage divider, working as input wont work good, and output is also not optimal i think. 
I just don`t changed the initialization for it, and wrote the input functionality very quick to show the idea (knowing that probably it will  not work on quanton). And i agree that some testing has to be done before thinking of merging it. I just wanted to discuss the principal functionality, and if it should be a part of this pull request, or a separate pull request in the future.
A additional bugfix is also already in this input branch. 

I agree that GPIO should be only used for really free pins. But for example on my quanton i use the serial SBUS as receiver input, so in GCS the RcvrPort could be "dynamically" (without recompiling) set to "Disabled". In  PicoC a user should can initiallize and use these pins as GPIO.
In future it my be better to have for the RcvrPort additional options, where not all Pins are used, for example only the ADCs used, and the rest is free. Or have special GPIO options in GCS for PWM input and output pins. 
But all this wont work with the predefined arrays of GPIO Pins.

ernieift:
Just tested GPIO_Read() on PWM input 2. These pins are pulled to high by default. You can simply switch them to ground otherwise you read 1. Just add these pins to the GPIO_PORTS/PINS. You don't need PIOS_GPIO_IN_xxx. Only something like this:
```
//-------------------------
// GPIO
//-------------------------
#define PIOS_GPIO_PORTS				{ GPIOA, GPIOC }
#define PIOS_GPIO_PINS				{ GPIO_Pin_4, GPIO_Pin_6 }
#define PIOS_GPIO_NUM				2
```
It works the way you implemented PIOS_GPIO_Read(). No need to call PIOS_GPIO_Enable() which would make it dangerous. This way you can easily build an robot which avoids obstacles.

Trex4Git:
Thank you for your testing, nice to hear that it works ;) 
Did you change from  PullUp to PullDown only for testing?  I think it might break the PWM Input functionality if it would be changed permanently.
I do not need the Input function for myself, just thought it would not be much work to implement it, and than have a complete GPIO functionality. 
So i am still unsure how people might use it, and what is the best configuration that already maybe enables some GPIO Pins, and do not break other functionality.

Putting all the definitions into GPIO_PORTS/PINS reduces overhead, but the user must know which pin could be used as input or output. The user can not do dangerous things if he uses it in a wrong way, but it will not work as he expects it to do.
And there are other things  that can be done with PicoC, and would cause more problems. For example easily create 100% CPU load ;)

Trex4Git:
Here an example for F4 PIOS how i thought to implement the advanced GPIO functionality for experienced users if wrapped for use in PicoC. If it is used for example with disabled RcvrPorts Pins, or disbaled UART Pins, it should work without security problems. But when used with a wrong configuration it could cause a lot of problems.

```
static GPIO_TypeDef *PIOS_GPIO_Port_ALL[] = {GPIOA,GPIOB,GPIOC,GPIOD,GPIOE,GPIOF,GPIOG,GPIOH,GPIOI};
static uint32_t *PIOS_GPIO_Pin_ALL[]      = {GPIO_Pin_0,GPIO_Pin_1,GPIO_Pin_2,GPIO_Pin_3,GPIO_Pin_4,GPIO_Pin_5,GPIO_Pin_6,GPIO_Pin_7,GPIO_Pin_8,GPIO_Pin_9,GPIO_Pin_10,GPIO_Pin_11,GPIO_Pin_12,GPIO_Pin_13,GPIO_Pin_14,GPIO_Pin_15};


/**
* Enable a GPIO Pin
* \param[in] Mode: GPIO Mode: 0=GPIO_Mode_IN / 1=GPIO_Mode_OUT
* \param[in] Port: GPIO Port: 0=GPIOA.....8=GPIOI
* \param[in] Pin: GPIO Pin: 0=GPIO_Pin_0,.....15=GPIO_Pin_15
*/
void PIOS_GPIO_EnableX(uint8_t Mode, uint8_t Port, uint8_t Pin)
{
	GPIOMode_TypeDef gpio_mode;

	if(Port<9 && Pin<16)
	{
		if(Mode == 0) {
			gpio_mode = GPIO_Mode_IN;
		}
		else if(Mode == 1) {
			gpio_mode = GPIO_Mode_OUT;
		}
		else {
			return;
		}
		GPIO_InitTypeDef GPIO_InitStructure;
		GPIO_InitStructure.GPIO_Mode = gpio_mode;
		GPIO_InitStructure.GPIO_OType = GPIO_OType_PP;
		GPIO_InitStructure.GPIO_Speed = GPIO_Speed_2MHz;
		GPIO_InitStructure.GPIO_PuPd = GPIO_PuPd_NOPULL;
		GPIO_InitStructure.GPIO_Pin = PIOS_GPIO_Pin_ALL[Pin];
		GPIO_Init(PIOS_GPIO_Port_ALL[Port], &GPIO_InitStructure);

		/* GPIO's Reset */
		GPIO_ResetBits(PIOS_GPIO_Port_ALL[Port], PIOS_GPIO_Pin_ALL[Pin]);

	}
}


/**
* Set value of a GPIO Pin
* \param[in] Mode: GPIO Mode: 0=reset_pin , 1=set_pin, 2=toggle_pin
* \param[in] Port: GPIO Port: 0=GPIOA.....8=GPIOI
* \param[in] Pin: GPIO Pin: 0=GPIO_Pin_0,.....15=GPIO_Pin_15
*/
void PIOS_GPIO_WriteX(uint8_t Mode, uint8_t Port, uint8_t Pin)
{
	if(Mode==2) {
		GPIO_ToggleBits(PIOS_GPIO_Port_ALL[Port],PIOS_GPIO_Pin_ALL[Pin]);
	}
	else if (Mode==0 || Mode==1) {
		GPIO_WriteBit(PIOS_GPIO_Port_ALL[Port],PIOS_GPIO_Pin_ALL[Pin], Mode);
	}
}


/**
* Read value of a GPIO Pin
* \param[in] Port: GPIO Port: 0=GPIOA.....8=GPIOI
* \param[in] Pin: GPIO Pin: 0=GPIO_Pin_0,.....15=GPIO_Pin_15
*/
+uint8_t PIOS_GPIO_Read(uint8_t Port, uint8_t Pin)
+{
+	return GPIO_ReadInputDataBit(PIOS_GPIO_Port_ALL[Port],PIOS_GPIO_Pin_ALL[Pin]]);
+}


```

ernieift:
Yeah, this causes many problems. I'd suggest not to implement the "Enable"-function. So picoC cannot change the AF-settings. I don't expect any problems with access to PWM inputs. These pins are configured as AF by default. The writes to the output register are ignored, but the reads are always functional independent from the pin configuration.
A 100% CPU load with picoC is simple to generate. Just remove all delays from the script and program a while(1) loop. That's why the task priority of picoC is so low.
BTW I haven't touched the pin config for the little test. On Quanton, GPIO_PuPd_UP is set by default for these pins. As i said, just add the PWM inputs to the GPIO list without an extra GPIO_INPUT list. So you can also read back the last written state of the buzzer. If someone needs more outputs, there is also i2c.
The best place for the GPIO list may be the wiki. Each target differs here.

Trex4Git:
Thanks again for your feedback.
Maybe at first a discussion in the forum would be better to evaluate the interest in the defined GPIO pins for different targets.  As i don't have a personal use for them, it would be difficult to evaluate what makes sense here.
Otherwise the ReadX / WriteX functions would also give access to pins which could be used securely (a list in the wiki with an overview of those pins may be also a good idea), but are not defined in the arrays for all the targets.

I don't know if there is really interest in it, and if its worth to open a possible security hole with the "EnableX"-function. But experienced users could use free IO's on there flight control without the need of changes in the "flight" code or additional hardware like a i2c IO expander.

I have no personal preference for either way. When doing this additional GPIO stuff after my initial commits I just thought about what might be useful.

 It is good to get some feedback from you guys.

Trex4Git:
If the "EnableX" functionality should make it into PicoC, it should at least only be used  with a high enough accesslevel (at the moment level 2 to is the highest i think).

ernieift:
We can determine the forum the need for "Enable" and "Enablex". Without these functions but with a list of the target PWM inputs and "Read", the most users would be satisfied with it like it is now.

Trex4Git:
I opened a thread in the forum to evaluate the interest in it: http://forum.taulabs.org/viewtopic.php?f=17&t=715

And what do you think about the ReadX/WriteX functions? I thought they might be useful for not especial defined GPIO pins, but pins that can be securely used, like the PWM Inputs you mentioned.

ernieift:
Have a look at #1640. When I read the data sheet correctly writes to AF-Pins are ignored and only stored internal. If we extend the hardware selection for quanton to use normal outputs on the PWM_IN, it were possible to use the free ones as static output without issues. Just select the mode in the gcs and use the programmed functionality.

Trex4Git:
So as there is not much feedback yet, i think it makes sense to only merge the simple GPIO Read functionality into this pull request, with static definitions for quanton of buzzer and battery pins as output. And these 2 together with the PWM input pins in one array as you mentioned earlier.

But the extended hardware selection for quanton is out of my knowledge about taulabs at the moment. And if there are changes in this, maybe also the documentation for quanton has to changed.

I don't know if its worth to merge already the extended GPIO functionality (ReadX, WriteX, EnableX) to PIOS if there is no usage for it now, and no PicoC wrapping. 
Maybe its better to wait for this, if there will be ever demand for it, or not.

ernieift:
Could you add the read functionality and the PWM inputs on Quanton to the list. You can put it together in this PR. Later, I can add a selection to the hwsettings on quanton quite easily. So there is no need to use ENABLE on these pins.

Trex4Git:
Added the simple read functionality together with the extended static GPIO definitions for quanton already yesterday and merged now.

Tested also reading the value of a the buzzer output pin, and it works as expected.

ernieift:
@Trex4Git  Looks like a really good addition. I think, we can merge this finally. Sorry for the delay...
@peabody124 What do you think?

Trex4Git:
If someone wants to test this, i posted some simple example scripts for PicoC in the forums: http://forum.taulabs.org/viewtopic.php?f=17&t=715&p=5959#p5959

lolorc:
The quick test i did was conclusive, Thanks Trex

Trex4Git:
@peabody124 Do you see some issues, that need to be fixed before this can be merged ?

peabody124:
@Trex4Git I'll try and review tomorrow, but since it has @ernieift's blessing it should be a quick merge :)

peabody124:
Just pointing out this does reverse the polarity of the 'PIOS_GPIO_On' and _Off messages. Not an issue since they were never used, though.

Trex4Git:
Thanks for the feedback and support.
I reversed the thing because i have seen it was not used anywehre before, but On set the Pin to 0, and Off to Vcc before.

PR #1631, opened 2015-05-20
Simulation discrete target: supersedes #1599

mlyle:


mlyle:
I'm happy with this and it works.

Issue #1220, opened 2014-07-30
State machine to track flight status


peabody124:
We need a state machine that actually tracks when we are flying or not. I see lots of issues with other flight controllers and failsafe due to the absence of this (eg how you handle failsafe while flying should be much different than when grounded).

Ideally this would go into flight status, which means it should be evaluated in manual control. It is possible out armed signal is ok , but if so we should think this through and confirm it.

E.g. arm on ground, turn off transmitter, no option to failsafe.

Would a reasonable criterion be to have armed, run up motors, and not yet disarmed? How does this factor into gliders etc...

Such a system can also do a few other things 
- store the takeoff location
- store the takeoff altitude
- ???

peabody124:
see also #1410

PR #1635, opened 2015-05-23
ENH: FlightStats Module

BrainFPV:
Module that collects various statistics during the flight.

This is work in progress, suggestions for additional data to collect or how to improve things are welcome!


BrainFPV:
Please review

peabody124:
Starting to review now. Cause you update the title to remove WIP if you think this is merge worthy.

peabody124:
Also this will go a long way to the infrastructure for #1220

mlyle:
Read this, looks good to me.  I still prefer integrating velocity for distanceTraveled but either is reasonable.

flight/Modules/FlightStats/flightstatsmodule.c
```diff
@@ -0,0 +1,257 @@
...
+				}
+				break;
+			case FLIGHTSTATS_STATE_RESET:
+				memset((void*)&flightStatsData, 0, sizeof(flightStatsData));
+				flightStatsData.State = FLIGHTSTATS_STATE_INITIALIZING;
+				break;
+			case FLIGHTSTATS_STATE_INITIALIZING:
+				PositionActualGet(&lastPositionActual);
+				if (first_init || settings.StatsBehavior == FLIGHTSTATSSETTINGS_STATSBEHAVIOR_RESETONARM) {
+					if (FlightBatteryStateHandle()) {
```
peabody124: Not critical, but this logic took me a few minutes to grok and I think can be simplified.
1. create a nice little method setInitialBattery() that does what this block of code does. Call it before entering the while loop.
2. make _STATE_RESET explicitly only reset the fields it wants to (basically everything but InitialBatteryVoltage
3. get rid of first_init variable, and in conditional of L137 go straight to _COLLECTING state (because essentially already initialized). the rest of the code can go.
4. delete the _INITIALIZING state, and call setInitialBattery() from _RESET

IMO that would be a fair bit more readable (except more lines at L152 to set the fields individually, but that's more typical of how we normally do it). Not a blocker though.

flight/Modules/FlightStats/flightstatsmodule.c
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * Check whether FC is armed
+ * \return true if armed
+ */
+static bool isArmed()
+{
+	uint8_t arm_status;
+	FlightStatusArmedGet(&arm_status);
+	if (arm_status == FLIGHTSTATUS_ARMED_ARMED) {
```
peabody124: Why not
``` return arm_status == FLIGHTSTATUS_ARMED_ARMED```

File: flight/Modules/FlightStats/flightstatsmodule.c (hunk 1 above)
Comment (peabody124) [inline flight/Modules/FlightStats/flightstatsmodule.c]: Alternatively you can even keep the memset and simply create another variable initial_battery_voltage and set that. Then just write it in collect_stats.

File: shared/uavobjectdefinition/flightstats.xml [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+	<field name="MaxDescentRate" units="m/s" type="float" elements="1"/>
+	<field name="MaxGroundSpeed" units="m/s" type="float" elements="1"/>
+	<field name="MaxAirSpeed" units="m/s" type="float" elements="1"/>
+	<field name="MaxRollRate" units="deg/s" type="uint16" elements="1"/>
+	<field name="MaxPitchRate" units="deg/s" type="uint16" elements="1"/>
+	<field name="MaxYawRate" units="deg/s" type="uint16" elements="1"/>
+	<field name="ConsumedEnergy" units="mAh" type="uint16" elements="1"/>
+	<field name="InitialBatteryVoltage" units="mV" type="uint16" elements="1"/>
+
+	<field name="State" units="" type="enum" elements="1" options="IDLE, RESET, INITIALIZING, COLLECTING"/>
```
peabody124: Drop this INITIALIZING state if you do what I described above.

peabody124:
Looks good, although I'd love to see those changes I suggested. Great work @BrainFPV

BrainFPV:
@peabody124 thanks for the review. I modified the code, I didn't do exactly what you suggested, but I think it makes the code more readable. There was actually a bug in the previous code, i.e., if the mode was `_RESETONBOOT`, the consumed energy also included the energy consumed while disarmed. 

@mlyle regarding integrating velocity, what I assume is that this is already done by the navigation filter to get the NED location. The filter runs much faster, so it will be more accurate than if we do the same integration here. This makes me think summing up the location difference is more accurate. Alternatively, the distance traveled could be computed by the navigation filter, which would be the most accurate but would introduce possibly unnecessary computations there..

flight/Modules/FlightStats/flightstatsmodule.c
```diff
@@ -0,0 +1,264 @@
...
+	GyrosData gyros;
+	GyrosGet(&gyros);
+	stats->MaxPitchRate = MAX(stats->MaxPitchRate, fabsf(gyros.y));
+	stats->MaxRollRate = MAX(stats->MaxRollRate, fabsf(gyros.x));
+	stats->MaxYawRate = MAX(stats->MaxYawRate, fabsf(gyros.z));
+
+	// Consumed energy
+	if (FlightBatteryStateHandle()) {
+		FlightBatteryStateConsumedEnergyGet(&tmp);
+		stats->ConsumedEnergy += (tmp - initial_consumed_energy);
```
BrainFPV: This is even more wrong now ;).. will think of a way to do it correctly

mlyle:
So, kind of.  It's really complicated.

The navigation filters both on us and the GPS feed both ways.  And seemingly contradictory things can be happening, like the GPS can be pretty sure from carrier phase information that we're moving northwards, and report that to us in the ground velocity information... while moving our position southwards as it refines away past errors in the position estimate and integrates pseudoranges over longer lengths of time.

flight/Modules/FlightStats/flightstatsmodule.c
peabody124: Is that you saying this is still WIP? :)

peabody124:
@BrainFPV do you want this merged as is? The title still says "WIP"

flight/Modules/FlightStats/flightstatsmodule.c
mlyle: Shouldn't this just be = not +=?

BrainFPV:
@peabody124 sorry for the delay. The bug in the consumed energy calculation has been fixed. The rationale is that the flight stats module should only use data while flying (and it assumes armed means flying). In the "reset on boot" mode, this means the energy consumed while disarmed should not be included in the calculation, but it should be combined with the energy consumed in the previous flight. I hope this makes sense.

BrainFPV:
Ping @peabody124

peabody124:
So I tested this on Sparky2, although I don't think my comment is specific. I enabled the logging module (but didn't arm) and I'm seeing 72% of the CPU utilization going to the logging task (while not even logging).

To simplify things, I erased settings and only enabled the logging module. Same thing.

shared/uavobjectdefinition/flightstats.xml
```diff
@@ -0,0 +1,25 @@
...
+	<field name="MaxPitchRate" units="deg/s" type="uint16" elements="1"/>
+	<field name="MaxYawRate" units="deg/s" type="uint16" elements="1"/>
+	<field name="ConsumedEnergy" units="mAh" type="uint16" elements="1"/>
+	<field name="InitialBatteryVoltage" units="mV" type="uint16" elements="1"/>
+
+	<field name="State" units="" type="enum" elements="1" options="IDLE, RESET, COLLECTING"/>
+
+	<access gcs="readwrite" flight="readwrite"/>
+	<telemetrygcs acked="false" updatemode="manual" period="0"/>
+	<telemetryflight acked="false" updatemode="manual" period="1000"/>
```
peabody124: Could you make this updatemode="periodic" and maybe set a longer period? 5000?

BrainFPV:
@peabody124 done

PR #1636, opened 2015-05-24
[Not for merge] PicoC Extensions: add additional functions in PicoC library

elafargue:
This enables extended UAV Object control from PicoC.

This is a PR for review and discussion. Code was written by V Getin.


elafargue:
Looks like the contributor removed all I2C related calls, checking why that it...

ernieift:
When it is the intension to generate a transforming UAV (multirotor/wing) why not do this within the firmware? There are much more and better possibilities to do this secure. Who wants to blame picoC if it doesn't work like expected?
The StabilizationXSettings are already in picoC. With access level 2 you have write access during flight. Changing from VTOL to Wing and back requires more attention. When somebody want this, it were better to do it in a single call.

ernieift:
After a night i have some comments. I think the conversion from multirotor into wing can work, but needs more attention. It is absolutely necessary to protect this with an access level >=1. 1 means in disarmed state 2 allows changes during flight. The path follower and other modules must be reviewed or adjusted, I guess. There are some which read their configuration only at boot time.
Has V. Getin already had success with the changes?

mlyle:
Closing for now; perhaps open an issue for discussion if there's an intention to proceed here.

PR #1633, opened 2015-05-22
Python scripts

peabody124:
Various scripts using the python tools

PR #1638, opened 2015-05-26
python-api: cleanup, functionality

mlyle:
- Performance 20-30% better depending on workload
- Buffering model changes to not "scrunch down" the packet buffer all the time and instead burn a little RAM.
- Use the precomputed structure unpack even for nested uavos-- now just keep a depth map and restructure the output tuple on demand.
- UAVO class eliminated; additional cleanup in class creation is needed, though (there's a scary method-of-death)
- Various indirection through the UAVO class eliminated too
- Add a simple function to request UAVOs.
- Teach the minimal_connection.py to crudely retrieve configuration from the simulator!

I think I'm getting close to not need nearly so much refactoring at each functional step and I should be able to mostly focus on functional API work from this point.

### DIFF
--- a/python/taulabs/uavtalk.py
+++ b/python/taulabs/uavtalk.py
@@ -206,9 +206,19 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
             timestamp = timestamp_fmt.unpack_from(buf, header_fmt.size + buf_offset)[0]
 
             # handle wraparound
+
+            # This is currently hacked because the ordering of instance id and
+            # time is messed up in this code-- hacked so at least timestamps
+            # don't jump.
             if timestamp < last_timestamp:
-                timestamp_base = timestamp_base + 65536
-            last_timestamp = timestamp
+                if timestamp > 200:
+                    timestamp_base = timestamp_base + 65536
+                    last_timestamp = timestamp
+                else:
+                    timestamp += 65536
+            else:
+                last_timestamp = timestamp
+
             timestamp += timestamp_base
         else:
             timestamp = last_timestamp
--- a/python/taulabs/uavtalk.py
+++ b/python/taulabs/uavtalk.py
@@ -57,29 +57,46 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
 
     received = 0
 
-    packet_bytes = ''
+    buf = ''
+    buf_offset = 0
+
+    pending_pieces = []
 
     while True:
+        # If we don't have data buffered, join up any chunks we've been given
+        # to ensure pending_pieces is empty for the rest of this loop.
+        #
+        # in other words, don't mix the pending_pieces drain model and the
+        # buffer concatenation model within a loop iteration.
+        #
+        # 10k chosen here to be bigger than any plausible uavo
+        if len(buf)-buf_offset < 10240:
+            pending_pieces.insert(0, buf[buf_offset:])
+            buf_offset = 0
+
+            buf = ''.join(pending_pieces)
+            pending_pieces = []
+
         if gcs_timestamps:
-            while len(packet_bytes) < logheader_fmt.size:
+            while len(buf) < logheader_fmt.size + buf_offset:
                 rx = yield None
 
                 if rx is None:
                     return
 
-                packet_bytes = packet_bytes + rx
+                buf = buf + rx
 
-            overrideTimestamp, logHdrLen = logheader_fmt.unpack_from(packet_bytes,0)
+            overrideTimestamp, logHdrLen = logheader_fmt.unpack_from(buf,buf_offset)
 
-            packet_bytes = packet_bytes[logheader_fmt.size:]
+            buf_offset += logheader_fmt.size
 
         # Ensure we have enough room for all the
         # plain required fields to avoid duplicating
         # this code lots.
         # sync(1) + type(1) + len(2) + objid(4) 
 
-        while (len(packet_bytes) < header_fmt.size) or (packet_bytes[0] != chr(SYNC_VAL)):
-            #print "waitingsync len=%d"%(len(packet_bytes))
+        while (len(buf) < header_fmt.size + buf_offset) or (buf[buf_offset] != chr(SYNC_VAL)):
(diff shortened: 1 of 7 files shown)


mlyle:
Just imported peabody124's #1633 changes, fixed up so they can land as well.

peabody124:
@solidgoldbomb are you tempted to review this? I'm happy to test it for functionality but am less qualified to read this code.

solidgoldbomb:
Yeah, I'll try to get this reviewed on the weekend.

python/taulabs/uavtalk.py
```diff
@@ -57,29 +57,46 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
...
     received = 0
 
-    packet_bytes = ''
+    buf = ''
+    buf_offset = 0
+
+    pending_pieces = []
 
     while True:
+        # If we don't have data buffered, join up any chunks we've been given
```
solidgoldbomb: "don't have sufficient data buffered"

python/taulabs/uavtalk.py
```diff
@@ -57,29 +57,46 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
...
+        # to ensure pending_pieces is empty for the rest of this loop.
+        #
+        # in other words, don't mix the pending_pieces drain model and the
+        # buffer concatenation model within a loop iteration.
+        #
+        # 10k chosen here to be bigger than any plausible uavo
```
solidgoldbomb: Using a large constant here is fine, but you could also compute it (once) from the uavo_defs list that was passed in as the generator was being started.

python/taulabs/uavtalk.py
```diff
@@ -57,29 +57,46 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
...
+        if len(buf)-buf_offset < 10240:
```
solidgoldbomb: spacing around ```-```

python/taulabs/uavtalk.py
```diff
@@ -57,29 +57,46 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
 
     received = 0
 
-    packet_bytes = ''
+    buf = ''
+    buf_offset = 0
```
solidgoldbomb: Now that you have an offset, you can probably remove the comment above that says "This could be made considerably faster if it kept an offset."

python/taulabs/uavtalk.py
```diff
@@ -120,11 +137,13 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
 
         # Before there used to be code to handle instance offsetting
```
solidgoldbomb: Probably don't need this comment any more.

python/taulabs/uavtalk.py
```diff
@@ -155,36 +174,37 @@ def process_stream(uavo_defs, use_walltime=False, gcs_timestamps=False):
...
+
+            buf_offset += 1
 
-            packet_bytes = packet_bytes[1:]
             continue
         
         if timestamp_len:
             # pull the timestamp from the packet
-            timestamp = timestamp_fmt.unpack_from(packet_bytes, header_fmt.size)[0]
+            timestamp = timestamp_fmt.unpack_from(buf, header_fmt.size)[0]
```
solidgoldbomb: This looks like it should be ```buf[buf_offset:]```.

python/visualize/vid_overlay_position.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/python
```
solidgoldbomb: Is this file related to the python API changes in this PR or was this included by mistake?

solidgoldbomb:
A few minor things to fix up.  One possible bug in uavtalk.py.  The uavo parsing function and the process_stream generator are getting a bit too long but they appear to be functionally correct.

I haven't reviewed the vid_overlay_position.py file.  @mlyle did you review it before adding it to this PR?

@peabody124 can you run some tests on this branch?  I ran a few basic tests of logview and shell, but I don't have any large/complex log files around.

Good to merge after the noted items have been fixed and test results are reported.  Thanks for all of the work you're doing on the python code.

python/taulabs/uavtalk.py
mlyle: header_fmt.size + buf_offset, but very, very good catch

mlyle:
Thanks for the review (and spotting the defect!).  I have hope for making the uavo parsing function cleaner/shorter, but the process_stream thing is unfortunately just how it has to be.

mlyle:
And I've read the animator; I don't love everything about it stylistically but it looks like correct code that will not become a maintenance problem.

solidgoldbomb:
Ok, just waiting on testing from @peabody124 then.  Almost there :)

peabody124:
Alright. Tested visualizing a path and making an overlay. Works fine for normal logs.

However, there is a regression when parsing overo logs (aside: we've also taken a big speed hit on those in next versus master). 

On next if I load a log (with ```shell.py -t -g master```) and then say ```aa = uavo_list.as_numpy_array(UAVO_AttitudeActual)``` it spends a while and eventually says
```received 2240000 objs```

then ```In [3]: aa.size
Out[3]: 238163```

On this PR if I do the same thing it never says all the received XXX objs, and the ```aa.size``` return 469. 

In addition on **both** this PR and next, but not master, the timestamps are messed up. There seems to be an endian issue or something going on, because the timestamps are more increasing smoothly. I remember this was something I had to handle differently when the timestamps were embedded in the uavtalk stream and CRC'd.

@mlyle I'll upload the log I'm playing with and send you a link.

peabody124:
I wonder if we should create a unit test with two or three log files and just sanity check things like the number of a certain object

mlyle:
So, there was a bug with the Overo 0xff padding and desync and end of file--- this is now handled properly.

There's also a bug with the ordering of instance id versus timestamp in the code.  This was triggering timestamp issues.  I have hacked around it crudely for this pull (it's not new to this pull request) and will make this better in next API push.

peabody124:
Ok tested and works.

Issue #1617, opened 2015-05-16
UBX should reconfigure while "NoGps" or if messages missing


peabody124:
If you apply power via USB before the GPS or disconnect/reconnect the GPS then you will not have a configuration. The system should repeatedly send the autoconfiguration messages (e.g. once every 10 seconds) if the status is NoGps. possibly also if some messages (e.g. NEDVEL) are never seen.

DTFUHF:
Probably better idea is a simple state machine that can detect and recover
from configuration and communication problems. That can also handle
autobauding. Simpler FCs like baseflight already do this.

peabody124:
We already have autobauding. It is just deciding when to call it.

DTFUHF:
Yes, I meant a GPS communication state machine could handle this as well.

elafargue:
This is fixed by #1641

PR #1641, opened 2015-05-26
gps: reinit GPS at runtime / issue #1617

mlyle:
Fixes issue #1617.  Confirmed to properly reinit at runtime on Sparky2.

### DIFF
--- a/flight/Modules/GPS/GPS.c
+++ b/flight/Modules/GPS/GPS.c
@@ -81,9 +81,6 @@ static struct pios_thread *gpsTaskHandle;
 
 static char* gps_rx_buffer;
 
-static uint32_t timeOfLastCommandMs;
-static uint32_t timeOfLastUpdateMs;
-
 static struct GPS_RX_STATS gpsRxStats;
 
 // ****************
@@ -190,6 +187,9 @@ static void gpsTask(void *parameters)
 
 	GPSPositionData gpsposition;
 	uint8_t	gpsProtocol;
+
+	uint32_t timeOfLastUpdateMs;
+	uint32_t timeOfLastCommandMs;
 
 	ModuleSettingsGPSDataProtocolGet(&gpsProtocol);
 
--- a/flight/Modules/GPS/GPS.c
+++ b/flight/Modules/GPS/GPS.c
@@ -177,27 +177,9 @@ int32_t GPSInitialize(void)
 MODULE_INITCALL(GPSInitialize, GPSStart);
 
 // ****************
-/**
- * Main gps task. It does not return.
- */
 
-static void gpsTask(void *parameters)
+static void gpsConfigure(uint8_t gpsProtocol)
 {
-	uint32_t xDelay = GPS_COM_TIMEOUT_MS;
-
-	GPSPositionData gpsposition;
-	uint8_t	gpsProtocol;
-
-	uint32_t timeOfLastUpdateMs;
-	uint32_t timeOfLastCommandMs;
-
-	ModuleSettingsGPSDataProtocolGet(&gpsProtocol);
-
-#if defined(PIOS_GPS_PROVIDES_AIRSPEED)
-	gps_airspeed_initialize();
-#endif
-
-reinit:
 #if !defined(PIOS_GPS_MINIMAL)
 	switch (gpsProtocol) {
 #if defined(PIOS_INCLUDE_GPS_UBX_PARSER)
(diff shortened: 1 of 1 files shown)


flight/Modules/GPS/GPS.c
```diff
@@ -261,20 +260,22 @@ static void gpsTask(void *parameters)
...
-		timeNowMs = PIOS_Thread_Systime();
+		uint32_t timeNowMs = PIOS_Thread_Systime();
 		if ((timeNowMs - timeOfLastUpdateMs) >= GPS_TIMEOUT_MS) {
 			// we have not received any valid GPS sentences for a while.
 			// either the GPS is not plugged in or a hardware problem or the GPS has locked up.
 			uint8_t status = GPSPOSITION_STATUS_NOGPS;
 			GPSPositionStatusSet(&status);
 			AlarmsSet(SYSTEMALARMS_ALARM_GPS, SYSTEMALARMS_ALARM_ERROR);
+			/* Don't reinitialize too often. */
+			if ((timeNowMs - timeOfLastCommandMs) >= GPS_TIMEOUT_MS) {
```
elafargue: Maybe we should use another (larger) timeout there, like 5 seconds, to make sure we don't try to reinit the GPS too often ?

mlyle: So, I slightly increased the timeout from the default 500ms in the original pull. It requires not receiving any GPS messages for 750ms and being 750ms from the last init attempt.

If this ever happens in flight, 750ms is a long, long time.  We expect to receive a new set of nav messages every 100ms or 200ms (and several messages during each update).   A delay of 600ms+ should never happen without a failure, and I don't think reiniting takes too much resources / makes anything worse.

flight/Modules/GPS/GPS.c
```diff
@@ -234,6 +230,9 @@ static void gpsTask(void *parameters)
 	}
 #endif /* PIOS_GPS_MINIMAL */
 
+	timeOfLastUpdateMs = 0;
+	timeOfLastCommandMs = PIOS_Thread_Systime();
```
peabody124: This variable is not updated ever again... this seem like a mistake.

flight/Modules/GPS/GPS.c
```diff
@@ -234,6 +230,9 @@ static void gpsTask(void *parameters)
 	}
 #endif /* PIOS_GPS_MINIMAL */
 
+	timeOfLastUpdateMs = 0;
```
peabody124: Also now these two variables are only manipulated in the task, they no longer need to be static to the whole file (good improvement).

peabody124:
Good improvement, but the way it is coded I don't understand that nested set of conditional since the inner one is probably always true. Can you maybe throw a debug printf in the re-init code and make sure it is firing only at the right time.

Re-initing too often is preferable to having no data, but I think with the auto-bauding it's probably not terribly quick, so let's make sure this isn't false alarming.

flight/Modules/GPS/GPS.c
mlyle: I added a goto... so we can go back up above this point and update the time to the last initialization.

flight/Modules/GPS/GPS.c
mlyle: OK, I didn't actually make that improvement-- it was already so :)  But moving their declaration.

mlyle:
The inner one isn't true because when we goto upwards the command time gets updated.  The inner conditional is there so that we wait at least a little while for the reinit to have an effect and don't just autobaud over and over and over.

I've watched logs to make sure it does the right thing (alarms are asserted every time it reinits).

There is a small behavior change-- GPS now alarms at startup instead of staying grey for as long-- I did this on purpose by initializing the update time to 0 instead of now.

flight/Modules/GPS/GPS.c
peabody124: Ah, I see what this variable```timeOfLastCommandMs``` does now. Could you actually call it something like timeOfLastReinitMs ?

flight/Modules/GPS/GPS.c
```diff
@@ -261,20 +260,22 @@ static void gpsTask(void *parameters)
...
+		uint32_t timeNowMs = PIOS_Thread_Systime();
 		if ((timeNowMs - timeOfLastUpdateMs) >= GPS_TIMEOUT_MS) {
 			// we have not received any valid GPS sentences for a while.
 			// either the GPS is not plugged in or a hardware problem or the GPS has locked up.
 			uint8_t status = GPSPOSITION_STATUS_NOGPS;
 			GPSPositionStatusSet(&status);
 			AlarmsSet(SYSTEMALARMS_ALARM_GPS, SYSTEMALARMS_ALARM_ERROR);
+			/* Don't reinitialize too often. */
+			if ((timeNowMs - timeOfLastCommandMs) >= GPS_TIMEOUT_MS) {
+				goto reinit;
```
peabody124: I know goto isn't considered an absolute prohibition anymore, but in this case it does make the code structure fairly hard to read. Would you mind doing this?

 1. move the initialization code into it's own function (this will be useful anyway when you start reparameterizing it)
 2. call that at startup and also here
 3. update the timeOfLastCommandMs / timeOfLastReinitMs here

That should make it quite a bit more readable

mlyle:
Here's an alternate way to do it.  I only use two patterns of goto -- goto retry (near the top) and goto fail (near the bottom) of functions.  If the former, along with changing variables, is not OK, there's a whole lot of code in PIOS that seems to follow this pattern that should probably be changed.

Note that I spotted a new bug in the original-- if the GPS rebooted, and we were constantly receiving characters (say the baud rate is wrong) that don't parse, we'd never get down to init the alarms.  That's why I now set xDelay to 0 after the first loop iteration.

I don't have as much confidence in the changeset anymore factored this way, and it needs me to read it again and to put it through test again.

flight/Modules/GPS/GPS.c
```diff
@@ -229,15 +207,44 @@ static void gpsTask(void *parameters)
...
+	gps_airspeed_initialize();
+#endif
 
 	GPSPositionGet(&gpsposition);
+
 	// Loop forever
-	while (1)
-	{
+	while (1) {
+		uint32_t xDelay = GPS_COM_TIMEOUT_MS;
```
peabody124: Not related to your changes, but I'd love to see this changed to a descriptive name like ```characterTimeoutMs```

flight/Modules/GPS/GPS.c
```diff
@@ -261,20 +268,24 @@ static void gpsTask(void *parameters)
...
-		timeNowMs = PIOS_Thread_Systime();
-		if ((timeNowMs - timeOfLastUpdateMs) >= GPS_TIMEOUT_MS) {
+		if ((loopTimeMs - timeOfLastUpdateMs) >= GPS_TIMEOUT_MS) {
 			// we have not received any valid GPS sentences for a while.
 			// either the GPS is not plugged in or a hardware problem or the GPS has locked up.
 			uint8_t status = GPSPOSITION_STATUS_NOGPS;
 			GPSPositionStatusSet(&status);
 			AlarmsSet(SYSTEMALARMS_ALARM_GPS, SYSTEMALARMS_ALARM_ERROR);
+			/* Don't reinitialize too often. */
+			if ((loopTimeMs - timeOfConfigAttemptMs) >= GPS_TIMEOUT_MS) {
```
peabody124: Won't this always happen at the same time? Is this second nesting doing anything?

mlyle:
rebased to next, a couple of bugs fixed, and verified to reinitialize nicely.

Issue #1637, opened 2015-05-25
Oneshot: Issues in GCS after newest modifications in #1615


DTFUHF:
On the Output config widget:

1) The tooltip for the Type: field, refers to HPWM, which is not a selectable option.

2) The Max field is resized by GCS to 3 digits. Steps to reproduce:

-Set an output channel to update rate = 2000 or higher when Resolution is at either setting of 1us or 80ns.
-Click Save.

3) There is some upper limit on the value that can be set. Example: 12kHz, 80ns resolution. For a 0-100% duty cycle output e.g. a brushed motor the min should be 0 and the max should be approx. 1041 (1041 * 80ns = 0.08328ms or about 12kHz) but this value can't be set higher than 833 in the GCS menu. It can be set higher in the UAVO menu, however.

Also, the GCS allows the setting of nonsensical values: for example, 1kHz / 1us / 9000 count max. I believe these are limited in actual flight code to 100% duty cycle values but it makes sense to also limit them here and show the user the proper value for max duty cycle.

ernieift:
Let me recalculate this. 12kHz PWM make 83,33...µs period. The maximum pulse width in integer are 84µs then.
You are right the 80ns are not quite correct. 83ns would be closer, but this is a small mistake.

mlyle:
#1642 corrected this.

PR #1642, opened 2015-05-26
GCS: Hpwm fixes

DTFUHF:
Proposed solution for #1637.

### DIFF
--- a/ground/gcs/src/plugins/config/outputchannelform.cpp
+++ b/ground/gcs/src/plugins/config/outputchannelform.cpp
@@ -387,18 +387,18 @@ void OutputChannelForm::updateMaxSpinboxValue(UAVObject *obj)
             // ... and if there's a match, set the maximum values and return
             if (channel-1 == m_index) {
                 if (actuatorSettingsData.TimerUpdateFreq[i] == 0)
-				{
-					// SyncPWM has been selected, limit values to maximum allowed by UAVO
-					ui.actuatorMin->setMaximum(std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max());
-					ui.actuatorMax->setMaximum(std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max());
+                {
+                    // SyncPWM has been selected, limit values to maximum allowed by UAVO
+                    ui.actuatorMin->setMaximum(std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max());
+                    ui.actuatorMax->setMaximum(std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max());
                     return;
-				}
+                }
 
-				double pwm_freq = 0;
-				if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_1MHZ)
-					pwm_freq = 1000000;
-				else if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_12MHZ)
-					pwm_freq = 12000000;
+                double pwm_freq = 0;
+                if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_1MHZ)
+                    pwm_freq = 1000000;
+                else if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_12MHZ)
+                    pwm_freq = 12000000;
                 double maxPulseWidth = round(pwm_freq / actuatorSettingsData.TimerUpdateFreq[i]);
 
                 // Saturate at the UAVO's maximum value
--- a/ground/gcs/src/plugins/config/outputchannelform.cpp
+++ b/ground/gcs/src/plugins/config/outputchannelform.cpp
@@ -389,7 +389,12 @@ void OutputChannelForm::updateMaxSpinboxValue(UAVObject *obj)
                 if (actuatorSettingsData.TimerUpdateFreq[i] == 0)
                     return;
 
-                double maxPulseWidth = round(10000000.0 / actuatorSettingsData.TimerUpdateFreq[i]);
+				double pwm_freq = 0;
+				if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_1MHZ)
+					pwm_freq = 1000000;
+				else if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_12MHZ)
+					pwm_freq = 12000000;
+                double maxPulseWidth = round(pwm_freq / actuatorSettingsData.TimerUpdateFreq[i]);
 
                 // Saturate at the UAVO's maximum value
                 if (maxPulseWidth > std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max())
--- a/ground/gcs/src/plugins/config/outputchannelform.cpp
+++ b/ground/gcs/src/plugins/config/outputchannelform.cpp
@@ -386,25 +386,30 @@ void OutputChannelForm::updateMaxSpinboxValue(UAVObject *obj)
         foreach(qint32 channel, channelBank) {
             // ... and if there's a match, set the maximum values and return
             if (channel-1 == m_index) {
-                if (actuatorSettingsData.TimerUpdateFreq[i] == 0)
+                
+                double maxPulseWidth = 0;
+                
+                if (actuatorSettingsData.TimerUpdateFreq[i] != 0)
                 {
-                    // SyncPWM has been selected, limit values to maximum allowed by UAVO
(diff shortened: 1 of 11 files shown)


peabody124:
Jenkins: test this please

peabody124:
Jenkins: add to whitelist

peabody124:
I slightly prefer the resolution to be in ns rather than MHz. Despite the 3ns inaccuracy ;-), it's a more intuitive metric I think.

ground/gcs/src/plugins/config/outputchannelform.cpp
```diff
@@ -387,9 +387,19 @@ void OutputChannelForm::updateMaxSpinboxValue(UAVObject *obj)
...
-
-                double maxPulseWidth = round(10000000.0 / actuatorSettingsData.TimerUpdateFreq[i]);
+                }
+
+                double pwm_freq = 0;
+                if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_1MHZ)
+                    pwm_freq = 1000000;
+                else if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_12MHZ)
+                    pwm_freq = 12000000;
+                double maxPulseWidth = round(pwm_freq / actuatorSettingsData.TimerUpdateFreq[i]);
```
peabody124: This looks like wrong math. The pulse width is now in true microseconds and the resolution doesn't matter (since we round to microseconds, untill someone adds a mode with a resolution of 10s of ms.

The right math would be to always ```round(pwm_freq / actuatorSettingsData.TimerUpdateFreq[i])```. However, if ```actuatorSettingsData.TimerUpdateFreq==0``` which indicates SyncPWM then this code produces a divide by 0 error. Here, the maximum pulse duration is a function of the sensor period. Most sensors are running between 300-500 Hz, some people go as high as 1khz. It might be appropriate to hardcode the max to 1000us for SyncPWM to cover all these cases robustly...

DTFUHF: The TimerUpdateFreq==0 is handled earlier on, where I set it to the
max allowed by the UAVO (uint16).

DTFUHF: I suppose ns vs. MHz is a matter of preference. But I used the 80ns value to calculate duty cycle for my application which resulted in an incorrect max pwm setting. I don't think it's a good idea to show incorrect values to the user.

peabody124: Yeah. Would you mind polling people on IRC and deferring to the preference there?  I agree it is inelegant to be inaccurate. We could even rename it 83ns to be pedantic.

peabody124: Oops. Your talking about both issues in the same thread :) now my response is confusing.

Ok I see the handling below. Did you check that? If you have a divide by 0 you get NaN. Normally comparisons against NaN are problematic (in that they are always false). I'd prefer to see it explicitly handled here - and see my comment about perhaps making maxpulsewidth 1000 in sync mode, although I'm less sure on that.

DTFUHF:
Cleaned up the logic slightly and added 1ms limit in SyncPWM mode.

peabody124:
Ok. My attempt at a poll on IRC did not garner support for ns. I'd like to review this another time and test it myself, unless someone else can. Should be good to merge though from initial review.

mlyle:
While we're talking about aesthetics-- a separate thing, not for this pull --- I think there should be a clear label on the units on the output pane being microseconds.

I think these used to be in units of the resolution, for awhile, and now they're not.   It confused me for a bit.

peabody124:
That would be nice. @mlyle you have a chance to test this works correctly? I'll merge if so.

mlyle:
I tested this and it seems to mostly do the right thing.

One concern I have is this-- the ranges for validation of the below inputs are only updated when you click "apply".  This basically forces you to have an intermediate state on the flight board where the values don't make sense together... and worse, may actually result in an intermediate state where they spin.

Is there any way to make them immediately take effect for validation?

ground/gcs/src/plugins/config/outputchannelform.cpp
```diff
@@ -386,15 +386,30 @@ void OutputChannelForm::updateMaxSpinboxValue(UAVObject *obj)
...
+                    // Saturate at the UAVO's maximum value
+                    if (maxPulseWidth > std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max())
+                        maxPulseWidth = std::numeric_limits<__typeof__(actuatorSettingsData.ChannelMax[0])>::max();
+                } else {
+                    // SyncPWM has been selected, the actual pulse limit is a function of sensor rate.
+                    // Simply limit to 1ms here to account for 1000Hz sensor rate.
+                    if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_1MHZ)
+                        maxPulseWidth = 1000;
+                    else if (actuatorSettingsData.TimerPwmResolution[i] == ActuatorSettings::TIMERPWMRESOLUTION_12MHZ)
+                        maxPulseWidth = 12000;
```
ernieift: The maxPulseWidth = 12000µs = 12ms in higher resolution? This can't be right. Why do want to limit the pulse width in sync mode. The pulse width is given in µs independent from the base frequency. If you want to limit anything, you have to look at the sensor rate, but why? Limiting the pulse width to 1ms makes normal ESCs (without OneShot) unusable in synced mode. 
BTW I wouldn't use round() here better choose ceil().

mlyle: The maximum pulse length has to be shorter than our maximum possible flight loop update rate, unless the underlying code correctly will inhibit a pulse if a pulse is already in progress.  I've not read the underlying code to know.

DTFUHF: Per Peabody's comments the pwm time should be limited to the max sensor
update rate. It was previously limited only to the size of the UAVO (16 bit
unsigned).

Yes it effectively makes sync-PWM and low resolution unusable. But this is
not a mode that is typically used by anybody. If desired, and if low enough
sensor rates are used, the user can set whatever pwm time he chooses with
the UAVOs directly.

A PWM count of 12000 at 12MHz is still 1 millisecond. It's my understanding
the values here are not in milliseconds but counts of 83.3 nanoseconds.

ernieift: The sensor rate isn't part of the output settings. It is defined in the hardware page and depends on the possibilities of the target.

DTFUHF: I probably should add this, so I will, and will assume the max rate is 1kHz
in the case sensor rate is unable to be found.

ernieift: PS: We have rewritten the output code to use physical units again. All values in the output page (MIN/NEUTRAL/MAX) are in µs. There is no fixed loop time here. The rate is based on the sensor interrupt.

DTFUHF: Hm, if this is the case, why expose resolution setting at all? high res
could be chosen automatically for SyncPWM or timer periods 400Hz and below.

ernieift: Do you mean higher not below? It's better to have the choice. Future CPUs can have other/more settings.

peabody124: Let's not get into the firmware auto detecting things when possible. It's nice to have the flexibility of explicit settings and then if needed cover them up in the GCS (e.g. the "OneShot" button in the wizard sets up three things).

peabody124: Oh, also this line _is_ a bug. The latest code uses true values (in microseconds). This is setting the maximum pulse duration to 12000 us.

ground/gcs/src/plugins/config/outputchannelform.ui
```diff
@@ -71,21 +69,6 @@ For OneShot set the frequency to zero too.</string>
      </item>
      <item>
       <property name="text">
-       <string>HPWM</string>
```
peabody124: Thanks for catching this.

ground/gcs/src/plugins/config/outputchannelform.cpp
peabody124: @mlyle made me think more about this. the firmware should allow longer pulses, even in syncpwm because I _think_ we do not disable the timer until the pulses are low. So essentially it will skip updates appropriate. Let me check this on hardware/review the code today. If that is the case we can remove this check entirely and I apologize for triggering this thread.

DTFUHF: No prob, I will update the other things in the time being. SyncPWM mode
could just be set to UAVO value limit.

mlyle:
A couple other things I noticed, in addition to the top things not taking effect until after 'apply':

neutral

Even after apply, and min and max have changed, the neutral value can be selected between 1000 and 2000.  I think there are some issues if the resultant clipped values min==max.

grid

mlyle:
It's not just if min==max, actually.  The range of the neutral is not always getting updated.  I don't understand when it does and when it doesn't.

PR #1645, opened 2015-05-27
gps: enable SBAS/DGPS, and select correct dynamics

mlyle:
FIRST: The dynamic model we were using is the <1g, lightweight dynamics air model.  This is a position filter that tries to figure out how we could possibly be moving.  Normal, full-scale fixed wing airplanes are suggested to use the 2G model, and I've selected this.  We are still more agile than this and should possibly consider using an even more aggressive dynamics model.

Note that this is a tradeoff-- the GPS having an aggressive dynamics model means we will have a less accurate position when sitting / completely still.  But the GPS will be able to react to aircraft movements in a realistic amount of time and will not introduce as much latency and rounding on observations.

SECOND: SBAS/WAAS/DGPS was disabled on Neo-7 and Neo-8 modules.  This is code inherited from autoquad, and no other projects do it (checked Ardupilot, Cleanflight, Openpilot).  There is no vendor update or substantiation of a problem with SBAS with Neo-7 modules.  Note the openpilot platinum receiver is a Neo-7 module, so it can't be completely broken.  This can't be good for our navigation performance.  The code has been changed to explicitly enable SBAS.

### DIFF
--- a/flight/Modules/GPS/ubx_cfg.c
+++ b/flight/Modules/GPS/ubx_cfg.c
@@ -166,7 +166,7 @@ static void ubx_cfg_set_mode(uintptr_t gps_port) {
         0x00,                  // length msb
         0b0000101,             // mask LSB (fixMode, dyn)
         0x00,                  // mask MSB (reserved)
-        0x06,                  // dynamic model (6 - airborne < 1g)
+        0x07,                  // dynamic model (7 - airborne < 2g)
         0x02,                  // fixmode (2 - 3D only)
         0,0,0,0,0,0,0,0,
         0,0,0,0,0,0,0,0,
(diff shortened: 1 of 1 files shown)


mlyle:
I will have test on this in the next couple hours, but only on Neo-8M.

mlyle:
Test looks wonderful.  Despite increased dynamics (less effective GPS filtering) and poor sky view in my backyard because of my house and trees, it stayed within a random walk that was 1/4th the size seen previously, and was 1/5th the distance to the position as shown by google maps "satellite" view.   Please, please, please pull. :p

flight/Modules/GPS/ubx_cfg.c
```diff
@@ -263,12 +263,22 @@ static void ubx_cfg_poll_version(uintptr_t gps_port) {
...
+        // 10Hz for ver 8
+        // TODO: 5 Hz if using GLONASS
+        ubx_cfg_set_rate(gps_port, (uint16_t)100);
+    } else if (ver == 7) {
+        // 10Hz for ver 7
         ubx_cfg_set_rate(gps_port, (uint16_t)100);
 
-        // SBAS screwed up on v7 modules w/ v1 firmware
-        ubx_cfg_set_sbas(gps_port, 0);
+        // History: Code here used to disable SBAS.  A survey of other projects
```
peabody124: No need to keep this comment or historical code. We have it in git history if needed. Just delete if we are going to.

peabody124:
This looks good to me once comment is cleaned up, but I'd like to get a flight report of it using default settings to make sure things still hold reasonably. I expect it to be fine.

peabody124:
Good catch on the SBAS by the way. I was trusting the Autoquad comments for the v1 Ublox7 having issues as that guy is pretty good, but definitely didn't mean to exclude Ublox8 from SBAS.

We _might_ want to add an option for optionally toggling SBAS though. I've heard some people really not liking it because when you gain and lose an SBAS correction (which apparently will happen in flight) it causes a substantial deviation in the location estimate. I believe we can see this in the GPS logs in the status going to and from "Diff3D"

mlyle:
Removed comment-- I had it there just in case someone had a problem and could see what it might be sifting through the code.  I really don't think it will be an issue because OP and others sell Neo-7 modules and don't have anything like that in their codebase.

In a subsequent pull I will add configuration tunables for whether to use GLONASS (everyone else seems to, including DJI and other commercial vendors), and whether to use SBAS.  I think the only reason you'd want to turn SBAS off is if you're at the edge of coverage and there's a geomagnetic storm-- which for example dumped availability down to 85% in Canada

mlyle:
I'll gather a flight test in the next few hours.

mlyle:
It works.  It's twitchier-- some of that was wind during my flight, some is almost certainly the dynamics changes.  There's a need for further work in the INS and deadbands to fix that.



North/East position according to GPS vs. time.

peabody124:
Looks good. It might take a bit to optimize other parameters to exploit this, but seems like a good move an we shouldn't wait for them.

BrainFPV:
@mlyle I think what happens when you enable the 2G model is that there is less lag between the estimated an the true velocity due to different filtering inside the uBlox module. However, we really should be using the 1G model. Apparently, its is possible to use the 1G model but configure the filter to have less lag. I will contact uBlox and try to find out how to do this, as it isn't documented.

BrainFPV:
Another thing we could experiment with is to disable GLONASS and use 18Hz updates using GPS only instead.

mlyle:
Why should we be using the 1G model?  We can clearly pull much more than 1G.

PR #1646, opened 2015-05-28
Urgent: Revert "vtolpathfollower: use peabody's suggestion for norm"

mlyle:
This reverts commit 89ca09988e1df3e7d1ac6ed7b4e2bd4367135740.

We can't do it this way because it ends up normalizing the vertical component of the vector-- e.g. breaking altitude hold.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_control.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_control.c
@@ -105,9 +105,8 @@ static float vtol_magnitude(const float *v, int n)
  * @param[in] desired the setpoint of the system
  * @param[out] the resultant error vector desired-actual
  * @param[in] normalize True if it's desired to normalize the output vector
- * @returns the norm, for fun.
  */
-static float vtol_calculate_distances(const float *actual,
+static void vtol_calculate_distances(const float *actual,
 	const float *desired, float *out, bool normalize)
 {
 	out[0] = desired[0] - actual[0];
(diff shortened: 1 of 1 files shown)


mlyle:
That is, breaking altitude hold/throttle control during navigation.  It robs us of half of our control authority (which isn't really sufficient to begin with) during position hold or following.

mlyle:
Confirmed with flight test-- flyaway to space over hot, updrafty parking lot doesn't happen without this code.

peabody124:
I'm trying to understand what the relevant change is. Returning the normalization parameter from the function itself and not recomputing it seems benign. Is it the unrelated change on line 249?

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -330,8 +325,10 @@ static int32_t vtol_follower_control_simple(const float dT,
 
 	/* Calculate the difference between where we want to be and the
 	 * above position */
-	float horiz_error_norm = vtol_calculate_distances(cur_pos_ned,
-		hold_pos_ned, errors_ned, true);
+	vtol_calculate_distances(cur_pos_ned, hold_pos_ned, errors_ned, 0);
+
+	float horiz_error_mag = vtol_magnitude(errors_ned, 2);
```
peabody124: Ok. Or is this the important change. The fact that using vtol_calculate_distances is doing a 3d distance with normalization? Should it be, or should it also take a dimension parameter?

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -251,7 +246,7 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
 
 		// Waypoint heights are thus treated as crossing restrictions-
 		// cross this point at or above...
-		if (criterion_altitude || current_leg_completed) {
+		if (criterion_altitude) {
```
peabody124: or is this this important change?

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Basically, the change you had me make was to take the difference, calculate the norm and normalize, before scaling back up with a deadband on the norm.

My code instead takes the difference, calculates the magnitude of the first 2 elements, and adjusts the deadband.  The Z component should not be involved in normalization of the vector or the deadband.

Issue #1629, opened 2015-05-19
GCS: Default log output on OSX doesn't work


mlyle:
If you start GCS by double clicking on it or open "Tau Labs GCS.app" at the command line, the default save location for logs is Macintosh HD.  By default users do not have write permission to this location on OSX so the log fails, silently.

The default should be changed and if there is a failure opening the log for writing, a dialog should pop up to inform the user of this fact.

elafargue:
This is now fixed.

PR #1649, opened 2015-06-02
Fixed logging directory so it defaults to HOME env variable + fixed error prompt

tgreer:
Tested on OS X by running Start Logging and prompt opens in relevant HOME dir.

Also fixed error prompt if GCS can't write to the file location/name specified. - Tested by trying to open logging file in Macintosh HD.
screen shot 2015-06-02 at 21 06 11
screen shot 2015-06-02 at 21 17 16

Thanks to mlyle


mlyle:
Expected to fix #1629

solidgoldbomb:
jenkins: add to whitelist

mlyle:
Confirmed to work (both error messages and default save location) on windows.  I think this should merge.

PR #1651, opened 2015-06-03
gcs: add autotune to modules tab

mlyle:
I agree the enabling button should stay at the bottom of the current autotune page.  But it's nice to see when you're enabling modules whether or not it's enabled (so you don't blow through resources).. and it has caused some user confusion / support requests as to why autotune is not in the modules list.

(Confused me as a new user, as well).


elafargue:
Wouldn't it be better make the checkbox disabled and add a UI element telling users to go to the Autotune tab to modify it ?

mlyle:
OK, changed to have a disabled checkbox (that reports the actual state.  Tooltip reports the correct thing.

autotunemod

peabody124:
Looks good to me.

PR #1652, opened 2015-06-04
stabilization: Add some integral.

mlyle:
Surveyed the autotune settings.  This is very conservative and doesn't
look like it would overcontrol anything in the autotune thread-- but
might be high enough to allow people to do a good autotune.


mlyle:
This corresponds to 20/15/0 P/I/D for pitch/roll rate loops.

peabody124:
Totally reasonable to me and will give better initial performance. Any objections?

DTFUHF:
Imo this is a very necessary change.

Trex4Git:
I also thing is a very good idea.

In my first test with my quad and the standard pid's  my quad was drifting in Leveling mode, and i thought the i have to use trims on my taranis. 
But after some tests, and using also some small Ki it became sample without trim.

So especially for beginners and for autotuning it should be very helpful.

Issue #366, opened 2013-02-27
Deprecate classic PFD


kubark42:
The QML PDF is more advanced in functionality and takes less processor power. Any objections to retiring the former PFD?

PR #1654, opened 2015-06-05
GCS: remove a lot of stuff

mlyle:
* First, possibly most controversial: removal of the QML PFD.  Spending resources to maintain two PFDs, and to have two in the user interface when the UI feels large, bloated, and slow seems wasteful.
* Second, remove osgearthview.  Not built by default, almost no activity in 2 years.
* Third, there's a lot of flags to "use opengl" on widgets that appear to be copy and pasted and have no effect.  Remove these.  I'd hope this would be relatively uncontroversial.

This just leaves one thing that explicitly uses GL in GCS-- the model viewer.  It doesn't work in a purely ANGLE or dynamic GL QT, but I hope to just dynamically disable it if an appropriate driver isn't available.


elafargue:
100% OK on the QML PFD, as far as I am concerned. Same for OSGEarthView: cool looking, but does is have an actual use?

brat002:
Agreed.

peabody124:
I'd prefer to keep them, actually. In the long run the QML PFD might be more easily extensible, and works fine. I've found OSGEarth useful when doing simulations as a visualization tool. It's disabled by default.

mlyle:
Then should we remove the non-QML PFD?

If OSGEarth is going to stay in there, it should build by default so that it doesn't break.   Or we should pick up the newer code.  It doesn't really look like the UI is encapsulated correctly the way it's written.

peabody124:
So we've had debates in the past about removing both PFDs (see #366). From an agile perspective, trashing the less used one ASAP is a great idea. My hesitation largely comes from the fact that the ability to switch back and forth has been useful when we've had various bugs etc. If anything I also prefer the non-QML one - historically we've sometimes had performance issues with the QML based one, but as the Qt versions have progressed that has lessened. Currently the configuration files even use different ones - I believe the dev config uses non-QML but the non-dev uses QML.

In the longer run, I think the QML based approach is the more future-oriented approach and makes it easier to add new features (since most of them are coded in a way that is easy for UI people to write. If we were to trash just one I'd say it should be non-QML based. It's just whenever there are packaging issues with the QML things .... well a fallback is useful.

**OSG**
The issue with building it by default is it actually pulls in a lot of dependencies (e.g. OSG) that only a handful of dev used. I can try getting it going again and see what happens. You'd probably enjoy it for simulations since it easily lets you see the flight path and attitude etc.

If you really want to drop it, I can always maintain a branch with it though.

mlyle:
Still testing this a little.  I notice PFD QML doesn't display as much GPS info and is a little blurry.  Will be working on that in subsequent pulls, but don't want to get too much functional stuff in here.

mlyle:
So I'm happy with this / would like this to land.  Only known regressions are the blurriness on the scales and the fact we can't access the satellite count from QML, which I'd like to work on going forward.  There is a noticeable improvement in GCS load time, which surprises me.

mlyle:
This now represents all the change, with the exception of one library change, needed to build on the Tver-Soft Dynamic mingw64 (for i386/32 bit builds) on Windows.   It still builds in the old toolchain, slims out of a lot of duplicative/unused code, and makes things noticeably slightly faster on startup.

elafargue:
Any way we can backport the old PFD artwork to the QML PFD ? I have not looked at the code for the QML PFD, but on the old one, it was just a matter of naming SVG elements the right way and they would get animated/updated:
     - Background: background
     - Foreground: foreground (contains all fixed elements, including plane)
     - earth/sky : world
     - Roll scale: rollscale
     - compass frame: compass (part of the foreground)
     - compass band : compass-band
     - Home point: homewaypoint
     - Next point: nextwaypoint
     - Home point bearing: homewaypoint-bearing
     - Next point bearing: nextwaypoint-bearing
     - Speed rectangle (left side): speed-bg
     - Speed scale: speed-scale.
     - Black speed window: speed-window.
     - Altitude rectangle (right site): altitude-bg.
     - Altitude scale: altitude-scale.
     - Black altitude window: altitude-window.
     - GCS Telemetry status arrow: gcstelemetry-XXXX
     - Telemetry link rate: linkrate
     - GPS status text: gps-txt
     - Battery stats: battery-txt

BrainFPV:
I also vote for keeping the QML PFD, in the long run, it should replace the non-QML one. Also, the `mocap` GCS plugin is very useful to have for development, so I don't think it should be removed.

Instead of removing many plugins from the GCS, I wonder if it would make sense to create a minimal GCS target with only a minimal set of plugins.  This GCS would be less resource hungry and could be used by regular users to configure their vehicles. I know we have a `LIGHTWEIGHT_GCS` option, but I think it still includes most of the plugins.

mlyle:
I think they might already be using the same artwork with a few more things turned on.  The QML files in ground/gcs/share/pfd load pfd.svg, and move elements around and add text.  I believe this is the same file loaded by the normal pfd.

I really like the idea of doing mocap, but it's my understanding that the project member with access to the specific mocap room that's written for has not been involved in Tau for a couple of years; the code has sat and it's not even clear it would still work.  If someone is using or could reasonably use this mocap module sometime in the next year, I agree we should keep it around.

If not, ROS has maintained drivers for Motion Analysis Cortex, VICON, and Optitrak rigs.  We can trivially hook that in through our Python API as I add bindings.

elafargue:
Actually, it's a different file - doesn't matter too much really. The QML PDF is a bit more involved than the old one - you need to mix SVG and the QML javascript to achieve the results, vs just creating a SVG with the right element tags, but I guess that's fine. Someone needs to extend the QML and SVG to support the additional elements if we want to have them displayed on the PFD.

How do both PFDs compare performance-wise? The QML PFD adds a javascript layer that will probably be a big performance hog on lower powered machines?

mlyle:
QML PFD actually measures faster; basically, that's how most of rest of the UI renders and there's a lot of acceleration / a pretty high performance javascript vm.

What elements are missing?  I can certainly spend time after this pull on making the QML PFD a little crisper, etc.

There's also a bunch of Inkscape attributes on the SVG that should probably get stripped off.

elafargue:
OK, sounds like we should just go for the QML PFD. I'll shed a tear on my old SVG PFD source code and soldier on ;-) That was one of the oldest GCS plugins, actually!

Don't remove the Inkscape attributes, they make it easier to edit the file in Inkscape for future changes.

peabody124:
Very nice. 8k LOC deleted :D 
@elafargue you think you can review this? I think most of the changes are simple but it touches on a lot of your old code.

elafargue:
Yep, planning to do it during the week-end

elafargue:
ok then: you have my blessing to put the old PFD plugin to rest ;-) Thanks for removing the rest of the code that is unused.

All looks good to me, we can most probably merge this.

The caveat are:
- People will need to update their configurations
- The QML PDF does not display all the same information as the old one, we should work at implementing what is missing.

Comments ?

mlyle:
Yes, that sounds fair.  QML needs work-- not all types in uavo seem to be exposed to it, so I'm planning to figure that out too.

The QML PFD shows a few things the original PFD doesn't-- like navigation points and rates.  But the GPS display is a little messed up.  I've changed the default configuration; people should be able to just select the PFD from the pulldown to get their PFD back.

PR #1655, opened 2015-06-06
WIP: Support RFM22B 915MHz hardware

mlyle:
What happens with openlrs (and perhaps openlrs configuration in general) needs to get smarter, and I need to look at ISM band boundaries and channel computation.  Mostly for discussion.


mlyle:
Will have to wait till I'm back to finish this.

Issue #1656, opened 2015-06-06
"zero gyros while arming" not always working.


ernieift:
A few days ago, i had an issue with my light weight racing copter. After arming and liftoff the copter was spinning and simply uncontrollable. I was outside and couldn't check any variables. Yesterday i had the same issue. Now i have found the reason: "zero Gyros while arming" doesn't work when you use a switch for arming.
You can reproduce the bug this way. Connect your vehicle via usb and move it around for about 10 seconds. You can see in the gcs all gyros have offsets now. If you set your arming settings via stick movements, these offsets were zeros while arming. If you use a switch for arming, the offsets are untouched.

peabody124:
Fixed by #1657

PR #1657, opened 2015-06-06
ARMING: add selectable delay to switch

ernieift:
This adds a selectable delay to arming by switch(+thottle). Zeroing the
gyros while arming the vehicle is not functional in switch mode only.
The result may be a poor flight performance on aggressively tuned
copters. Now you can select between immediately and delayed
arming. In delayed arming the gyro calibration is functional again.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -523,10 +523,10 @@ static bool arming_position(ManualControlCommandData * cmd, ManualControlSetting
 			(cmd->Pitch > ARMED_THRESHOLD);
 			// Note that pulling pitch stick down corresponds to positive values
 	case MANUALCONTROLSETTINGS_ARMING_SWITCH:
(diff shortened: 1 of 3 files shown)


ernieift:
This is the fix for the issue problem #1656 .

mlyle:
This looks good.

Just for discussion-- there's no realistic prospect of fixing the gyro calibration on instant arm?  If not, do we need the GCS to warn the user on combinations of the zeroing and arming setting that don't work?  This is an important but obscure issue that people are likely to miss in any doc we write.

ernieift:
Oh, you just have to wait >7 seconds without any move your copter after power up. But in some scenarios this isn't possible and you're getting a better gyro calibration after a while waiting. The chips warm up a little bit ;).

mlyle:
I'm happy with the note; in favor of merging this.

peabody124:
Yeah, this is a nice improvement. Thank you @ernieift

PR #1661, opened 2015-06-07
python-ins: Move py INS code out of python subtree

mlyle:
Also add GPL notice to README.


peabody124:
Why?

mlyle:
Because I assume you don't want to relicense it LGPL.

peabody124:
Oh gotcha :) thanks. Ugh, I don't like adding lots of top level directories either though. Good attention to detail though.

@solidgoldbomb any thoughts on organization here?

solidgoldbomb:
I don't think this makes sense as a new top-level directory.

mlyle:
Do you want to move it into flight, because it's generating flight code?

mlyle:
Closed per discussion.

PR #1664, opened 2015-06-09
syncpwm UI updates

mlyle:
This has seen cursory but not systematic test.  The problems I saw were fixed.  Also fixed was a uavo overflow.

### DIFF
--- a/ground/gcs/src/plugins/config/configoutputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configoutputwidget.cpp
@@ -120,6 +120,14 @@ ConfigOutputWidget::ConfigOutputWidget(QWidget *parent) : ConfigTaskWidget(paren
         res->clear();
         res->addItems(resolutions->getOptions());
         addWidget(res);
+        connect(res, SIGNAL(currentIndexChanged(int)), this, SLOT(refreshWidgetRanges()));
+    }
+
+    QList<QComboBox*>::iterator rateIter;
+    for (rateIter = rateList.begin(); rateIter != rateList.end(); rateIter++) {
+        QComboBox *rate = *rateIter;
+
+        connect(rate, SIGNAL(currentIndexChanged(int)), this, SLOT(refreshWidgetRanges()));
     }
 
     disconnect(this, SLOT(refreshWidgetsValues(UAVObject*)));
--- a/ground/gcs/src/plugins/config/configoutputwidget.cpp
+++ b/ground/gcs/src/plugins/config/configoutputwidget.cpp
@@ -255,6 +255,8 @@ void ConfigOutputWidget::assignOutputChannels(UAVObject *obj)
         quint16 type = actuatorSettingsData.ChannelType[outputChannelForm->index()];
         outputChannelForm->setType(type);
     }
+
+    refreshWidgetRanges();
 }
 
 /**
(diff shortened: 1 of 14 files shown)


mlyle:
This supersedes #1642

DTFUHF:
Much better.

The buffer overflow - I thought the number of elements field in a UAVO enum was meant to describe the number of elements in that enum. Is this the case? Or is this fix used to fix a separate bug.

mlyle:
@DTFUHF -- thanks for the initial refactor.  I see why things were the way they were-- the UI code for this output actuator stuff is very fiddly and hard to make do the right thing.  Half of my change is hack, unfortunately. :/

Re: buffer overflow-- It's not the number of possible values for the enum, but the number of instances of that thing in an array.  I missed it in code review and just happened to get lucky to see it scroll by on compile output.

(I think I'm planning to eliminate GCS warnings and try and make them fatal so that it's harder to do something like this).

ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -455,15 +465,78 @@ void ConfigOutputWidget::refreshWidgetsValues(UAVObject * obj)
...
+                //
+                QComboBox* ccmb = rateList.at(i);
+                quint32 timerFreq = timerStringToFreq(ccmb->currentText());
+
+                QComboBox* res = resList.at(i);
+                quint32 timerRes = res->currentIndex();
+
+                double maxPulseWidth = 0, uavoMaxPulseWidth = 0;
+
+                // Saturate at the UAVO's maximum value
```
peabody124: This comment/variable is inaccurate I think. The UAVO max is in microseconds. However, you are actually calculating the maximum number of microseconds on the physical timer, assuming 16 bits (some aren't, but a reasonable conservative assumption).

So it is really timerMaxPeriodUs

ground/gcs/src/plugins/config/configoutputwidget.cpp
```diff
@@ -455,15 +465,78 @@ void ConfigOutputWidget::refreshWidgetsValues(UAVObject * obj)
...
+                uavoMaxPulseWidth = 65535;
+
+                if (timerRes == ActuatorSettings::TIMERPWMRESOLUTION_12MHZ)
+                    uavoMaxPulseWidth = uavoMaxPulseWidth / 12;
+
+                if (timerFreq != 0)
+                {
+                    maxPulseWidth = 1000000 / timerFreq;
```
peabody124: and if we are changing names :D could this be timerPeriodUs

ground/gcs/src/plugins/config/outputchannelform.cpp
```diff
@@ -86,14 +86,6 @@ OutputChannelForm::OutputChannelForm(const int index, QWidget *parent, const boo
     ui.actuatorLink->setChecked(false);
     connect(ui.actuatorLink, SIGNAL(toggled(bool)), this, SLOT(linkToggled(bool)));
 
-    // Trigger when autopilot is connected
-    connect(this, SIGNAL(autoPilotConnected()), this, SLOT(onAutopilotConnect()));
```
peabody124: why was this removed? isn't it calling a method on the parent that refreshes the initial values (e.g. neutral?)

mlyle: No.  The parent already gets an autopilot connect event and has it connected.  We were using that to calculate minimum/maximum values, but now the parent does it when it processes the bank.  We don't need to know down here.

PR #1667, opened 2015-06-10
pathplanner: Don't follow invalid waypoints

mlyle:
I've run this through several cycles and it looks happy / like it eliminates the flyaway bug.


peabody124:
1. confirmed with next that having an invalid waypoint triggers a PH and the error state for path planner (so at least it isn't a flyaway potential).

peabody124:
ok and with this setting a middle waypoint to invalid, it still stops at the last valid waypoint but there is no longer an error condition (because we are explicitly handling this and not letting it slip) so the firmware behavior changes seem good.

flight/Modules/PathPlanner/pathplanner.c
```diff
@@ -286,6 +284,30 @@ static void holdLastPosition()
...
+	// Get the activated waypoint
+	WaypointData waypoint;
+	WaypointInstGet(idx, &waypoint);
+
+	// Perhaps should fully validate here..
+	if (waypoint.Mode == WAYPOINT_MODE_INVALID) {
+		return false;
+	}
+
+	if (waypoint.Mode == WAYPOINT_MODE_STOP) {
```
peabody124: Does this stop mode have any meaning? Should we drop it too?

mlyle: I think we should but wasn't sure that was the minimal fix.

mlyle:
Note it still is potentially flyaway because of the GCS half of the problem..

flight/Modules/PathPlanner/pathplanner.c
peabody124: Good point. Maybe add this to #1669

ground/gcs/src/plugins/pathplanner/modeluavoproxy.cpp
```diff
@@ -69,16 +69,25 @@ bool ModelUavoProxy::modelToObjects()
...
+    int x;
+
+    // Figure out the total amount of work to do.
+    int progressMax = myModel->rowCount();
+
+    if (instances > progressMax) {
+        progressMax = instances;
+    }
+
+    for(x=0;x<myModel->rowCount();++x)
```
peabody124: spacing around operators (although i see this was code from L72)

ground/gcs/src/plugins/pathplanner/modeluavoproxy.cpp
```diff
@@ -124,6 +132,25 @@ bool ModelUavoProxy::modelToObjects()
             m_eventloop.exec();
         }
     }
+
+    for (; x < instances; x++) {
+        Waypoint *wp = Waypoint::GetInstance(objManager, x);
+                // shadows wp above
```
peabody124: what does this mean?

ground/gcs/src/plugins/pathplanner/modeluavoproxy.cpp
```diff
@@ -124,6 +132,25 @@ bool ModelUavoProxy::modelToObjects()
             m_eventloop.exec();
         }
     }
+
+    for (; x < instances; x++) {
```
peabody124: could you add a comment that this runs over any pre-existing waypoints the model doesn't correspond to (or not - author's preference :D)

peabody124:
Ok. GCS side changes are good too and test correctly. There are few cosmetic changes that would be nice at the author's discretion, but good to merge.

ground/gcs/src/plugins/pathplanner/modeluavoproxy.cpp
mlyle: It's a for loop without initializing x first.  e.g. continue counting up from where we were.

mlyle:
I'll touch this up in a few minutes.

mlyle:
OK for merge.

PR #1669, opened 2015-06-11
paths: simplify constants/remove "DRIVE"

mlyle:
"FLY" was used even in the GroundPathFollower; DRIVE was never valid.


peabody124:
This seems reasonable, although while changing things can we drop the "STOP" Waypoint type too, since that means nothing.

I'd like to maintain the option for restoring this in the future if we ever do things differently when driving a path versus flying it - but for now this was just cruft.

mlyle:
Passes dev testing with STOP removed.  Would be nice if someone could test loading/saving from flash before merge, but it's probably OK (OK in sim).

PR #1671, opened 2015-06-13
Prevent flyaway from invalid curves

mlyle:
Both code to validate things in GCS, and to fix up bad paths on the flight side.

### DIFF
--- a/flight/Libraries/paths.c
+++ b/flight/Libraries/paths.c
@@ -7,7 +7,7 @@
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2012.
  * @brief      Path calculation library with common API
- * 
+ *
  * Paths are represented by the structure @ref PathDesired and also take in
  * @ref PositionActual.  This library then computes the error from the path
  * which includes the vector tangent to the path at the closest location
@@ -40,15 +40,15 @@
 #include "pathdesired.h"
 
 // private functions
-static void path_endpoint(const float * start_point, const float * end_point, 
+static void path_endpoint(const float * start_point, const float * end_point,
                           const float * cur_point, struct path_status * status);
-static void path_vector(const float * start_point, const float * end_point, 
+static void path_vector(const float * start_point, const float * end_point,
                         const float * cur_point, struct path_status * status);
-static void path_circle(const float * center_point, float radius, 
+static void path_circle(const float * center_point, float radius,
                         const float * cur_point, struct path_status * status,
                         bool clockwise);
 static void path_curve(const float * start_point, const float * end_point,
-                       float radius, const float * cur_point, 
+                       float radius, const float * cur_point,
                        struct path_status * status, bool clockwise);
 
 /**
@@ -203,7 +203,7 @@ static void path_vector(const float *start_point,
 static void path_circle(const float * center_point,
                         float radius,
                         const float * cur_point,
-                        struct path_status * status, 
+                        struct path_status * status,
                         bool clockwise)
 {
 	float diff_north, diff_east;
--- a/flight/Libraries/paths.c
+++ b/flight/Libraries/paths.c
@@ -290,7 +290,7 @@ static void path_curve(const float * start_point,
 		if (fabsf(radius) < min_radius) {
 			// Whoops! Radius was not close.  Convert to (nearly)
 			// straight line.
-			radius=min_radius * 1000;
+			radius = min_radius * 1000;
 		}
 	}
 
--- a/flight/Libraries/paths.c
+++ b/flight/Libraries/paths.c
@@ -44,11 +44,11 @@ static void path_endpoint(const float * start_point, const float * end_point,
                           const float * cur_point, struct path_status * status);
 static void path_vector(const float * start_point, const float * end_point, 
                         const float * cur_point, struct path_status * status);
-static void path_circle(const float * center_point, const float radius, 
+static void path_circle(const float * center_point, float radius, 
                         const float * cur_point, struct path_status * status,
                         bool clockwise);
 static void path_curve(const float * start_point, const float * end_point,
-                       const float radius, const float * cur_point, 
+                       float radius, const float * cur_point, 
                        struct path_status * status, bool clockwise);
 
 /**
@@ -134,7 +134,6 @@ static void path_endpoint(const float *start_point,
 	// Compute direction to travel
 	status->path_direction[0] = diff_north / dist_diff;
 	status->path_direction[1] = diff_east / dist_diff;
-
 }
 
 /**
@@ -202,7 +201,7 @@ static void path_vector(const float *start_point,
  * @param[out] status Structure containing progress along path and deviation
  */
 static void path_circle(const float * center_point,
-                        const float radius,
+                        float radius,
                         const float * cur_point,
                         struct path_status * status, 
                         bool clockwise)
@@ -210,6 +209,10 @@ static void path_circle(const float * center_point,
 	float diff_north, diff_east;
 	float cradius;
 	float normal[2];
+
+	if (radius < 0.10f) {
+		radius = 0.10f;		// Never try a circle less than 10cm
+	}
 
 	// Current location relative to center
 	diff_north = cur_point[0] - center_point[0];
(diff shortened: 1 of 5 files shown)


mlyle:
There's a remaining flight bug with 180 degree arcs where you get stuck in an infinite circle.  Will be in a subsequent pull.

mlyle:
@elafargue can you look at this please when you get a chance?

flight/Libraries/paths.c
```diff
@@ -264,15 +267,37 @@ static void path_circle(const float * center_point,
...
+		} else {
+			radius -= 0.05f;
+		}
+
+		radius *= 1.005f;
+
+		if (fabsf(radius) < min_radius) {
+			// Whoops! Radius was not close.  Convert to (nearly)
+			// straight line.
+			radius=min_radius * 1000;
```
guilhermito: spaces around "="

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+
+    for (int i = 0; i < rowCount(); i++) {
+	PathPlanData * row = dataStorage.at(i);
+
+        bool dirty = false;
+
+        struct FlightDataModel::NED thisNED = getNED(row);
+
+        if (i == 0) {
+            row->mode        = Waypoint::MODE_VECTOR;
```
guilhermito: looks like an extra tab

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+		    break;
+		default:
+		    break;
+	    }
+
+        } else {
+	    double distance = sqrt(pow(thisNED.North - prevNED.North, 2) +
+		pow(thisNED.East - prevNED.East, 2));
+
+	    if (distance > 600) {
```
guilhermito: is this valid for vector mode to?

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+		// If the distance is more than 600m, presume it's invalid.
+		// Distances larger than this begin to become problematic
+		// with local tangent plane approximation.
+
+		thisNED = prevNED;
+		setNED(row, thisNED);
+		distance = 0;
```
guilhermito: Can you explain the correction being made here? are you defining this waypoint to be the same as the previous one or deleting it?

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+
+		showErrorDialog("Waypoint corrected",
+		        "Over-long leg shortened.");
+
+                dirty=true;
```
guilhermito: spaces

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+	    switch (row->mode) {
+                case Waypoint::MODE_CIRCLELEFT:
+                case Waypoint::MODE_CIRCLERIGHT:
+                    if (row->mode_params < (distance / 2 + 0.1f)) {
+                        row->mode_params = distance / 2 + 0.2f;
+
+                        showErrorDialog("Waypoint corrected",
+                                "Radius of circle increased to minimum");
+
+                        dirty=true;
```
guilhermito: spaces

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+                    } else if (row->mode_params > 300) {
+                        row->mode_params = 300;
+
+                        showErrorDialog("Waypoint corrected",
+                                "Radius of circle decreased to maximum");
+
+                        dirty=true;
```
guilhermito: spaces

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -452,8 +456,118 @@ void FlightDataModel::showErrorDialog(const char* title, const char* message)
...
+                default:
+                    break;
+            }
+
+            if (dirty) {
+                /* Let anyone listening know we changed it */
+                QModelIndex leftIndex, rightIndex;
+
+                leftIndex = this->index(i, LATPOSITION);
+                rightIndex = this->index(i, LASTCOLUMN-1);
```
guilhermito: How does this relate to the set of data that was changed?

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -470,26 +584,30 @@ void FlightDataModel::readFromFile(QString fileName)
     QDomDocument doc("PathPlan");
     QByteArray array=file.readAll();
     QString error;
+
+    file.close();
```
guilhermito: File open success should also be checked

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
```diff
@@ -570,12 +689,14 @@ void FlightDataModel::readFromFile(QString fileName)
             data->lngPosition = wpLLA[1];
             data->altitude = wpLLA[2];
 
-        beginInsertRows(QModelIndex(),dataStorage.length(),dataStorage.length());
-        dataStorage.append(data);
-        endInsertRows();
+            beginInsertRows(QModelIndex(),dataStorage.length(),dataStorage.length());
```
guilhermito: spaces after comas

ground/gcs/src/plugins/pathplanner/flightdatamodel.h
```diff
@@ -79,8 +80,13 @@ class PATHPLANNER_EXPORT FlightDataModel : public QAbstractTableModel
     //! Replace a model data with another model
     bool replaceData(FlightDataModel *newModel);
 
+    //! Prevent validation/correction of data
+    void pauseValidation(bool pausing);
+
 private:
     QList<PathPlanData *> dataStorage;
+    
+    bool valPaused;
```
guilhermito: class variable not initialized

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
mlyle: Woops!  Extra line I had copied for reference from below actually.

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
mlyle: Yes.  We use something called the local tangent plane which treats the earth like it's a flat plane around the home location.  It gets increasingly inaccurate as we get away from home.  So validation rejects overly-long legs right now until that code can catch up (we'll probably want to warn in that case, too, because half a kilometer is an awful long leg).

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
mlyle: I'm setting it to be the same.  This both works with the loop factoring of iterating forward over the points (deleting a point from the middle of an iterator is problematic) and allows the user to fix the waypoint.  I added a comment to this effect.

While I was just looking at it, it occurred to me that I should do this test for the first iteration of the loop, so I took all of this out of an else block.

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
mlyle: If we change any of the NED stuff, latitude, longitude, and altitude change too.  So many of the changes change 6 columns.  Some of the changes change the mode or params column.  So I considered it safer to just let UI know that the whole row may have changed.

ground/gcs/src/plugins/pathplanner/flightdatamodel.h
mlyle: Very good catch!

ground/gcs/src/plugins/pathplanner/flightdatamodel.cpp
mlyle: Added.

mlyle:
OK, that review feedback is incorporated.  Thanks for the feedback!

I also found a section where I had tabs and shouldn't have, so I replaced those with spaces.

guilhermito:
Has the code been flight tested?

mlyle:
The original issue was produced in the simulator-- this has been flown in the posix simulator and confirmed to do the right thing for invalid curves (convert to lines).  The GCS code has been tested extensively in combination with the simulator.

There is a remaining bug where 180 degree arcs become an infinite-loop-of-circling, but I'm planning to handle that in another pull.  It takes some careful thought about how not to overcomplicate the math and is a separate issue.

I'm working on INS changes also at the moment, so I am not able to flight test path following on my nav platform.

PR #1673, opened 2015-06-14
uavobjectparser: improvements to validation, error messages

mlyle:
It's less confusing now when something is wrong with a uavo xml file.

### DIFF
--- a/ground/uavobjgenerator/uavobjectparser.cpp
+++ b/ground/uavobjgenerator/uavobjectparser.cpp
@@ -170,12 +170,12 @@ QString UAVObjectParser::parseXML(QString& xml, QString& filename)
 
         // Process child elements (fields and metadata)
         QDomNode childNode = node.firstChild();
-        bool fieldFound = false;
-        bool accessFound = false;
-        bool telGCSFound = false;
-        bool telFlightFound = false;
-        bool logFound = false;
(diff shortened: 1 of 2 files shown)


ground/uavobjgenerator/uavobjectparser.cpp
```diff
@@ -117,6 +118,26 @@ bool fieldTypeLessThan(const FieldInfo* f1, const FieldInfo* f2)
...
 
 /**
+ * Returns a human-meaningful error message
+ * @param fileName The filename we found an error in
+ * @param errMsg A human readable description of what's wrong
+ * @param errorLine The line number where this problem occurred
+ * @param errorCol The column number
+ * @returns A formatted error string.
+ */
+QString UAVObjectParser::genErrorMsg(QString& fileName,
```
guilhermito: operator should be next to variable name not type

BrainFPV:
LGTM

ground/uavobjgenerator/uavobjectparser.cpp
mlyle: I agree-- I like it that way.

Unfortunately, that whole file, and all of the QT code we have puts the referencing operator next to the type name.

mlyle:
Cool, thanks for the merges.

PR #1674, opened 2015-06-14
FIX: GCS airspeed sensor selection

BrainFPV:
Fixes the airspeed sensor type selection, which had been completely broken.

### DIFF
--- a/ground/gcs/src/plugins/config/configmodulewidget.cpp
+++ b/ground/gcs/src/plugins/config/configmodulewidget.cpp
@@ -102,12 +102,6 @@ ConfigModuleWidget::ConfigModuleWidget(QWidget *parent) : ConfigTaskWidget(paren
     connect(ui->cbVoltagePin, SIGNAL(currentIndexChanged(int)), this, SLOT(toggleBatteryMonitoringGb()));
     connect(ui->cbCurrentPin, SIGNAL(currentIndexChanged(int)), this, SLOT(toggleBatteryMonitoringGb()));
 
-    // Link the fields
-    addUAVObjectToWidgetRelation(airspeedSettingsName, "GPSSamplePeriod_ms", ui->sb_gpsUpdateRate);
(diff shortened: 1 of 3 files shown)


mlyle:
Awesome!  I've wondered why this was.

I know nothing about the specific pitot sensors out there.  Is it safe to assume if someone checks that box they want the eagletree sensor and not the diydrones options we support?

ground/gcs/src/plugins/config/configmodulewidget.cpp
```diff
@@ -620,21 +618,46 @@ void ConfigModuleWidget::updateAirspeedGroupbox(UAVObject *obj)
...
+ */
+void ConfigModuleWidget::enableAirspeedTypeGPS(bool checked)
+{
+    if (checked){
+        AirspeedSettings *airspeedSettings;
+        airspeedSettings = AirspeedSettings::GetInstance(getObjectManager());
+        AirspeedSettings::DataFields airspeedSettingsData;
+        airspeedSettingsData = airspeedSettings->getData();
+        airspeedSettingsData.AirspeedSensorType = AirspeedSettings::AIRSPEEDSENSORTYPE_GPSONLY;
+        airspeedSettings->setData(airspeedSettingsData);
```
guilhermito: Aren't you entering in an endless loop here, this is called whenever the airspeedsettings are updated and you are updating them inside this.

BrainFPV: Fortunately this isn't the case; the `clicked` signal only gets emitted when the user clicks the checkbox, not when it is changed using `setChecked`.

BrainFPV:
@mlyle the default Pitot selection is arbitrary but was biased by the fact that I have one of the ET sensors ;). We could change the default to the `MPXV7002` sensor, which can be bought from [3DR](https://store.3drobotics.com/products/airspeed-kit-with-mpxv7002dp).

mlyle:
Whatever you think is best.  I'm OK with this to merge.

Issue #1647, opened 2015-05-30
Expo in Advanced Stabilization


fujin:
Hi,

I think it would be cool to have an expo box in advanced stabilization tab beneath the degrees/sec value for pitch/roll/yaw.

Coming from previous flight controllers, I pretty much decided how much I wanted to roll (known in c.f. as RC rate, a command multiplier) and then reduced sensitivity with p/r/y rates (stick attenuation).

I was able to replicate the in-FC feeling that I am accustomed to by setting my p/r to 720deg/sec (two rolls per second) and y to 360deg/sec (one per) then setting P/R stabilizationsettings expo to 45%, Y expo to 70%.

cheers,

--aj

Trex4Git:
As i am just beginning with development for taulabs and have no experience with QT GUI development yet, only an early gui test: 
![gcs_expo_test](https://cloud.githubusercontent.com/assets/9142182/7982680/7a2b7dbe-0ab9-11e5-87b7-41292fa768bb.png)

Trex4Git:
A small update. the expo settings located close to the range limits, so little bit more compact.
But i am not shure how the interest of others is in it, so i did not make a pull request yet.

![gcs_expo_test2](https://cloud.githubusercontent.com/assets/9142182/8002495/b498e732-0b6f-11e5-9db2-38f3a89e243b.png)

fujin:
Nice, the 2nd definitely looks more compact.

Trex4Git:
Just to learn a little bit more about GCS coding and QT GUI development i implemented a test version of dynamic plots, that shows the combination of the expo value together with the full stick value on y-axis. And the normalized stick input on x-axis. (Layout got rearranged, line width, color etc only a early test; so still work to do, just showing the principal idea).

Plots are dynamically changed when the values in the boxes are changed.
Not compact anymore, but may be useful.

But still no feedback from the main developers on this topic.

![gcs_expo_test3](https://cloud.githubusercontent.com/assets/9142182/8025277/df8b3b02-0d4f-11e5-86ae-bf894e9f67cf.png)

Trex4Git:
A small optical update and fixed rearranged Layout issue.
![gcs_expo_test4](https://cloud.githubusercontent.com/assets/9142182/8045053/0982865e-0e31-11e5-92b6-0aaea40643b8.png)

I am interested in opinions regarding the design. Which is preferred, and does someone has a good idea for another layout?
The shown ones are just my first basic ideas, because of easy integration.

[...]

PR #1676, opened 2015-06-17
GCS : Expo settings and plots in Advanced Stabilization

Trex4Git:
As there was the request for doing the Expo setup not only in the ObjectBrowser but also in the configuration widget (Stabilization/Advanced) and i wanted to take take a look at the GCS development and also QT i tried to implement this functionality.

The issue with the request: https://github.com/TauLabs/TauLabs/issues/1647
Also some different versions of the implementation including screenshots  are presented there.
And also mentioned earlier here: https://github.com/TauLabs/TauLabs/pull/1493#issuecomment-75871884

The scope of this PR:
* making the RateExpo and HorizonExpo setting available in the config widget
* adding plots, that show the result
	* here i made the decision to plot the stick input not only scaled by the expo function, but also by the "Full Stick" setting. As Taulabs uses physical units i think its more informative this way then only plotting the relative scaling like it is done in e.g. in the Taranis or other FC software
* modifying the layout to fit the new elements

For the Rate mode (inner loop) i think its easy to understand, but for the HorizonExpo setting shown in the outer loop it gets a bit more complicated.

It tried to visualize the information how the the Horizon modes works in a compact format (based on how i understand it from reading the sources). 
The HorizonExpo setting has different effects:
* the mixing for the transition from Attitude to Rate mode (for stick stick inputs from 0 to the threshold value) is  expo scaled
* the "Horizon Attitude" input data is also expo scaled  with the HorizonExpo setting (no expo scaling i normal Attitude mode)
* the "Horizon Rate" input data is expo scaled with the HorizonExpo setting (scaling in normal Rate mode is done with the RateExpo setting)

I grouped the HorizonExpo setting and plot to the outer loop panel in the UI layout, but for normal attitude mode the curves might be misleading, as the shown data applies only for normal attitude mode, if the HorizonExpo is set to 0 (which is not the default setting). With using labels on the plot and tooltips i tried to give a better understanding what is shown there.
This plot shows data only for the Horizon mode, not the Attitude mode: 
* For each axis there are 2 curves. One for the"Horizon Attitude" part (horizon angle) and in darker color one for the "Horizon Rate" part (horizon rate).
* The curves are scaled both with HorizonExpo.
* There are markers on the curves which show the threshold point.
	* For stick inputs above its purely  the "Horizon Rate" mode  with expo scaling by  HorizonExpo.
	* For stick inputs below there is a dynamic transition from "Horizon Attitude" to "Horizon Rate" mode. In the example screenshot you can see that for different HorizonExpo settings the threshold value changes its position regarding the normalized stick input on x-axis.

gcs_expo_merge_10

I think its not as easy to understand this plot as the plot for the inner loop, but advanced users should can get information from it that can help tuning the Horizon mode.
And as this is in the advanced tab, i think it might be OK to keep the HorizonExpo settings and plot there.

To reduce the information in the plots, curves can be disabled/enabled with a click on each legend entry. Maybe especially useful with the 6 curves in the HorizonExpo plot.


Trex4Git:
I have not tested the Horizon mode myself and don't know if its worth to do the following:
* It might be useful to add another parameter to the UAVObject: AttitudeExpo
* This should be applied like the name says to normal Attitude mode, like RateExpo is used for normal Rate mode. 
* Then in horizon mode:
	* Use  this AttitudeExpo for the "Horizon Attitude" instead of the HorizonExpo. So it is like normal Attitude mode.
	* Use the RateExpo for the "Horizon Rate" instead of the HorizonExpo. So it is like normal Rate mode.
	* And use the HorizonExpo only for scaling the transition from Attitude to Rate mode.
* In GCS Advanced Stabilization UI: 
	* In the Outer loop panel the HorizonExpo settings and plot could be replaced with the AttitudeExpo settings and plot.
	* And the HorizonExpo setting and plot could be moved from the Advanced to the Expert widget.

This might be a cleaner implementation and also gives the user more tuning parameters with less coupled effects of the HorizonExpo setting.

It is at the moment not in the scope of this PR. But if you think it should be, i can add it. 
Otherwise i think,  i (or someone else) could do that (if wished) in a later PR.

ground/gcs/src/plugins/config/expocurve.h
```diff
@@ -0,0 +1,104 @@
...
+    double *y_data;
+
+    ExpoPlotElements_t rollElements;
+    ExpoPlotElements_t pitchElements;
+    ExpoPlotElements_t yawElements;
+
+    QMutex mutex;
+
+    //! Inverse expo function
+    double invers_expo3(double y,int g);
```
mlyle: spacing between args -- double y, int g

ground/gcs/src/plugins/config/expocurve.h
```diff
@@ -0,0 +1,104 @@
...
+      QwtPlotCurve Curve;
+      QwtPlotMarker Mark;
+      QwtPlotMarker Mark_;
+      QwtPlotCurve Curve2;
+      QwtPlotMarker Mark2;
+      QwtPlotMarker Mark2_;
+    } ExpoPlotElements_t;
+
+    //! Set label for the stick channels
+    void init(int lbl_mode,int horizon_transistion,int roll_value,int pitch_value,int yaw_value,int roll_max,int pitch_max,int yaw_max,int roll_max2,int pitch_max2,int yaw_max2);
```
mlyle: spacing

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,304 @@
...
+// Inverse expo3 function from /flight
+// @param[in] y   expo output data from [-1,1]
+// @param[in] g   sets the exponential amount [0,100]
+// @return  rescaled expo output to input data
+//
+// The math: http://www.wolframalpha.com/input/?i=Solve%5By%3D%28%28g%2F100%29*x%5E3%2B%28%28100-g%29%2F100%29*x%29%2Cx%5D
+// e.g. y=0.85 expo = 50% : http://www.wolframalpha.com/input/?i=Solve%5B0.85%3D%28%2850%2F100%29*x%5E3%2B%28%28100-50%29%2F100%29*x%29%2Cx%5D
+double ExpoCurve::invers_expo3(double y,int g)
+{
+    double temp1 = pow((2700*pow(g,2)*y + sqrt(7290000*pow(g,4)*pow(y,2)+108*pow((100-g),3)*pow(g,3))),(1.0/3.0));
```
mlyle: spaces around all the operators and commas, break into multiple lines.

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,304 @@
...
+        plot_elements.Curve2.show();
+    }
+
+    if (HorizonTransition != 0) {
+        marker_x = invers_expo3((double)(HorizonTransition/100.0),value);
+        marker_y = max*HorizonTransition/100;
+
+        // additional scaling (*0.985) of positive x marker position for better visual fit with the curve
+        if (mode == 1) {
+            plot_elements.Mark.setValue(marker_x*0.985,marker_y);
```
mlyle: spacing between args and operators

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,304 @@
...
+ * @param roll_max max for initial roll curve
+ * @param pitch_max max for initial pitch curve
+ * @param yaw_max max for initial yaw curve
+ */
+void ExpoCurve::init(int lbl_mode,int horizon_transistion,int roll_value,int pitch_value,int yaw_value,int roll_max,int pitch_max,int yaw_max,int roll_max2,int pitch_max2,int yaw_max2)
+{
+    switch (lbl_mode)
+    {
+        case 0:
+            rollElements.Curve.setTitle("Roll rate (deg/s)");
```
mlyle: do these need tr() around the strings?

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,304 @@
...
+        pitchElements.Mark2.attach(this);
+        pitchElements.Mark2_.attach(this);
+        yawElements.Mark2.attach(this);
+        yawElements.Mark2_.attach(this);
+    }
+    else {
+        CurveCnt = 1;
+    }
+
+    plotDataRoll(roll_value,roll_max,1);
```
mlyle: spaces after commas

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,304 @@
...
+    this->enableAxis(QwtPlot::yRight);
+    this->setAxisTitle(QwtPlot::xBottom, " normalized stick input");
+
+    STEPS = 1000;
+    x_data =  new double[STEPS];
+    y_data =  new double[STEPS];
+
+    double step   = 2*1.0 / (STEPS - 1);
+    for (int i = 0; i < STEPS; i++) {
+        x_data[i] = (i*step) - 1.0;
```
mlyle: This whole function-- spaces after commas and around *, etc.

ground/gcs/src/plugins/config/configstabilizationwidget.cpp
```diff
@@ -221,3 +262,57 @@ void ConfigStabilizationWidget::applyMWRateConvertDialog()
     }
 }
 
+void ConfigStabilizationWidget::showExpoPlot()
```
mlyle: Can this be factored any cleaner to reduce the amount of code duplication?  It's nice to minimize the amount plotted, but not absolutely essential.  Perhaps set flags for what is dirty in these conditionals and then if (horizonRollDirty) { m_stabilization->horizonStickExpoPlot->plotDataRoll...

Also, formatting and operator spacing.

ground/gcs/src/plugins/config/configstabilizationwidget.cpp
```diff
@@ -221,3 +262,57 @@ void ConfigStabilizationWidget::applyMWRateConvertDialog()
...
+{
+
+    if(QObject* obj = sender()) {
+        if ( (obj==m_stabilization->horizonRollExpo) ) {
```
mlyle: Spacing around ==

ground/gcs/src/plugins/config/configstabilizationwidget.cpp
```diff
@@ -73,9 +73,50 @@ ConfigStabilizationWidget::ConfigStabilizationWidget(QWidget *parent) : ConfigTa
...
 
+    // at first save the rate max values for the horizon rate plot
+
+    int roll_expo = m_stabilization->rateRollExpo->value();
+    int pitch_expo = m_stabilization->ratePitchExpo->value();
+    int yaw_expo = m_stabilization->rateYawExpo->value();
+    int roll_max = m_stabilization->fullStickRateRoll->value();
+    int pitch_max = m_stabilization->fullStickRatePitch->value();
+    int yaw_max = m_stabilization->fullStickRateYaw->value();
+    m_stabilization->rateStickExpoPlot->init(0,0,roll_expo,pitch_expo,yaw_expo,roll_max,pitch_max,yaw_max,0,0,0);
```
mlyle: Spacing after comma.  Also, is there any way you can call showExpoPlot below-- or factor its contents out-- to eliminate code duplication?

ground/gcs/src/plugins/config/expocurve.h
mlyle: also consistent capitalization on the above variables.  There's multiple schemes in use, but probably most common is lowercase with underbars, e.g. curve_cnt.

ground/gcs/src/plugins/config/expocurve.cpp
elafargue: they do. Well, someone needs to work on translation too...

Trex4Git:
Tried to implement almost all of your review comments.

Just regarding the translation stuff at the tool-tips i did not change anything ( tr() ).
It works without that, it tested it.
But as i have seen that it is used in already existing taulabs code i also used it here. But no personal preference on it.

ground/gcs/src/plugins/config/configstabilizationwidget.cpp
```diff
@@ -73,9 +73,46 @@ ConfigStabilizationWidget::ConfigStabilizationWidget(QWidget *parent) : ConfigTa
...
+    connect(m_stabilization->horizonRollExpo, SIGNAL(valueChanged(double)), this, SLOT(showExpoPlot()));
+    connect(m_stabilization->horizonPitchExpo, SIGNAL(valueChanged(double)), this, SLOT(showExpoPlot()));
+    connect(m_stabilization->horizonYawExpo, SIGNAL(valueChanged(double)), this, SLOT(showExpoPlot()));
+    connect(m_stabilization->rateRollKp_3, SIGNAL(valueChanged(double)), this, SLOT(showExpoPlot()));
+    connect(m_stabilization->ratePitchKp_4, SIGNAL(valueChanged(double)), this, SLOT(showExpoPlot()));
+    connect(m_stabilization->rateYawKp_3, SIGNAL(valueChanged(double)), this, SLOT(showExpoPlot()));
+
+
+    // set the flags for all Expo Plots, so that they get updatet after initialization
+    update_exp.RateRoll = 1;
```
mlyle: true/false

mlyle: (use bool, and assign to true false rather than 1/0)

ground/gcs/src/plugins/config/configstabilizationwidget.cpp
```diff
@@ -221,3 +258,95 @@ void ConfigStabilizationWidget::applyMWRateConvertDialog()
...
+        }
+
+        else if ( (obj==m_stabilization->horizonYawExpo) || (obj==m_stabilization->rateYawKp_3) ) {
+           update_exp.HorizonAttitudeYaw = 1;
+           if ( (obj==m_stabilization->horizonYawExpo) ) {
+            update_exp.HorizonRateYaw = 1;
+           }
+        }
+
+        else if ( (obj==m_stabilization->rateRollExpo) || (obj==m_stabilization->fullStickRateRoll) ) {
```
mlyle: spaces around equality operator ==

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,303 @@
...
+ * @param index
+ */
+void ExpoCurve::showCurve(const QVariant & itemInfo, bool on, int index)
+{
+    Q_UNUSED(index);
+    QwtPlotItem * item = QwtPlot::infoToItem(itemInfo);
+    if (item)
+        item->setVisible(!on);
+
+    mutex.lock();
```
mlyle: This mutex should be unnecessary.

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,303 @@
...
+ * @brief ExpoCurve::init //! Set label for the stick channels
+ * @param lbl_mode 0: Label for rate mode, 1: Label for horizon mode
+ * @param horizon_transitions 0: no marker, >0: % horizon transitions defined in /flight/Modules/Stabilization/stabilization.c
+ */
+void ExpoCurve::init(int lbl_mode, int h_transistion)
+{
+    switch (lbl_mode)
+    {
+        case 0:
+            roll_elements.Curve.setTitle("Roll rate (deg/s)");
```
mlyle: I think all of these setTitles need tr( ) around the strings.

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,303 @@
...
+    QPalette pal = m_legend->palette();
+    pal.setColor(m_legend->backgroundRole(), QColor(100, 100, 100));	// background colour
+    pal.setColor(QPalette::Text, QColor(0, 0, 0));			// text colour
+    m_legend->setPalette(pal);
+
+    insertLegend(m_legend, QwtPlot::TopLegend);
+
+    // axis
+    this->enableAxis(QwtPlot::yRight);
+    this->setAxisTitle(QwtPlot::xBottom, " normalized stick input");
```
mlyle: tr()

Trex4Git:
These undone small things makes me nervous ;)
So again some small changes based on your feedback.

ground/gcs/src/plugins/config/expocurve.h
```diff
@@ -0,0 +1,103 @@
...
+#include "qwt/src/qwt_plot.h"
+#include "qwt/src/qwt_plot_curve.h"
+#include "qwt/src/qwt_scale_draw.h"
+#include "qwt/src/qwt_scale_widget.h"
+#include "qwt/src/qwt_plot_grid.h"
+#include "qwt/src/qwt_legend.h"
+#include "qwt/src/qwt_legend_label.h"
+#include "qwt/src/qwt_plot_marker.h"
+#include "qwt/src/qwt_symbol.h"
+#include <QMutexLocker>
```
mlyle: remove mutex

Trex4Git:
shit, forgot that i had to include that to use the mutex.

I think its bad to have now so much small commits in this after the initial PR because of my faults. 

There are some functions in git to reset/change the commit history. For privat development it should be ok, but its often not recommend for public stuff.

As this is not merged yet and only a PR, do you think it may makes sense to change the commit history a little bit, or just make a new commit?

I thought about putting all the code style and smaller things together in one commit, refactoring/reduced code duplication stays as is, and all the mutex changes together in one commit. 

If there appear some problems in the future with the removed lock, it would be easier to revert it.

ground/gcs/src/plugins/config/expocurve.h
Trex4Git: this has also to be removed: https://github.com/Trex4Git/TauLabs/blob/GCS_Expo_merge/ground/gcs/src/plugins/config/expocurve.h#L97

BrainFPV:
@Trex4Git Nice work. I haven't yet reviewed the code in detail, but regarding making a new PR with fewer commits, I would keep working here until everything is good and then you can always decide to make a new PR with all  commits squashed into one (make a new branch based on this one and use `git rebase --interactive`).

mlyle:
So, first, no matter what you do, fix and commit.

Then you could even do it here-- just git rebase -i and change the last few commits to be 'fixup' in the text editor.  Then git push --force

This rewrites history, so you'll probably want to comment that you've done that here and use this sparingly... but it creates a fresh/clean history for us to bring into the project's revision log.

Anyways, it's not essential.  I'd be happy to merge this even with a few extra fix commits.  It's more something for you to consider for future PRs.

Trex4Git:
Removed the 2 unused mutex codes. Changed history only a little bit, so that all the removed mutex stuff is grouped in one commit.
And added some more information to the commit messages.

Trex4Git:
As most of the discussion here was regarding code optimization i am still interested what you think about the 3 effects that the HorizonExpo has, and how the data is visualized.

Before the HorizonExpo Parameter was implemented there was no difference between "Attitude" and "Horizon Attitude" or "Rate" and "Horizon Rate". 
I am not sure if it really was the intention to also scale the "Horizon Attitude" and "Horizon Rate" curves with HorizonExpo or if only the transition should be scaled: https://github.com/TauLabs/TauLabs/pull/1493

Do you also would support for such a additional functionality ? https://github.com/TauLabs/TauLabs/pull/1676#issuecomment-112931666

mlyle:
Let's just work on getting the minimal set of stuff in that makes this work, and then work on additional functionality.

I am probably going to send you a pull request with some feedback.

Trex4Git:
As you mentioned  on IRC you don't like it that the markers don't hide when the corresponding curve is hidden, i implemented it.
But want to collect more input & changes before doing more commits, so here you can se the implementation:

gcs_expo_hidemarker

Trex4Git:
I also would like to get your and others feedback and implement/change the set of stuff that needs to be done.

Trex4Git:
mlyle, you also mentioned it, that it may not look good on smaller displays. 
So i  will also change the font size of the axis titles as they are big in default configuration.

mlyle:
OK.  I think we could safely turn down some of the paddings of the enclosing box and the spacing between the two columns. 

One other thing I saw that isn't perfect is that on my display, the series legends on the right side show up on two lines-- the first 4 on one line, and the next 2 on the second (unlike your screenshot of 3 & 3).  I don't know if that can be easily fixed.

Trex4Git:
I could reproduce your described behavior regarding the legend entries when i scaled the widget to a larger width. 
Limited now now the number of legend entries to max. 3 columns,  and it seems to work ;) 

For the padding size i just don't changed the settings that was used before my changes to the UI. 
Do you have a special ones  that should be changed, maybe you can mark it in the screenshot?
Or do you thin i should just turn down all margins?

For the outer loop column there is also a additional spacing because there are only 2 PID setting rows, and i wanted to align the following elements in the two columns.
I think it looks nicer with this, even with some unused space between the PID settings and stick scaling group.

Trex4Git:
An update regarding the UI changes:
* limited the legend entries to max. 3 columns
* set a smaller font for the axis titles
* set a higher value for the min. height of the plots
* reduced some spaces and margins around the boxes

For the screenshot i reduced the window to a size, where the plots should be shown with their minimal height. 
For smaller window sizes the scrolling starts, and UI elements are not scaled down further:

gcs_expo_hidemarker_margins

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,301 @@
...
+    }
+    else {
+        curve_cnt = 1;
+    }
+}
+
+/**
+ * @brief ExpoCurve::plotData Show expo data for one of the stick channels
+ * @param value The expo coefficient; sets the exponential amount [0,100]
+ * @param curve The curve that has to be plot (roll,nick,yaw)
```
guilhermito: doxygen parameters don't follow function signature

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,301 @@
...
+ * @param mark The horizon marker that has to be plot (roll,nick,yaw) if != 0
+ *
+ * The math here is copied/ the same as in the expo3() function in /flight/Libraries/math/misc_math.c
+ * Please be aware of changes that are made there.
+ */
+void ExpoCurve::plotData(int value, int max, ExpoPlotElements_t &plot_elements, int mode)
```
guilhermito: Can't mode be an enum here and on the other functions where it is used?

ground/gcs/src/plugins/config/expocurve.cpp
```diff
@@ -0,0 +1,301 @@
...
+    }
+    else {
+        return y;
+    }
+}
+
+// Slot function to show/hide a curve
+/**
+ * @brief ExpoCurve::showCurve
+ * @param itemInfo
```
guilhermito: param description?

Trex4Git:
Thanks for your feedback. As i am a newbie, i have not much experience and got "inspired" by some existing code in Taulabs. 
For example the function to show/hide a curve was more or less only copy&paste in the first version.

But i am glad that i can learn a lot of things here. So i updated the doxygen documentation as you described, and also added some more documentation for other functions.

I also changed the mode from int to a custom enum. I think that provides easier to read/ understand code.

Trex4Git:
I pushed the commit with the above described changes. An overview is also in the commit message.

mlyle:
Merged!  Thanks for all your work!

Trex4Git:
Thanks for all your support.

PR #1678, opened 2015-06-21
attitude/ins: Use GPS pos/vel accuracy estimates

mlyle:
This also weighs the GPS Z information more highly, but probably not as
high as we ultimately should.

Also we should move to these being real-world units.  Originally in my changeset I had done this, but I had problems with my baro on my frame that led me to removing this change (that turned out to be unrelated / also happened on base code).  Next time around.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -1057,6 +1057,8 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 		INSSetAccelVar(insSettings.AccelVar);
 		INSSetGyroVar(insSettings.GyroVar);
 		INSSetBaroVar(insSettings.BaroVar);
+		/* This is more optimistic than in the actual flight loop, where
+		 * ublox accuracy data is added.  But that seems OK */
 		INSSetPosVelVar(insSettings.GpsVar[INSSETTINGS_GPSVAR_POS], insSettings.GpsVar[INSSETTINGS_GPSVAR_VEL], insSettings.GpsVar[INSSETTINGS_GPSVAR_VERTPOS]);
 
 		// Initialize the gyro bias from the settings
@@ -1206,6 +1208,32 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 		vel[2] = gpsVelData.Down;
 
 		gps_vel_updated = false;
+	}
+
+	// If either vel or pos is updated, update the variances.
+	if (gps_updated || gps_vel_updated) {
+		// Typical good pos 'accuracy' values are 2.5-3.5.  Early
+		// flight is near 4.0.
+		// Typical bad pos 'accuracy' values are 10+
+		// The values here are fudged.  Basically, when GPS is
+		// "good" we'll have a lower variance than the nominal
+		// setting.  But from there it will increase really quickly.
+		// sqrt(.5) / 3.5 =~ 0.181f 
+		float pos_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_POS] * 
+			(0.6f + powf(gpsData.Accuracy * 0.180f, 2));
+
+		// A good value of speed accuracy in flight is 0.35-0.5. 
+		// sqrt(.5) / .5 =~ 1.414
+		float speed_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_VEL] *
+			(0.5f + powf(gpsVelData.Accuracy * 1.414f, 2));
+
+		float v_pos_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_VERTPOS] +
+			(0.7f + powf(gpsData.Accuracy * 0.167f, 3.0));
+
+		// We trust the vertical much less as accuracy gets worse.
+		// cuberoot(.3)/4.0 =~ .167
+
+		INSSetPosVelVar(pos_var, speed_var, v_pos_var);
 	}
 
 	// Update fake position at 10 hz
@@ -1429,6 +1457,7 @@ static void settingsUpdatedCb(UAVObjEvent * ev)
 		INSSetAccelVar(insSettings.AccelVar);
 		INSSetGyroVar(insSettings.GyroVar);
 		INSSetBaroVar(insSettings.BaroVar);
+		/* Don't set GPS variance here, because the flight loop does */
 	}
 	if(ev == NULL || ev->obj == HomeLocationHandle()) {
 		uint8_t armed;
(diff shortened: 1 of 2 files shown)


mlyle:
So just to expound-- the primary purpose of this pull request is that we receive speed and velocity accuracy information from the GPS.  Generally when the solution changes the position accuracy is poor, and this tells the INS to take that position less seriously.  Accuracy also falls a bit in quick flight. This effectively means the INS will respond more slowly to the possibly faulty data and take other sources of information more seriously.

It's also important to note that speed and position accuracy have nothing to do with each other, and you can even have situations where the gps velocity vector is going south, and the position is moving northwards while all of the accuracies are improving-- that is, it can be confident we're moving south but realize that it had given us a position solution too far to the south before.

peabody124:
Awesome. I try and review this in the next few days.

brat002:
I did test of these changes yesterday. Looks, like it work.

flight/Modules/Attitude/attitude.c
```diff
@@ -1208,6 +1210,32 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
...
+	if (gps_updated || gps_vel_updated) {
+		// Typical good pos 'accuracy' values are 2.5-3.5.  Early
+		// flight is near 4.0.
+		// Typical bad pos 'accuracy' values are 10+
+		// The values here are fudged.  Basically, when GPS is
+		// "good" we'll have a lower variance than the nominal
+		// setting.  But from there it will increase really quickly.
+		// sqrt(.5) / 3.5 =~ 0.181f 
+		float pos_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_POS] * 
+			(0.6f + powf(gpsData.Accuracy * 0.180f, 2));
```
fujin: your comment on line 1221 says `0.181f` but you have entered `0.180`, is this a mistake?

flight/Modules/Attitude/attitude.c
```diff
@@ -1208,6 +1210,32 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
...
+
+		// A good value of speed accuracy in flight is 0.35-0.5. 
+		// sqrt(.5) / .5 =~ 1.414
+		float speed_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_VEL] *
+			(0.5f + powf(gpsVelData.Accuracy * 1.414f, 2));
+
+		float v_pos_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_VERTPOS] +
+			(0.7f + powf(gpsData.Accuracy * 0.167f, 3.0));
```
fujin: This 'magic value' has no comment, is it based on one of the comments above?

fujin: I see, right below, my bad.

flight/Modules/Attitude/attitude.c
```diff
@@ -1208,6 +1210,32 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
...
+
+		// We trust the vertical much less as accuracy gets worse.
```
fujin: consider moving this comment block to line 1229 above v_pos_var initialization

flight/Modules/Attitude/attitude.c
mlyle: It's different rounding and should probably be fixed, but the precise values don't really matter here.

shared/uavobjectdefinition/inssettings.xml
```diff
@@ -6,7 +6,7 @@
 		<field name="AccelVar" units="(m/s)^2" type="float" elementnames="X,Y,Z" defaultvalue="0.003"/>
 		<field name="GyroVar" units="(deg/s)^2" type="float" elementnames="X,Y,Z" defaultvalue="0.00001,0.00001,0.0001"/>
 		<field name="MagVar" units="mGau^2" type="float" elementnames="X,Y,Z" defaultvalue="10,10,100"/>
-		<field name="GpsVar" units="m^2" type="float" elementnames="Pos,Vel,VertPos" defaultvalue="0.001,0.01,10"/>
+		<field name="GpsVar" units="m^2" type="float" elementnames="Pos,Vel,VertPos" defaultvalue="0.001,0.01,0.5"/>
```
fujin: Do any tooltips need updating here?

fujin:
LGTM :+1:

CheBuzz:
Reviewed.  Changes look good.  Anybody opposed to merging?

flight/Modules/Attitude/attitude.c
peabody124: for simplicity, you can factor out the 0.18^2 and adjust the default GpsVar settings. then all of these will take a simpler (both math and computationally) form of:

   var = setting * (constant + accuracy^2)

peabody124:
It looks sensible although I'd like to wait until we have merged next to master for final release before merging this. 

I would prefer to see the slope and minimum separated into two settings, because I think we might end up empirically tweaking these. In the next few weeks I'll get Freedom flying again and acquire some high speed logs with the accuracy parameter so we can (ideally) refine them.

flight/Modules/Attitude/attitude.c
```diff
@@ -1208,6 +1210,32 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
...
+		// "good" we'll have a lower variance than the nominal
+		// setting.  But from there it will increase really quickly.
+		// sqrt(.5) / 3.5 =~ 0.181f 
+		float pos_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_POS] * 
+			(0.6f + powf(gpsData.Accuracy * 0.180f, 2));
+
+		// A good value of speed accuracy in flight is 0.35-0.5. 
+		// sqrt(.5) / .5 =~ 1.414
+		float speed_var = insSettings.GpsVar[INSSETTINGS_GPSVAR_VEL] *
+			(0.5f + powf(gpsVelData.Accuracy * 1.414f, 2));
```
peabody124: so e.g. here we can make it:
speed_var = insSettings.GpsVarMin + insSettings.GpsVarScale * powf(gpsVelData.Accuracy, 2)

set GpsVarMin to 0.0005 and GpsVarScale to 0.002

that produces the identical mapping from accuracy to speed_var but has a bit more flexibility if we want it later

peabody124:
Ok, after discussing this with Icee it can merge as is. I'll likely reparameterize it later with some code that provides a way of quantifying the likelihood of the data, since there is a concern with more parameters right now we won't have a good way to improve them.

PR #1679, opened 2015-06-21
zippyloiter: directly feed attitude control through when loitering

mlyle:
This is a feel-optimized GPS-augmented loiter algorithm.  When sticks are neutral, it holds position.  When they are deflected it is a in a body-frame "do what I mean" mode.

A decay IIR (instant attack, tunable decay) is kept of recent command magnitude to know how much the position controller should be neutered / how much existing error should be removed.  The desired hold position is updated based on present stick deflection and current velocity tangent to that deflection.  A portion of the stick input is mixed into the position controller's attitude output to provide instant responsiveness.


brat002:
Thanks for your great job! Today/tommorow i will test these changes on my sparky2.

mlyle:
Cool!  Test feedback is always welcome.

There's a chance you may want the fbb9697 version instead.  There's a lot that affects the feel of the mode and I'm going to be thinking in the next couple of weeks how to make it easy to tune this between a lazy aerial photography mode and a slightly slower attitude mode.

BrainFPV:
@mlyle can you explain what "zippyloiter" does?

mlyle:
It's a rewrite of the code that handles roll/pitch stick movements in position hold mode.

It comes into play on a smaller stick movement (1/5th stick deflection), and has expo.  It takes into account current velocity in the commanded direction, so holding stick forward will accelerate, up to a point.  It uses a different control scheme that isn't likely to get stuck and stutter and stop like the previous loiter mode.  And it blends in direct attitude movement so that it can respond to user input quicker than the position hold loop can.

fbb9697 was the original cut, but pug said it was too fast and had problems with smoothly braking from high speeds.  This one is slower/closer to the original loiter mode.  Ultimately there needs to be a couple more tunables, but their values and the velocity loop gains interact in nonintuitive ways so this will take some thought.

brat002:
How to enable this mode for testing? I can`t find it in flight modes.

mlyle:
It's position hold when you move the sticks.

You may want the previous version mentioned above.

brat002:
I did tested it right now. Works pretty good.

BrainFPV:
@mlyle this is a much needed feature. Let us know if we can help testing or review the code (I assume it is too early for this as it is still WIP).

mlyle:
@BrainFPV I'll try and prioritize it.  It has to wait for the release to get out the door.  I think it's mostly good to go, it's mostly a question of tuning and exposing tunables.  (I may slightly adjust the control scheme to improve braking behavior0.

peabody124:
We might want to touch base with some of the people from @LibrePilot about
their GPS assist. I think they did some work on braking. @AlessioMorale
@m-thread

mlyle:
While all feedback is welcome, this is a different control architecture.  I was/am hoping to get braking for free just as an emergent behavior of this controller, and may still depending on tuning, but if not I can put the couple lines to explicitly feedforward a little bit more on decel.

mlyle:
Awaiting flight test and code review.

mlyle:
Got some feedback from pug about a critical bug here.  Will get this fixed up imminently.

pug398:
Bug appears to be squashed and testing continues :) Both slow and fast loiter from the stick and braking is improved at higher speed.

mlyle:
@brat002 Any chance you could try flying this and comment on whether it works for you and how you like the tuning?  I'd really appreciate it.

pug398:
This has been working well for me and haven't seen any more issues :)

brat002:
My quadcopter was in repair :) I will test these changes on this week.
28 Июл 2015 г. 0:53 пользователь "pug398" <notifications@github.com>
написал:

mlyle:
OK.  This needs a code review pass.

Additional flight test is appreciated, too.

This is ready for merge other than I may hear from @pug398 that he wants to change the default tuning in the uavos slightly, though what is there flies fine.

peabody124:
This is high on my todo list! Thanks for all the work on it and to the testers. @mlyle you mind updating the title of the message and perhaps the PR message with a overview of the algorithm? That would help me understand the code a touch.

mlyle:
Top comment updated.

pug398:
After testing this I think maybe increase the LoiterAttitudeFeedthrough default from 12 to 15 degrees. 12 degrees limits the max loiter speed too much. You can then use LoiterLookaheadTimeconstant to adjust the max loiter speed from slow to fast. There is always slow positioning loiter at 1/2 stick available. Other default change might be PositionFeedForward bumped back up to 0.45. This has worked well on several frames/fc with the other settings I have.

mlyle:
@pug398 -- thank you, now tuned per your settings.  Just waiting for code review.

brat002:
My Sparky 2.0 has died. I can not test new changes until new MPU9250 will recieved and soldered. Sorry.

mlyle:
Sorry to hear that @brat002

mlyle:
There's a preview build at http://jar.lyle.org/~mlyle/TauLabs-20150813-782dc755-install.exe (rebased to today's next)

flight/Modules/ManualControl/tablet_control.c
```diff
@@ -97,8 +97,9 @@ int32_t tablet_control_select(bool reset_controller)
 
 			// Command to not move. This code is identical to that in manualcontrol
 			LoiterCommandData loiterCommand;
-			loiterCommand.Forward = 0;
-			loiterCommand.Right = 0;
+			loiterCommand.Pitch = 0;
+			loiterCommand.Roll = 0;
+			loiterCommand.Throttle = 0.5f;
```
peabody124: I'm sure this is safe but could you confirm that to me :) this isn't going to make things shoot up, right?

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1107,20 +1107,12 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
...
-	                        0;
-	// Note the negative - forward pitch is negative
-	loiterCommand.Forward *= -MAX_SPEED / (1.0f - CMD_THRESHOLD);
-
-	loiterCommand.Right = (cmd->Roll > CMD_THRESHOLD) ? cmd->Roll - CMD_THRESHOLD :
-	                      (cmd->Roll < -CMD_THRESHOLD) ? cmd->Roll + CMD_THRESHOLD :
-	                      0;
-	loiterCommand.Right *= MAX_SPEED / (1.0f - CMD_THRESHOLD);
+
+	loiterCommand.Pitch = cmd->Pitch;
```
peabody124: so now the loiter command truly means pitch and roll again? what are the units? deg instead of m/s ? should we be using the values from stab settings here so it has meaningful units? the answer might be clear as i review.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -302,8 +183,9 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
  * @param[in] landing whether to descend
  * @param[in] update_status does this update path_status, or does somoene else?
  */
-static int32_t vtol_follower_control_simple(const float dT,
-	const float *hold_pos_ned, bool landing, bool update_status) {
+static int32_t vtol_follower_control_impl(const float dT,
```
peabody124: can you update the oxygen header to explain what a null err does (or non null)

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -325,22 +207,32 @@ static int32_t vtol_follower_control_simple(const float dT,
...
+		err[2] = errors_ned[2];
+	}
+
 	/* Apply a cubic deadband; if we are already really close don't work
 	 * as hard to fix it as if we're far away.  Prevents chasing high
 	 * frequency noise in direction of correction.
 	 *
 	 * That is, when we're far, noise in estimated position mostly results
 	 * in noise/dither in how fast we go towards the target.  When we're
+	 * r
```
peabody124: ?

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -408,9 +300,10 @@ static int32_t vtol_follower_control_simple(const float dT,
  * Takes in @ref PositionActual and compares it to @ref PositionDesired
  * and computes @ref VelocityDesired
  */
-int32_t vtol_follower_control_endpoint(const float dT, const float *hold_pos_ned)
+int32_t vtol_follower_control_endpoint(const float dT, const float *hold_pos_ned,
```
peabody124: same - update doxygen

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -496,22 +389,69 @@ static int32_t vtol_follower_control_accel(float dT)
 	return 0;
 }
 
+
+static float vtol_follower_control_altitude(float downCommand) {
```
peabody124: so this is the algorithm from ```altitudehold.c``` copy and pasted, right?

peabody124: oops nevermind, same also but just copied from further down this file.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -522,82 +462,44 @@ int32_t vtol_follower_control_attitude(float dT)
...
 
+	StabilizationDesiredData stabDesired;
+
 	// Set the angle that would achieve the desired acceleration given the thrust is enough for a hover
 	stabDesired.Pitch = bound_min_max(RAD2DEG * atanf(forward_accel_desired / GRAVITY),
-	                   -guidanceSettings.MaxRollPitch, guidanceSettings.MaxRollPitch);
+	                   -guidanceSettings.MaxRollPitch, guidanceSettings.MaxRollPitch) + att_adj[1];
 	stabDesired.Roll = bound_min_max(RAD2DEG * atanf(right_accel_desired / GRAVITY),
-	                   -guidanceSettings.MaxRollPitch, guidanceSettings.MaxRollPitch);
+	                   -guidanceSettings.MaxRollPitch, guidanceSettings.MaxRollPitch) + att_adj[0];
```
peabody124: so here i think we should consider applying the max angles allows by the stabilization settings. by default these are greater than the navigation ones, but if nav is maxed at 20 and their range for att_adj is 35 (i still need to see that) they might get 55 deg which is well outside the default max stab settings.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -496,22 +389,69 @@ static int32_t vtol_follower_control_accel(float dT)
...
+	if (!att_adj) {
+		att_adj = default_adj;
+	}
+
 	AccelDesiredData accelDesired;
 	AccelDesiredGet(&accelDesired);
 
-	StabilizationDesiredData stabDesired;
+	StabilizationSettingsData stabSet;
+	StabilizationSettingsGet(&stabSet);
```
peabody124: it's probably best to move updating this to the settings update method below (make sure you register the callback). It increases the static memory usage (really all the settings should move into a dynamically allocated structure) but avoids additional semaphore locks here.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -627,6 +529,142 @@ int32_t vtol_follower_control_attitude(float dT)
...
+	input /= (1 - CMD_THRESHOLD);	// Normalize to -1 to 1 range.
+
+	/* Confine to desired range */
+	if (input > 1.0f) {
+		input = 1.0f;
+	} else if (input < -1.0f) {
+		input = -1.0f;
+	}
+
+	return expo3(input, 40);	// And apply 40% expo
```
peabody124: this should be exposed in a setting

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -627,6 +529,142 @@ int32_t vtol_follower_control_attitude(float dT)
...
+ * @param[in] dT the time since last evaluation
+ * @param[in,out] hold_pos the hold position
+ * @param[out] att_adj an adjustment to be made to attitude for responsiveness.
+ */
+
+bool vtol_follower_control_loiter(float dT, float *hold_pos, float *att_adj) {
+	LoiterCommandData cmd;
+	LoiterCommandGet(&cmd);
+
+	// XXX TODO reproject when we're not issuing body-centric commands
```
peabody124: want to remove this and create an issue. having TODO in the code just clutters with comments but doesn't really track things.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -627,6 +529,142 @@ int32_t vtol_follower_control_attitude(float dT)
...
+	float commands_rp[2] = {
+		cmd.Roll,
+		cmd.Pitch
+	};
+
+	float command_mag = vectorn_magnitude(commands_rp, 2);
+	float deadband_mag = loiter_deadband(command_mag);
```
peabody124: just out of curiosity, why are you not using the cubic deadband method here and using a hard cutoff deadband?

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -627,6 +529,142 @@ int32_t vtol_follower_control_attitude(float dT)
...
+
+	static float historic_mag = 0.0f;
```
peabody124: normally i don't ask for comments, but could you document this structure. 

there a peak hold detection and decay going on, and we only do things below when there was a historical stick magnitude that falls within a certain historical envelope. i'm guessing because there is further processing to do after the stick goes back to neutral for some time window...

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -420,31 +406,6 @@ static int32_t vtol_fsm_static()
...
  */
 
-//! The currently configured navigation mode. Used to sanity check configuration.
-static enum vtol_nav_mode vtol_nav_mode;
-
-/**
- * General methods which based on the selected @ref vtol_nav_mode calls the appropriate
- * specific method
- */
-static int32_t do_default()
```
peabody124: ok i'll need some time to work through the implications of this change. this is where i wish we had unit tests on the whole nav system :(

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -627,6 +529,142 @@ int32_t vtol_follower_control_attitude(float dT)
...
+}
+
+/**
+ * Receives loiter commands and makes appropriate adjustments.
+ * @param[in] dT the time since last evaluation
+ * @param[in,out] hold_pos the hold position
+ * @param[out] att_adj an adjustment to be made to attitude for responsiveness.
+ */
+
+bool vtol_follower_control_loiter(float dT, float *hold_pos, float *att_adj) {
```
peabody124: could this be renamed calculate loiter? i was getting a bit confused because i expected this to actually implement some control but it gets the values prepared for the actual control  ```vtol_follower_control_attitude```

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -52,12 +52,14 @@
 #include "vtol_follower_priv.h"
 
 #include "attitudeactual.h"
-#include "loitercommand.h"
 #include "pathdesired.h"
 #include "positionactual.h"
 #include "stabilizationdesired.h"
+#include "velocityactual.h"
```
peabody124: was this required in the end? i don't think so

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: actually this change wasn't needed. vtol_follower_control_endpoint never is _not_ passed NULL for the last parameter

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -546,46 +507,22 @@ static int32_t do_land()
...
-	if (loiterCommand.Frame == LOITERCOMMAND_FRAME_BODY) {
-		north_offset = (loiterCommand.Forward * cosf(yaw) - loiterCommand.Right * sinf(yaw)) * DT;
-		east_offset = (loiterCommand.Forward * sinf(yaw) + loiterCommand.Right * cosf(yaw)) * DT;
-	} else {
-		north_offset = loiterCommand.Forward * DT;
-		east_offset = loiterCommand.Right * DT;
+		hold_position(vtol_hold_position_ned[0],
+				vtol_hold_position_ned[1],
+				vtol_hold_position_ned[2]);
 	}
```
peabody124: should there be an else that forces att_adj to 0 to be safe here?

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -546,46 +507,22 @@ static int32_t do_land()
...
-	const float LOITER_LEASH = 4;
-
-	LoiterCommandData loiterCommand;
-	LoiterCommandGet(&loiterCommand);
-
-	float yaw;
-	AttitudeActualYawGet(&yaw);
-	yaw *= DEG2RAD;
+	float att_adj[2] = { 0, 0 };
+	if (vtol_follower_control_loiter(DT, vtol_hold_position_ned, att_adj)) {
```
peabody124: here we should be using a temp variable instead of ```vtol_hold_position_ned```. a) this is for safety since that is one of the core control variables and probably should only be altered by ```hold_position``` called immediately below and b) to ensure when it doesn't do anything this variable isn't actually changed.

peabody124:
Overall this is looking good
 - code refactoring the vector methods and moving them into libraries are good, same with the changes shuffling around the altitude hold code
 - First pass the calculation for the position adjustment/attitude feedforward looks reasonable but I don't fully understand the algorithm it is trying to implement. Is this written up anywhere or drawn from anything? I need to review this a second pass.
 - I need to verify the changes to the state machine still. has this been tested with path following and RTH? I'm assuming so but that needs to happen.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: I don't want a small deviation from stick center to result in any sustained (even very slow) drift.  An intended 0 from the user needs to be a real 0.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Yah.  I want to eliminate a lot of this copy and paste but often I end up just moving it around in preparation for that.

flight/Modules/ManualControl/transmitter_control.c
mlyle: Units means "loiter effort" on a scale of -1 to 1.  It's partially dimensioned as an angle but the position controller does work too.

flight/Modules/ManualControl/tablet_control.c
mlyle: Right now throttle is ignored.  I have a pending issue to enable user control of altitude, in which case 0.5 will be inside the deadband which means "don't climb or descend".

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: OK.  This is a style maintained from the old loiter code.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: The do_default was scary and there were a few cases where it could do the right thing.  I made sure each state machine had a static_fn defined for all states.  It doesn't use more memory and it's good to be explicit.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: It's explicitly initialized to 0 just above.  Should I document that control_loiter leaves it alone?

mlyle: Actually it's not impossible it may want to tweak attitude even when not updating the hold position.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: It's more of a marker for the correct place, and this is the next work item in the sequence of stuff to do.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: Er, where it could do the -wrong- thing heh

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: So I'm not sure I understand.

Before we used to contend the semaphore a whole lot, getting individual settings out; I centralized this to the top of the loop instead so it's just contended once.

Is lock contention a big concern?  It was my understanding that the actual atomic operations on ArmV7 are cheap, and the actual locks should basically be uncontended-- the only time they'll come into play is if we task switch in the middle of one of these copies, and even then, only if we task switch to someone who immediately wants to copy that thing.

I really like the idea of callbacks if we're going to do a lot of computation on something and we can instead do that once instead of per-iteration.  But otherwise I'd kinda prefer to get them each time.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: removed as suggested.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Yah, I thought about that some.  Let me think about it a little more.

I don't really suggest an att_adj that high, and I think the position controller will be fighting rather than augmenting a big roll/pitch like that without insane gains (at least in the long term; it might help in the very short acceleration phase).

Also these should probably be symmetric bounds.  I'm not sure why this is this way.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Weren't you just talking about too many control knobs? ;)

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: I don't love it but will do to make you happy ;)

mlyle: Actually I'd like to change my mind on this one.  

1. Everything the file vtol_follower_control.c exports has that prefix.
2. It's almost a crude controller ;)

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
peabody124: yeah i prefer the always explicit style

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: still for this one we should so people can play with that feel and just make the default 40 if you feel that is a good value

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: stylistically i prefer this way too. it isn't using the arm style lock-write-unlock type of atomic operation. this ends up having to bind up the global semaphore that is used amongst all the tasks for accessing the uavobject manager which isn't wonderful or free. We should also stick to one style in a module (in this one it is using the callback to update global variables) rather than mixing styles.

the thing that sold me on this model was actually pointed out by @lilvinz - the most appropriate thing is the updated callback actually sets a flag. then at the top of the while loop we consume that flag and update the settings. this avoids any race conditions where a setting can change mix-execution with potentially unpredicted side effects.

flight/Modules/ManualControl/tablet_control.c
peabody124: Could you make this something really low then like 0.1 and then update it in the subsequent PR. I'd hate this to get forgotten and one day cause an unexpected climb. Sorry for being pedantic, but flyaway are my biggest fear.

flight/Modules/ManualControl/transmitter_control.c
peabody124: Ok cool. One thing I'm not sure there is a good solution for. I think analyzing log files with this might be a bit tricky since we aren't really exposing all the internal things that alter the control scheme. Not sure if there is a good solution to this.

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: alright :)

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: good point, duh :) thanks

flight/Modules/VtolPathFollower/vtol_follower_control.c
peabody124: sure thing - could you just elaborate in the doxygen then since it did make the code a bit confusing later on.

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
peabody124: wouldn't that be safer to do by returning the "changed" flag as true while not changing the hold position itself? it just seems more robust against future bugs to make sure when it says it isn't doing anything it physically can't do anything.

flight/Modules/VtolPathFollower/vtol_follower_control.c
mlyle: Hm.  I'm not sure I buy this.  And if it is an overhead problem I should just fix the uavobject manager.

>  the most appropriate thing is the updated callback actually sets a flag. then at the top of the while loop we consume that flag and update the settings. this avoids any race conditions where a setting can change mix-execution with potentially unpredicted side effects.

Wait, I don't understand.  We take a copy of it while protecting accesses to the master with a lock, right?  So getting it at the top of the function doesn't risk it changing mid-execution...

In any event, this was a simple change in factoring-- I changed the later, pre-existing calls to  StabilizationSettingsManualRateGet(manual_rate); to instead look at a copy of stabSet retrieved at the beginning of the function, which if anything should be safer in these respects.  I'd rather not tangle up this pull with additional cleanup.

mlyle:
@peabody124 You don't like the description at the top of the pull request?  I'm glad to elaborate further.

mlyle:
New build artifact at http://jar.lyle.org/~mlyle/TauLabs-20150816-ec55be7b-install.exe

(hopefully) Fixes a several-month old regression in RTH spotted by @pug398

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -426,7 +308,7 @@ int32_t vtol_follower_control_endpoint(const float dT, const float *hold_pos_ned
 int32_t vtol_follower_control_land(const float dT, const float *hold_pos_ned,
 	bool *landed)
 {
-	return vtol_follower_control_simple(dT, hold_pos_ned, false, true);
+	return vtol_follower_control_impl(dT, hold_pos_ned, true, true);
```
peabody124: Ah. So this is why @pug398 was having issues with landing. That makes sense, I couldn't really understand that.

I wonder if we should actually apply this fix to master and make a hotfix release. Having RTH broken is pretty bad...

peabody124: great catch btw!

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -142,8 +129,7 @@ static void hold_position(float north, float east, float down);
 
 /**
  * The state machine for landing at home does the following:
- * 1. enable holding at the current location
- * 2  TODO: if it leaves the hold region enable a nav mode
+ * enable holding at the current location
```
peabody124: see you don't like TODO statements in the code either :-p

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -182,13 +166,14 @@ const static struct vtol_fsm_transition fsm_follow_path[FSM_STATE_NUM_STATES] =
...
 };
+
 /**
  * The state machine for landing at home does the following:
  * 1. holds where currently at for 10 seconds
  * 2. ascend to minimum altitude
  * 3. flies to home at 2 m/s at either current altitude or 15 m above home
  * 4. holds above home for 10 seconds
- * 5. descends to ground
+ * 5. descends to ground (This step currently does not complete)
```
peabody124: remove this comment now - this is fixed now, right?

peabody124:
Just some simulator testing results:
 - flying path still works, lands at correct location. explicitly tested that stick movements don't kick you off path
 - RTH sequencer works - also ignores stick inputs, hovers over home, lands correctly
 - loitering works - i'm still trying to tease apart how much of that is due to the feedforward versus moving the hold position - do you know @mlyle? Is this something you calibrated or computed some way?

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
peabody124: ok reviewed this part and it all makes sense, nice improvement

pug398:
Several flight tests and huge improvement. Fixing that took care of all four RTH problems mentioned. Slow return speed, overshoot, large initial altitude drop and uncontrollable fast descent rate. Even the motors stop again when it touches down. Very nice  :)

peabody124:
The drop and uncontrollable descent rate (and motors not spooling down) make sense. Those got broken in the refactor and are now fixed.

A change in return speed and overshoot doesn't (unless a tuning setting were changed). I look at the RTH log from the other day you sent me and that one didn't show much overshoot (unless GPS was not consistent with reality, which is possible).

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: It's more that there's never any end to descending / no landing detection

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
mlyle: I disagree with the concept-- not sure why we'd transition from endpoint following to path following in response to wind or whatever. :P

peabody124: outside of a certain range the endpoint control doesn't produce as good results as the path navigation. you could also rephrase it as the endpoint controller itself being more intelligent for large errors

mlyle:
@peabody124 It's all heuristic.

At small stick deflections loitering is almost entirely due to moving the hold position.  The feedforward makes it slightly more responsive, but the position controller is fully enabled and is free to undo the feedforward (and it ultimately does-- it tends to pull back a bit from the initial attitude).

As stick deflections increase, we start to significantly decay the error in the position controller.  This is exactly equivalent to when we turn down the gains in MWRate or Horizon mode, except that when we stop decaying them the set point persists.  This makes the attitude component much more important.  I tried to pick an input architecture for the position controller (with the consideration of acceleration) which will not fight the stick input much, even before this attenuation.  This is also important to get the braking behavior we want-- the position setpoint has to be in front of the multicopter when the decay process begins for smooth braking.

mlyle:
For future investigation, we may want to end up decaying normal error more slowly than tangent error.  It's important to decay both, so that we don't end up flying sideways at the end of a loiter and so that the user has primary control authority.  But we could do a little more to fight crosswinds than we do now.

mlyle:
Temp hold on committing this- @pug398 feedback

peabody124:
Ok I finally 85% understand how it moves the setpoint. One day some debugging code to log quite how those two interact would be cool, but mostly an academic exercise.

Ok to merge with me once @pug398 rechecks. Hit the button yourself if I'm not around.

mlyle:
Cool, thank you.  Will work to get us unstuck on whatever is regressing pug.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -627,6 +517,148 @@ int32_t vtol_follower_control_attitude(float dT)
 	return 0;
 }
 
+
+static float loiter_deadband(float input) {
+	const float CMD_THRESHOLD = 0.2f;
```
BrainFPV: I think this is a bit much, i.e., quite a bit of stick deflection is needed for the UAV to start moving. Given that the sticks are spring centered, it should IMO just be big enough to make sure that an imperfect calibration doesn't cause drift. How about 0.1 or even 0.05?

BrainFPV:
Finally had a chance to test this today (Quanton on a 550 flamewheel hex). Definitely an improvement over the previous version. Will be really nice once it is combined with altitude control :).

ducminhkt:
I just updated 744c79e to my FlyingF4. When i change my MPU 6050 Gyro rate from 1000 to 2000, CPU load rise from 53% to 92% and hold there. I think may be a bug somewhere.

pug398:
I was just checking CPU load on Quanton F4 at at gyro rate of 500-2000. With INS enabled but not unlocked (indoors), cpu load is holding at 20%. No oneshot.

ducminhkt:
I just recheck. With no oneshoot, the same result. With MPU rate 500 - CPU load 32%, with rate 1000 - CPU load 53%, with rate 2000 - CPU 89%, with rate 4000 - CPU full load 99-100%. My FlyingF4 with GPS, Telemetry, PPM, hardware no change. Before update this version, CPU load at about 58% with MPU rate 2000 or even higher.

mlyle:
@ducminhkt please open an issue rather than discuss in a closed pull request.

Please include what version you were running before.

mlyle:
@ducminhkt Please also include from what task is active from the TaskInfo.RunningTime UAVO in the UAVO browser.

ducminhkt:
Today I do some test and I think i found. When I leave MPU6050 LPF at 256 (default) - issue come. When I change LPF to 188 or lower - issue gone.
@mlyle: sorry.

mlyle:
@ducminhkt LPF of 256Hz forces an update rate of 8KHz which we really don't tolerate well.  We should really take the option away until @Trex4Git 's work gets in to let us deal with it better.  No problem.

PR #1680, opened 2015-06-21
autotune: Reduce stack usage by ~344 bytes.

mlyle:
I have only tested this in simulator, not on hardware, and me getting to do this before I depart is unlikely.  So I'm putting this up as a useful WIP that could be proved innocent and merged.

Stack analysis from -fstack-usage reported that before the function AutotuneTask used 1280 bytes.  Now it's 936.  I only reduced the stack size of the task by 300 bytes.

I hope this could make autotune work on CC3D again.


guilhermito:
Looks good

BrainFPV:
Dito. Nice to see that that for once the stack size of a thread goes down not up ;)

fujin:
See all you needed was a little push :)

fujin:
Unfortunately can confirm this code still induces a soft boot loop on the coptercontrol target, tested on Windows built from:

```
AJ@HAXSTATION /c/TauLabs (mpl-autotoon)
$ git log -n1
commit 873629c14f5fecb9a1fbcbdaa7d5187eba0265ef
Author: Michael Lyle <mlyle@lyle.org>
Date:   Sun Jun 21 11:01:26 2015 -0700

    autotune: Reduce stack usage by ~344 bytes.
```

Looks like a step in the right direction though!

VArcht:
Tried this on CC3D , got the reboot loops too :(

mlyle:
Disabling mainport telem is enough to prevent reboot loops according to IRC, but not to arm.

So I scraped up a few slivers of memory from stabilization hoping it helps.

mlyle:
OK, so-- I aggressively inlined af_predict, which increased this to a total stack savings of 416 bytes, of which I claimed 396-- 96 more than the original commit that people were able to barely get to work.

I also probably found 30-40 bytes elsewhere.  Please let this work. :stuck_out_tongue_winking_eye:

mlyle:
Just rebased out the stabilization changes because I don't trust them and based on that 184 bytes..  Also just turned off lighttelemetry.

Hopefully what's left is still enough to run autotune.

mlyle:
I'm thinking this is good if it gets positive test feedback in its current form.

VArcht:
Bench tested successfully (ran AT with props off) on CC3D with Telemetry disabled. Heap shows 80 bytes free. Memory is in warning state but seems to be working fine. I will flight test when I get a chance.

mlyle:
Woot!  Thanks so much for testing :)

ernieift:
The memory problem is only on the f1 target. To avoid other issues, i suggest this:

In autotune.c
```
#if defined(PIOS_AUTOTUNE_STACK_SIZE)
#define STACK_SIZE_BYTES PIOS_AUTOTUNE_STACK_SIZE
#else
#define STACK_SIZE_BYTES 1900
#endif
```
and in coptercontrol/fw/pios_config.h
```
#define PIOS_AUTOTUNE_STACK_SIZE        1504
```
So we can adjust the stacksize for coptercontrol manually as low as we can.

mlyle:
Ok, we can...  This change wasn't based on a reduction of the auto tune stack to the bare minimum, though-- through rearranging the stack we require less stack space at once.   The new value actually has more padding on all targets than the previous one.

I can't make any further changes myself until I am back home in early July.

mlyle:
I've thought about this and would like this to please merge as-is.  I freed up 416 bytes of stack on all targets; I claimed 396 back on all targets, too, leaving 20 bytes more padding than there used to be in the autotune thread.  Why the weird/scary value of 1504?  So that the stack could be a multiple of 8 bytes, as is a best practice for ChibiOS.

If there's a justification for getting 396 bytes more padding for autotune than there used to be on targets other than CC3D, we should discuss it elsewhere.

mlyle:
Cool! Thank you.

peabody124:
So now it works on CC3D again? Nice. Was it flight tested on an F4?

mlyle:
Flight tested on F4 (tune not applied).  Bench tested with motors doing weird auto-tuney things on F1 by varcht.  No one's applied a tune from this yet so that's a little concerning but what I saw looked sane.

zsantolim:
I have tested this commit successfully on a CC3D and a 250 size quad. Started with a clean configuration. Before enabling the Auto Tune module, the system has around 1170 bytes remaining on heap.
Enabling Auto Tune and rebooting didn't caused any failed boot, but remaining heap dropped down to only 60 bytes and a mem warning appeared.
A flight test was successfully made. After that, Auto Tune was triggered and completed also without any flaws.
Let me know if you need more details on the test.

peabody124:
@zsantolim thanks. quite tight but at least it works again, which is nice. if you want to be super thorough it would be awesome if you could test on the RC: https://github.com/TauLabs/TauLabs/releases/tag/20150705

mlyle:
@zsantolim Super! Thanks for the test!

zsantolim:
@peabody124 @mlyle I have just tested Pre-Release 20150705 as suggested, and It was the same git hash I have tested before. But I have started all over with the provided release package.
Started with a clean configuration, flashed the release firmware. Using PPM as input, 4 motor outputs, 6 flightmodes, default PID configuration.
I have to correct myself: before enabling the Auto Tune, the remaining heap is exactly 1744 bytes. After enabling the module and rebooting, it gets down to 60 bytes only and we get a memory warning.
Anyways, flight tested the CC3D on the 250 size quad, than switched the Auto Tune. It took about 90 seconds to complete successfully. 
Landed, checked the values, everything ok.
Disabled the Auto Tune module and the remaining heap went back to the previous value, 1744 bytes.
Tested the windows release, the linux one was the one I have tested before, compiling myself.
Everything looks good for CC3D.

mlyle:
@zsantolim Great!  I've closed the related issue.

PR #1682, opened 2015-06-22
Logging module improvements

### LINKED ISSUE #1507: Logging module: file header invalid
I have been trying to use the new logging module on the Brain FPV FC, but when I download the log and try to open it using the Python parser, I get an error stating the header of the file is invalid. Has anyone else experienced this problem or are extra steps needed to open the file?

BrainFPV:
Adds the following improvements to the Logging module:

- Writes correct TauLabs log file header. Fixes #1507
- Adds the option to log all settings UAVOs at the start of the log file. This information will be very useful when analyzing logs. Also, the space needed for the settings is small, so it is enabled by default.
- Logs some UAVOs on change (currently only `FlightStatus` and `WaypointActive`)
- Adds some additional UAVOs to be logged
- Fixes a bug in `pios_streamfs` that limited the log file size to 1MB (`uint8_t` was used for `arena_id`)

### DIFF
--- a/flight/Libraries/timeutils.c
+++ b/flight/Libraries/timeutils.c
@@ -5,6 +5,7 @@
  *
  * @file       timeutils.c
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2015
+ *             Free Software Foundation, Inc. (C) 1991-2015
  * @brief      Time conversion functions
  *
  * @see        The GNU Public License (GPL) Version 3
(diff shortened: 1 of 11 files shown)


flight/Libraries/timeutils.c
```diff
@@ -0,0 +1,99 @@
...
+				dayOfWeek -= 7;
+			++year;
+		}
+		else
+		{
+			dayOfWeek  += days;
+			dayOfWeek  %= 7;
+
+			/* calculate the month and day */
+			static const uint8_t daysInMonth[12] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
```
mlyle: static keyword isn't meaningful here is it?

BrainFPV: Yes, I think you are right. Will fix it in the first round of cleanups.

flight/Libraries/timeutils.c
```diff
@@ -0,0 +1,99 @@
...
+	/* calculate days */
+	days     = hours   / 24;
+	hours   -= days    * 24;
+
+	/* Unix time starts in 1970 on a Thursday */
+	year      = 1970;
+	dayOfWeek = 4;
+
+	while(1)
+	{
```
mlyle: Time code makes me nervous because it's so easy to get subtly wrong.  This appears correct, but I'm not 100% confident.  This is also slow, but OK.

For reference the glibc version is at: https://sourceware.org/git/?p=glibc.git;a=blob;f=time/offtime.c;h=a437266e95eaaf7bb99e1b4512a633f9dfd906d4;hb=6b142b3a1d007d7e6f50c26710de7177bc4aca74 

It runs in constant-ish time so maybe we want to bring this in instead.

BrainFPV:
Note: The amount of data logged is currently about 290kb/min, which may be a bit too high. As even with 8MB of flash, it is only enough to log about 28 minutes. I think it would be good to have enough space for about 2h or so of data. So we should log some UAVOs less frequently or remove them.

flight/Modules/Logging/logging.c
```diff
@@ -285,6 +352,81 @@ static void loggingTask(void *parameters)
...
+	int pos;
+	char tmp_str[45];
+	char *info_str;
+	char this_char;
+	DateTimeT date_time;
+
+	const struct pios_board_info * bdinfo = &pios_board_info_blob;
+
+	sprintf(tmp_str, "%s\n", "Tau Labs git hash:");
+	send_data((uint8_t*)tmp_str, strlen(tmp_str));
```
mlyle: Why not directly send the constant?  #define LOG_HEADER "Tau Labs git hash:\n"
send_data(LOG_HEADER, strlen(LOG_HEADER)) will evaluate the strlen nicely at compile time etc.

flight/Modules/Logging/logging.c
```diff
@@ -285,6 +352,81 @@ static void loggingTask(void *parameters)
...
+	for (int i = 0; i < 4; i++){
+		this_char = *(char*)(bdinfo->fw_base + bdinfo->fw_size + 7 - i);
+		tmp_str[pos++] = DIGITS[(this_char & 0xF0) >> 4];
+		tmp_str[pos++] = DIGITS[(this_char & 0x0F)];
+	}
+	send_data((uint8_t*)tmp_str, pos);
+
+	// Date
+	date_from_timestamp(*(uint32_t *)(bdinfo->fw_base + bdinfo->fw_size + 8), &date_time);
+	sprintf(tmp_str, " %d%02d%02d\n", 1900 + date_time.year, date_time.mon + 1, date_time.mday);
```
mlyle: snprintf is nice, both because it precludes any chance of an overflow and because it returns the length, avoiding a strlen.

flight/Libraries/timeutils.c
mlyle: I like the test though, so if you prefer this implementation I'm OK to merge as-is.

mlyle:
In the future, if we want more logging, we could do: http://oku.edu.mie-u.ac.jp/~okumura/compression/lzss.c

BrainFPV:
@mlyle thanks for the code review!

I was thinking about using compression as well; due to the large header overhead, the compression ratio would be quite good. However, one problem is that when the log wraps around and the start of the file is overwritten, it will no longer be decodable. We would need to do something smart, e.g., store the dictionary for each file in a separate flash arena that doesn't get overwritten automatically.

I addressed all your comments and made some additional improvements: Made the logging speed configurable, and added some more UAVOs to be logged. The log file size depending on the rate is approximately:

- 10Hz: 90kB/min
- 25Hz: 200 kB/min
- 100Hz: 550 kB/min

I set the default rate to 25Hz, which will fill up a 8Mbit flash chip in about 40 minutes.

peabody124:
I love this. Awesome.

PR #1683, opened 2015-06-22
Python: Fix multi-instance parsing

BrainFPV:
Fixes Python log parsing for logs that have multi-instance UAVOs present.

### DIFF
--- a/python/taulabs/telemetry.py
+++ b/python/taulabs/telemetry.py
@@ -54,7 +54,9 @@ class TelemetryBase():
         if githash:
             uavo_defs.from_git_hash(githash)
         else:
-            uavo_defs.from_uavo_xml_path("shared/uavobjectdefinition")
+            xml_path = os.path.join(os.path.dirname(__file__), "..", "..",
+                                    "shared", "uavobjectdefinition")
+            uavo_defs.from_uavo_xml_path(xml_path)
 
         self.githash = githash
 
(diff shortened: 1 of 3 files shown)


python/taulabs/telemetry.py
```diff
@@ -54,7 +54,9 @@ def __init__(self, githash=None, service_in_iter=True,
         if githash:
             uavo_defs.from_git_hash(githash)
         else:
-            uavo_defs.from_uavo_xml_path("shared/uavobjectdefinition")
+            xml_path = os.path.join(os.path.dirname(__file__), "..", "..",
+                                    "shared", "uavobjectdefinition")
+            uavo_defs.from_uavo_xml_path(xml_path)
```
BrainFPV: Added this so the `taulabs` module can be imported from any directory.

mlyle:
LGTM-- sorry I didn't get to this earlier.

As long as you're doing this, though, you should revert 722ac0beef1d59d333c0289ea1067a5f3b8087f in the same PR, as that's a workaround for reading instanceid as a timestamp.

mlyle:
Also consider renaming the 'cleanup' log message.

BrainFPV:
@mlyle thanks for reviewing. Should be good now :)

mlyle:
Tests good to me.. i'll factor this down into the objects later to pick up some speed.

Issue #1404, opened 2014-11-28
GCS: PicoC Widget Button to transmit Test Value


cGiesen:
To transmit Test Value to FC we must use the browser for today.
It would be nice to auto transmit the value when change in GCS or have a transmit button

Trex4Git:
I have done a quick implementation. Compiles and works in a first test.Before making a PR some Feedback would be nice.

![gcs_testvalue_send2](https://cloud.githubusercontent.com/assets/9142182/8192889/8d24a38e-1471-11e5-829b-e2248c062c63.png)

Trex4Git:
Instead a text button, a icon ToolButton with an already existing and used icon:

![gcs_testvalue_send3](https://cloud.githubusercontent.com/assets/9142182/8195000/ac8b96ac-147f-11e5-911f-922e448a63f8.png)

Trex4Git:
As alternative 2 new Icons. Interested in your opinion.

![up_edit](https://cloud.githubusercontent.com/assets/9142182/8251933/3b71878e-1681-11e5-908b-5bc18f5089eb.png) or ![refresh_edit](https://cloud.githubusercontent.com/assets/9142182/8251936/468b005a-1681-11e5-81ac-c491e173e1ac.png)

Trex4Git:
Icon 1 was my favorite, mainly because of better consistency with the other icons in the Widget.

But for better looking in the real widget i had to change small things.
When there is no more feedback i think i will make a pull request with this.

![gcs_picoc_send_testvalue_2](https://cloud.githubusercontent.com/assets/9142182/8291534/86fb7a56-192a-11e5-91d7-ecccbd1061d0.png)

Trex4Git:
Created a pull request: https://github.com/TauLabs/TauLabs/pull/1686

mlyle:
Resolved by #1686.

PR #1686, opened 2015-06-24
GCS:  added TestValue send button to PicoC Widget

Trex4Git:
As requestet in this issue: https://github.com/TauLabs/TauLabs/issues/1404
Implementation of a Button to send the TestValue directly from the PicoC Widget to the UAV. For easy interaction with scripts. 
Before you had to got to the UAV Object Browser to do this.

gcs_picoc_send_testvalue_2


ground/gcs/src/plugins/picoc/picocgadgetwidget.cpp
```diff
@@ -385,6 +385,17 @@ void PicoCGadgetWidget::sleep(int msec)
 }
 
 /**
+* @brief PicoCGadgetWidget::on_tbTestValueSend_clicked  Send the Test Value
+ */
+void PicoCGadgetWidget::on_tbTestValueSend_clicked()
```
guilhermito: These signal/slots connections based on widget name while easier/faster to code are IMO harder to read and aren't usually used on the GCS code. Can you change it please?

Trex4Git: Thanks for your feedback,
i also had to search a little bit how that auto connect for "Signal/Slot" works as i am new to GUI and QT coding. 
I think this is a official description:  http://doc.qt.io/qt-5/designer-using-a-ui-file.html#widgets-and-dialogs-with-auto-connect

I used it, because all other Slot functions in this widget use the same functionality, and i wanted to keep it consistent in this file.

Maybe adding a comment to this file with a short description and a link to the auto connect documentation is better, than just changing only this function.

Trex4Git:
Some feedback how i should proceed with that would be nice.

guilhermito:
Sorry, yes that file uses the auto connection mechanism throughout so I guess keeping it is ok. Merging

PR #1691, opened 2015-06-26
ENH: Move things to fast heap

### LINKED ISSUE #1684: Free memory on F4 targets by moving things to fastheap?
The memory on the STM32F405 targets is segmented into 128kB of normal SRAM and 64kB of CCM-SRAM, the latter can only be accessed by the CPU and not the DMA controller. We are currently using the first 128kB for most things, while the CCM-SRAM is used for "fastheap", which is largely empty (about 8kB out of ~60kB are used).

It seems to me that there should be things that could move to the fastheap, i.e., as long as a buffer is not accessed by the DMA controller, it could be moved to fastheap. This could free up a lot of the normal heap. One example are the PicoC process stacks, which AFAIK don't use DMA.

Has anyone looked into this before? Thoughts?

BrainFPV:
Moves some things to fast heap (heap on F4 targets that can only be accessed by the CPU and not the DMA controller). I only moved things in `PiOS` and not drivers, as it is easier to test and has a more significant impact (as queues etc. are allocated in many places).

On Brain, this reduces the heap utilization by about 5.5kB.

Fixes #1684

### DIFF
--- a/flight/PiOS/Common/pios_streamfs.c
+++ b/flight/PiOS/Common/pios_streamfs.c
@@ -174,7 +174,7 @@ static struct streamfs_state *streamfs_alloc(void)
 {
 	struct streamfs_state *streamfs;
 
-	streamfs = (struct streamfs_state *)PIOS_malloc(sizeof(*streamfs));
+	streamfs = (struct streamfs_state *)PIOS_malloc_no_dma(sizeof(*streamfs));
 	if (!streamfs) return (NULL);
 
 	streamfs->magic = PIOS_FLASHFS_STREAMFS_DEV_MAGIC;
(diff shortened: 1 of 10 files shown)


fujin:
Code looks straight forward :+1: I'll bench test to verify the heap utilization. Anything else that we need to test, @BrainFPV?

flight/PiOS/Common/pios_streamfs.c
```diff
@@ -542,7 +542,7 @@ int32_t PIOS_STREAMFS_Init(uintptr_t *fs_id, const struct streamfs_cfg *cfg, enu
 		goto out_exit;
 	}
 
-	streamfs->com_buffer = (uint8_t *)PIOS_malloc(cfg->write_size);
```
solidgoldbomb: This specific malloc looks like it would result in breakage if this filesystem is ever used on top of a flash driver that uses DMA (ie. where ```PIOS_FLASH_write_data()``` uses DMA underneath.

solidgoldbomb:
The streamfs com_buffer change seems like it could be a future trap.  The rest of the changes look good.  @BrainFPV do you have a sense for which of these changes yielded the largest saving?

Would be nice if we had a way for drivers that use DMA to enforce the DMA safety of any parameters passed to it.  I'm not aware of a way to do that though.

mlyle:
So, I'm in favor of this merging.  We can have a DMA-safety assert in the future, while this provides an immediate win in bus contention, performance, and available memory.

PR #1692, opened 2015-06-27
FIX Sparky2 Accel LPF selection

BrainFPV:
Something I noticed when stealing some code to use on Brain ;)


peabody124:
Good catch, thanks!

PR #1694, opened 2015-06-28
Logging fix

peabody124:
Fixes a bug introduced by #1682 that connected to objects even if they might not be
registered by the ObjectManager. This would cause CPU usage of the logging module
to skyrocket.

### DIFF
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -207,6 +207,8 @@ static void loggingTask(void *parameters)
 			case LOGGINGSETTINGS_MAXLOGRATE_100:
 				PIOS_Thread_Sleep(10);
 				break;
+			default:
+				PIOS_Thread_Sleep(1000);
 		}
 
 		LoggingStatsGet(&loggingData);
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -289,7 +289,7 @@ static void loggingTask(void *parameters)
 				flightstatus_updated = false;
 			}
 
-			if (waypoint_updated){
+			if (waypoint_updated && WaypointActiveHandle()){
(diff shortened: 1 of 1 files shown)


peabody124:
@BrainFPV can you look over this and test it on your system. The code worked on brain because all objects are registered all the time for the OSD (btw I can send you a patch to that code later that checks for their existence too).

flight/Modules/Logging/logging.c
```diff
@@ -307,18 +310,21 @@ static void loggingTask(void *parameters)
...
 
 			// Log slow
-			if ((i % 50) == 2) {
+			if ((i % 50) == 2 && GPSTimeHandle()) {
 				UAVTalkSendObjectTimestamped(uavTalkCon, GPSTimeHandle(), 0, false, 0);
 			}
 
 			// Log very slow
-			if ((i % 500) == 3) {
+			if ((i % 500) == 3 && GPSPositionHandle()) {
```
BrainFPV: should be `GPSSatellitesHandle()`

BrainFPV:
@peabody124 sorry for the breakage and thanks for the fix. Looks good to me otherwise.

The problem didn't occur on the Brain because the OSD module initializes all UAVOs it needs when initializing. No need to send a patch; I meant to clean up that code anyways.

BrainFPV:
PS: With the fix for the OSD below, I can reproduce the problem. See also #1695

https://github.com/BrainFPV/TauLabs/commit/a99a5264cdb85aa2366b43b4562829afbe777c9b

peabody124:
@BrainFPV fixed that line. If this is good can @guilhermito merge this? Then we should be up for an -RC

BrainFPV:
@peabody124  what about #1635 any chance it can make it into the RC?

BrainFPV:
Someone should merge this, as the logging module is currently broken in `next`.

peabody124:
I'll merge this since you reviewed it and it fixes a known breakage.

PR #1696, opened 2015-06-29
GCS & Flight: Expo functionality rework

Trex4Git:
As the HorizonExpo effects are a little bit complicated, and i grouped them in the fist version to the attitude settings in Advanced stabilization 8what is not ideal), i did some rework on this topics.
I think its now less confusing for non experts.

- add a separate attitude exponential setting (GCS/flight) 
- extended expocurve functionality for AttitudeExpo
- in Advanced Stabilization Changed HorizonExpo settings and plot to new AttitudeExpo
- Moved HorizonExpo settings and plot to Expert Stabilization
- changed Expert Stabilization UI a little bit to fit the HorizonExpo settings and plot
- changed UI for TPA & added Default button for consistent layout

I did not change the Horizon mode functionality, as this needs proven advantages from flight tests and some user request would also be good. 
So that (https://github.com/TauLabs/TauLabs/pull/1676#issuecomment-112931666) may be done in the future if there is demand.


Trex4Git:
Advance Stabilization:


screenshot from 2015-06-29 22 17 18

Expert Stabilization:

screenshot from 2015-06-29 22 18 25

Trex4Git:
- changed UI/layout for the "Axis lock"/"Weak Leveling" settings
- rebased to latest next, because of some changes in stabilization.ui
- rearranged the tab order according to the other tab order changes in next

screenshot from 2015-07-14 19 40 39

Trex4Git:
This should not merge before the last release, as it was done during the "feature freeze". 
But that is already some time ago, and I am interested if I should make some changes.

Or is there no more interest, than its maybe better to close this.

mlyle:
@Trex4Git I'm sorry that this hasn't had my attention yet.

This looks like a good change.  I would really like it if on the advance stabilization page you could put the graphs on the right instead of the bottom.  That will help make all the nested boxes look bigger, and also allow us to set the graphs to be able to be shrunk when the window shrinks.  1366x768 screens are still common and it's much more natural to scroll up and down than left to right.

Trex4Git:
Just a quick test, on a different branch.
does this look like, what you preferred?

Fullscreen (in my VM):
screenshot from 2015-08-01 16 28 37

reduced windows size:
screenshot from 2015-08-01 16 28 24


I am not sure if the Expert tab should be modified the same.

mlyle:
That looks beautiful.

It would be nice if the expert tab could be too, but I think there's only one graph so I don't see how it could work as well.

mlyle:
I have one other small request if it is easy while you're changing the rest of this.  If not you can feel free to ignore it (not mandatory for this pull).

arrows

Basically, it would be nice if we could remove the innermost boxes (indicated by arrows) on these tabs.  They use space and clutter the screen.

Trex4Git:
I am also not a real fan of this boxes. But they seem to have also a layout functionality, and if the are removed it seems to me a lot of work is needed to align everything again.
And i don't know if just the border can be made invisible(combined with some reduced layout margins).
They were already they before I started my work on this.

And 2 versions of changed layout for expert tab, you need to scroll up and down more than before in both versions (both  with reduced window size):

- Similar layout of the plot to the advanced tab, but i am not a big fan of this
screenshot from 2015-08-01 20 06 15

- A little bit more scrolling needed, and for the plot not the same layout as on advanced tab. But the horizon plot is the most complex one, so more space for it might be a good thing.
screenshot from 2015-08-01 20 24 57

mlyle:
OK, if that's hard (I didn't know) don't bother with it).

mlyle:
I see your expert mode one.  It looks like the horizon is at the bottom... and if so that looks great for now.

Trex4Git:
For both test for a new expert layout I placed the horizon settings at the bottom, but I can also palce them at  top or somewhere in between, thats a small change.

And you prefer the expert mode page with the smaller plot?

mlyle:
I think at bottom is great.    And no, I like this one: https://cloud.githubusercontent.com/assets/9142182/9023054/5058d080-388c-11e5-8d6e-bcd4d507e170.png

Trex4Git:
Thats also my preferred solution, and I misunderstood you, sorry.

So from your opinion its good to include these changes into this PR?

mlyle:
Yes.

mlyle:
Looks good.  I will try and test and do detailed code review in the next couple of days.

mlyle:
Looks great!

Trex4Git:
Thank you.

Issue #477, opened 2013-04-27
sin_lookup does not correctly handle negative angles


peabody124:
Because the % operator preserves the sign of the input, sin_lookup can reference into a random region of flash for negative angles.  Need to do a similar thing to the circular modulus in #469.

When this is done it should also be included in the math unit test coverage (see #476 )

mlyle:
Removed.

PR #1701, opened 2015-07-04
math: remove sine table

mlyle:
It wasn't used, and it wasn't accurate at all (7-8 bit resolution).  It has problems with negative angles (e.g. issue #477.


peabody124:
Yeah this was used for the older autotune to compute some fast frequency transforms. If it is no longer used I'm happy to trash it and we can bring it back if ever required. Let's wait until after release is finalized, just to have a clean window for testing (if you don't mind, there is no real reason deleting this should change anything though)

mlyle:
Needed rebase because it touched freedom makefile.

PR #1702, opened 2015-07-04
Flight / UAVObject: added S.Bus option for RSSI type

Trex4Git:
For example with the taranis you can send the RSSI value also as one of the rc channels.
So when using S.Bus you don't need extra cabels, filters etc. Just select the configured channel.

Signed-off-by: Trex4Git <trex_dev@gmx.net>


mlyle:
Looks good!  Thanks.

Trex4Git:
A way to setup the Taranis for getting as one of the rc channels can be seen there: https://www.youtube.com/watch?v=d71VdSQmpJo

Issue #1442, opened 2015-01-13
macdeployqt broken in Qt 5.4 which breaks packaging


peabody124:
As reported by balu, the welcome screen is broken again. There is a regression in macdeployqt for Qt 5.4:

https://bugreports.qt.io/browse/QTBUG-43100
https://bugreports.qt.io/browse/QTBUG-42952
https://bugreports.qt.io/browse/QTBUG-42954

http://forum.taulabs.org/viewtopic.php?f=18&t=480#p3846

peabody124:
For now there is a workaround for this. Compile Qt5.4.0 with the patch from here https://bugreports.qt.io/browse/QTBUG-42954 when packaging OSX. This is more appealing than continuing to hold OSX back a version compared to the other operating systems, and the fix should be applied in Qt 5.4.1.

peabody124:
I'm removing the blocker label on this because there is a workaround currently (recompiling Qt with the patch on macdeployqt). I'll leave it open until Qt 5.4.1 is released and we confirm that fixes the root issue.

mlyle:
Hack workaround merged in #1703; when we pick up 5.5.0 it should be reverted.

mlyle:
Fix merged, yaaay.

PR #1703, opened 2015-07-04
packaging: QML PFD: work around QTBUG-42954

mlyle:


mlyle:
Added fix for welcome screen.

peabody124:
Can you see if this is needed with #1704? That might be a better solution if they fixed the packaging bug.

mlyle:
This is definitely an interim fix/hack.  It should not be needed with 5.4.2 or 5.5.0.

It's just a question of risk-- does picking up QT 5.5 introduce new issues?  The OpenGL build will break on windows assuming the QT 5.5 is OpenGL dynamic like I think it is.  I have a branch somewhere with most of the required changes that I could prepare and submit.  Alternatively we could go to QT 5.4.2 which should fix this and be less invasive.

In either case it'd take Jenkins changes.

peabody124:
I can make some test 5.5 builds tomorrow and we can see if they have breakage. If so I agree 5.4.2 is better.

mlyle:
Issue #1442

peabody124:
This works. Using this until we upgrade to either 5.4.2 or 5.5. Then we should remember to revert it.

PR #1705, opened 2015-07-04
GCS/ExpoCurve: use a new QwtSymbol instance for each QwtPlotMarker

guilhermito:
Fixes crash on exit


mlyle:
Looks good!  Thanks for running this down.  @Trex4Git FYI this is our interim solution.

PR #1706, opened 2015-07-05
GCS: Expo fix for segmentation fault error on closing

Trex4Git:
There is a little bit strange behaviour in QWT Marker, as it wants to delete the symbol.
But the symbol is allocated elsewhere, and only a pointer is passed to the marker.

I tried different things, but the only one that seems to fix this bug, is to use a separate symbol for each marker.
Even if they should look the same.

Signed-off-by: Trex4Git <trex_dev@gmx.net>


Trex4Git:
Sorry, 
had yesterday not much freetime to fix this, and after i created this PR i have seen that it was already fixed with more or less the same solution.

Should i just close this PR?

mlyle:
Yes, sorry.

PR #1707, opened 2015-07-05
make: upgrade qt install package to qt5.4.2

solidgoldbomb:
This is supposed to include a fix for:
* qtbug-42954 (https://bugreports.qt.io/browse/QTBUG-42954)
* qtbug-42952 (https://bugreports.qt.io/browse/QTBUG-42952)
which broke packaging on mac.

Builds and runs (no board attached yet) on Linux 64-bit.

Needs testing on Windows and (especially) OSX.

### DIFF
--- a/make/tools.mk
+++ b/make/tools.mk
@@ -22,7 +22,7 @@ ifdef OPENOCD_FTDI
 endif
 
 # Set up QT toolchain
-QT_SDK_DIR := $(TOOLS_DIR)/Qt5.4.1
+QT_SDK_DIR := $(TOOLS_DIR)/Qt5.4.2
 
 ifdef LINUX
   ifdef AMD64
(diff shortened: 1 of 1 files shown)


solidgoldbomb:
@lilvinz Could you please install qt 5.4.2 on jenkins so this can build?

I'm not sure if we'll use 5.4.2 in the next release or if we'll upgrade all the way to 5.5.

mlyle:
Issue #1442

solidgoldbomb:
@mlyle what do you mean by "Issue #1442"?  Does this fix the issue or not fix the issue?

mlyle:
Just wanted to make sure it was linked so the issue could be closed.

It fixes the issue, but so does the previously accepted PR.  I recommend we close this PR and keep the hack, and then immediately after release stabilize 5.5.0 and revert PR #1703.

solidgoldbomb:
Ok.  @lilvinz, no need to install qt 5.4.2 on jenkins.

PR #1708, opened 2015-07-05
do not panic on external i2c errors

lolorc:
I'm using an ATtiny as a I2C slave. My quanton panics when the attiny is not powered (when the FC is only powered by the usb port)

16:25 < peabody124> laurent: yeah i agree throwing error instead of panic makes more sense for external sensors

I took your words for it ;-) here is a proposal I tested with one quanton.
I'm not too sure about flyingf3/f4, by definition i think all sensors are external ones on these platforms.

### DIFF
--- a/flight/targets/colibri/fw/pios_board.c
+++ b/flight/targets/colibri/fw/pios_board.c
@@ -692,7 +692,7 @@ void PIOS_Board_Init(void)
 		}
 
 		if (PIOS_I2C_CheckClear(pios_i2c_usart1_adapter_id) != 0)
-			panic(6);
+			AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
 
 #if defined(PIOS_INCLUDE_HMC5883)
 		{
@@ -705,9 +705,9 @@ void PIOS_Board_Init(void)
 				if (PIOS_HMC5883_Init
 				    (pios_i2c_usart1_adapter_id,
 				     &pios_hmc5883_external_cfg) != 0)
-					panic(8);
+					AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
 				if (PIOS_HMC5883_Test() != 0)
-					panic(8);
+					AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
 			}
 		}
 #endif /* PIOS_INCLUDE_HMC5883 */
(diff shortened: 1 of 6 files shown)


carbogit:
a most welcome improvement!

flight/targets/colibri/fw/pios_board.c
```diff
@@ -692,7 +692,7 @@ void PIOS_Board_Init(void)
 		}
 
 		if (PIOS_I2C_CheckClear(pios_i2c_usart1_adapter_id) != 0)
-			panic(6);
+			AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
```
peabody124: Checking the I2C is clear should still be a panic. If nothing is there the bus should still panic. This should be fixed throughout.

peabody124:
@mlyle I know you are planning some HAL abstraction on all the ports. Would merging this be an issue for that, or just demonstrate the need? :D

mlyle:
I haven't vetted this for safety.  I have no problem with it merging.

But it's another thing that is drifting the ports further apart.

lolorc:
not panicking on external checkclear was initially the point of this PR.
checkclear will always fails if the i2c slave is not powered and sda/sdc are pull down

lolorc:
@mlyle what ports are you talking about ?

jihlein:
I tried to test this on an F3, but checkclear on the F3 never makes it to panic (or the alarmset if so modified) when the slave is not powered, it hangs in the routine.  Can someone else verify this to make sure it's not a quirk of my setup?

BrainFPV:
For mags, it would make sense to have an error if no mag update has been received for some time, like we do for the barometer in the Sensors module. At the moment, it is possible that an external I2C disconnects mid flight and no alarm is raised whatsoever. The cleanest way to handle this would be to add a mag alarm.

peabody124:
@BrainFPV yeah I agree. the INS will error until it gets a mag update but not fail if the stop coming.

@lolorc @jihlein this might be board specific. most of the time i would expect the pullup on the FC to work when the sensor is powered off still...

lolorc:
@peabody124 so what would be acceptable ? a simple PR for quanton making it not panic on either false condition (checkckear, mag init, mag test)
Or should I let all other platforms but quanton panic on !checkclear and leave alarmset on !mag test or !mag init ?

peabody124:
I don't know that this will be a critical issue. @mlyle will likely
refactor all of this code in not too long with a nice configuration API
that is reused across boards. It should be as 'assertive' as possible while
still functioning, though. In the sense that do the FlyingF3/4 platforms
have pulls up on the I2C pins internally? If so, can you verify they panic
on the check clear when sensor is disconnected? What about connected but
unpowered.

If the standard sensors (e.g. mag on external GPS) when unpowered pull down
the I2C bus, that will be annoying. Checking it is clear is a useful
distinction from failing to initialize a particular sensor and I've found
that panic quite useful when debugging hardware.

lolorc:
no worries, this is not critical at all, I've just encountered this with quanton and to be honest I have no clue about flyingf3/f4, sparky 1 and 2.
I'm closing this PR. We'll see what comes out of the refactor.I'd just say make it the most complex and configurable as possible. (but a target define would probably be enough)
for now i'll add that to my port.

Thanks for the comments and suggestions (especially @BrainFPV's one, that would make sense to failsafe in that case)

PR #1710, opened 2015-07-07
target: Naze32 support

mlyle:
Initial support; needs further work.


tracernz:
I've started working on this in the last few days. I think this currently breaks the pipxtreme target due to the hse_value variable introduced in the RCC driver. I fixed it in my branch https://github.com/tracernz/TauLabs/commit/4807b2e9e0c5ea06c32bab15580a05be1138dca3

PR #1713, opened 2015-07-09
ManualControl: Use rising edge to detect switch arming

tracernz:
Currently when using switch arming (switch or switch+throttle) with an arming timeout delay set, the arming state will toggle between armed and disarmed every time the inputs are processed while the throttle is low (throttle command less than 0, i.e. below neutral). This is because the disarm timeout is not initialised for switch arming.

This PR fixes that by detecting a rising edge on the arming position and initialising the disarm timeout correctly (Note: it is necessary to use a rising edge as otherwise every time the disarm timeout caused a disarm, the switch would cause a re-arm on the next iteration). Perhaps there is a better name for the last_arm and I can change it if there's already a convention in the code.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -621,15 +621,22 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
 	{
 		set_armed_if_changed(FLIGHTSTATUS_ARMED_DISARMED);
 
+		// last_arm used for detecting a rising edge to trigger switch arming
+		static bool last_arm = false;
 		bool arm = arming_position(cmd, settings) && valid;
 
 		if (arm && (settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCH ||
 				settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCHTHROTTLE)) {
-			arm_state = ARM_STATE_ARMED;
+			if (!last_arm) {
+				armedDisarmStart = lastSysTime;
+				arm_state = ARM_STATE_ARMED;
+			}
 		} else if (arm) {
 			armedDisarmStart = lastSysTime;
 			arm_state = ARM_STATE_ARMING;
 		}
+
+		last_arm = arm;
 	}
 		break;
 


peabody124:
jenkins: test this please

peabody124:
jenkins: add to whitelist

peabody124:
Nice catch. Since this is a bugfix, we should probably merge it before the
final tagged release.

peabody124:
Problem makes sense and fix looks good. Let me reproduce and test fix then will merge.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -621,15 +621,21 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
 	{
 		set_armed_if_changed(FLIGHTSTATUS_ARMED_DISARMED);
 
+		static bool last_arm = false;
```
peabody124: could we move this to the scoping at 628 to make it clear this only alters the behavior of switch arming?

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -621,15 +621,21 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
...
 		bool arm = arming_position(cmd, settings) && valid;
 
 		if (arm && (settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCH ||
 				settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCHTHROTTLE)) {
-			arm_state = ARM_STATE_ARMED;
+			if(!last_arm) {
```
peabody124: insert space before ```(```

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -621,15 +621,21 @@ static void process_transmitter_events(ManualControlCommandData * cmd, ManualCon
...
+				armedDisarmStart = lastSysTime;
+				arm_state = ARM_STATE_ARMED;
+			}
 		} else if (arm) {
 			armedDisarmStart = lastSysTime;
 			arm_state = ARM_STATE_ARMING;
 		}
+
+		last_arm = arm;
```
peabody124: ditto about scoping. please move above.

peabody124:
tested and it fixes the problem. just a code style issue and code clarity issue and it should be good to merge.

tracernz:
Is this what you meant? last_arm will never be set to false after the first time you arm this way... Then you couldn't re-arm..
```C
{
		set_armed_if_changed(FLIGHTSTATUS_ARMED_DISARMED);

		
		bool arm = arming_position(cmd, settings) && valid;

		if (arm && (settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCH ||
				settings->Arming == MANUALCONTROLSETTINGS_ARMING_SWITCHTHROTTLE)) {
			static bool last_arm = false;
			if (!last_arm) {
				armedDisarmStart = lastSysTime;
				arm_state = ARM_STATE_ARMED;
			}
			last_arm = arm;
		} else if (arm) {
			armedDisarmStart = lastSysTime;
			arm_state = ARM_STATE_ARMING;
		}

	}
```

peabody124:
well that is a darn good point, thanks for not blindly doing what i suggested :D

peabody124:
jenkins: test this please

Issue #1716, opened 2015-07-13
Strange tab input selection order in Stabilization / Advanced screen


dustin:
Tabbing between inputs, I get this really strange order:

![tau_labs_gcs](https://cloud.githubusercontent.com/assets/1779/8644084/c59c9972-28f0-11e5-8dc0-fe0c06ad4342.png)

Also, to even get there from the previous ones, it jumps through a couple odd places (system viewer on the right and module selection on the left).

Trex4Git:
That is good to understand, thanks for the feedback.

I hope that today after work i find the time to take a closer look at it, and i think fixing it would not be much work.

peabody124:
There is a button at the top of the designer that lets you edit tab order:
![screen shot 2015-07-13 at 11 09 53 am](https://cloud.githubusercontent.com/assets/618346/8654412/c0517b90-294f-11e5-9e9d-105ee99ff8b5.png)

Also, when saving the .ui file, make sure to go back to the basic tab first. It preserves this as the default tab when you go there. It would be nice to have this fix in the final release.

Trex4Git:
Already had seen that this morning, and now had time to fix it.

![advanced1_new](https://cloud.githubusercontent.com/assets/9142182/8659069/563bc6bc-29a8-11e5-93c8-d72be0f24647.png)
![advanced2_new](https://cloud.githubusercontent.com/assets/9142182/8659072/5a7d360c-29a8-11e5-8fe9-ef141cf9a94b.png)

And as not only the advanced tab with the new expo settings had strange tab order, i changed it in the complete stabilization widget:

![basic_new](https://cloud.githubusercontent.com/assets/9142182/8659103/9be7db06-29a8-11e5-81ec-3b883954ba0e.png)
![mwrate_new](https://cloud.githubusercontent.com/assets/9142182/8659109/a209080c-29a8-11e5-8a8c-3c0331dbc09d.png)
![expert_new](https://cloud.githubusercontent.com/assets/9142182/8659111/a60279ac-29a8-11e5-925e-89580d8a5c3a.png)

When this is OK i can make a PR from this branch: https://github.com/Trex4Git/TauLabs/commit/a4351d6d479798b569e759e43b179e4067be4d81

mlyle:
Fixed by #1718

PR #1718, opened 2015-07-13
GCS: Tab Order fix for Stabilization Config

Trex4Git:
Rearranged the tab order in complete Stabilization Config.
It was not only strange in advanced tab because of exp changes, but also in other tabs.

Regarding this issue: https://github.com/TauLabs/TauLabs/issues/1716

### DIFF
--- a/ground/gcs/src/plugins/config/stabilization.ui
+++ b/ground/gcs/src/plugins/config/stabilization.ui
@@ -498,7 +498,7 @@
       <enum>QTabWidget::Rounded</enum>
      </property>
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <property name="documentMode">
       <bool>false</bool>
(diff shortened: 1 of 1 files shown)


Trex4Git:
Not only changed the advanced tab with the new expo settings that had strange tab order:

advanced1_new
advanced2_new

I also changed it in the complete stabilization widget, because there were also some strange tab orders:

basic_new
mwrate_new
expert_new

mlyle:
Looks great!   Thanks for tidying this up.

ground/gcs/src/plugins/config/stabilization.ui
```diff
@@ -498,7 +498,7 @@
       <enum>QTabWidget::Rounded</enum>
      </property>
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
```
guilhermito: This looks strange

Trex4Git: I think this element describes which tab (Basic, Advanced, MultiWii Rate, Expert) is the default one.

I changed that unintentionally to Advanced (1) when i did the expo changes; because when you save the ui file it sets automatically the currently opened tab as default.
Peabody gave me that hint.

So  this is now changed back to basic (0).

PR #1719, opened 2015-07-13
revolution: remove base revolution target

mlyle:
This is probably unused by any actual users.  Revomini is what's actually used.


ground/gcs/src/plugins/boards_openpilot/boards_openpilot.pro
```diff
@@ -13,15 +13,12 @@ HEADERS += \
     openpilotplugin.h \
     coptercontrol.h \
     config_cc_hw_widget.h \
-    revolution.h \
-    revomini.h \
```
solidgoldbomb: Removing ```revomini.h``` here doesn't fit the pattern.  Too aggressive with the delete or was this file just unused or something?

solidgoldbomb:
Other than the slightly odd removal of revomini.h, this looks ok to me.

mlyle:
You're right-- I accidentally deleted that line/dependency.  It should be OK now-- OK to merge when Jenkins says it's OK?

PR #1723, opened 2015-07-16
GPS: Change GPS error when no port is assigned to GPS

DTFUHF:
When no port is assigned to GPS but the GPS module is enabled, the GPS alarm is set to Critical (filled red box), which gives the user the indication that the GPS is receiving data but has zero satellites. This request changes the alarm to Error, which is indicated with a red X, and the tooltip is updated to include this case.


mlyle:
Looks good, thanks!

PR #1725, opened 2015-07-17
Flt: Sparky2 makefile changes for chibios/windows build

jihlein:
Sparky2 build fails under windows after change to chibios.  Makefile needs same update the other F4/chibios targets required.


mlyle:
Good catch!

Issue #1724, opened 2015-07-16
Attitude task doubles in CPU usage with high frequency (250hz) gyro lpf


fujin:
![2015-07-17-105220_1920x1080_scrot](https://cloud.githubusercontent.com/assets/6873/8737697/3be038d8-2c79-11e5-9a27-1a6cc1429740.png)

Reproduced on Brain-20150714. Dropping back to 184hz brings the cpu down to 7% on the Attitude task.

fujin:
@BrainFPV :ocean:

BrainFPV:
@fujin this is a limitation of the MPU9250; when the lpf is 250Hz, the MPU9250 is running at 8KHz, thus causing crazy CPU usage. I removed the 250Hz option for Brain, the same should probably be done for sparky2. 

I know you would probably like it, as it should give a super low tau in auto tune ;).

fujin:
when #1726 lands we can consider this resolved I think. We should maybe document the issue for future MPU9250 targets?

https://github.com/BrainFPV/TauLabs/commit/af84efa74130d1afdf08712206d8b4fef5520e0f

mlyle:
Fixed for now with #1726

PR #1726, opened 2015-07-17
sparky2: remove 250Hz gyro LPF.

mlyle:
Issue #1724 / From pios_mpu9250_spi.c:
    int32_t PIOS_MPU9250_SetSampleRate(uint16_t samplerate_hz)
    {
            // mpu9250 ODR divider is unable to run from 8kHz clock like mpu60x0 :(
            // check if someone want to use 250Hz DLPF and don't want 8kHz sampling
            // and politely refuse him

I don't know if this is true or not-- the datasheet is ambiguous in its wording and I have not tested on bare hardware.  But in any case the current 250Hz DLPF code is broken, doesn't respect the user-configured sample rate, and trying to run the flight loop at 8KHz is unreasonable.

### DIFF
--- a/flight/targets/sparky2/fw/pios_board.c
+++ b/flight/targets/sparky2/fw/pios_board.c
@@ -1219,7 +1219,6 @@ void PIOS_Board_Init(void) {
 	uint8_t hw_mpu9250_dlpf;
 	HwSparky2MPU9250GyroLPFGet(&hw_mpu9250_dlpf);
 	enum pios_mpu9250_gyro_filter mpu9250_gyro_lpf = \
-	    (hw_mpu9250_dlpf == HWSPARKY2_MPU9250GYROLPF_250) ? PIOS_MPU9250_GYRO_LOWPASS_250_HZ : \
 	    (hw_mpu9250_dlpf == HWSPARKY2_MPU9250GYROLPF_184) ? PIOS_MPU9250_GYRO_LOWPASS_184_HZ : \
 	    (hw_mpu9250_dlpf == HWSPARKY2_MPU9250GYROLPF_92) ? PIOS_MPU9250_GYRO_LOWPASS_92_HZ : \
 	    (hw_mpu9250_dlpf == HWSPARKY2_MPU9250GYROLPF_41) ? PIOS_MPU9250_GYRO_LOWPASS_41_HZ : \
--- a/shared/uavobjectdefinition/hwsparky2.xml
+++ b/shared/uavobjectdefinition/hwsparky2.xml
@@ -77,7 +77,7 @@
 		<field name="GyroRange" units="deg/s" type="enum" elements="1" options="250,500,1000,2000" defaultvalue="2000"/>
 		<field name="AccelRange" units="*gravity m/s^2" type="enum" elements="1" options="2G,4G,8G,16G" defaultvalue="8G"/>
 		<field name="MPU9250Rate" units="Hz" type="enum" elements="1" options="200,250,333,500,1000" defaultvalue="500"/>
-		<field name="MPU9250GyroLPF" units="Hz" type="enum" elements="1" options="250,184,92,41,20,10,5" defaultvalue="184"/>
+		<field name="MPU9250GyroLPF" units="Hz" type="enum" elements="1" options="184,92,41,20,10,5" defaultvalue="184"/>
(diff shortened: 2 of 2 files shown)


fujin:
WRT: https://github.com/TauLabs/TauLabs/issues/1724 ?

fujin:
:+1: @BrainFPV Just did the same for Brain target @ https://github.com/BrainFPV/TauLabs/commit/af84efa74130d1afdf08712206d8b4fef5520e0f

Trex4Git:
I think 250Hz DLPF means disabling the DLPF in the MPU. And from the documentation of the MPU9250 this also means the sample rate divider does not work, so we have constant 8 kHz sampling rate. 
(When DLPF is enabled the sample rate divider should work, but on a 1kHz base)

There is also a LPF in our code.
I think i have read somewhere in the forums that the DLPF in the MPU has a little bit strange behavior and that's why disabling it and only using the software LPF is preferred. But that was regarding MPU6000/6050, and with them you could use the sample rate divider also with DLPF disabled on 8kHz base.

So when this 250Hz option is removed, the setting of the sample rate should work, but we always will use the DLPF in the MPU additional to our software LPF.

As a temporary fix it should work, but for future something that i already thought about when thinking about buying a sparky 2, and that i have also read on IRC is, to run the MPU with no DLPF at 8kHz, but don't trigger our control loop with every interrupt, but do a sub-sampling on the MPU data, and run the control loop only every n interrupt.

mlyle:
Doing our own lowpass and decimation would be nice--- and I've been suspicious that we have aliasing problems in general.

For a fast decimator filter, look at CIC filters (though we'd need to do fixed point).

But an 8KHz interrupt rate is still rather high.

Trex4Git:
I thought about using the already existing IIR LPF filter: https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Stabilization/stabilization.c#L296-L298

mlyle:
So, the problem with a first-order IIR is that it doesn't have a sharp enough stopband -and- it has a rather large group delay.  Both are a problem when decimating.

Trex4Git:
I opened a new issue regarding the sub-sampling development. 
There is a link to my first implementation and first results, but as I am not satisfied with that, I want to do some more development before opening a PR.
But I also don`t want to to do more off-topic discussion here, so I thought a issue might be a good place for that at this stage.

peabody124:
@Trex4Git I was going to post in this last night but wanted to do some tests first. I suspect simply accumulating samples and averaging them over the downsampling window will work well when running at 8khz to bring it down to the 1-2 khz range. Especially since for the attitude estimation we integrate the gyro results (and they are already antialiased) this is a fairly sensible thing to do.

I also might prefer to see a "proper" fix to this rather than dropping the 250Hz mode. I wanted to test that last night - if it still flies when running at 8khz (I know the CPU usage is high). DIdn't get that far, though.

Trex4Git:
As describe in https://github.com/TauLabs/TauLabs/issues/1730 the next thing i want to test, is doing the sub-sampling in the MPU task instead of the stabilization task.
For first tests i also wanted to do a simple averaging, and test the performance.
Later i thought it would be nice to have the sub-sampling unified, and not a averaging in the MPU driver, and additional a LPF filter in the stabilization task.
And as the IIR LPF is already running every time and tested, i thought about a little refactoring and moving the LPF filter from the stabilization task into the MPU task.

Trex4Git:
Next implementation tested: https://github.com/TauLabs/TauLabs/issues/1730#issuecomment-122659335

mlyle:
I think we should merge this pr.  We are dropping samples and using excessive cpu.  The way this works it is going to create aliasing too.

mlyle:
Look, @tracernz and I are running down issues multiple millisecond delays / jitter on sparky2 at 500Hz and 1KHz rates.  We should not be running at 8KHz.  This should merge.

Trex4Git:
As for the subsampling/downsamplig stuff more functionality was requested, it will take some time to get finished.

So merging this now, as a temporary security patch should be fine. And its easy to revert it later, if the downsampling might get merged sometime.

Trex4Git:
A additional note:
If someone has done some logging when MPU was running at 8kHz, it would be nice if I could get those logs. To have a look at the gyro data, for filter development.

Otherwise if someone has the time to make some measurements/logs with 8kHz before it gets removed would be nice. Hovering and flying would be great.

peabody124:
I'm OK with it merging then

mlyle:
OK, merging myself then.

PR #1731, opened 2015-07-18
UAVOFrSkySPortBridge: left only one device ID

xdandys:
With newer FrSky receiver firmwares there is no longer significant bandwidth
penalty when using only single one sensor device ID. It also makes new OpenTx (>=2.1.0)
happy, because when there is multiple sensor devices it treats each as separate
instance, which causes difficulties and also quickly fill all available telemetry slots on Taranis.


peabody124:
Any backward compatibilities with older versions of OpenTX?

xdandys:
This change should be absolutely compatible with older OpenTX versions. Tested with 2.0.12

PR #1728, opened 2015-07-18
20150718 - Prevent F3 I2C Init Lockup

jihlein:
I've observed F3 targets locking up when an external I2C device, such as an external mag, is connected, but not powered,  The F4 targets do not exhibit this behavior.

Looking at the PIOS I2C drivers, the F3 version is missing the infinite loop protection in 4 while loops in the i2c_adapter_reset_bus function.

Adding the infinite loop protection solves this issue.

In working thru this issue, I've discovered that external mag support in Sparky is broken.  I'll fix that under a separate pull request.

PR #1732, opened 2015-07-20
do not panic quanton on external i2c errors

lolorc:
following #1708 comments and #1728

### DIFF
--- a/flight/targets/quanton/fw/pios_board.c
+++ b/flight/targets/quanton/fw/pios_board.c
@@ -624,7 +624,7 @@ void PIOS_Board_Init(void) {
 		}
 
 		if (PIOS_I2C_CheckClear(pios_i2c_usart1_adapter_id) != 0)
-			panic(6);
+			AlarmsSet(SYSTEMALARMS_ALARM_I2C, SYSTEMALARMS_ALARM_CRITICAL);
 
 #if defined(PIOS_INCLUDE_HMC5883)
 		{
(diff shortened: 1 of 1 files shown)


flight/targets/quanton/fw/pios_board.c
```diff
@@ -624,7 +624,7 @@ void PIOS_Board_Init(void) {
 		}
 
 		if (PIOS_I2C_CheckClear(pios_i2c_usart1_adapter_id) != 0)
-			panic(6);
```
peabody124: i still don't entirely see this one. Can you confirm that if you have a sensor plugged in but not powered it pulls the bus down? I would think an I2C peripheral would be smart enough not to do that but perhaps not.

xdandys:
Coincidentally I tested this last weekend and can confirm that unpowered sensor pulls the bus down :( At least HMC5983. But it actually make sense, because there are internal protective diodes from I2C pins to positive power supply.

flight/targets/quanton/fw/pios_board.c
lolorc: I just tested again with a attiny as a i2c slave and also with a an external mag (HMC5883 GY-273).
If unpowered they pull the bus down and checkclear fails.

peabody124:
OK thanks for for humoring me and testing this. Will review this later. Should be easy then.

ernieift:
Not tested this, but makes sense. The HMC5883 check would fail without the device. This is quite ok. I think the changes are made for picoc. No one wants the FC in a panic loop when you use an unpowered external display or LED stripe.

flight/targets/quanton/fw/pios_board.c
mlyle: @peabody124 The ESD protection diodes on the parts pull the bus down.

mlyle:
Looks good.

flight/targets/quanton/fw/pios_board.c
theothercliff: I looked at a GPS/mag unit that I have this problem with, and it has pull up resistors on SCL and SDA.  They effectively become pull downs if the device is connected, but unpowered.

PR #1735, opened 2015-07-21
python: make program to do the 'doiap' bootloader sequence

mlyle:
Also some changes to make serial more reliable, simplify python code.

This supports the Naze effort among other things.


tracernz:
I have used this successfully on OS X and Linux with CC3D and Naze.

PR #1737, opened 2015-07-22
WIP: gcs: move GCS to C++11

mlyle:
This is infrastructure that is helpful for future changes, and will allow me to do a bunch of things more efficiently and with cleaner looking code.  Probably the biggest immediate win from the C++11 standard is the ability to use initializer lists with objects / QLists.


mlyle:
WIP because jenkins on linux showed some issues that didn't show up on mac or linux, so working those.

mlyle:
Actually will resubmit.

PR #1738, opened 2015-07-22
GCS: Build with C++11

mlyle:
This is infrastructure to make GCS more efficient and allow more elegant code.

Biggest immediate win is initializer lists, but it goes beyond that.

Apologies for the previous broken PR.  This one should be pretty good.  There's a little nuance in getting this right for all platforms.

### DIFF
--- a/ground/gcs/gcs.pri
+++ b/ground/gcs/gcs.pri
@@ -57,8 +57,6 @@ isEmpty(GCS_BUILD_TREE) {
 }
 GCS_APP_PATH = $$GCS_BUILD_TREE/bin
 macx {
-#    QMAKE_CFLAGS_X86_64 += -mmacosx-version-min=10.7
-#    QMAKE_CXXFLAGS_X86_64 = $$QMAKE_CFLAGS_X86_64
     GCS_APP_TARGET   = "Tau Labs GCS"
     GCS_LIBRARY_PATH = $$GCS_APP_PATH/$${GCS_APP_TARGET}.app/Contents/Plugins
     GCS_PLUGIN_PATH  = $$GCS_LIBRARY_PATH
--- a/ground/gcs/gcs.pri
+++ b/ground/gcs/gcs.pri
@@ -57,8 +57,8 @@ isEmpty(GCS_BUILD_TREE) {
 }
 GCS_APP_PATH = $$GCS_BUILD_TREE/bin
 macx {
-    QMAKE_CFLAGS_X86_64 += -mmacosx-version-min=10.7
-    QMAKE_CXXFLAGS_X86_64 = $$QMAKE_CFLAGS_X86_64
+#    QMAKE_CFLAGS_X86_64 += -mmacosx-version-min=10.7
+#    QMAKE_CXXFLAGS_X86_64 = $$QMAKE_CFLAGS_X86_64
     GCS_APP_TARGET   = "Tau Labs GCS"
     GCS_LIBRARY_PATH = $$GCS_APP_PATH/$${GCS_APP_TARGET}.app/Contents/Plugins
     GCS_PLUGIN_PATH  = $$GCS_LIBRARY_PATH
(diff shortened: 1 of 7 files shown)


mlyle:
Rebased because of conflict with merged gcov code.

ground/gcs/gcs.pri
```diff
@@ -67,6 +67,7 @@ macx {
     GCS_DATA_BASENAME = Resources
     GCS_DOC_PATH     = $$GCS_DATA_PATH/doc
     GCS_BIN_PATH     = $$GCS_APP_PATH/$${GCS_APP_TARGET}.app/Contents/MacOS
+    QMAKE_MACOSX_DEPLOYMENT_TARGET=10.9
```
peabody124: Does this affect backward compatibility?

ground/gcs/gcs.pri
```diff
@@ -57,8 +57,8 @@ isEmpty(GCS_BUILD_TREE) {
 }
 GCS_APP_PATH = $$GCS_BUILD_TREE/bin
 macx {
-    QMAKE_CFLAGS_X86_64 += -mmacosx-version-min=10.7
-    QMAKE_CXXFLAGS_X86_64 = $$QMAKE_CFLAGS_X86_64
+#    QMAKE_CFLAGS_X86_64 += -mmacosx-version-min=10.7
```
peabody124: We don't really like commented code. If this isn't needed please trash these lines.

ground/gcs/gcs.pri
mlyle: Yes.  This ends support for Mac OS X 10.7 and 10.8.  While I don't like dropping support for something that shipped 3 years ago (10.8 shipped in July 25, 2012), it's been unsupported since late 2013.   Basically all hardware that ran 10.7/10.8 can run 10.9 and 10.10.

There's a big installed base stuck on 10.6 because of hardware compatibility, but 10.7 and 10.8 are much more rare and falling much more quickly: https://stats.wikimedia.org/archive/squid_reports/2015-06/SquidReportOperatingSystems.htm
(down to .36+.36% from .39+.40% in the past month).

mlyle:
Done

ground/gcs/gcs.pri
peabody124: What is the upside of this then? Sounds like it's not common but I'd still like to have a benefit if we break the software for 1% of mac users.

mlyle: It lets us use modern language features.  C++11 is a pretty big win.  And the proportion of users on older releases continues to fall.  Note that the QT guys have already dropped official support for 10.7 in 5.5 and have indicated 10.8 will probably go soon.

constexpr and initializer lists let us do a lot more work at compile time instead of startup and runtime.  It also makes a lot of code more compact (removing millions of .append(constant) lines, for instance).  It improves compatibility with C code.  There's an auto keyword that can infer return types so that you can handle several kinds of lists with the same code.  And it catches more bugs at compile time.

I'm working pretty deeply on the uavobjectgenerator,

peabody124: that's what i was trying to established, it was required for C++11. I'm just somewhat surprised since I'd think the language you compile in wouldn't change the binaries that are output and where they can run. Perhaps it has requires some libraries that are otherwise not supported.

mlyle: Well, one library in particular-- the C++ standard library has changed between C++03 and C++11.

10.8 and before don't ship with a fully C++11 compliant standard library.

mlyle: (So it breaks the build if you don't have this, because xcode-utils dynamic links against the version of the libraries that matches your deployment target).

mlyle:
Please just let me know which way to go-- to proceed with uavogen work assuming C++11, or to assume this can't be committed for legacy OSX support and that I should go the other way.

peabody124:
It sounds like the pros outweigh losing 10.8/7 support but it would be nice to give it a day for anyone to speak up.

Is it possible to compile for 10.7 from 10.7 with the code changes?

mlyle:
I don't think so, but I don't have a 10.7 rig to try.  Basically, QT when compiling in C+0x/C+11 mode expects to have all the C+11 std lib.

mlyle:
Cool, thank you!

PR #1739, opened 2015-07-23
Make the GCS Compile with MSVC

guilhermito:
This seems more stable and faster on windows.
Also added a feature to download the mesa drivers if needed. This feature is only available when compiled with MSVC since it crashes just by checking OpenGL flags with mingw.


mlyle:
Hi-- I have two questions before I go into detailed review.

First, can we separate out the mesa downloader from this?  The new OpenGL dynamic selection algorithm in 5.4 (which now applies to mingw builds as of 5.5) https://blog.qt.io/blog/2014/11/27/qt-weekly-21-dynamic-opengl-implementation-loading-in-qt-5-4/ allows us to automatically switch to a Mesa implementation if ANGLE and OpenGL32 are missing.  All we need to do is move a couple blocks of our code to use the QT wrappers of OpenGL functions.  This is on my to-do list.  Anyways, I'm open to all approaches but I'd rather consider that separately from the Visual Studio changes.

Second, is anything necessary to build with C++11?  I'm going to be doing some work that really benefits from C++11 and so I'd hate to regress MSVC builds because they don't use C++11.  ( #1738 ).

mlyle:
In particular, it looks like we have conflicting code on the math stuff in hitl / pathplanner.

guilhermito:
Dropped the mesa stuff. These are not Visual Studio related changes though, as I'm sure you know that is an IDE which can even use non MS compilers.
About the HITL conflicts I'm sure I can make the two PRs work together although I don't understand what your changes on that particular file have to do with C++0x, can you please explain (on your PR if you prefer).

ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.h
```diff
@@ -175,7 +176,7 @@ namespace mapcontrol
         */
         TLMapWidget(QWidget *parent=0,Configuration *config=new Configuration);
         ~TLMapWidget();
-
+        internals::PointLatLng pointkkkkkkkkkkkkkkkkk;
```
mlyle: Is this deliberate?

ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.h
```diff
@@ -45,6 +45,7 @@
 #include "mapcircle.h"
 #include "waypointcurve.h"
 #include "waypointitem.h"
+#include "../core/corecommon.h"
```
mlyle: I don't like this-- perhaps a better place can be found for this header?  (I'd prefer to not use relative include paths like this that presume a certain repository layout, and it doesn't really seem to be specific to the core plugin).

ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.h
```diff
@@ -388,6 +389,7 @@ namespace mapcontrol
         void setSelectedWP(QList<WayPointItem *> list);
 
         void setWindVelocity(double windVelocity_NED[3]);
+        internals::PointLatLng dummy() {return internals::PointLatLng();}
```
mlyle: I assume this is to suppress a warning?

ground/uavobjects/uavobjects.pro
```diff
@@ -38,9 +38,9 @@ win32|unix {
 win32 {
     # Windows sometimes remembers working directory changed from Makefile, sometimes not.
     # That's why pushd/popd is used here - to make sure that we know current directory.
-
+!exists(../../build/uavobject-synthetics ) {
```
mlyle: I suggest mkdir -p which doesn't care if it already exists.

mlyle:
Does not build on Mac OS X:

In file included from /Users/mlyle/taulabs/ground/gcs/src/libs/glc_lib/3rdparty/quazip/quazip.cpp:35:
In file included from /Users/mlyle/taulabs/ground/gcs/src/libs/glc_lib/3rdparty/quazip/quazip.h:39:
/Users/mlyle/taulabs/ground/gcs/src/libs/glc_lib/3rdparty/quazip/zip.h:54:10: fatal error: 
      'QtZlib/zlib.h' file not found
#include <QtZlib/zlib.h>

ground/gcs/src/plugins/uploader/bl_messages.h
```diff
@@ -85,6 +85,103 @@ enum bl_commands {
  * Note: These enum values MUST NOT be changed or backward
  *       compatibility will be broken
  */
+#ifdef _MSC_VER
```
mlyle: We need to figure out how to get rid of all this duplication.  Perhaps factoring out the union declarations to be outside the struct would help.

ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.h
guilhermito: nope probably fell asleep on the keyboard

ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.h
guilhermito: This refers to the core part of the library so it is independent of the repo layout. I can add this path to the INCLUDEPATH var if you prefer.

ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.h
guilhermito: Nope, it was to make sure the dynamic library exported the functions it imported from the static lib. It is a common workaround. But since I dropped the static/dynamic arrangement and made it into a single dynamic lib this is not needed anymore.

ground/uavobjects/uavobjects.pro
guilhermito: Sure

ground/gcs/src/plugins/uploader/bl_messages.h
mlyle: @guilhermito If it would help, I could take a pass at de-duplicating this and send you a pull request to your branch.

guilhermito: I would appreciate it, thanks

mlyle: It's not tested but I sent you a pull request already.

mlyle:
Thanks for this change!  I'll give some thought to continuous integration so we hopefully don't inadvertently lose the ability to build on MSVC in the future.

PR #1741, opened 2015-07-23
Produce and process code coverage data from the GCS

guilhermito:
Also created a small script https://gist.github.com/guilhermito/842e0dc215fb3a64a8f5 to make it simpler to use.
Code coverage is usually used with unit tests but it is also good to find stale unused code.


mlyle:
Nice!  Find any potentially dead large sections of code / files?  Files that aren't executed at all or only initialization sections are of particular interest to me in GCS.

PR #1742, opened 2015-07-24
Flight: FrSky Sport Telem modifications to match FrSky Sensor Hub

jihlein:
Modifications to FrSky Sport Telem to match FrSky Sensor Hub rpm, t1, and t2 values.  This module has been tested to compile, link, and run, but I can't verify operation due to lack of FrSky Sport Telem compatible receivers,

### DIFF
--- a/flight/Modules/UAVOFrSKYSPortBridge/UAVOFrSKYSPortBridge.c
+++ b/flight/Modules/UAVOFrSKYSPortBridge/UAVOFrSKYSPortBridge.c
@@ -287,73 +287,86 @@ static bool frsky_encode_cells(uint32_t *value, bool test_presence_only, uint32_
 }
 
 /**
- * Encode temperature of barosensor as T1
+ * Encode GPS status as T1 value
+ * Right-most two digits means visible satellite count, left-most digit has following meaning:
+ * 1 - no GPS connected
+ * 2 - no fix
+ * 3 - 2D fix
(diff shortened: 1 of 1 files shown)


xdandys:
This looks good. Give me day or two and i will test it. 

This also leads me to think about more meaningful usage of other telemetry fields. For example AccX, AccY, AccZ. It's currently mapped to raw accelerometer outputs. It actually have very limited information value when you consider how noisy accelerometers are and how low sample rate we can achieve with frsky telemetry. What about map them to attitude roll, pitch, yaw angles instead? It would for example allow us to have some fancy custom lua telemetry screen with artificial horizon on Taranis.

peabody124:
@xdandys there already is an NEDAccel object, although we might have stopped using it. It is the LPF version of acceleration in a north/east/down frame.

mlyle:
I don't think he means nedaccel, he means to take attitudeactual.

xdandys:
Ok, I've just tested this and it's working as expected.

mlyle:
Merging based on positive test feedback and that it's same as other code.

lolorc:
@xdandys something like this ? https://github.com/TauLabs/TauLabs/compare/next...lolorc:frsky_sport-encode_angle
I've never PR'ed it because I wanted it to be runtime configurable, and I never spent the time adding a new object...

PR #1743, opened 2015-07-24
sparky2: enable frsky smartport telemetry

xdandys:
This enables SmartPort telemetry on sparky2.


xdandys:
Holzleim tested this (http://forum.taulabs.org/viewtopic.php?t=223&p=6423#p6423) and it seems to work.

PR #1745, opened 2015-07-24
remove REVOLUTION define from all modules / targets

mlyle:


peabody124:
Removing outdated inappropriately named variable: nice

Can you confirm that simulation still flies and that you can get it to not have an error in configuration.

mlyle:
@peabody124 made me nervous there! I thought you'd seen something wrong.

I just built my killrevolutiondefine branch and it flies/navigates fine, at least for minimal cases.

peabody124:
Just wanted to confirm it since I remembered those flags having a hook for simulation at some point. Thanks for doing so!

mlyle:
Of course, thanks for the merge!

PR #1746, opened 2015-07-24
simulation/posix: port to architectures besides x86.

mlyle:
This is work towards Raspberry Pi2 target support.  Verified that simulation works / flies nav successfully on RPi2.


mlyle:
Cool! TYVM.

PR #1749, opened 2015-07-25
Fix german language support

LinuxJohannes:
My systen 64bit Opensuse generate
not the language extension de. It is de_DE.

### DIFF
--- a/ground/gcs/share/taulabs/translations/translations.pro
+++ b/ground/gcs/share/taulabs/translations/translations.pro
@@ -1,6 +1,6 @@
 include(../../../gcs.pri)
 
-LANGUAGES = de es fr ru zh_CN
+LANGUAGES = de de_DE es fr ru zh_CN
 
 # var, prepend, append
 defineReplace(prependAll) {
--- a/ground/gcs/share/taulabs/translations/translations.pro
+++ b/ground/gcs/share/taulabs/translations/translations.pro
@@ -1,6 +1,6 @@
 include(../../../gcs.pri)
 
-LANGUAGES = de de_DE es fr ru zh_CN
+LANGUAGES = de es fr ru zh_CN
 
 # var, prepend, append
 defineReplace(prependAll) {


mlyle:
Does anyone else understand how the translation infrastructure works?  It seems like it would be unfortunate to have to enumerate all country combinations (de_AT, de_BE, de_CH, ...) here, at least before we have different translations for them.

LinuxJohannes:
Yes,
have more investigate it today. Its a enumeration in ground/gcs/src/libs/tlmapcontrol/src/core/languagetype.h and ground/gcs/src/libs/tlmapcontrol/src/core/languagetype.cpp.

ground/gcs/src/app/main.cpp calls the qttranslater with this country-string, here called as locale:

main line 262
QString locale = QLocale::system().name();
//returns "de_DE" on my system

will be overwrited possible
main line 271
locale = settings.value("General/OverrideLanguage", locale).toString();

and the language will be loaded possible, if aviable:
   if (translator.load(QLatin1String("taulabs_") + locale, creatorTrPath)) {

The translations are in ground/gcs/share/taulabs/translations/

My vision is to make a change in main.cpp co cut locale to 2 char:
main line 262
QString locale = QLocale::system().name();
//returns "de_DE" on my system
// cut it to "de"

mlyle:
What would be best is if we could try de_DE first and then fall back to de, so we can have country specific translations in the future.

LinuxJohannes:
OK, fine with me. Is it fine for you

@@ -259,6 +259,8 @@ int main(int argc, char **argv)
     SharedTools::QtSingleApplication app((QLatin1String(appNameC)), argc, argv);
 
     QString locale = QLocale::system().name();
+    // patch for german Linux systems, cutting de_DE to de
+    if(leftSide == "de") locale = locale.left(2);
 
     // Must be done before any QSettings class is created
     QSettings::setPath(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,

LinuxJohannes:
Sorry, forgot a     QString leftSide; to declare the string before.

LinuxJohannes:
tested and solved with  Fix german language support #1756 
have deleted the first change here in this patch, but to add a new patch not possible ??

mlyle:
Closed to discuss further in #1756

PR #1750, opened 2015-07-26
FIX REGRESSION LAST COMMIT: gps: always init objects except on small tgts

mlyle:


mlyle:
Resulted in restart loop on Colibri when GPS module not enabled.  Only found after resetting completely because of dependency.  Was my bad / an obvious code mistake.

PR #1756, opened 2015-07-29
Fix german language support

LinuxJohannes:
changed this file for fix #1749
How to add this change to #1749 ??

### DIFF
--- a/ground/gcs/src/app/main.cpp
+++ b/ground/gcs/src/app/main.cpp
@@ -260,6 +260,9 @@ int main(int argc, char **argv)
 
     QString locale = QLocale::system().name();
 
+    QString leftSide;
+    if(leftSide == "de") locale = locale.left(2);
+
     // Must be done before any QSettings class is created
     QSettings::setPath(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
             QCoreApplication::applicationDirPath()+QLatin1String(SHARE_PATH));


mlyle:
Is there any chance you could write this to do the following:

* try to load the locale file as specified
* if this fails, trim everything after the underscore, and try and load that ?

mlyle:
If not, I can add it to my to-do list and get something together for you to test.

mlyle:
@LinuxJohannes Can you please try mlyle@SHA: 1d4379a5d39562d20df72b2c2ae61c90363785e4 to see if it works for you?  I have not tested this, but it seems like the correct fix.

mlyle:
@LinuxJohannes sorry, I meant mlyle@SHA: 9acc961db34c43e4edcb87e13c7500f74e4bc56a.

LinuxJohannes:
Sorry, it doesn't work.

setupTranslatorForLocale will be called with "de_DE"  and returns with TRUE, but no german translation is loaded

LinuxJohannes:
Have added debug code:
```
static bool setupTranslatorForLocale(QTranslator &trans, const QString& prefix,
        const QString& locale) {
	qDebug()<<"1.setupTranslatorForLocale";
	qDebug()<<locale;
    if (loadTranslation(trans, prefix, locale)) {
	  	qDebug()<<"2.setupTranslatorForLocale return true";
        return true;
    }
	qDebug()<<"3.setupTranslatorForLocale try to split";
    QStringList splitLocale = locale.split("_");
	qDebug()<<splitLocale;
    // transform DE_de to DE, and try again.
    if (splitLocale.length() > 1) {
	  qDebug()<<splitLocale[0];
        if (loadTranslation(trans, prefix, splitLocale[0])) {
            return true;
        }
    }
}

static void setupTranslations(const QString& locale) {
    QTranslator translator;
	qDebug()<<"1.setupTranslations";
	qDebug()<<locale;
    if (setupTranslatorForLocale(translator, QLatin1String("taulabs_"),
                locale)) {
        QTranslator qtTranslator;

        if (setupTranslatorForLocale(qtTranslator, QLatin1String("qt_"),
                    locale)) {
            QCoreApplication::installTranslator(&translator);
            QCoreApplication::installTranslator(&qtTranslator);
        }
    else qDebug()<<"2. setupTranslations else !!"; 
    }
}
```
Thats the output:
./taulabsgcs
1.setupTranslations
"de"
1.setupTranslatorForLocale
"de"
2.setupTranslatorForLocale return true
1.setupTranslatorForLocale
"de"
2.setupTranslatorForLocale return true
"fusion"
Starting the Udev client

mlyle:
Hm, that's unfortunate.

It looks like the translator tries to do the correct thing to load an appropriate file, too, from the documentation:

http://doc.qt.io/qt-5.5/qtranslator.html#load

LinuxJohannes:
Yes, see it also.  Very interesting : can select "System language" or "German" in Options of GCS. German returns "de", System language returns "de_DE"

LinuxJohannes:
Hi. You called this function with a wrong translator 
```
+static bool loadTranslation(QTranslator &trans, const QString& prefix,
+        const QString& locale) {
+    const QString &fileName = prefix + locale;
+
+    const QString &creatorTrPath = QCoreApplication::applicationDirPath()
+                                   + QLatin1String(SHARE_PATH "/translations");
+    const QString &qtTrPath = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
+
```
qttranslator have to be only for the QT-Path and translator only for the gcs translation
```
+    if (trans.load(fileName, qtTrPath) || trans.load(fileName, creatorTrPath)) {
+        return true;
+    } 
+
+    return false;
+}
```

mlyle:
@LinuxJohannes I'm not sure I follow.  I know the QT translations won't be found in the application dir path, but the if statement will try and load in the other place.

I think I see it looking at the code actually.

mlyle:
I repushed the branch and confirmed it works with things in the preferences.  I had a bug with scoping of the translators, so that's fixed now.

Can you please try again?

mlyle:
I also confirmed this works on Mac OS X: 

LC_ALL=de_DE ./build/ground/gcs/bin/Tau\ Labs\ GCS.app/Contents/MacOS/Tau\ Labs\ GCS

mlyle:
Actually I think we need a clear problem report.  GCS without any code change opens correctly in German when invoked with ```LC_ALL=de_DE ./build/ground/gcs/bin/Tau\ Labs\ GCS.app/Contents/MacOS/Tau\ Labs\ GCS ```

This also seems to work on Linux.

Can you please open an issue with what operating system you are using and the version of the GCS you're seeing this on?  If you're on a Linux, a copy of your environment obtained by running env at the command line could be helpful.

Issue #1730, opened 2015-07-18
[WIP] Flight: adding gyro sub-samplng functionality


Trex4Git:
I have read somewhere on the forum, that it is preferred to disable the DLPF filter in the MPU an use the software IIR LPF filter (https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Stabilization/stabilization.c#L296-L298) instead.

But with the MPU9250 this is not possible, as the sample rate divider in the MPU only works when the DLPF is used (than the base frequency is 1kHz). When the  DLPF is disabled the output sample rate is 8Khz, what generates a high CPU load.
See https://github.com/TauLabs/TauLabs/pull/1726 and https://github.com/TauLabs/TauLabs/issues/1724

I have already thought about, and also read on IRC about, to run the MPU with no DLPF at 8kHz, but don't trigger our control loop with every interrupt, but do a sub-sampling on the MPU data, and run the control loop only every n interrupt.

That's why i already did a simple implementation: https://github.com/Trex4Git/TauLabs/commit/a88e63f64d66bd51c478c8ffabba813bae71dc4c?diff=split
The diff view seems to be a little bit strange. What I have done is that in the stabilization task directly after new gyro data is recognized, the data is read and the already existing gyro IIR lLPF filter is run,  and a counter ("gyroSubsample") is incremented. This data reading and filtering is run with the full MPU rate. 
But the rest of the stabilization task is run only every "GyroSubSampling" (a new UAV Object).

To test it, I set the outputs to SyncPWM (to check the control rate frequency with a logic analyzer). 
Reducing the control rate with a fixed MPU rate seems to work as expected, but the CPU load is higher then hoped. 

Tested with a Quanton. It has the MPU6000, but i don't have a FC with MPU9250, an as the MPU6000 also works with a sample rate divider in the MPU,no DLPF and 8Khz base frequency, I think its OK to do the test with it.

MPU rate | sub-sampling | control rate | CPU load
------------ | ------------- | ------------- | -------------
2 kHz | 1 | 2 kHz | 52%
2 kHz | 2 | 1 kHz | 46%
2 kHz | 3 | 666,67 Hz | 44%
2 kHz | 4 | 500 Hz | 43%
2 kHz | 5 | 400 Hz | 42.5%
2 kHz | 6 | 333,33 Hz | 42%
1 kHz | 1 | 1 kHz | 28%
500 Hz | 1 | 500 Hz | 16%

[...]

PR #1758, opened 2015-07-30
flight/PIOS: SPI speed enhancements for MPU6000 & MPU9250_SPI

Trex4Git:
Enhancement for the SPI clock to speed up data transfer after the configuration is done.
The data register in MPU can be read up to 20MHz +-10%.
For F4 µC based FC (e.g. Quanton, Sparky 2) this should result in 21MHz SPI clock, for F1 based µC it should result in 18MHz.

First tests on Quanton were successful.
CC3D and Sparky 2 need testing.

This will also be included when the new sub-samlping / filter functionality is done and might get merged sometime: https://github.com/TauLabs/TauLabs/issues/1730

But it maybe helps also with other problems, and should reduce CPU load in general. So i extracted the enhancement from the other work (https://github.com/Trex4Git/TauLabs/commit/0a46d5748eec4fa38b504d7efc5420f2a525d277), to have a small PR with this enhancement already available.

The enhancement on CPU load can be seen in the table there: https://github.com/TauLabs/TauLabs/issues/1730#issuecomment-123429613

When there is no interest yet, i can close this PR.

### DIFF
--- a/flight/PiOS/Common/pios_mpu6000.c
+++ b/flight/PiOS/Common/pios_mpu6000.c
@@ -44,6 +44,13 @@
 #define MPU6000_TASK_PRIORITY	PIOS_THREAD_PRIO_HIGHEST
 #define MPU6000_TASK_STACK		484
 
+#ifdef PIOS_MPU6000_SPI_HIGH_SPEED
+#define MPU6000_SPI_HIGH_SPEED              PIOS_MPU6000_SPI_HIGH_SPEED
+#else
+#define MPU6000_SPI_HIGH_SPEED              21100000	// slightly higher then 21MHz so that it is really 21MHz after prescaler calculations
+#endif
+#define MPU6000_SPI_LOW_SPEED               1000000
+
 /* Global Variables */
 
 enum pios_mpu6000_dev_magic {
@@ -76,8 +83,8 @@ static struct mpu6000_dev *pios_mpu6000_dev;
 static struct mpu6000_dev *PIOS_MPU6000_alloc(void);
 static int32_t PIOS_MPU6000_Validate(struct mpu6000_dev *dev);
 static void PIOS_MPU6000_Config(const struct pios_mpu60x0_cfg *cfg);
-static int32_t PIOS_MPU6000_ClaimBus();
-static int32_t PIOS_MPU6000_ReleaseBus();
+static int32_t PIOS_MPU6000_ClaimBus(bool lowspeed);
+static int32_t PIOS_MPU6000_ReleaseBus(bool lowspeed);
 static int32_t PIOS_MPU6000_SetReg(uint8_t address, uint8_t buffer);
 static int32_t PIOS_MPU6000_GetReg(uint8_t address);
 static void PIOS_MPU6000_Task(void *parameters);
@@ -157,9 +164,9 @@ int32_t PIOS_MPU6000_Init(uint32_t spi_id, uint32_t slave_num, const struct pios
 	pios_mpu6000_dev->cfg = cfg;
 
 	/* Configure the MPU6000 Sensor */
-	PIOS_SPI_SetClockSpeed(pios_mpu6000_dev->spi_id, 100000);
+	PIOS_SPI_SetClockSpeed(pios_mpu6000_dev->spi_id, MPU6000_SPI_LOW_SPEED);
 	PIOS_MPU6000_Config(cfg);
-	PIOS_SPI_SetClockSpeed(pios_mpu6000_dev->spi_id, 3000000);
+	PIOS_SPI_SetClockSpeed(pios_mpu6000_dev->spi_id, MPU6000_SPI_HIGH_SPEED);
 
 	pios_mpu6000_dev->threadp = PIOS_Thread_Create(
 			PIOS_MPU6000_Task, "pios_mpu6000", MPU6000_TASK_STACK, NULL, MPU6000_TASK_PRIORITY);
@@ -230,9 +237,9 @@ static void PIOS_MPU6000_Config(const struct pios_mpu60x0_cfg *cfg)
 	 * on all different targets.
 	 */
 
-	PIOS_MPU6000_ClaimBus();
+	PIOS_MPU6000_ClaimBus(true);
 	PIOS_DELAY_WaitmS(1);
-	PIOS_MPU6000_ReleaseBus();
+	PIOS_MPU6000_ReleaseBus(true);
 	PIOS_DELAY_WaitmS(10);
 
 	// Reset chip
@@ -369,9 +376,10 @@ void PIOS_MPU6000_SetLPF(enum pios_mpu60x0_filter filter)
 
 /**
  * @brief Claim the SPI bus for the accel communications and select this chip
+ * \param[in] flag controls if low speed access for control registers should be used
  * @return 0 if successful, -1 for invalid device, -2 if unable to claim bus
  */
-static int32_t PIOS_MPU6000_ClaimBus()
+static int32_t PIOS_MPU6000_ClaimBus(bool lowspeed)
 {
 	if (PIOS_MPU6000_Validate(pios_mpu6000_dev) != 0)
 		return -1;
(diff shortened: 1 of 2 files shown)


flight/PiOS/Common/pios_mpu9250_spi.c
```diff
@@ -406,6 +406,9 @@ int32_t PIOS_MPU9250_SPI_Init(uint32_t spi_id, uint32_t slave_num, const struct
 	/* Configure the MPU9250 Sensor */
 	if (PIOS_MPU9250_Config(cfg) != 0)
 		return -2;
+		
+	/* Configure a higher SPI clock speed for faster data transfer, after configuration is done */	
+	PIOS_SPI_SetClockSpeed(dev->spi_id, 21100000);
```
xdandys: There is already MPU9250_SPI_HIGH_SPEED constant, why not modify it instead?

Trex4Git: Thank you very much for the feedback.

I did only work on the MPU6000 driver, than i saw the issue with Sparky2 and thought i could be easy ported to MPU9250 and make a quick edit on github.

But the pios driver for MPU9250 is already more advanced, and at my quick look at it I haven't noticed that.

Will change this when I am back from work.

Trex4Git:
Changed Sparky2 commit to already existing functionalityas mentioned by xdandys, thanks for that.
Also ported some of the SPI clock changing functionality from the MPU9250_SPI driver, to the MPU6000 driver.

The speed improvements on SPI transfers should be:
Sparky 2: 10.5MHz -> 21MHz (factor 2)
Quanton: 2.625MHz -> 21MHz (factor 8)
CC3D: 2.25MHz -> 18MHz (factor 8)

mlyle:
I don't think we should run over the 20MHz specified clock.  I understand there is margin, but looking at the data sheet for the MPU9250 section 3.6.1, SCLK clock frequency is specified as 0.9MHz minimum, 20MHz maximum.   If anything, we should be a few megahertz under the specified clock to add margin for board signal integrity issues and temperature/voltage variation.  (The value there in the datasheet is noted to be based on characterization of 5 parts, not guaranteed by production test.)

It sounds like it'd be a big win to get Quanton to 10.5MHz and CC3D to 18MHz if it otherwise works, though.

mlyle:
Also note to run it at full clock (20MHz), we'd need to carefully read the datasheets and evaluate setup and hold times and trace lengths.  If we were positive that was safe, it might be worth clocking down to 160MHz to get a 20MHz clock-- it might improve performance overall.

On the other hand, I like margin.  The 18MHz for CC3D is already getting very close to what I'm comfortable with.

jihlein:
FWIW, I've used the MPU6000 at 21 MHz SPI clock reads for the data registers on a different STM32F4 project with no issues what so ever.

mlyle:
I bet you could run it at 30MHz and it would usually work, too.  But it's outside of specification and not a good idea, especially for something critical to keeping an aircraft in the air.

jihlein:
It's within the +/- 10% spec, but I certainly can't argue that a lower clock speed gives more margin against parts that are just within spec.

Trex4Git:
Also OP seems to be using it since a long time without problems.

mlyle:
Yah, I see the +/- 10% in the other table, but it's not in the table I referenced, which just says 20MHz max.

Absent a statement from InvenSense that the part can be safely operated over 20MHz across the range of temperature and voltage, I can't merge this PR as is.  I'm sure it usually works.

Maybe other people from core will have a different opinion, though.

Trex4Git:
I looked at the MPU datasheet on page 15/54 section 6.4.
There the sensor data register are specified as typical (not MAX) 20 MHz +-10%, whereas the I2C is specified with 400 kHz MAX in fast mode.

mlyle:
spitiming

Trex4Git:
I have looked at that table after you posted it earlier. Before i had looked closer to the MPU6000 datasheet, and just verified that I can find the same specification in the MPU9250 datasheet.

I can not argue against the values that are in table 8,but i am not sure what is the intention of this table and how much margin we need (e.g. I2C is also often used at 400kHz and maybe makes more problems then SPI at 21MHz).
For MPU6000 there is not such a table, only the timing specification  for 1MHz like in 3.6 (table 7) for the MPU9250. For MPU6000 there are no additional timing specification for above 1MHz operation.

I think  table 8 should update some timing specification for higher SPI clocks then 1MHz. Therefore the Min frequency for this timings is specified as 0.9MHz, and overlapping with the 1MHz definitions in 3.6.
But I think the Max 20MHz in 3.6.1 doesn't say that 20 MHz is the maximum SPI clock that can be used (because than 900Hz would be the minimum allowed clock), but says that these timing specification are meant for this frequency range.
Earlier in the datasheet typical (not Max) 20MHz <sup>+</sup><sub>-</sub>10% were specified, and for MPU6000 there are no specification above 1MHz, but taulabs is using currently 2.625MHz on the Quanton.
I think its a little bit strange, that they defined it here not for the specification they made earlier, but i don`t know how critical that is. 

I hope i have at the weekend some time to do some CPU load comparisons.

But it would be nice to get more feedback from other people that have worked with the MPU.

lilvinz:
>Also note to run it at full clock (20MHz), we'd need to carefully read the datasheets and evaluate setup and hold times and trace lengths. If we were positive that was safe, it might be worth clocking down to 160MHz to get a 20MHz clock-- it might improve performance overall.

Running the CPU at 160 MHz will probably not work as a 48 MHz clock is required for USB.

Trex4Git:
I did some tests with the SPI clock reduced to 10.5MHz:

MPU rate | IRQ sub-sampling (Drop)  | Task sub-sampling (LPF) | control rate | CPU load 10.5MHz  |  CPU load 21MHz
------------ | ------------ | ------------- | ------------- | ------------- | -------------
8 kHz | 1 | 4 | 2 kHz | 58 | 55%
8 kHz | 1 | 8 | 1 kHz | 38 | 36%
8 kHz | 1 | 16 | 500 Hz | 33 | 29%
8 kHz | 1 | 5 | 1.6 kHz | 50 | 

The higher CPU load is not as large as i have thought.
I still think that 21 MHz is in the specifications of Invense (only the SPI timing are not explicit defined for it; but MPU6000 does not have any timing specification for spi clocks above 1 MHz ), but if a large margin is  really needed i will change that.

@lilvinz As you are the designer of the Quanton, do you have any experiences  regarding the SPI clock?
(Offtopic: has production&selling of the quanton stopped?)

flight/PiOS/Common/pios_mpu6000.c
```diff
@@ -379,21 +387,28 @@ static int32_t PIOS_MPU6000_ClaimBus()
 	if (PIOS_SPI_ClaimBus(pios_mpu6000_dev->spi_id) != 0)
 		return -2;
 
+	if (lowspeed)
+			PIOS_SPI_SetClockSpeed(pios_mpu6000_dev->spi_id, MPU6000_SPI_LOW_SPEED);
```
mlyle: Indentation

flight/PiOS/Common/pios_mpu6000.c
```diff
@@ -379,21 +387,28 @@ static int32_t PIOS_MPU6000_ClaimBus()
...
 	return 0;
 }
 
 /**
  * @brief Release the SPI bus for the accel communications and end the transaction
+ * \param[in] must be true when bus was claimed in lowspeed mode
  * @return 0 if successful
  */
-static int32_t PIOS_MPU6000_ReleaseBus()
+static int32_t PIOS_MPU6000_ReleaseBus(bool lowspeed)
```
mlyle: Alternatively we could remember the speed in a global and restore it.  But I'm fine with this way too.

flight/PiOS/Common/pios_mpu6000.c
```diff
@@ -44,6 +44,13 @@
 #define MPU6000_TASK_PRIORITY	PIOS_THREAD_PRIO_HIGHEST
 #define MPU6000_TASK_STACK		484
 
+#ifdef PIOS_MPU6000_SPI_HIGH_SPEED
+#define MPU6000_SPI_HIGH_SPEED              PIOS_MPU6000_SPI_HIGH_SPEED
+#else
+#define MPU6000_SPI_HIGH_SPEED              21100000	// slightly higher then 21MHz so that it is really 21MHz after prescaler calculations
```
mlyle: If we change these to 20000000 I'd be glad to merge this.

mlyle:
Cool, thanks!

Trex4Git:
Thank you too.

Maybe I will open an issue for further discussion, but it's nice to have some improvements already.

PR #1759, opened 2015-07-30
HAL refactor phase 1

mlyle:
This branch is intended to start the process of removing a lot of code duplication from targets.

First, it adjusts UAV Object generation of enums to allow "parents".  Child enums are guaranteed to have the same constant values as their parent for all of their values.  (Child enum values must be a strict subset of their parent's values).

Next, it creates a source file called pios_hal where initialization of ports (USBVCP, USBHAL, hardware USART/I2C, and radios) is handled.  This is not maximally clean yet but it is an important step on the way.

Finally, it moves three targets to this code: pipxtreme, sparky2, and revo mini.  These were selected because they're the three with radios, reasonably different form each other, and it enables upcoming frequency configuration work on the radio code.


mlyle:
Remaining punch list: 

* object/hardware id/handles
  * Consistent location of them
  * Conditionalize their existance
* Make compile-time errors for combinations that don't make sense, e.g. USB & no COM, RFM & no COM... and remove the inner conditionals
* Catch errors where we have a conflict / multiple things setting an id, rcvr, etc.  At least work towards this
* Minimal DSM cleanup
  * only allow one DSM receiver to be consistent with other ones
  * reuse other configuration info to have less things passed into port stuff
* Code style / capitalization
* Port at least a couple other targets to make sure this is consistent/serviceable for them
* Python inherited UAVO support
* Android inherited UAVO support

BrainFPV:
Nice work; I like the simplification the shared UAVObject definition brings. I just glanced over the code and so far it looks good to me.

mlyle:
Made some corrections and adapted revomini.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -808,14 +808,8 @@ static bool updateRcvrActivityCompare(uintptr_t rcvr_id, struct rcvr_activity_fs
...
-			case MANUALCONTROLSETTINGS_CHANNELGROUPS_DSMMAINPORT:
-				group = RECEIVERACTIVITY_ACTIVEGROUP_DSMMAINPORT;
-				break;
-			case MANUALCONTROLSETTINGS_CHANNELGROUPS_DSMFLEXIPORT:
-				group = RECEIVERACTIVITY_ACTIVEGROUP_DSMFLEXIPORT;
-				break;
-			case MANUALCONTROLSETTINGS_CHANNELGROUPS_DSMRCVRPORT:
-				group = RECEIVERACTIVITY_ACTIVEGROUP_DSMRCVRPORT;
+			case MANUALCONTROLSETTINGS_CHANNELGROUPS_DSM:
+				group = RECEIVERACTIVITY_ACTIVEGROUP_DSM;
```
peabody124: This was the way it was before intentionally - it allowed supporting multiple DSM inputs. for example when you have one TX for the pilot and another being relayed to a gimbal.

I'm not sure anyone is using that feature so it is probably ok to drop for now (although please anyone using it feel free to mention so). We should have a plan with the refactor to generalize having two DSM though and just calling it RECEIVERACTIVITY_ACTIVEGROUP_DSM1 and RECEIVERACTIVITY_ACTIVEGROUP_DSM2 and the HAL can take care of ordering. Sound reasonable?

flight/Modules/RadioComBridge/RadioComBridge.c
```diff
@@ -379,10 +378,10 @@ static void radioRxTask( __attribute__ ((unused))
 #ifdef PIOS_INCLUDE_WDG
 		PIOS_WDG_UpdateFlag(PIOS_WDG_RADIORX);
 #endif
-		if (PIOS_COM_RADIO) {
+		if (PIOS_COM_RFM22B) {
```
peabody124: To be honest, I would love to see all these magic constants for port IDs vanish and be passed as variable to the init method. Might be appropriate for this refactoring.

flight/PiOS/Common/pios_dsm.c
```diff
@@ -107,7 +107,11 @@ static void PIOS_DSM_Bind(struct pios_dsm_dev *dsm_dev, uint8_t bind)
 	const struct pios_dsm_cfg *cfg = dsm_dev->cfg;
 
 	GPIO_InitTypeDef GPIO_InitStructure = cfg->bind.init;
-	GPIO_InitStructure.GPIO_PuPd  = GPIO_PuPd_UP;
+#ifdef SMALLF1
```
peabody124: thank you so much for merging these :)

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
+#include <pios.h>
```
peabody124: file needs header / doxygen and a brief explanation what it does would be great

peabody124: it also looks like this only applies to serial ports - it might be worth adding that to the name space to make it specific. as opposed to ADC ports, CAN ports, etc.

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+				uintptr_t pios_ppm_rcvr_id;
+				if (PIOS_RCVR_Init(&pios_ppm_rcvr_id, &pios_ppm_rcvr_driver, pios_ppm_id)) {
+					PIOS_Assert(0);
+				}
+				pios_rcvr_group_map[MANUALCONTROLSETTINGS_CHANNELGROUPS_PPM] = pios_ppm_rcvr_id;
+			}
+#endif  /* PIOS_INCLUDE_PPM */
+			break;
+
+		case HWSHARED_PORTTYPES_DISABLED:
```
peabody124: trivial aesthetics - can disabled go at top :)

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+			break;
+		case HWSHARED_PORTTYPES_TELEMETRY:
+			PIOS_HAL_configure_com(usart_port_cfg, PIOS_COM_TELEM_RF_RX_BUF_LEN, PIOS_COM_TELEM_RF_TX_BUF_LEN, com_driver, &port_driver_id);
+			// Name of this should change XXX
+			target = &pios_com_telem_rf_id;
```
peabody124: I'm still not loving passing around global variables, but of course this isn't something you've worsened.

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+			break;
+		case HWSHARED_PORTTYPES_SBUS:
+#if defined(PIOS_INCLUDE_SBUS) && defined(PIOS_INCLUDE_USART)
+			if (sbus_cfg && sbus_rcvr_cfg) {
+				uintptr_t usart_sbus_id;
+				if (PIOS_USART_Init(&usart_sbus_id, sbus_rcvr_cfg)) {
+					PIOS_Assert(0);
+				}
+				uintptr_t sbus_id;
+				if (PIOS_SBus_Init(&sbus_id, sbus_cfg, com_driver, usart_sbus_id)) {
```
peabody124: and ideally one day SBus init can tell if there is a hardware inverter available and fall back on a soft usart implementation? this currently allows quite a few invalid configurations.

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+#endif /* PIOS_INCLUDE_PICOC */
+			break;
+        } /* port_type */
+
+        if (port_type != HWSHARED_PORTTYPES_SBUS && sbus_toggle) {
+                GPIO_Init(sbus_cfg->inv.gpio, (GPIO_InitTypeDef*)&sbus_cfg->inv.init);
+                GPIO_WriteBit(sbus_cfg->inv.gpio, sbus_cfg->inv.init.GPIO_Pin, sbus_cfg->gpio_inv_disable);
+        }
+
+	// TODO: catch configuration errors of duplicated channels here
```
peabody124: you mean if ```*target != 0``` ?

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+				PIOS_Assert(0);
+			}
+			pios_rcvr_group_map[MANUALCONTROLSETTINGS_CHANNELGROUPS_OPENLRS] = pios_rfm22brcvr_rcvr_id;
+		}
+#endif /* PIOS_INCLUDE_OPENLRS_RCVR */
+	} else if (radio_type == HWSHARED_RADIOPORT_DISABLED ||
+			max_power == HWSHARED_MAXRFPOWER_0) {
+		// When radio disabled, it is ok for init to fail. This allows
+		// boards without populating this component.
+		if (PIOS_RFM22B_Init(&pios_rfm22b_id, PIOS_RFM22_SPI_PORT, rfm22b_cfg->slave_num, rfm22b_cfg) == 0) {
```
peabody124: this is the bit that should be reworked for not reininting the modem when not in use. I need a different method that PIOS_RFM22B_Init. I won't fix it now since it will clash with your work.

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+#endif  /* PIOS_INCLUDE_USB_RCTX */
+			break;
+
+	}
+}
+#endif  /* PIOS_INCLUDE_USB_HID */
+
+
+#if defined(PIOS_INCLUDE_RFM22B)
+void PIOS_HAL_configure_RFM22B(HwSharedRadioPortOptions radio_type,
```
peabody124: doxygen headers please - applies to all methods

flight/targets/coptercontrol/fw/pios_board.c
```diff
@@ -47,7 +47,7 @@
 
 
 /* One slot per selectable receiver group.
- *  eg. PWM, PPM, GCS, DSMMAINPORT, DSMFLEXIPORT, SBUS
+ *  eg. PWM, PPM, GCS, DSM, DSM, SBUS
```
peabody124: find/replace typo

flight/targets/discoveryf4/fw/pios_board.c
```diff
@@ -44,7 +44,7 @@
 #include "modulesettings.h"
 
 /* One slot per selectable receiver group.
- *  eg. PWM, PPM, GCS, DSMMAINPORT, DSMFLEXIPORT, SBUS
+ *  eg. PWM, PPM, GCS, DSM, DSM, SBUS
```
peabody124: find/replace typo

flight/targets/naze32/fw/pios_board.c
```diff
@@ -213,7 +213,7 @@ static /*const*/ struct pios_mpu60x0_cfg pios_mpu6050_cfg = {
 
 /**
  * One slot per selectable receiver group.
- *  eg. PWM, PPM, GCS, DSMMAINPORT, DSMFLEXIPORT, SBUS
+ *  eg. PWM, PPM, GCS, DSM, DSM, SBUS
```
peabody124: find/replace typo

flight/targets/pipxtreme/board-info/pios_board.h
```diff
@@ -165,22 +165,18 @@ extern uint32_t pios_i2c_flexi_adapter_id;
...
-extern uintptr_t pios_com_telem_vcp_id;
-extern uintptr_t pios_com_telem_uart_telem_id;
-extern uintptr_t pios_com_telem_uart_main_id;
-extern uintptr_t pios_com_telemetry_id;
-extern uintptr_t pios_com_rfm22b_id;
-extern uintptr_t pios_com_radio_id;
+extern uintptr_t pios_com_telem_usb_id; 
+extern uintptr_t pios_com_vcp_id; 
+extern uintptr_t pios_com_radio_id; 
+extern uintptr_t pios_com_telem_rf_id; 		// POORLY NAMED
```
peabody124: feel free to fix it. let's not have commentary throughout the code :)

flight/targets/pipxtreme/fw/pios_board.c
```diff
@@ -266,148 +170,13 @@ void PIOS_Board_Init(void) {
...
-
-		// Reinitilize the modem to affect te changes.
-		PIOS_RFM22B_Reinit(pios_rfm22b_id);
-	} else {
-		rfm22bstatus.LinkState = RFM22BSTATUS_LINKSTATE_DISABLED;
-	}
-
-	// Update the object
-	RFM22BStatusSet(&rfm22bstatus);
+	PIOS_HAL_configure_RFM22B(hwTauLink.Radio, bdinfo->board_type,
```
peabody124: so this is where the abstraction bumps into issues. the radio types supported are different here. HwTauLink does not support OpenLRS so that is an invalid port option.

Also it doesn't look like there is a compilation error, even though the Radio port in this .xml is not a shared type. This could allow some pretty nasty bugs if the compiler isn't actually checking for the common enums. Not sure if this is fixable.

shared/uavobjectdefinition/hwtaulink.xml
```diff
@@ -11,15 +11,15 @@
 		<field name="CoordID" units="hex" type="uint32" elements="1" defaultvalue="0"/>
```
peabody124: Radio field above should be common type ```HwShared.RadioPort```

flight/PiOS/Common/pios_port.c
```diff
@@ -0,0 +1,623 @@
...
+
+		PIOS_Assert(rx_buffer);
+		PIOS_Assert(tx_buffer);
+		if (PIOS_COM_Init(&pios_com_rf_id, &pios_rfm22b_com_driver, pios_rfm22b_id,
+					rx_buffer, PIOS_COM_RFM22B_RF_RX_BUF_LEN,
+					tx_buffer, PIOS_COM_RFM22B_RF_TX_BUF_LEN)) {
+			PIOS_Assert(0);
+		}
+
+#ifndef PIOS_NO_TELEM_ON_RF
```
peabody124: ouch the double negative hurts my head :)

flight/PiOS/Common/pios_port.c
peabody124: never mind - looks like namespace in file is PIOS_HAL which is reasonable. can the file name match that?

shared/uavobjectdefinition/hwshared.xml
```diff
@@ -0,0 +1,41 @@
...
+				<option>MavLinkTX_GPS_RX</option>
+				<option>HoTT SUMD</option>
+				<option>HoTT SUMH</option>
+				<option>HoTT Telemetry</option>
+				<option>FrSKY Sensor Hub</option>
+				<option>FrSKY SPort Telemetry</option>
+				<option>LighttelemetryTx</option>
+				<option>PicoC</option>
+<!-- These last couple are special cases, really.  le sigh -->
+				<option>I2C</option>
```
peabody124: for this the top level code will need to know whether a port is set to I2C in order to know whether to try probing for an external mag etc. that being said this isn't done currently so not a huge issue :)

peabody124:
Overall this is going in a really nice direction and has a good LOC balance :)

My preference would be to get this applied to all targets (or even defer that) and tested and leave any increased scope to later PRs. I added a few checkboxes to the top of things that would/should be tested -although if we don't include more targets F3 testing isn't necessary.

I didn't review the UAVO Generator code yet. I also didn't test that android will actually connect using this. I'm totally ok delaying full android support, but don't want to see the hash detection etc broken (probably isn't).

I still don't see a way to address things like certain ports don't support SBus. This is going to lead to users having invalid configurations. Also things like CC3D not supporting a large number of serial port options but those will be presented to the user. Do you have any thoughts on this? I thought previously we had discussed perhaps having variants of the port type, but I don't see a way for that to fit into the code.

flight/targets/pipxtreme/fw/pios_board.c
mlyle: The types in hwshared include all the possibilities and the children can include any subset.  The stuff in pios_port must check for all types and this is enforced by the compiler.  If someone does a switch statement on the child type in the individual hardware module it enforces them to the subset.

flight/targets/pipxtreme/board-info/pios_board.h
mlyle: note this is WIP and on my task list above is variable name cleanup and conditionalization of these...

flight/targets/discoveryf4/fw/pios_board.c
mlyle: Once they're adjusted it'll go away.

flight/PiOS/Common/pios_port.c
mlyle: Yup, also still not named correctly per my todo list.

flight/PiOS/Common/pios_port.c
mlyle: Yup, pretty much.

flight/PiOS/Common/pios_port.c
mlyle: Right now the child port type would not include SBUS in those cases.  If it erroneously did, we would be prevented from configuring it by the if block.

I don't know about soft usart, but I don't know why sbus can't be "bring your own inverter" like the frsky modes etc.  Just takes documentation.  But that's for another PR.

flight/Modules/RadioComBridge/RadioComBridge.c
mlyle: The problem is no one calls the module init methods.  It's done from putting pointers in an initcall section automatically, so there's not really a way to say what port you mean.

There's other ways but they use a few bytes of ram, and probably aren't worth it

flight/Modules/ManualControl/transmitter_control.c
mlyle: Why just DSM though?  It's a total hack specific to one receiver type. 

We could transition to have the channelgroup always specify the port instead.  Would let you duplicate any of these and would use less memory.  Probably not good for this PR though.

flight/PiOS/Common/pios_port.c
mlyle: Yah.  It's the lesser of evils since we can't really pass things to the modules.  You either need to break that, or to have a broker that you can ask "give me the id of the thing assigned to... telemetry" that modules call.  In the future the hal is going to own all of these and modules can share them.

shared/uavobjectdefinition/hwtaulink.xml
mlyle: Good catch.

mlyle:
Just to clarify:

> I still don't see a way to address things like certain ports don't support SBus. This is going to lead to users having invalid configurations. Also things like CC3D not supporting a large number of serial port options but those will be presented to the user. Do you have any thoughts on this? I thought previously we had discussed perhaps having variants of the port type, but I don't see a way for that to fit into the code.

Sparse children look like this in the generated uavo:

```
typedef enum { HWTAULINK_MAINPORT_DISABLED=HWSHARED_PORTTYPES_DISABLED, HWTAULINK_MAINPORT_GPS=HWSHARED_PORTTYPES_GPS, HWTAULINK_MAINPORT_TELEMETRY=HWSHARED_PORTTYPES_TELEMETRY, HWTAULINK_MAINPORT_COMBRIDGE=HWSHARED_PORTTYPES_COMBRIDGE }  __attribute__((packed)) HwTauLinkMainPortOptions;
```

vs. 

```
typedef enum { HWSPARKY2_MAINPORT_DISABLED=HWSHARED_PORTTYPES_DISABLED, HWSPARKY2_MAINPORT_TELEMETRY=HWSHARED_PORTTYPES_TELEMETRY, HWSPARKY2_MAINPORT_GPS=HWSHARED_PORTTYPES_GPS, HWSPARKY2_MAINPORT_DSM=HWSHARED_PORTTYPES_DSM, HWSPARKY2_MAINPORT_DEBUGCONSOLE=HWSHARED_PORTTYPES_DEBUGCONSOLE, HWSPARKY2_MAINPORT_COMBRIDGE=HWSHARED_PORTTYPES_COMBRIDGE, HWSPARKY2_MAINPORT_MAVLINKTX=HWSHARED_PORTTYPES_MAVLINKTX, HWSPARKY2_MAINPORT_MAVLINKTX_GPS_RX=HWSHARED_PORTTYPES_MAVLINKTX_GPS_RX, HWSPARKY2_MAINPORT_HOTTSUMD=HWSHARED_PORTTYPES_HOTTSUMD, HWSPARKY2_MAINPORT_HOTTSUMH=HWSHARED_PORTTYPES_HOTTSUMH, HWSPARKY2_MAINPORT_HOTTTELEMETRY=HWSHARED_PORTTYPES_HOTTTELEMETRY, HWSPARKY2_MAINPORT_FRSKYSENSORHUB=HWSHARED_PORTTYPES_FRSKYSENSORHUB, HWSPARKY2_MAINPORT_FRSKYSPORTTELEMETRY=HWSHARED_PORTTYPES_FRSKYSPORTTELEMETRY,HWSPARKY2_MAINPORT_LIGHTTELEMETRYTX=HWSHARED_PORTTYPES_LIGHTTELEMETRYTX, HWSPARKY2_MAINPORT_PICOC=HWSHARED_PORTTYPES_PICOC }  __attribute__((packed)) HwSparky2MainPortOptions;
```

The right things also happen in the uav object browser.  Hashes take into account skipped elements in a robust way, etc.

mlyle:
I deleted the test checklist above because i'm planning to put together a test matrix on the wiki.

flight/PiOS/Common/pios_port.c
mlyle: Note that I have frequency-dependent stuff to do also.. i could take this over easily enough if you're not attached to doing the work.

ground/gcs/src/plugins/uavobjects/uavobjectfield.cpp
```diff
@@ -877,12 +882,14 @@ QVariant UAVObjectField::getValue(quint32 index)
...
         quint8 tmpenum;
         memcpy(&tmpenum, &data[offset + numBytesPerElement*index], numBytesPerElement);
-        //            Q_ASSERT((tmpenum < options.length()) && (tmpenum >= 0)); // catch bad enum settings
-        if(tmpenum >= options.length()) {
-            qDebug() << "Invalid value for" << name;
-            return QVariant( QString("Bad Value") );
-        }
-        return QVariant( options[tmpenum] );
+	// Too slow?
+	for (int i = 0; i < indices.length(); i++) {
```
guilhermito: Why not use a QHash<QString, int> instead of a QStringList and a QList<int> ?

mlyle: Basically because there's existing stuff that relies on the options being a list-- to generate combo boxes, export things, etc.  I didn't want to break it.

Anyways, the other direction this data structure is fast.  It just requires a scan the other way.  Could add a hash to speed this way up but I've not seen any performance penalty on save.

mlyle:
Rebased: (trivial) conflict with coptercontrol uavo change.

peabody124:
Hey @mlyle since I think this is targeting the next release you want to clean up the header? could you also add back a few of the bullets that should be tested here (or link to a test matrix if so) That way we can know where we stand on merging this.

mlyle:
Test matrix is on the release wiki page at: https://github.com/TauLabs/TauLabs/wiki/Endomorphism-(September)-release

peabody124:
great. want to update the title too since at this point it isn't for review but just testing and merge, right?

mlyle:
http://jar.lyle.org/~mlyle/TauLabs-20150815-f18976be-install.exe is the current test artifact.

mlyle:
The header of this is updated to get it ready for merge.  The test matrix is getting nicely filled in.  

I would like this to merge soon.  Work on the test matrix can continue.  I think once DSM is verified to work on CC the remainder of the work items can open as issues.  We can fill the test matrix in a little more afterwards, but it's not a good reason to block other work (frequency config, TLM support, Naze32 serial).

flight/PiOS/Common/pios_hal.c
```diff
@@ -0,0 +1,748 @@
...
+			target = &pios_com_bridge_id;
+			break;
+		case HWSHARED_PORTTYPES_MAVLINKTX:
+#if defined(PIOS_INCLUDE_MAVLINK)
+			PIOS_HAL_ConfigureCom(usart_port_cfg, 0, PIOS_COM_MAVLINK_TX_BUF_LEN, com_driver, &port_driver_id);
+			target = &pios_com_mavlink_id;
+#endif          /* PIOS_INCLUDE_MAVLINK */
+			break;
+		case HWSHARED_PORTTYPES_MAVLINKTX_GPS_RX:
+#if defined(PIOS_INCLUDE_MAVLINK)
```
tracernz: Should this configure GPS rx even if MAVLINK is not included?

peabody124: I think so. It is a more graceful way to fail.

mlyle: Maybe.  It's not what we did before.

mlyle:
Merging to unblock other work.

PR #1763, opened 2015-08-02
gcs: combine gcscontrolgadget and gcscontrolwidget into one plugin

mlyle:
Slight benefit to compile parallelism and GCS load time; tree cleanliness.


mlyle:
Missing one change here, so please hold.

mlyle:
This is good now, I believe.

Issue #1709, opened 2015-07-06
Issues with next TL7_5_15


pug398:
PFD does not report # of satellites correctly. Instead of numbers has strange characters. Reported correctly in System>Satellite widget but not in PFD GPS sat count. Reported PDOP and ACC ok. Started last few releases and config files deleted and reinstalled. Opengl32.dll installed and graphics have been working fine. Windows error on closing of GCS appears fixed. Looked back and seems to have started with 6_20_15 release.

mlyle:
Yes.  I believe this is a bug / limitation in QML that QT 5.5.0 may be expected to fix.  Before we had two PFD implementations-- the C implementation and the javascript/qml future one;  I removed the old one and there are definitely some regressions / limitations in functionality there because it seems QML can't get to some types.

pug398:
Ok Mike. Just wanted to make sure that it was known and didn't get overlooked before going to release. Looks like the other issues I ran into have been resolved.

mlyle:
Believed to be resolved.

PR #1764, opened 2015-08-02
gcs: pfd: correct satellite count display

mlyle:
Work around upstream QT bug 37241 in QML:

https://bugreports.qt.io/browse/QTBUG-37241

### DIFF
--- a/ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
+++ b/ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
@@ -39,14 +39,13 @@ Item {
     // GPS status text
     Text {
         id: gps_text
-        text: "GPS: " + GPSPosition.Satellites + "\nPDP: " + GPSPosition.PDOP.toFixed(2) + "\nACC: " + GPSPosition.Accuracy.toFixed(2)
+        text: "GPS: " + GPSPosition.Satellites.charCodeAt(0) + "\nPDP: " + GPSPosition.PDOP.toFixed(2) + "\nACC: " + GPSPosition.Accuracy.toFixed(2)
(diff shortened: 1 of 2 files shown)


mlyle:
Resolves issue #1709

ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
```diff
@@ -39,14 +39,13 @@ Item {
     // GPS status text
     Text {
         id: gps_text
-        text: "GPS: " + GPSPosition.Satellites + "\nPDP: " + GPSPosition.PDOP.toFixed(2) + "\nACC: " + GPSPosition.Accuracy.toFixed(2)
+        text: "GPS: " + GPSPosition.Satellites.charCodeAt(0) + "\nPDP: " + GPSPosition.PDOP.toFixed(2) + "\nACC: " + GPSPosition.Accuracy.toFixed(2)
```
BrainFPV: Doesn't this just work for 0..9 sats?

mlyle: This takes the first character of the "Satellites" string and takes its ascii value.  Unfortunately because of the upstream QT bug of treating any character types as a string we have to do this ordinal value here.

It works for values 0..127, maybe for 0..255.

BrainFPV: OK, makes sense.

BrainFPV:
LGTM; I would merge it if I had merge rights ;)

tracernz:
LGTM++; No indication of the upstream bug being fixed anytime soon.

Issue #168, opened 2013-01-06
Provide global sensor dT


peabody124:
This depends on #163

Right now the dT for each sensor update is computed per sample.  This is going to introduce artificial jitter into the dT estimation.  The PIOS_SENSORS interface provides a natural place to accululate a smoothed estimate of the sampling rate.  This can either be done per sensor or if that is not necessary the fastest sensor (typically gyro) can be the one tracking it.

Suggested API
    PIOS_SENSOR_AcquiredSample() -- this stores a timestamp for smoothing
    PIOS_SENSOR_GetDt() -- returns the smoothed time step

mlyle:
@Trex4Git is looking at this

mlyle:
#1765 did the extent of what we wanted to do for the Endo release-- we can make better with time.

Trex4Git:
I think maybe a strategical decision on sensor dT may be helpful.

In my PR only for the gyro LPF in stabilization a averaged dT is calculated after start (where before a constant fake dT was used). 
All other things there like the PID control an also the LPF the derivative term still use latest dT.
When they should also use a filtered dT I am not sure if only calculating it once after startup would be sufficient, or a continuous filtering of dT is needed.

I have seen that the complementary filter for attitude uses also the same filter (&math with fake dT) for accel and gyro data, as the gyro LPF in stabilization before that PR https://github.com/TauLabs/TauLabs/pull/1765.
So i am thinking about changing it the same way as done for stabilization.

If just adapting we would have than 2 filtered dT calculations, and the suggested API at the beginning of this iusse also looks more like a central place with filtered dT calculations.
The sensors task might be place where PIOS_SENSOR API could be used. Because from there Attitude task and Stabilization task seems to get triggered. 
But as not all tasks there have the same priority, this might be a place for jitter problems (I described there my impression: https://github.com/TauLabs/TauLabs/issues/1786).

So it might be better to have separate filtered dT calculations for attitude and stabilization task, as they might have different delays.
I also seen ideas of running attitude and stabilization loops at different rates. That would also be a reason to use separate dT, I think.

mlyle:
I think we should have a single global dT.

This is because a lot of the subsequent uses of dT really have to do with the duration the sensor data is interpreted over.  And this doesn't have so much to do with when we're looking at the information but instead when it's measured.

[...]

PR #1765, opened 2015-08-02
[WIP] flight/stabilization: enhancement of the gyro LPF functionality

Trex4Git:
I noticed that the current gyro LPF in stabilization.c uses a constant fake delta Time (https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Stabilization/stabilization.c#L941-L950). From the comments it seems this was done because of jitter in the real dT, and different control rates at that time were 300Hz & 475Hz. 
Currently on many targets 500Hz is default rate, Quanton has 666Hz, and 1kHz and above is also possible on some targets.
And there is also work on higher rates.

Here are some results of the LPF magnitude response with 2kHz, 1kHz and 500Hz rate:
taulabs_gyro_filter 

But i think that the principal behavior (cut off frequency) of the LPF should not change with the control loop rate. It should be adjusted depending of the noise in the UAV.

So i started with some initial work as a base for further discussion.

Thats what i have done so far:
- changed the calculation of the LPF coefficients, so that 1/settings.GyroTau is now equal to the cut off frequency of the filter
 - I think it would be easier to understand the functionality if we rename the Parameter in the UAVObject from GyroTau to GyroFcut. But for the first steps i didn't want to change to much, that might be not used later.
- using the real dT for the gyro LPF, otherwise the behavior (cut off frequency) is changed when rate of the control loop (e.g. MPU rate) is changed
 - in the comments at the source code it was said that using a constant fakeDt was because of jitter. I don't know how big this problem is. For first implementation i used the real(dynamic) dT. Is this really problematic?
 - when we really need a dT without any jitter, than it would be good to provide the information of the desired gyro sampling rate from the gyro driver (e.g. MPU6000/9250) to the stabilization module
- changed the default value of GyroTau in a way, that the LPF behaves for all rates now like the old default settings with 500Hz MPU rate (51Hz cut off)

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -231,21 +231,19 @@ static void stabilizationTask(void* parameters)
 		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
 		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
 		// run it only at the beginning for the first samples, to reduce CPU load, and the value should converge to a constant value
-		if (dT_filter_iteration-- > 2) {
+		if (dT_filter_iteration >= 1) {
+			dT_filter_iteration--;
 			dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
-
-			if (gyro_filter_enabled  == GYRO_LPF_DISABLED)
-				gyro_alpha = 0;   // Gyro LPF should be disabled
-			else
-				gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
 		}
 
 
-		// new calculation of gyro_alpha when cutoff frequency has changed, and dt_filtered is already calculated
-		if ( (gyro_filter_enabled  == GYRO_LPF_ENABLED_UPDATE) && (dT_filter_iteration == 1) ) {
+		// new calculation of gyro_alpha when cutoff frequency has changed, or dt_filtered is not already completly calculated
+		if ( (gyro_filter_enabled  == GYRO_LPF_ENABLED_UPDATE) || (dT_filter_iteration > 0) ) {
 			gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
 			gyro_filter_enabled = GYRO_LPF_ENABLED;
 		}
+		else
+			gyro_filter_enabled = GYRO_LPF_DISABLED;
 
 		FlightStatusGet(&flightStatus);
 		StabilizationDesiredGet(&stabDesired);
@@ -982,7 +980,10 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
 		//fakeDt now only used for vbar_decay, for gyro filter now a real, but filtered, dT is used
 		const float fakeDt = 0.0025f;
 
-		if (settings.GyroCutoff < 0.0001f) {	// not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
+		//TODO: if the infrastructure for alarms will be more advanced in the future,
+		//      we should assert a alarm here when GyroCutoff > 90% of F_nyquist and LPF gets disabled
+
+		if (settings.GyroCutoff < 1.0f) {	// disable LPF when GyroCutoff < 1.0 Hz
 			gyro_filter_enabled = GYRO_LPF_DISABLED;
 			settings.GyroCutoff = 0;
 		}
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -108,7 +108,7 @@ static StabilizationSettingsData settings;
 static TrimAnglesData trimAngles;
 static struct pios_queue *queue;
 float gyro_alpha = 0;
-static float gyro_alpha_temp = 0;
+static float gyro_filter_enabled = 0;
 static float dT = 0;
 static float dT_filtered = 0;
 float axis_lock_accum[3] = {0,0,0};
@@ -200,6 +200,8 @@ static void stabilizationTask(void* parameters)
 	const uint32_t SYSTEM_IDENT_PERIOD = 75;
 	uint32_t system_ident_timeval = PIOS_DELAY_GetRaw();
 
+	uint16_t dT_filter_iteration = 1e3;
+
 	// Main task loop
 	zero_pids();
 	while(1) {
@@ -222,12 +224,22 @@ static void stabilizationTask(void* parameters)
 		
 		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
 		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
-		dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
+		// run it only at the beginning for the first samples, to reduce CPU load, and the value should converge to a constant value
+		if (dT_filter_iteration-- > 1) {
+			dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
 
-		if (gyro_alpha_temp  < 0.0001f) // not trusting this to resolve to 0
-			gyro_alpha = 0;   // Gyro LPF should be disabled
-		else
-			gyro_alpha = expf(-gyro_alpha_temp * dT_filtered);
+			if (gyro_filter_enabled  < 0.0001f) // not trusting this to resolve to 0
+				gyro_alpha = 0;   // Gyro LPF should be disabled
+			else
+				gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
+		}
+
+
+		// new calculation of gyro_alpha when cutoff requency has changed, and dt_filtered is already calculated
+		if ( (gyro_filter_enabled  > 1.0001f) && (dT_filter_iteration == 1) ) {
+			gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
+			gyro_filter_enabled = 1.0f;
+		}
 
 		FlightStatusGet(&flightStatus);
 		StabilizationDesiredGet(&stabDesired);
@@ -303,9 +315,18 @@ static void stabilizationTask(void* parameters)
 		local_attitude_error[2] = circular_modulus_deg(local_attitude_error[2]);
 
 		static float gyro_filtered[3];
-		gyro_filtered[0] = gyro_filtered[0] * gyro_alpha + gyrosData.x * (1 - gyro_alpha);
-		gyro_filtered[1] = gyro_filtered[1] * gyro_alpha + gyrosData.y * (1 - gyro_alpha);
-		gyro_filtered[2] = gyro_filtered[2] * gyro_alpha + gyrosData.z * (1 - gyro_alpha);
+		if (gyro_alpha < 0.0001f) {	// not trusting this to resolve to 0
+			gyro_filtered[0] = gyrosData.x;
+			gyro_filtered[1] = gyrosData.y;
+			gyro_filtered[2] = gyrosData.z;
+		}
+		else {
+			gyro_filtered[0] = gyro_filtered[0] * gyro_alpha + gyrosData.x * (1 - gyro_alpha);
+			gyro_filtered[1] = gyro_filtered[1] * gyro_alpha + gyrosData.y * (1 - gyro_alpha);
+			gyro_filtered[2] = gyro_filtered[2] * gyro_alpha + gyrosData.z * (1 - gyro_alpha);
+		}
+
+
 
 		// A flag to track which stabilization mode each axis is in
 		static uint8_t previous_mode[MAX_AXES] = {255,255,255};
@@ -956,13 +977,13 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
 		const float fakeDt = 0.0025f;
 
 		if (settings.GyroCutoff < 0.0001f)
-			gyro_alpha_temp = 0;   // not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
+			gyro_filter_enabled = 0;   // not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
 		else if ( (dT > 0) && (settings.GyroCutoff > (0.45f / dT)) ) { // disable LPF when GyroCutoff > 90% of F_nyquist (= Fs/2 = 1/(2*dT)); using here not the filtered dT, because this is at the start to low
-			gyro_alpha_temp = 0;
-			StabilizationSettingsGyroCutoffSet(&gyro_alpha_temp); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
+			gyro_filter_enabled = 0;
+			StabilizationSettingsGyroCutoffSet(&gyro_filter_enabled); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
 		}
 		else
-			gyro_alpha_temp = 2.0f * (float)(M_PI) * settings.GyroCutoff;
+			gyro_filter_enabled = 2; // indicating that the gyro LPF is enabled, but the Cutoff parameter has changed, so the recalculation is needed
 
 
 		// Compute time constant for vbar decay term based on a tau
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -100,6 +100,12 @@ enum {
 	PID_MAX
 };
 
+enum {
+	GYRO_LPF_DISABLED  = 0,
+	GYRO_LPF_ENABLED,
+	GYRO_LPF_ENABLED_UPDATE,
+		};
+
 
 // Private variables
 static struct pios_thread *taskHandle;
@@ -108,7 +114,7 @@ static StabilizationSettingsData settings;
 static TrimAnglesData trimAngles;
 static struct pios_queue *queue;
 float gyro_alpha = 0;
-static float gyro_filter_enabled = 0;
+static uint8_t gyro_filter_enabled = 0;
 static float dT = 0;
 static float dT_filtered = 0;
 float axis_lock_accum[3] = {0,0,0};
@@ -200,7 +206,7 @@ static void stabilizationTask(void* parameters)
 	const uint32_t SYSTEM_IDENT_PERIOD = 75;
 	uint32_t system_ident_timeval = PIOS_DELAY_GetRaw();
 
-	uint16_t dT_filter_iteration = 1e3;
+	uint16_t dT_filter_iteration = 1001;
 
 	// Main task loop
 	zero_pids();
@@ -225,20 +231,20 @@ static void stabilizationTask(void* parameters)
 		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
 		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
 		// run it only at the beginning for the first samples, to reduce CPU load, and the value should converge to a constant value
-		if (dT_filter_iteration-- > 1) {
+		if (dT_filter_iteration-- > 2) {
 			dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
 
-			if (gyro_filter_enabled  < 0.0001f) // not trusting this to resolve to 0
+			if (gyro_filter_enabled  == GYRO_LPF_DISABLED)
 				gyro_alpha = 0;   // Gyro LPF should be disabled
 			else
 				gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
 		}
 
 
-		// new calculation of gyro_alpha when cutoff requency has changed, and dt_filtered is already calculated
-		if ( (gyro_filter_enabled  > 1.0001f) && (dT_filter_iteration == 1) ) {
+		// new calculation of gyro_alpha when cutoff frequency has changed, and dt_filtered is already calculated
+		if ( (gyro_filter_enabled  == GYRO_LPF_ENABLED_UPDATE) && (dT_filter_iteration == 1) ) {
 			gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
-			gyro_filter_enabled = 1.0f;
+			gyro_filter_enabled = GYRO_LPF_ENABLED;
 		}
 
 		FlightStatusGet(&flightStatus);
@@ -315,7 +321,7 @@ static void stabilizationTask(void* parameters)
 		local_attitude_error[2] = circular_modulus_deg(local_attitude_error[2]);
 
 		static float gyro_filtered[3];
-		if (gyro_alpha < 0.0001f) {	// not trusting this to resolve to 0
+		if (gyro_filter_enabled  == GYRO_LPF_DISABLED) {
 			gyro_filtered[0] = gyrosData.x;
 			gyro_filtered[1] = gyrosData.y;
 			gyro_filtered[2] = gyrosData.z;
@@ -976,14 +982,17 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
 		//fakeDt now only used for vbar_decay, for gyro filter now a real, but filtered, dT is used
 		const float fakeDt = 0.0025f;
 
-		if (settings.GyroCutoff < 0.0001f)
-			gyro_filter_enabled = 0;   // not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
-		else if ( (dT > 0) && (settings.GyroCutoff > (0.45f / dT)) ) { // disable LPF when GyroCutoff > 90% of F_nyquist (= Fs/2 = 1/(2*dT)); using here not the filtered dT, because this is at the start to low
-			gyro_filter_enabled = 0;
-			StabilizationSettingsGyroCutoffSet(&gyro_filter_enabled); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
+		if (settings.GyroCutoff < 0.0001f) {	// not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
+			gyro_filter_enabled = GYRO_LPF_DISABLED;
+			settings.GyroCutoff = 0;
+		}
+		else if ( (dT > 0) && (settings.GyroCutoff > (0.45f / dT)) ) {	// disable LPF when GyroCutoff > 90% of F_nyquist (= Fs/2 = 1/(2*dT)); using here not the filtered dT, because this is at the start to low
+			gyro_filter_enabled = GYRO_LPF_DISABLED;
+			settings.GyroCutoff = 0;
+			StabilizationSettingsGyroCutoffSet(&settings.GyroCutoff); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
 		}
 		else
-			gyro_filter_enabled = 2; // indicating that the gyro LPF is enabled, but the Cutoff parameter has changed, so the recalculation is needed
+			gyro_filter_enabled = GYRO_LPF_ENABLED_UPDATE; // indicating that the gyro LPF is enabled, but the Cutoff parameter has changed, so the recalculation is needed
 
 
 		// Compute time constant for vbar decay term based on a tau
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -108,6 +108,7 @@ static StabilizationSettingsData settings;
 static TrimAnglesData trimAngles;
 static struct pios_queue *queue;
 float gyro_alpha = 0;
+static float dT = 0;
 float axis_lock_accum[3] = {0,0,0};
 uint8_t max_axis_lock = 0;
 uint8_t max_axislock_rate = 0;
@@ -202,8 +203,6 @@ static void stabilizationTask(void* parameters)
 	while(1) {
 		iteration++;
 
-		float dT;
-		
 		PIOS_WDG_UpdateFlag(PIOS_WDG_STABILIZATION);
 		
 		// Wait until the AttitudeRaw object is updated, if a timeout then go to failsafe
@@ -944,10 +943,11 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
 		// update rates on OP (~300 Hz) and CC (~475 Hz) is negligible for this
 		// calculation
 		const float fakeDt = 0.0025f;
+
 		if(settings.GyroTau < 0.0001f)
 			gyro_alpha = 0;   // not trusting this to resolve to 0
 		else
-			gyro_alpha = expf(-fakeDt  / settings.GyroTau);
+			gyro_alpha = (1 / (1 +  ((2.0f * (float)(M_PI) * dT) / settings.GyroTau) ));
 
 		// Compute time constant for vbar decay term based on a tau
 		vbar_decay = expf(-fakeDt / settings.VbarTau);
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -108,7 +108,9 @@ static StabilizationSettingsData settings;
 static TrimAnglesData trimAngles;
 static struct pios_queue *queue;
 float gyro_alpha = 0;
+static float gyro_alpha_temp = 0;
 static float dT = 0;
+static float dT_filtered = 0;
 float axis_lock_accum[3] = {0,0,0};
 uint8_t max_axis_lock = 0;
 uint8_t max_axislock_rate = 0;
@@ -218,6 +220,15 @@ static void stabilizationTask(void* parameters)
 		dT = PIOS_DELAY_DiffuS(timeval) * 1.0e-6f;
 		timeval = PIOS_DELAY_GetRaw();
 		
+		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
+		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
+		dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
+
+		if (gyro_alpha_temp  < 0.0001f) // not trusting this to resolve to 0
+			gyro_alpha = 0;   // Gyro LPF should be disabled
+		else
+			gyro_alpha = expf(-gyro_alpha_temp * dT_filtered);
+
 		FlightStatusGet(&flightStatus);
 		StabilizationDesiredGet(&stabDesired);
 		AttitudeActualGet(&attitudeActual);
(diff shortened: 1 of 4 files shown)


mlyle:
Removing jitter is important.  Any nonlinearities in the gain of this will create mixing products.  E.g. if we have even harmonic vibration at 20Hz and 40Hz, and there is jitter causing gain, will create an offset at 0Hz (40-20Hz = 0Hz mixing product).  Offsets at 0Hz are really bad.

We could use the configured gyro rate, or we could run a lowpass filter on the actual achieved sampling rate that removes the jitter.  That might be better because it will behave better in the case of lost samples.  It's worth noting that we have bugs that are currently causing jitter.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -944,10 +943,11 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
 		// update rates on OP (~300 Hz) and CC (~475 Hz) is negligible for this
 		// calculation
 		const float fakeDt = 0.0025f;
+
 		if(settings.GyroTau < 0.0001f)
 			gyro_alpha = 0;   // not trusting this to resolve to 0
 		else
-			gyro_alpha = expf(-fakeDt  / settings.GyroTau);
+			gyro_alpha = (1 / (1 +  ((2.0f * (float)(M_PI) * dT) / settings.GyroTau) ));
```
mlyle: This doesn't seem to be the correct form.  exp(-T / tau) is the correct way to get an IIR alpha for a given time constant.

mlyle:
Just to explain what I'm saying in a different way--- note that dt is not the time between samples, but the time between when the microcontroller is evaluating them.  If no samples are dropped, the time between sampling is uniform.  But any jitter becomes random variations in dT.  Effectively, then, we're multiplying the samples by a noise spectrum with an exponential distribution and an inverse correlation between successive samples, which is really bad.

Trex4Git:
I already also had the idea of filtering the actual achieved sampling rate. But I am not sure if filtering some data for getting parameters for another filter is the best way, but it would be a "standalone" solution, and no changes at others parts of the code is needed.

Otherwise I thought about adding a additional parameter to the UAVObject. Than the default value will have a similar behavior. But the gyro driver can change that value depending on the configured  gyro rate. So that not every gyro driver has to be changed at the same time.

flight/Modules/Stabilization/stabilization.c
Trex4Git: Yes, that is the right. But i thought defining the filter in time domain is maybe more difficult for people, than doing it in frequency domain.

Thats why i have written in this implementation it means  tau  = 1/Fcut based on this description: https://en.wikipedia.org/wiki/Low-pass_filter#Discrete-time_realization
And when there is interest in merging this, i prefer to rename the parameter (and also adapt the math) from GyroTau to GyroFcut (as already written).

mlyle:
Well, if you filter it, it won't remove that noise spectrum entirely but greatly attenuate it.  The other reason I like that approach is that it makes it a little better if we are dropping samples.  Perhaps we default to a time constant of about 1 second for that filter-- then any jitter would be attenuated by >100x.

Of course, it's really bad to not keep up and lose 20% of samples-- that introduces its own distortion.  But at least here the filter would adapt a little.

If you really don't like that approach, plumbing it through the UAVO is the correct way.

mlyle:
One other thing-- we should rename GyroTau or change the capitalization.  Because otherwise people with existing settings will have the same UAVO and will keep their existing value and you'll effectively be doubling their gyro bandwidth.

Trex4Git:
As written above in response to your code comment:

"Yes, that is the right. But i thought defining the filter in time domain is maybe more difficult for people, than doing it in frequency domain.

Thats why i have written in this first test implementation it means tau = 1/Fcut based on this description: https://en.wikipedia.org/wiki/Low-pass_filter#Discrete-time_realization
And when there is interest in merging this, i prefer to rename the parameter (and also adapt the math) from GyroTau to GyroFcut (as already written)."

flight/Modules/Stabilization/stabilization.c
mlyle: Ah, I'm sorry, I missed that.

mlyle:
I think using the 1/Fcut form is nonintuitive.  We should either use a conventional time constant or the cutoff frequency itself.

Trex4Git:
That is also my opinion, i have just used that notation in my quick written matlab scripts to test the filter.
And than used it in the quick first implementation. 
If there would have been no interest in this, i don't wanted to change to much.
GyroFcut is also just a working name for me that might be used later, not a final solution that i want to pretend.


gyro_alpha = (1 / (1 +  ((2.0f * (float)(M_PI) * dT) / settings.GyroTau) ));
will change to
gyro_alpha = (1 / (1 +  (2.0f * (float)(M_PI) * dT * settings.GyroFcut) ));

In comparison with the wiki page that I posted, they use also alpha as the parameter for the filter, like in the taulabs filter, but its not the same:
alpha_wiki = 1 - alpha_taulabs

so we have a little bit different formula to calculate the alpha from Fcut compared to wikipedia. I hope i have not done a mistake in the math.

Trex4Git:
Regarding what dT to use:
It's not that i don't like the filtering approach. 
I just thought it was not elegant and might be not accepted by you/the advanced developer. The UAVObject approach would be more work, as even if not absolutely needed with a default behavior like the current implementation all gyro driver (MPU6000/6050/9150/9250,LS3GD20,...?) should be updated to this functionality to have a clean / consistent behavior. 

But as you prefer also the filtering approach, and i have not thought about the advantages you mentioned, that might me the way to go.
Do you think we could use the same IIR LPF, just with a constant cutoff frequency / time constant?

mlyle:
Yah, I like the IIR best, actually.   Technically it won't be a "time constant", but an alpha corresponding to 200 samples or whatever.

It's too bad how limited SMALLF1 is, etc.  The gyro IIR is unfortunate-- it adds lots of phase shift at the frequencies we're most concerned about for oscillation.

Trex4Git:
There I posted some example plots of the magnitude response and group delay for 2 setups of the 1tap IIR: http://www.rcgroups.com/forums/showthread.php?t=2439428&page=41

I did also some small investigations what might be the best LPF filter if we have to much noise in the passband of the upcoming CIC solution.
- using a higher order and/or differential delay on the CIC
- using a 1. or higher Order IIR as additional filter (chebyshev type1&2, butterworth, bessel)
- using a 1. or higher Order FIR as additional filter (different design method /windows like Kaiser window)

The higher order CIC is not  a good solution, as the delay is bigger, than a "smaller" CIC combined with the existing 1tap IIR for the same damping performance. That advantage of the CIC might be the constant group delay over frequency.

The other IIR I looked at, also could not achieve a better lowpass performance compared to the existing IIR, when the delay should not be  larger.

FIR filter need a higher order for the same lowpass behavior, so more computational power.
Advantage, like the CIC, might be the constant group delay, but its also not smaller than the group delay of the 1tap IIR at small frequencies.

So even with the CIC implementation, on noise setups the 1tap IIR should be useful so i did some more investigations on our implementation what lead to this PR.

mlyle:
@fujin flew two tunes.  First was with default gyrotau:
17.9ms tau=17.9ms

Second was with gyrotau=0:
13.0ms tau=13.0ms

Definitely the low frequency group delay of the IIR can be measured.

Trex4Git:
Also read that on IRC. I think that was done with 500Hz MPU rate.

So i generated a plot (magnitude response, phase, group delay) with default GyroTau setup (0.005) for 500Hz, 1kHz and 2kHz MPU/Control Rate. Also curves for 500Hz but modified GyroTau to have a similar filter behavior like 1KHz rate (GyroTau = 0.00265) and 2KHz rate (GyroTau = 0.00161).

taulabs_1tapiir_500hz_compare

You can see that the current default setup with 500Hz should have a delay of ~3.1ms. So @fujin results show even more improvement when the LPF is disabled, but i don't know how exact the autotune measurements are, or if all of my math is correct.

You can also see that with higher cuttoff frequencies of the LPF, the delay gets smaller.
And for having a similar cutoff Frequency but on a lower MPU/Control Rate you get also lower delay. 

In principal it should be equal if its a IIR or FIR, the delay depends on the order of the filter, the cutoff frequency and the sample/nyquist frequency. Or better a ratio between the cutoff and the nyquist frequency.

With this PR it should be easier in future to tune the LPF to the requirements of the Quad (e.g. how much noise is there).


I am very interested what will be the best setup in future.
When a additional LPF is needed, maybe a compromise for the Control rate must be found. 
Keep it as high as necessary for a good system/control behavior, but as low as possible for a lower delay at a given LPF cutoff frequency.

Trex4Git:
So when the formula from the Wiki is used, that i also want to use, the results look different:

taulabs_1tapiir_500hz_compare_new

At least for low frequencies there seems to be the same delay, when the parameters of the filter are calculated based on the formula for the cutoff frequency.
And i have to mention that when using that formula the magnitude response did not look so similar as i hoped, e.g. the -3dB frequency is quiet different for the different sample rates.

Trex4Git:
The math from the wiki that i used seems to be a approximation, so with here an update to the original math, but more looptime independent and based on Fcut:

gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroFcut * dT);   
taulabs_1tapiir_500hz_compare_new2

For small cut off frequencies it seems still not perfect, as you can see e.g. for 100Hz desired cutoff, 90Hz cutoff @ 500Hz  is closer to the others. I think its still a aproximation.
I found also another formula that gives better results, but is more complex to calculate:
GyroAlpha = 2 - cos(2 * Fcut * dT) - sqrt(cos(2 * Fcut * dT)^2 - 4 * cos(2 * Fcut * dT) + 3);

Trex4Git:
Pushed to latest changes to the PR. Description in the commit message.

A filter for the dT is included, but i am not sure if its really necessary for the Gyro LPF, as all the other things in stabilization run with the unfiltered dT, even the LPF for the derivative term in PID calculations (https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Stabilization/stabilization.c#L322 & https://github.com/TauLabs/TauLabs/blob/f81a193f00f62c601761c8e653d829b3bca49d6f/flight/Libraries/math/pid.c#L154-L155). 
That is in principal the same filter (EMA, exponential moving average), but uses the more simplified math that i also had in the first implementation. But as the cutoff frequency for the derivative term is relative low and is set from autotune, the difference should be very small.

I have also seen that the complementary filter for attitude uses also the same filter (&math with fake dT) for accel and gyro data, as the gyro LPF in stabilization before this PR.
https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Attitude/attitude.c#L1486-L1494 & 
https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Attitude/attitude.c#L1486-L1494 & 
https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Attitude/attitude.c#L558-L567


So I think we should apply the changes here also to that filter, but I am not sure if it should be done in this PR or a extra one.

And if the dT_filtered is necessary  I am not sure if its better to have 2 separate filter in stabilization and attitude, or one common in a different place. 
The 2 separate filter have the advantage that also the delay/jitter of the processing in the µC and the RTOS overhead is considered.
One common filter could be placed in the sensor task, that "collects" the data from the sesors and does some processing before the later tasks like stabilization or attitude are called.

According to this https://github.com/TauLabs/TauLabs/issues/168, the filtered dT can than the saved in the pios sensor driver like PIOS_SENSOR_SetDtFiltered(), and when other tasks need that information:  PIOS_SENSOR_GetDtFiltered()

flight/Modules/Stabilization/stabilization.c
```diff
@@ -941,13 +951,19 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
...
+			gyro_alpha_temp = 0;
+			StabilizationSettingsGyroCutoffSet(&gyro_alpha_temp); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
+		}
 		else
-			gyro_alpha = expf(-fakeDt  / settings.GyroTau);
+			gyro_alpha_temp = 2.0f * (float)(M_PI) * settings.GyroCutoff;
+
 
 		// Compute time constant for vbar decay term based on a tau
 		vbar_decay = expf(-fakeDt / settings.VbarTau);
```
mlyle: How about we do this one too?  We can ignore the ones in stabilization for now-- the cutoff frequencies are lower and it's after other filtering, so the effects of jitter are lessened some.

shared/uavobjectdefinition/stabilizationsettings.xml
```diff
@@ -30,7 +30,7 @@
 	<field name="VbarPiroComp" units="" type="enum" elements="1" options="FALSE,TRUE" defaultvalue="FALSE"/>
 	<field name="VbarMaxAngle" units="deg" type="uint8" elements="1" defaultvalue="10"/>
 
-	<field name="GyroTau" units="" type="float" elements="1" defaultvalue="0.005"/>
+	<field name="GyroCutoff" units="Hz" type="float" elements="1" defaultvalue="40.6"/>
```
mlyle: What are your assumptions on this value?  This configured to be the same for how many Hz of loop rate?

I think we should pick a compromise value that is higher than the default with the lower loop rates, but lower than the default with the higher rates that some people run.  I also would prefer a round number, like 45.

Trex4Git: In cleanflight with the same filter 60Hz seems to be the favorite.
But as i have not much flight experience, and my quad can't fly at the moment, i wanted to start with a conservative default value.

But when there are recommendations for better default values, i like to change this.

flight/Modules/Stabilization/stabilization.c
Trex4Git: I don't really understand what you mean.

I do there a pre calculation of the LPF coefficient that only changes on UAVObject changes, so not everything has to be calculated every run of the looptime.

Or do you mean i should also change the virtual bar coefficient? As i have absolutely no experience with this flightmode, i don't wanted to touch this, if there is no need.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -219,6 +220,15 @@ static void stabilizationTask(void* parameters)
...
 		timeval = PIOS_DELAY_GetRaw();
 		
+		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
+		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
+		dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
+
+		if (gyro_alpha_temp  < 0.0001f) // not trusting this to resolve to 0
+			gyro_alpha = 0;   // Gyro LPF should be disabled
+		else
+			gyro_alpha = expf(-gyro_alpha_temp * dT_filtered);
```
mlyle: One thing I need to test before this merges is to make sure that this isn't too expensive for F1 targets.

shared/uavobjectdefinition/stabilizationsettings.xml
mlyle: OK.  Maybe we make it 45, then.  This is going to need flight test anyways.

mlyle:
So this is definitely planned for this release.  It has test dependencies though that will likely make it take a little while to get merged.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -219,6 +220,15 @@ static void stabilizationTask(void* parameters)
 		dT = PIOS_DELAY_DiffuS(timeval) * 1.0e-6f;
 		timeval = PIOS_DELAY_GetRaw();
 		
+		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
+		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
+		dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
```
BrainFPV: This is a good idea. I think it would make sense to also pass this to `pid_apply_setpoint` instead of `dT`.

flight/Modules/Stabilization/stabilization.c
BrainFPV: Maybe it could be calculated once after the FC has been running for a few seconds. Re-calculating over and over again is kind of silly as the gyro sampling rate is constant (and so `dT_filtered` will converge to a constant value quite quickly).

flight/Modules/Stabilization/stabilization.c
```diff
@@ -941,13 +951,19 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
...
+		//fakeDt now only used for vbar_decay, for gyro filter now a real, but filtered, dT is used
 		const float fakeDt = 0.0025f;
-		if(settings.GyroTau < 0.0001f)
-			gyro_alpha = 0;   // not trusting this to resolve to 0
+
+		if (settings.GyroCutoff < 0.0001f)
+			gyro_alpha_temp = 0;   // not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
+		else if ( (dT > 0) && (settings.GyroCutoff > (0.45f / dT)) ) { // disable LPF when GyroCutoff > 90% of F_nyquist (= Fs/2 = 1/(2*dT)); using here not the filtered dT, because this is at the start to low
+			gyro_alpha_temp = 0;
+			StabilizationSettingsGyroCutoffSet(&gyro_alpha_temp); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
```
BrainFPV: Using a settings UAVO for this is ugly..

Trex4Git: I also thought about this. 
But it will cause just 1 callback to all the registered functions, and no endless loop is triggered.
 
And as the settings should not be changed too often, I thought it is nice to get feedback in GCS, when the gyro LPF is disabled because of to high cutoff frequency.

flight/Modules/Stabilization/stabilization.c
Trex4Git: I already noticed that all other functions in stabilization run with the unfiltered dT. 
So I am not sure if we really need filtered dT for the Gyro LPF, or if all other function should also use dT_filtered.

But as I have not much experiences with flying, changing also the PID control to this would need good testings, an I am not sure if this should be in the scope of this PR.

The implementation would be very easy, but testing might be critical.

flight/Modules/Stabilization/stabilization.c
Trex4Git: Thats a good idea.

Trex4Git:
A update with changed GyroCutoff default value.

Also the CPU load should be reduced as dT_filtered and gyro_alpha are not calculated permanently anymore.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -108,6 +108,9 @@ static StabilizationSettingsData settings;
 static TrimAnglesData trimAngles;
 static struct pios_queue *queue;
 float gyro_alpha = 0;
+static float gyro_filter_enabled = 0;
```
BrainFPV: why not use `bool` for this?

Trex4Git: Because when 0 it should indicate that LPF is disabled, 1 should indicate enabled, and 2 should indicate enabled, but Cutoff has changed, so recalculation of gyro_alpha is needed.
For that uint8_t should also be possible.
But as I want to inform GCS when Gyro LPF is disabled (because of to high cutoff frequency) i write to the UAVObject (what you also commented as critical, and maybe need further discussion), and therefore i need a float pointer. 
Maybe that would work also with casting from a uint8_t, have not tried that yet.

BrainFPV: An enum would be an elegant solution too.

Trex4Git:
@BrainFPV yes enum is nice idea, and it should be better to understand the source code. 
Also changed now the gyro_filter_enabled variable from float to uint8_t, that will save some bits

I have implemented it already, but want wait a little bit if other changes are requested, to combine them in one commit.

Trex4Git:
Pushed the latest changes to the PR.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -219,6 +228,25 @@ static void stabilizationTask(void* parameters)
 		dT = PIOS_DELAY_DiffuS(timeval) * 1.0e-6f;
 		timeval = PIOS_DELAY_GetRaw();
 		
+		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
+		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
+		// run it only at the beginning for the first samples, to reduce CPU load, and the value should converge to a constant value
+		if (dT_filter_iteration-- > 2) {
```
mlyle: Why the weird bounds here?

flight/Modules/Stabilization/stabilization.c
```diff
@@ -219,6 +228,25 @@ static void stabilizationTask(void* parameters)
...
+
+			if (gyro_filter_enabled  == GYRO_LPF_DISABLED)
+				gyro_alpha = 0;   // Gyro LPF should be disabled
+			else
+				gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
+		}
+
+
+		// new calculation of gyro_alpha when cutoff frequency has changed, and dt_filtered is already calculated
+		if ( (gyro_filter_enabled  == GYRO_LPF_ENABLED_UPDATE) && (dT_filter_iteration == 1) ) {
```
mlyle: I suggest if ( ( gyro_filter_enable == GYRO_LPF_ENABLED_UPDATE) || (dT_filter_iteration > 0)).    Then the computation of gyro_alpha can be removed from above.  Since you have a flag for GYRO_LPF_ENABLED, the conditional to set the gyro_alpha to 0 is not needed; the code doesn't look at it in the DISABLED state.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -941,13 +978,22 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
...
 		// based on a time (in ms) rather than a fixed multiplier.  The error between
-		// update rates on OP (~300 Hz) and CC (~475 Hz) is negligible for this
-		// calculation
+		// update rates on OP (~300 Hz) and CC (~475 Hz) is negligible for this calculation
+		//fakeDt now only used for vbar_decay, for gyro filter now a real, but filtered, dT is used
 		const float fakeDt = 0.0025f;
-		if(settings.GyroTau < 0.0001f)
-			gyro_alpha = 0;   // not trusting this to resolve to 0
+
+		if (settings.GyroCutoff < 0.0001f) {	// not trusting this to resolve to 0; disable LPF when GyroCutoff = 0
```
mlyle: Because it's now a frequency (reciprocal of time constant) the first part of the comment should be removed

Do we want to have a new minimum value that makes sense in these terms?  I can't picture why you'd want a lowpass frequency of less than, say, 5Hz.

flight/Modules/Stabilization/stabilization.c
Trex4Git: That is still a bug, but was also a bug in my brain.
Wanted to keep the variable = 1, but with eyry call the variable gets decimated.  

So needs a bugfix.

```
if (dT_filter_iteration > 1) {
			dT_filter_iteration--;
```

flight/Modules/Stabilization/stabilization.c
Trex4Git: good catch, thanks; was a relict from early implementation

```
if (dT_filter_iteration >= 1) {
			dT_filter_iteration--;
			dT_filtered = 0.01f * dT + (1.0f - 0.01f) * dT_filtered;
		}


		// new calculation of gyro_alpha when cutoff frequency has changed, and dt_filtered is already calculated
		if ( (gyro_filter_enabled  == GYRO_LPF_ENABLED_UPDATE) || (dT_filter_iteration > 0) ) {
			gyro_alpha = expf(-2.0f * (float)(M_PI) * settings.GyroCutoff * dT_filtered);
			gyro_filter_enabled = GYRO_LPF_ENABLED;
		}
```

flight/Modules/Stabilization/stabilization.c
Trex4Git: I thought the user has to decide what he wants.  
I also think low values are not very useful, as the delay gets very large, and we filter useful information I was not sure where a border for disabling shout be.

I implemented only 0 or very low cutoff frequencies disabled the LPF.
But when you want it to be at e.g. 5 Hz, i can to do it.

Trex4Git:
Replied to the code comments.
Want to wait with a commit at least when is clear what the new minimum for cutoff should be.


What do you think about moving the deltaT calculation from stabilization task to a implementation like mentioned there https://github.com/TauLabs/TauLabs/issues/168 that gets triggered from sensor task.


The data seems to flow from MPU task, to the sensor task, and than attitude and stabilization task gets triggered.
So when the time calculation is done in sensor task we don't have information about some delay/jitter in running of stabilization or attitude tasks, but we would have a central point for this calculation, and the data could be used for stabilization task and attitude task.




I also noted maybe a strange priority of that tasks.
The MPU task and stabilization task have highest priority, but sensor task and attitude task have only high priority.
But the data is send from MPU task to sensor task with a queue. Than in sensor task some data processing is done, and the final data, e.g. gyro are stored in a UAVObject. The stabilization task gets then "triggered" by the gyro UAVObject update event.
So when we want a fast processing with low delay/jitter, the sensor task also needs highest priority, i think. 
Also some flightmodes in stabilization task need data from attitude task, and the accel data update is extra done before the gyro data update [https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Sensors/sensors.c#L207-L211). Maybe the attitude task should also be highest priority.
But that is not really the scope of this PR, and maybe I should open an issue for that.

Trex4Git:
@mlyle regarding the min value for the gyro cutoff frequency.
I also think there is no need for really low cutoff frequencies as written already above.

But maybe there are some use cases that i don't think about at the moment.
So with the current implementation you can disable it with setting to 0, otherwise the user can choose what he wants.
Otherwise i think maybe 1.0Hz should be the lower limit, so that value of 0.xxx would disable it.

mlyle:
OK-- whatever value you prefer.

I had said 5.0Hz mostly because of how long the delay gets-- I'm not sure we could control even a very large quadcopter with that much delay.  But thinking about it, it's possible a value like 2.5Hz could still be useful for ground vehicles or boats.

Trex4Git:
So would you prefer 0.0001 or 1.0 as limit?
I think 0.0001 has no disadvantage other that the user could do useless stuff.

mlyle:
1.0 please

Trex4Git:
Pushed the latest changes.

Trex4Git:
Oh, i don't expected such a quick merge, but thanks for merging.

As my quad can't fly at the moment, I think it would be nice if someone else can do some flight test.

mlyle:
I'm confident it's functionally the same to the existing code, so I'm OK with picking up the risk.

The slight change in cutoff frequency that it does have should be nice for most people, too.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -100,6 +100,12 @@ enum {
 	PID_MAX
 };
 
+enum {
```
peabody124: make this a named enum and use it for the type of gyro_filter_enabled

flight/Modules/Stabilization/stabilization.c
```diff
@@ -108,6 +114,9 @@ static StabilizationSettingsData settings;
 static TrimAnglesData trimAngles;
 static struct pios_queue *queue;
 float gyro_alpha = 0;
+static uint8_t gyro_filter_enabled = 0;
```
peabody124: ```static enum gyro_lpf_state gyro_filter_enabled```

flight/Modules/Stabilization/stabilization.c
```diff
@@ -941,13 +976,25 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
...
+		//      we should assert a alarm here when GyroCutoff > 90% of F_nyquist and LPF gets disabled
+
+		if (settings.GyroCutoff < 1.0f) {	// disable LPF when GyroCutoff < 1.0 Hz
+			gyro_filter_enabled = GYRO_LPF_DISABLED;
+			settings.GyroCutoff = 0;
+		}
+		else if ( (dT > 0) && (settings.GyroCutoff > (0.45f / dT)) ) {	// disable LPF when GyroCutoff > 90% of F_nyquist (= Fs/2 = 1/(2*dT)); using here not the filtered dT, because this is at the start to low
+			gyro_filter_enabled = GYRO_LPF_DISABLED;
+			settings.GyroCutoff = 0;
+			StabilizationSettingsGyroCutoffSet(&settings.GyroCutoff); // Set GyroCutoff to 0 to signalize in GCS, that it is disabled
```
peabody124: whoa ABSOLUTELY NOT! the FC should not be changing settings like. If you want to communicate it create a data object to change.

flight/Modules/Stabilization/stabilization.c
```diff
@@ -941,13 +976,25 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
 		// The dT has some jitter iteration to iteration that we don't want to
 		// make thie result unpredictable.  Still, it's nicer to specify the constant
 		// based on a time (in ms) rather than a fixed multiplier.  The error between
-		// update rates on OP (~300 Hz) and CC (~475 Hz) is negligible for this
-		// calculation
+		// update rates on OP (~300 Hz) and CC (~475 Hz) is negligible for this calculation
+		//fakeDt now only used for vbar_decay, for gyro filter now a real, but filtered, dT is used
 		const float fakeDt = 0.0025f;
```
peabody124: you can just use the dt_filtered for vbar decay too

flight/Modules/Stabilization/stabilization.c
```diff
@@ -941,13 +976,25 @@ static void SettingsUpdatedCb(UAVObjEvent * ev)
...
-			gyro_alpha = 0;   // not trusting this to resolve to 0
+
+		//TODO: if the infrastructure for alarms will be more advanced in the future,
+		//      we should assert a alarm here when GyroCutoff > 90% of F_nyquist and LPF gets disabled
+
+		if (settings.GyroCutoff < 1.0f) {	// disable LPF when GyroCutoff < 1.0 Hz
+			gyro_filter_enabled = GYRO_LPF_DISABLED;
+			settings.GyroCutoff = 0;
+		}
+		else if ( (dT > 0) && (settings.GyroCutoff > (0.45f / dT)) ) {	// disable LPF when GyroCutoff > 90% of F_nyquist (= Fs/2 = 1/(2*dT)); using here not the filtered dT, because this is at the start to low
```
peabody124: this is risky logic. you are creating a dependency between when the settings method is called and what dT (which might be jittery) is

flight/Modules/Stabilization/stabilization.c
peabody124: sorry too strong phrasing. we generally don't like the practice of altering the settings from the FC. i don't prefer this but since it only happens for invalid settings it's not a huge risk in practice.

a better approach would have been setting a warning

flight/Modules/Stabilization/stabilization.c
Trex4Git: will be done

flight/Modules/Stabilization/stabilization.c
Trex4Git: As i don't have any experience with this flight-mode i didn't want to touch it, till someone thinks it should be changed.
But i will do that.

Should also all text be removed, or just modified to have some history documentation in there.

flight/Modules/Stabilization/stabilization.c
Trex4Git: As requested the filtered dT is not calculated continuously, but only for the first 1000 "runs" of the task.

So i can check here if the calculations already has finished, otherwise this check won't be done. Additional when the filtered dT calculation is finished, there will also be the check if cutoff frequency is to high.

Is this OK?

mlyle: What's the reason for using dT in this line instead of the filtered dT?

It looks like if dT ever resolves to 0 we'll do the update.

flight/Modules/Stabilization/stabilization.c
mlyle: We should decide how we deal with invalid configuration in general and set a pattern.  Overriding setting objects to put in the actual employed setting like this code would be one way.  Another way would be to build some kind of infrastructure to communicate it back.  Doing alarms on a case-by-case basis and building lots of enums is probably not a sane one.

Grepping the code, it appears there's several pieces of older code that follow the pattern of updating setting objects to communicate things.

flight/Modules/Stabilization/stabilization.c
Trex4Git: The intention was that filtered dT is at the first runs not converged yet, and  when this settings update functions is called before it has converged we get unwanted results.

So dT might have some jitter, but I thought it should be not to big for this check. I am also not checking if cutoff > f_nyquist, but put in some margin so cutoff >  90% * f_nyquist.

So the intention of this implementation is not to disable the LPF if cutoff  is > f_nyquist,
but disable it when its large and close to f_nyquist. 

But that is not exactly defined, so basing it on dt filtered, with some checks if dt filtered is already converged / calculations finished (like describe above) should be a better solution.

flight/Modules/Stabilization/stabilization.c
Trex4Git: Sorry , i was not aware we generally don't like the practice of altering the settings from the FC in taulabs.
I thought i had already seen functionality like that somewhere in the code (but maybe it was OP or something else), and that was where i got the idea of doing it like that. But I am not 100% sure about it.

Is was aware that when you set again a UAVObject from a settings update Callback, you should avoid triggering endless loops. And I think that should not happen with this implementation.

It seems now like a principal decision is needed how that should be done. And I will than implement the preferred solution.

One other solution might be just removing the feedback about disabling the LPF.

Are there some examples how the alarm/warning system should be used?
Can just variables be added to the alarm UAVObject (/shared/uavobjectdefinition/systemalarms.xml) and than set if a problem is detected?
There seems to be the possibility to set access level (read/write) separate for gcs and flight for the complete UAVObject. Is there also the possibility to set the access level for single elements?

mlyle:
Reverting for now.  According to A/B test (next vs. next with 1765 reverted) by @pug398 this causes regression on Quanton hover.  Changing the cutoff to 55Hz is said not to help.

Need to figure out root cause.

peabody124:
Yeah I don't like the whole approach to inferring dT - I think we should probably make PIOS_Sensor expose the gyro rate (which requires each driver supporting that) and doing it explicitly/deterministicly. For sure digging into the calculated gyro_alpha to see if that is coming out weirdly.

mlyle:
I like inferring dT because it's slightly more sane if we are dropping samples (still not very good.. just less bad).

peabody124:
we shouldn't ever be dropping samples. that would be a pretty big red flag. on the attitude estimation side it would also mess things up. You would also see that as a warning from the event system and you can look at the object ID that is being lost.

peabody124:
side note: we also do sample-by-sample dT in things like attitude. in this case that temporal jitter isn't going to actually be reflective of the times of the true sensor samples so is actually a source of noise rather than being more "true". It would also be good to replace that with a true dt.

that being said, having a bit of monitoring to make sure that those cycle-to-cycle dt are reasonable would be sensible, but if we aren't seeing missed samples in the event queue probably isn't necessary because that means we consumed the data before new was available.

mlyle:
Yah, we shouldn't ever be dropping samples, and it'll result in mixing products and aliasing.  If we adjust and infer the real sample rate (configured 1000 / achieving 850) we can still integrate and do subsequent filters properly though.  (Note I'm not proposing weighting based on actual dT, but based on 1/avgRate, as almost all variance in dT is just noise)

I agree with changing attitude to use actual achieved dTs-- I believe that's in the above discussion.

mlyle:
One tricky alternative, but that's probably objectively best--

Sensors could send up the time of acquisition quantized to, say, double the configured sensor rate.  Then we could act on that sample-by-sample.  The clock on the sensor is +/- 2%, so 20 times a second @500Hz we would have a sample weighted higher or lower because of clock skew.  Maybe it's even possible to fix that with some thought (a drift accumulator or something like that).

jihlein:
Haven't had time to keep up with all this, but how about using one of the unused hardware timers to keep track of the dt between samples??

Date: Wed, 19 Aug 2015 11:36:28 -0700

Trex4Git:
I think we don't have problems with getting dT. This is already based on hardware timers (some debug cpu cycle counter).

It should be discussed if we need a filtered dT or not (In my first version i used unfiltered dT, added filtering, as it was requested).
And if a central dT (e.g. in sensors task, as describe here: https://github.com/TauLabs/TauLabs/issues/168#issuecomment-131009936) should be calculated, or individual in stabilization and attitude, or if we should just get the gyro rate, without possibles delays in sensor task, or stabilization/attitude task.

Trex4Git:
I want to track down the problems. 

What is the best way to debug on taulabs ?
I have not setup hardware debugging yet, but with breakpoints I think there may be problems with the rtos, timeouts, watchdogs,...

Is there a simple way like a a debug console where I can use e.g printf to a serial port?
Or would it better to create temporary UAVObject variables where i can put debug output like dT, filterd dT gyro alpha and monitoring them on GCS?

mlyle:
Yes, there is a debug console on most targets.  The console can be assigned to USB VCP or a serial port, and then you can use DEBUG_MSG to send stuff to the console.  

```
//#define DEBUG_THIS_FILE

#if defined(PIOS_INCLUDE_DEBUG_CONSOLE) && defined(DEBUG_THIS_FILE)
#define DEBUG_MSG(format, ...) PIOS_COM_SendFormattedString(PIOS_COM_DEBUG, format, ## __VA_ARGS__)
#else
#define DEBUG_MSG(format, ...)
#endif
```

Trex4Git:
I maybe found the bug. Gyro LPF was most times disabled because of a logic error i think.
https://github.com/Trex4Git/TauLabs/blob/015791f327a3508b27a94aadaff4a69ccdd17be1/flight/Modules/Stabilization/stabilization.c#L245-L246

But still want to check if the rest is fine, and debug some variables.
Don't have much freetime at the moment, there was an accident in the family.

When that will be fixed, and peabodys requests are also implemented, should I create than a new PR?

Still we have to decide how the alarm/warning stuff should be handled.

The central dT filtering / calculation should be done i a different PR, I think.

mlyle:
Sorry to hear about the accident.  I hope everything is OK.

I may try and clean this one up myself for this release.  I don't think the other PR is going to make it unfortunately.

Trex4Git:
Thanks, in principal it seems OK now. But I have to help at some points.

On Sunday i should have more time, and can do the bugfixing, debugging and start with implementing the requested changes.
Already started with some of this.

PR #1771, opened 2015-08-05
RFM22b: use PIOS_Thread_Sleep instead of Delay

### LINKED ISSUE #1757: SyncPWM: Jitter Issue at High MPU Sample Rates
There seems to be a pretty serious issue with jitter when running SyncPWM at high rates (primarily 1 kHz but also evident at 500 Hz). I've done a little testing with Sparky2, MPU9250 @ 1 kHz sample rate, 184 Hz DLPF. This is the type of thing I'm seeing with output @ 250 us: Capture
This appeared to be happening approximately every 150 ms which is the RFM22b supervisor task rate (I had RFM22b radio disabled but the supervisor task still runs). The raw data I captured is available in [CSV](http://1drv.ms/1KyEUPS) or [MATLAB](http://1drv.ms/1KyExVC) format.

I'm now not so sure that the cause is RFM22b, further investigation required. I adjusted PIOS_RFM22B_SUPERVISOR_TIMEOUT to 230 ms and captured this: Capture
The times on the right side are the time between "glitches" like the one shown. The seem to be quite regular.

peabody124:
This avoids a problem where the modem resetting (when disabled)
was tying up in a high priority thread and blocking other system
processes.

Fixes #1757

### DIFF
--- a/flight/PiOS/Common/pios_rfm22b.c
+++ b/flight/PiOS/Common/pios_rfm22b.c
@@ -1362,7 +1362,7 @@ static enum pios_radio_event rfm22_init(struct pios_rfm22b_dev *rfm22b_dev)
 			break;
 		}
 		// Wait 1ms if not.
-		PIOS_DELAY_WaitmS(1);
+		PIOS_Thread_Sleep(1);
 	}
 
 	// ****************
@@ -1407,6 +1407,9 @@ static enum pios_radio_event rfm22_init(struct pios_rfm22b_dev *rfm22b_dev)
 		// incorrect RF module version
 		return RADIO_EVENT_FATAL_ERROR;
 	}
+
+	PIOS_Thread_Sleep(1);
+
 	// calibrate our RF module to be exactly on frequency .. different for every module
 	rfm22_write(rfm22b_dev, RFM22_xtal_osc_load_cap, OSC_LOAD_CAP);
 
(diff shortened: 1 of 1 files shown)


tracernz:
Good stuff. Any reason for all the additional sleeps added?

mlyle:
Also, we shouldn't continuously reset the modem when it's disabled. :P

mlyle:
@tracernz he just wants to yield more so that other stuff has more of a chance to run (at the cost of the reset taking longer).

peabody124:
Yeah I want to completely redo what we do with the modem when disabled. We only need to query it for the serial number and presence to allow the bind wizard to work. We shouldn't be starting it, but this is a hot fix for the main issue and should be there too (modem in init shouldn't be spinning in PIOS_DELAY statements).

mlyle:
Has this been tested and verified to fix the problem?

tracernz:
I should hopefully get a chance today. I'll report back when done.

Sent from Outlook




On Wed, Aug 5, 2015 at 10:20 AM -0700, "Michael Lyle" <notifications@github.com> wrote:
Has this been tested and verified to fix the problem?

---

peabody124:
Yeah I scoped it, reproduced the problem and verified the fix.

mlyle:
Ok..  Let's leave issue open for correct reinit fix.

PR #1773, opened 2015-08-05
pipxtreme: reindent pios_board

mlyle:


mlyle:
Voluntarily committing myself.  Needed for rebase of #1759 and can't just be cherry-pick'd without conflict.

peabody124:
Can you paste the gnu indent command you used to do this? that makes it easier for others to replicate this and merge conflicting changes at a later point. Otherwise things like this get clobbered over eventually.

mlyle:
I used == in vim with filetype indent plugin on, tabstop=8, shiftwidth=8.  It appears to follow our style.

Issue #1753, opened 2015-07-28
Pin out for CC3D PPM


MrSatoV:
Hey guys, I'm pulling my hair out over here.  I feel like I'm missing something super simple, but so far nobody has the answer.  
I see in this post that CC3D is working with PPM:
tracernz commented 8 days ago
Since fixing the stack sizes (5b98d5a) I have test flown this on both Naze32 rev. 5 and CC3D. Flew with a PPM RX in acro and horizon mode with full throttle punchouts etc. on a 4S powered 250 without issue. I'm happy with this now. :+1: I have also now reviewed all of the code and am satisfied. There is plenty of work to do before the target is fully supported but this PR will get us flyable.

But no matter what pin I use, I get no TX input.  I've seen some workarounds, like using sbus, but it sounds like tracernz isn't doing any kind of workaround, and just plain using PPM through the recvr port.  I am aware of the timer issue with using pin3, but that's easy enough to fix, and all other FC firmwares I've tried have an alternate pin assigned to use a different timer.  In OP it's 4 or 8 (8 for OneShot, which is what I want, and use with OP).  Of course I tried both, but neither give any results.  

What pin should I be using?  I tried them all, but at least if I know I have the right pin I can move on in the troubleshooting process.

tracernz:
I was using pin 3 with outputs set to 1 MHz (1000-2000 us pulse width), not "OneShot". I think using PPM with 12 MHz outputs on CC3D is broken, see #1754 .

MrSatoV:
Ahhh.  Thanks.  Sounds like it might get fixed in the next release.

peabody124:
See #1754 for relevant bug I made based on this.

Of course try this with the props off! But I think you should still be able to use the sync part of one shot (which is the important bit). Set them to OneShot but then change from 12MHz to 1MHz in the output configuration. You'll only have 125 "levels" of resolution but TBH I'm pretty confident you won't be able to tell the difference.

MrSatoV:
As soon as I get done with helping my daughter with my homework I'll give it a whirl and get back to you!

mlyle:
@MrSatoV did requesting 1MHz make it work?

soend:
Tried this with D4R-II connected to pin 3 in receiver port. Changed resolution in outputs to 1MHz but it didnt fix it.

mlyle:
Fixed by PR #1775.  It is now possible to use pin8 for PPM on CC3D, and this is now the default in GCS.

PR #1775, opened 2015-08-07
CC: support PPM on pin8, fix oneshot

mlyle:
Fixes issue #1753 .

Also a minor packaging fix from target names getting desynched in makefiles, fixes windows packaging.

~~Test binaries at http://jar.lyle.org/~mlyle/TauLabs-20150807-efb4c19b-install.exe~~

Revised test binaries at http://jar.lyle.org/~mlyle/TauLabs-20150808-e2a092cd-install.exe


mlyle:
Two known issues:

1. Doesn't arm ESCs on boot when in PIN8 settings--- we are obviously relying on some side effect from PPM configuration.  So this needs work/analysis.
2. Wiring diagram for PPM is bad in GCS.

SystemsGuy:
Works here on CC3D with PPM+8.  Connection diagram needs to be updated.  Flys + autotunes fine with FastPWM, moving the F/C to a Oneshot system now.

Oneshot arms and appears to work on the bench.

Memory Warning when Autotune is enabled.  It's probably worth updating the text description to reflect that as "normal".

soend:
Tested this on bench with CC3D and D4R-II connected to PIN 8.
In output configuration Update rate: SyncPWM and Resolution: 12Mhz.

Works fine.

flight/PiOS/STM32F10x/pios_tim.c
```diff
@@ -137,7 +137,7 @@ int32_t PIOS_TIM_InitChannels(uintptr_t * tim_id, const struct pios_tim_channel
 			RCC_APB2PeriphClockCmd(RCC_APB2Periph_GPIOB, ENABLE);
 			break;
 		case (uint32_t) GPIOC:
-			RCC_APB2PeriphClockCmd(RCC_APB2Periph_GPIOB, ENABLE);
+			RCC_APB2PeriphClockCmd(RCC_APB2Periph_GPIOC, ENABLE);
```
peabody124: good catch

flight/targets/coptercontrol/fw/pios_board.c
```diff
@@ -856,11 +856,13 @@ void PIOS_Board_Init(void) {
 #endif	/* PIOS_INCLUDE_PWM */
 		break;
 	case HWCOPTERCONTROL_RCVRPORT_PPM:
```
peabody124: want to call this PPMONPIN1 for consistency? maybe not worth breaking .uav files.

package/Makefile
```diff
@@ -45,7 +45,7 @@ CLEAN_FLIGHT := YES
 endif
 
 # Set up targets
-ALL_BOARDS        := coptercontrol pipxtreme revomini sparky sparky2 sparkybgc quanton colibri flyingf3 flyingf4 aq32
+ALL_BOARDS        := coptercontrol pipxtreme revomini sparky sparky2 sparkybgc quanton colibri flyingf3 flyingf4 aq32 naze32
```
peabody124: not good to mix unrelated things

peabody124:
LGTM if flight tested

package/Makefile
mlyle: Yah.. But I needed it anyway on the branch to get windows packaged builds.

Just as an aside packaging is broken everywhere-- on OSX by the QT5.5 change, by this.. That's something I'm going to look at systemically in a week or two.

mlyle:
Merging based on comment.

flight/targets/coptercontrol/fw/pios_board.c
mlyle: Yah I didn't want to break existing stuff.  It's not the default PPM anymore which is a little weird, but... meh.

Issue #1778, opened 2015-08-08
Rate Gyro Suppression (Acro+)


fujin:
Leading on some of the code I sketched out in response to https://github.com/TauLabs/TauLabs/issues/1776 (https://github.com/TauLabs/TauLabs/issues/1777) it should be possible to implement something akin to `Acro+`, that is Rate with Gyro Suppression/factor. "Insanity Factor". We may also wish to delete vbar, mwrate.

fujin:
acro insanity factor @ https://github.com/openpilot/OpenPilot/blob/29b9265531a808c8e337c4c9ff0091888bbd89da/flight/modules/Stabilization/innerloop.c#L290

fujin:
I'm fairly certain acro+ is just rate+gyro suppress, but would welcome enlightenment.

Trex4Git:
That was also my understanding when I had a look at the source code.
I had written in the acro thread on forums also a short impression of Acro+ http://forum.taulabs.org/viewtopic.php?f=12&t=704#p6000, as  I think it may work better then MWRate or VirtualFlybar, as the basic tuning is the same as normal rate mode, and autotune should work.

But as there was no feedback if its worth to implement this, i had no further look until now. I seemed to me that high gyro rates + expo was fine for the acro pilots .

fujin:
@Trex4Git Thanks for confirming. You have my feedback now: it is worth it to implement this, and if you see my previous works, it should be relatively simple to implement inside of stabilization_rate.

Trex4Git:
;) had no look at your previous works yet, but are they directly related to Acro+ functionality?
I mad no deep analysis of the OP code, but my first impression was, that it should be just a few new lines of code.

Most work might be implementing a new flight mode, or do you think that this should be integrated into normal rate mode?

mlyle:
So I really believe that acro+ and mwrate are equivalent.  And it's close to virtualbar.  I don't really like the idea of getting a third way to do the same thing.

Trex4Git:
I think virtualbar and mwrate are more similar to each other, then to acro+.
The first 2 try to control to keep the angle constant, and only the stick input with gyro suppression makes it rotating.
Acro+ is for lower stick inputs like normal rate mode, so already rotating with control, and adding than gyro suppression and mixing pure stick input to the output.
So for higher stick input all 3 should be similar, but for lower stick input acro+ might has a different behavior.

[...]

PR #1780, opened 2015-08-09
WIP Acro(plus) mode for rate gyro suppression

fujin:
Connects to TauLabs/taulabs#1778

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -301,13 +301,13 @@ static void stabilizationTask(void* parameters)
 		static uint8_t previous_mode[MAX_AXES] = {255,255,255};
 		bool error = false;
 
+
 		//Run the selected stabilization algorithm on each axis:
 		for(uint8_t i=0; i< MAX_AXES; i++)
 		{
 			// Check whether this axis mode needs to be reinitialized
 			bool reinit = (stabDesired.StabilizationMode[i] != previous_mode[i]);
 			previous_mode[i] = stabDesired.StabilizationMode[i];
-
 			// Apply the selected control law
 			switch(stabDesired.StabilizationMode[i])
 			{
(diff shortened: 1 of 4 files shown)


fujin:
This is definitely not working yet.

Trex4Git:
No deep analysis yet, just a few quick comments:
- Why a new bound function?
There is already bound functions (bound_min_max or bound_sym) that works with float, and are used often in the code: https://github.com/fujin/TauLabs/blob/acro-plus/flight/Libraries/math/misc_math.c#L35-L45
- You named it like in OP/LP. When its the same implementation  I think it is a good idea, so people don't get confused with different names for the same thing. And I think it would be nice to give some credits / comments to OP e.g. in the source code
- When the unit of the "AcroInsanityFactor" in stabilizationsettings.xml is percent it should better scale from 0-100 (the math in the function has to be changed than) and not 0-1. So when its it scaled from 0-1 a different unit might be better.
- is "STABILIZATIONDESIRED_STABILIZATIONMODE_ACRO" really the name you get from the uavobject-synthetics function ?
 - in "stabilizationdesired.h" should be a enum where you can see the used names

So here a quick sketch how I would have written it (a mix of a sketch that i already had prepared and your/the openpilot implementation). The functionality should be very similar to the Openpilot implementation, but the code is more similar to our Rate & MWrate flight modes. I also added the integral treatment from our MWRate flighmodet, but not sure if it is usable here, or a better adaption of the OP integral treatment is needed.
```
case STABILIZATIONDESIRED_STABILIZATIONMODE_ACRO:
	// this implementation is based on the Openpilot/Librepilot Acro+ flightmode 
    // and our existing rate & MWRate flightmodes
	if(reinit)
		pids[PID_RATE_ROLL + i].iAccumulator = 0;

	// The unscaled input (-1,1)
	float *raw_input = &stabDesired.Roll;

	// The factor for gyro suppression / mixing raw stick input into the output; scaled by raw stick input
	float factor = fabsf(raw_input[i]) * settings.AcroInsanityFactor; 

	// Store to rate desired variable for storing to UAVO
	rateDesiredAxis[i] = bound_sym(stabDesiredAxis[i], settings.ManualRate[i]);

	// Zero integral for aggressive maneuvers, like it is done for MWRate
	if ((i < 2 && fabsf(gyro_filtered[i]) > 150.0f) ||
	    (i == 0 && fabsf(raw_input[i]) > 0.2f)) {
		pids[PID_RATE_ROLL + i].iAccumulator = 0;
		pids[PID_RATE_ROLL + i].i = 0;
	}

	// Compute the inner loop
	actuatorDesiredAxis[i] = pid_apply_setpoint(&pids[PID_RATE_ROLL + i],  rateDesiredAxis[i],  gyro_filtered[i], dT);
	actuatorDesiredAxis[i] = bound_sym(actuatorDesiredAxis[i],1.0f);
	actuatorDesiredAxis[i] = factor * raw_input[i] + (1.0f - factor) * actuatorDesiredAxis[i];  

	break;
 ```

**Comment by fujin** [comment]
@Trex4Git thanks for your comments!!! I greatly appreciate it.

**Comment by fujin** [comment]
2015-08-10-224843_1920x1080_scrot

I implemented the feedback from @Trex4Git; I messed a bit with the 0~100 acro insanity factor stuff trying to get it right, I left it as 0.0~1.0 for now. The sim seems unusable; with acro+ selected incredibly amounts of deg/s are observed far beyond what I would have expected.

**Comment by fujin** [comment]
Closing this until it does something reasonable.

**Comment by Trex4Git** [comment]
I think the math for acro insanity factor in your implementation my quick sketch is for 0-1 range. When 0-100 is used in the UAVObject just:
```
float factor = fabsf(raw_input[i]) * settings.AcroInsanityFactor / 100; 
```

And when the sim maybe has no ideal physics, and quad has nearly unlimited power your results seems to be reasonable. 
So when e.g. the croInsanityFactor is set to 1.0 (0-1 range) that means at full stick input there is no control left, and the rotation of the quad should only be limited by its physics, like power of the quad, aerodynamics,...

fujin:
@Trex4Git thanks I'll update the code

PR #1781, opened 2015-08-09
[WIP] Flight: adding gyro down-sampling & additional filtering functionality

Trex4Git:
Adding different down-sampling methods & filter
Original intention was to give the Sparky2 with MPU9250 the possibility to disable the DLPF in the MPU but run at lower rates then 8kHz
Early discussion and development is in this issue: https://github.com/TauLabs/TauLabs/issues/1730
There are also performance measurements regarding the CPU load and theoretical investigations on the filter. 

Also its interesting if we can reduce the noise in gyro signal, or get less delay.
The way it currently works in Taulabs is that for higher Gyro rates, the gyro LPF has a wider passband and less delay, but more noise gets through the filter.
Especially the CIC filter can have a good anti-aliasing performance, and it will be interesting to see, if that is useful for us.
Because i am not sure how good the anti-aliasing filter is in the MPU, especially when the DLPF is disabled. There i posted some delays of the CIC filter and the DLPF in MPU.

So one interest of me is also to find a good setup regarding noise reduction, small delay and acceptable CPU load. 
With the settings I implemented at the moment we could do some tests with different setups. It would be nice if some people also have interest in testing and collecting data. Because my quad can't fly at the moment, and i have to fix some things. And i have also not much experience with flying.

Now the implementation of CIC filter is done, so I started this PR to see if there are principal concerns.
At the moment its only fully implemented for Quanton, but before enabling the config for other MPU6000 targets or porting the changes also to MP9250 and Sparky2 I want to be sure, if it is OK in principal.

The settings are in HwTargetXYZ UAVObject, but only implemented for Quanton at the moment (the other MPU6000  targets run with the shown default config).

hw_quanton_downsampling_settings_cut

The settings are grouped under GyroDownSampling:
- Drop: works at the lowest level in the MPU IRQ and just drops unused samples. Should be a replacement for the sample rate divider like in MPU6000, that don't work with MPU9250 when DLPF is disabled. Has the lowest CPU load, "but highest priority". So the rest of the system gets the output rate of this. Setting = 1 means use every sample, = 4 means down-sampling by 4 / use every 4th sample. Can be combined with LPF down-sampling
- LPF: works one level above the MPU IRQ, in the MPU task. Uses all samples that are output from the IRQ. So there are more information that we can work with, and that might have an advantage for the control system. But produces higher CPU load. The rest of the system gets  the output rate of this. Setting = 1 means use every sample and do no filte; = 4 means down-sampling by 4.
- LPF_Order: determines the order of the filter that is used for down-sampling. 
	- = 0: simple averaging of the collected samples (boxcar filtering?)
	- = 1-3: a CIC filter of this order is used
- LPF_DDelay: determines the differential delay if a CIC filter is used. 1 and 2 are supported at the moment

At the moment there are so many settings, as still test have to be done as described earlier.
Later they may be reduced, or stay in the  UAVObject, but in GCS there will be a default config for a "high speed sampling" mode.


flight/PiOS/Common/pios_mpu6000.c
```diff
@@ -51,6 +56,17 @@
 #endif
 #define MPU6000_SPI_LOW_SPEED               1000000
 
+// Private variables
```
mlyle: I suggest we conditionalize this on #ifndef SMALLF1 ... it's a lot of RAM and I doubt the little F1 targets can handle downsampling from high rates.

flight/Libraries/math/cic.c
```diff
@@ -0,0 +1,160 @@
+/**
+ ******************************************************************************
+ * @addtogroup TauLabsLibraries Tau Labs Libraries
+ * @{
+ * @addtogroup TauLabsMath Tau Labs math support libraries
+ * @{
+ *
+ * @file       pid.c
+ * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
```
mlyle: Not openpilot; 2015 for Tau

flight/Libraries/math/cic.c
```diff
@@ -0,0 +1,160 @@
...
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
+ * @brief      PID Control algorithms
```
mlyle: Appropriate brief, etc, too

flight/PiOS/Common/pios_mpu6000.c
```diff
@@ -596,138 +681,320 @@ static void PIOS_MPU6000_Task(void *parameters)
...
+				gyro_filter_xyz[Y1].integrateState0 += - (int16_t)(mpu6000_rec_buf[IDX2_GYRO_YOUT_H] << 8 | mpu6000_rec_buf[IDX2_GYRO_YOUT_L]);
+				gyro_filter_xyz[X1].integrateState0 += (int16_t)(mpu6000_rec_buf[IDX2_GYRO_XOUT_H] << 8 | mpu6000_rec_buf[IDX2_GYRO_XOUT_L]);
+				gyro_filter_xyz[Z1].integrateState0 += (int16_t)(mpu6000_rec_buf[IDX2_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX2_GYRO_ZOUT_L]);
+			}
 			break;
 		}
 
+		// 1st order CIC and boxcar (order = 0) integration already done in the conversion above
+		// for the higher order integration stages and all axes:
+	    for(uint8_t i=0; i< AXES_SIZE; i++) {
```
mlyle: Indentation.

flight/Libraries/math/cic.c
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+}
+
+/**
+ * @brief Get the decimated / down-sampled output of the CIC filter
+ * @param[in] filter_data Pointer to the CIC filter structure which stores the data / information
+ * @returns Output the computed gain of the filter
+ */
+int32_t cic_get_decimation_output(struct cic_filter_data *filter_data) {
```
mlyle: const

flight/Libraries/math/cic.c
```diff
@@ -0,0 +1,160 @@
...
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ */
+
+#include <math.h>
+#include "misc_math.h"
+#include "cic.h"
+
+
+//! Store the filter parameter
+static uint8_t cic_filter_order     = 1;
```
mlyle: everything else is stored in filter_data... can these be too?

Then A) we won't use the RAM when we're not using the decimaiton filter, and B) we can have multiple instances with different parameters.

flight/Libraries/math/cic.c
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * @brief Configure the common parameters of the CIC filter
+ * @param[in] cic_filter_order The order of the CIC filter
+ * @param[in] cic_filter_ddelay  The differential delay of the CIC filter
+ * @param[in] cic_filter_ddelay  The decimation factor of the CIC filter
+ */
+void cic_configure(uint8_t filter_order, uint8_t filter_ddelay,  uint8_t filter_decimation) {
+	cic_filter_order     = bound_min_max(filter_order, 0, 3);
+	cic_filter_ddelay    = bound_min_max(filter_ddelay, 1, 2);
+	cic_filter_decimation= bound_min_max(filter_decimation, 1, 255);
```
mlyle: We may want to verify that the amount of gain won't overflow int32's.

flight/PiOS/Common/pios_mpu6000.c
```diff
@@ -575,14 +614,60 @@ static void PIOS_MPU6000_Task(void *parameters)
...
+			IDX2_GYRO_ZOUT_L,
+			BUFFER2_SIZE,
+		};
+
+
+		static bool AllSamplingFlg = false; // flag that indicates that all data should be sampled
+		static bool DoOutputFlg    = false;   // flag that indicates that we have to compute an new output at the down sampled frequency
+
+		// buffer for sampling all data
+		uint8_t mpu6000_send_all_buf[BUFFER_SIZE] = { PIOS_MPU60X0_ACCEL_X_OUT_MSB | 0x80, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```
mlyle: I know these existed before, but const would be nice

mlyle:
This looks good.  I'd like to ultimately get rid of all the config and just have something automatically come into play when you select a high bandwidth mode, but the parameters are good while we're tuning.

It'd be nice to have this ported to other hardware.  Have you run this on hardware and made sure at high sampling rates CPU usage is acceptable and that the artificial horizon in GCS does the right thing?  Ideally you'd scope the PWM output and look for jitter, too.

flight/Libraries/math/cic.c
Trex4Git: I am editing with eclipse, and these parts were collapsed by default, and I did not noticed them.
But will be fixed

flight/Libraries/math/cic.c
Trex4Git: See comment above.

flight/PiOS/Common/pios_mpu6000.c
Trex4Git: I thought with the improvement of SPI Speed it could maybe work on CC3D when at first dropped down-sampling from 8kHz to a lower rate, and then filtering down-sampling to 1kHz or below might work.
But as you say it will consume some RAM, so its better to disable that.
I also got now more information about the DLPF implementation in the MPU, so when using max 1kHz output rate for control loop, the new CIC decimation might not have a big advantage.

flight/PiOS/Common/pios_mpu6000.c
Trex4Git: This looks fine in my editor, have to search where the problem might be.

flight/Libraries/math/cic.c
Trex4Git: You mean const pointers like
```
void cic_apply_higher_order_int_stage(const struct cic_filter_data *filter_data);
```

I have read that this might have speed improvements, so changed it for all the cic  function, and also at some
I have read that this might have speed improvements, so changed it for all the cic  function, and also at other place
```
void PIOS_MPU6000_SetGyroDownSamling(const uint8_t *gyro_downsampling)
```

flight/Libraries/math/cic.c
Trex4Git: I also thought about, but was not sure if we need a cic filter somewhere else.
So i implemented it like it is done for pid.c/pid.h
So we just have one CIC filter setup, and can't use different instances. But otherwise we would store the CIC setup for all 3 axes, so some overhead.

And for SmallF1 targets cic.h will not be included, so we don't lose RAM with that on these targets.

But if you like to have more modular, i will change this.

flight/Libraries/math/cic.c
Trex4Git: I have seen a formula for the width of variables:
bit_width_output = bit_width_input + filter_order * log2(differential_delay * decimation_factor)
                         = bit_width_input + M * log2(N*R) = bit_width_input + M * log2(D)
We have 16bit data, and with this implementation max. M=3,N=2 is possible. The output data is saved in 32bit variables.

So for this up to 20x down-sampling should work (16 + 3 * log2(2 * 20) = 31.97), and for the earlier thought 32kHz 24x down-samlping M=3,N=1 is possible (16+3 * log2(1 * 24) =29,7549).

I will add this as info as a comment also to the source code. Or should i implement a dynamic check?

flight/PiOS/Common/pios_mpu6000.c
Trex4Git: What should be const there?

flight/PiOS/Common/pios_mpu6000.c
mlyle: Can you share that information?  It looked like from our earlier analysis and the delay milliseconds in the table we'd be a little ahead on latency.

flight/Libraries/math/cic.c
mlyle: Yes.  Basically it's a promise that we won't change that structure, so it is OK for the code to freely be reorganized with respect to other things that look at that structure.

flight/Libraries/math/cic.c
mlyle: You never know when we'll might want to use this somewhere else, and there's no cost from putting it into the state structure.

flight/Libraries/math/cic.c
mlyle: > But what should we do if a dynamic check failed?

Right now you're bounding it already within what we consider "sane" parameters.  We can turn down a user's parameter to be within the gain limit.

Ideally we'd also assert an alarm, but the infrastructure for that is currently not so good (to communicate configuration errors).

flight/PiOS/Common/pios_mpu6000.c
mlyle: the mpu6000_send_all_buf.  It doesn't need to actually be stored on stack (can stay in flash) if it never changes.

Trex4Git:
Answered to all comments, but on some comments I need some feedback.
As written earlier I also think a easy config would be nice later:
> At the moment there are so many settings, as still test have to be done as described earlier.
Later they may be reduced, or stay in the UAVObject, but in GCS there will be a default config for a "high speed sampling" mode.
>

For other hardware that uses MPU6000 I already included the cic files in the makefile. To enable that on them just a few lines have to be done in each flight/targets/XYZ/fw/pios_board.c  and shared/uavobjectdefinition/hwXYZ.xml.
For MPU9250 and Sparky2 i also wnat to port it, before this gets merged. Other targets are at the moment not in the scope of this PR. Maybe later if there are some other SPI gyros. For I2C i think its not worth it.

That is also a reason why i called it [WIP], because not everything is 100% finished yet, but its already at a advanced progress and before doing more work i wanted to be sure that in principal it would be OK.

So had no time yet to do extensive testing on this, because i wanted to push this before the deadline for next release. Maybe there is enough time for final implementation & tests. 
So i had to rebase it partly manual to latest next, and did some refactoring and documentation.

Just cheked if it compiles for all MPU6000 targets, and works in principal on my Quanton.
Horizon and scopes in GCS looked fine so far. 
With 8kHz MPU rate, DLPF=256 (disabled), R=4 (2kHz output frequency), M=2 and N=2  CPU load is 57-58%.
Have not scoped the output yet with the last changes, but did it from time to time during the first implementations and different down-sampling approaches.  On first try with high CPU loads, it  looked not so fine: But also before my changes i noted some jitter with MPU rate of 1kHz and worse on 2 kHz.
After SPI clock enhancements it looked fine, but i will do further tests

flight/PiOS/Common/pios_mpu6000.c
Trex4Git: Had overseen that. 
I looked to often to the code, and that does not seemed like a critical part for me, so I think my brain filtered that.

Thanks for the hint.

flight/PiOS/Common/pios_mpu6000.c
Trex4Git: will post it in a new  comment at the bottom, so that it is not hidden when these diffs are outdated

Trex4Git:
Got now also some more information about the DLPF in the MPU, in a cooperation with cleanflight ;)
https://github.com/cleanflight/cleanflight/pull/1186#issuecomment-129844137
On register maps for older MPU is a more detailed description with plots.

It looks like a 2nd order LPF is always active, then the configurable DLPF adds a  additional 1st order LPF.
It can be also be assumed that for enabled DLPF there is a simple first order filtered (boxcar or CIC M=1,N=1,R=8)) decimation to 1kHz. 
Because there is q quicker decay around 1 kHz (the sampling frequency) and such 1st order decimation filter should have the first  zero in frequency response at the output sampling frequency.

So when we do 1kHz or below output frequency our CIC filtering might not have a big advantage against DLPF 188 setting of the MPU. But with DLPF 256 we can also have a anti-aliasing filter for sample rate above 1kHz for capable targets.

mlyle:
OK, now I understand what you mean-- we still think there's a delay advantage at DLPF=256.   Good- this agrees with what I saw from the filter delays in the data sheet table.

Trex4Git:
Regarding the "#ifndef SMALLF1"  stuff, to reduce RAM need for small targets:
I started with defining / not defining also other stuff that is than not needed anymore for F1 targets.
But code became really ugly, so i switched to do the F1 stuff in a different file in a smallf1 folder, similar as its done for attitude module.
Is this OK ?

And do you think all down-sampling functionality  should be removed for F1 targets, or a small lightweight implementation with just boxcar averaging may stay in there?
So maybe CC3D is capable of filtered down-sampling from e.g. 2 kHz to 666.67Hz, and that maybe has an advantage over the current implementation.

mlyle:
I see what you mean about it becoming a pain.

I'd rather not have separate files.  Can I take a crack at doing this cleanly and send you a PR?

Trex4Git:
I already started with the 2 files approach and seems to compile fine.

But if 1 file is preferred I can also do some refactoring.

But than i think maybe having 2 "complete" PIOS_MPU6000_Task implementation in there (one for SmallF1, one for others) might be the cleanest way.
As the original code was already a little bit blown, with the switch/case constructs for sensor orientation processing and definitions if accel should be used or not (#if defined(PIOS_MPU6000_ACCEL)).

When than also parts for smallf1 or others  will be added with defines in between the existing code, the code might get confusing, especially for other people then me (the author).

When doing separate implementation of the PIOS_MPU6000_Task we might get some source code duplication, but it should be easier to read and maintain. 
And as using preprocessor directives ("#ifndef SMALLF1" ) also the duplication is not in a way that we have equal functionality in different code parts, that are used together at the same time on a target.

There is only a little overhead of source code, and for each individual target only the corresponding part gets compiled and uses ROM space on the µC.

mlyle:
What I was thinking is we could always do the filtering in the sensor frame-- that is, don't reproject yet.

Then we could add a macro to consume a sample from the right place.  One version of the macro would consume from the filter on non-SMALLF1, and the other from the buffer on SMALLF1.  And we could use that macro in a reprojection/rotation step afterwards.

Trex4Git:
I not 100% sure how you would do it. 
It sound like doing the filtering (or not) before the sensor orientation stuff (https://github.com/TauLabs/TauLabs/blob/next/flight/PiOS/Common/pios_mpu6000.c#L602-L667).
But i think for the non-SMALLF1 at least the "2byto to 16bit" conversion 
```
(int16_t)(mpu6000_rec_buf[IDX_ACCEL_XOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_XOUT_L])
```
has to be done before applying the filter.

If you already have something implemented i can integrate that, sure.
Otherwise you could do it, when you have time, or I could also work on this.

Trex4Git:
Started already with some quick test changes for that. 
But there are still flags, counter,.. in there that are not needed for smallF1, so can be also "not defined", but that might also look not so nice.
But I will have a further look.

mlyle:
OK, I'm glad to do it if you want.  I'm pretty good at rearranging code to make it a little cleaner.  But I don't want to get in your way.

Trex4Git:
I am already implementing it. So when you don't like the way I have done it,  I am fine with merging your changes.
But if you already have done it, i would stop my work on this.

macros look like this:
```
/* Private macros */
#ifndef SMALLF1
#define RAW_GYRO_X  (cic_get_decimation_output(&gyro_filter_xyz[X1])) 
#define RAW_GYRO_Y  (cic_get_decimation_output(&gyro_filter_xyz[Y1])) 
#define RAW_GYRO_Z  (cic_get_decimation_output(&gyro_filter_xyz[Z1])) 
#else
#define RAW_GYRO_X  ((int16_t)(mpu6000_rec_buf[IDX_GYRO_XOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_XOUT_L])) 
#define RAW_GYRO_Y  ((int16_t)(mpu6000_rec_buf[IDX_GYRO_YOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_YOUT_L])) 
#define RAW_GYRO_Z  ((int16_t)(mpu6000_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_ZOUT_L])) 
#endif
```

mlyle:
I have not.  Looking forward to see what you put together :)

Trex4Git:
Tried to implement all requested changes.
Down-sampling functionality for smallF1 targets should be completely removed.

If simple boxcar averaging should be implemented for smallF1 targets (with less overhead compared to CIC) I can do it.

Trex4Git:
Since the latest commit with enhancements and refactoring something is broken.
There seems to be a strange "invers" coupling between accel x and y. But  i have not found the reason for this bug yet :( and don't had much free time on weekend.

Trex4Git:
Fix that, and squashed it to latest commit.

Trex4Git:
If some is interested to test it i have added the gyro down-sampling functionality to the other F4 targets with MPU6000 (AQ32, Colibri, RevoMini).

Also I am interested if the implementation is in principal OK after the last refactoring. Because if its fine, i want to adapt the changes to MPU9250/Sparky2

Trex4Git:
Did a first implementation of this for MPU9250 & Sparky2.

As I don't have a Sparky2 yet, i could only test if it compiles.
But not if it runs in principal, e.g. if sensor data or model view in GCS are OK.

Trex4Git:
Got some feedback that it seems to run fine on bench with Sparky 2.0. 
He want`s to do also some flight tests, when he has more time for it.

ducminhkt:
My FlyingF4 use MPU6050 sensor. How can I test this ?

Trex4Git:
It is only implemented for MPU6000 and MPU9250 at the moment, both have SPI interfaces.

MPU6050 has only I2C interface, and that is probably to slow for 8kHz readout. 
When reading the MPU data at a slower rate, you might get aliasing problems. So I implemented the high sampling only for targets capable of running the high sampling rate with a software anti-aliasing down-sampling. 

You can either try running higher MPU rates with maybe aliasing problems. But also you should test what MPU rate is possible with a acceptable CPU load.

Otherwise in cleanflight (as most targets have also only a I2C connection to the MPU, or the µC is to slow for 8kHz sampling) the recommendation for the Beta version from BorisB (which is more similar to Taulabs with gyro sync, software filtering,...)  is to set the DLPF in the MPU to 188 and the MPU ratre to 1kHz. Then you have a down-sampling with anit-aliasing in the MPU. But it adds some delay to the signal processing.

So you could test what is the max. MPU rate the flyingF4 is capable of with disbaled DLPF (0 / 256). When there are noise problems you could try the mentioned 188 DLPF and 1kHz MPU rate.

ducminhkt:
Thanks.
But, in my knowledge, max I2C speed is 400kHz. So I don't know why it too slow for 8kHz data read.

tracernz:
400 kHz is the clock rate. Each byte takes at least 10 clock cycles IIRC plus you have protocol overhead (address bytes). The MPU6xxx data registers are 16-bit so with just gyro and accel registers you have 2*6 bytes. You're already down to around 3 kHz max sample rate.

ducminhkt:
@tracernz, Thanks

kubark42:
@tracernz This branch has some merge conflicts.

You might find that @mluessi's logging improvements in #1928 give you enough speed to do off-line analyses of real-world data. And I'm happy to try this out in a motion-capture room to give ground-truth results.

Trex4Git:
A lot have changed in last time and this will need probably a manual rebase to current next.

I will work on this, but had not much free time.

Trex4Git:
- small bugfixes
- applied MPU orientation overflow changes
- rebase to latest next

PR #1788, opened 2015-08-12
Make: Allow QT make spec to be set by env variable

tracernz:
This allows gcs to be built by e.g. clang. 

Note: currently clang breaks the uploader plugin due to an undefined symbol in librawHID (_ZN10USBMonitorD1Ev).


BrainFPV:
Looks good to me

PR #1791, opened 2015-08-13
TauLinkModule

peabody124:
This provides support for the TauLinkModule hardware. That hardware is similar
to the TauLink/PipXtreme but adds a bluetooth module for relaying telemetry and
is designed to fit into a JR Module and provide data to the Taranis for display
purposes.

This also includes some refactoring of the FrSky Sport module to create a common
library that can handle the data packing. This should also be adoptable for the FrSky
Sensor Hub module since the main difference there is the byte that is used for escaping.


mlyle:
https://developer.android.com/about/dashboards/index.html

An API version bump from 14 to 21 excludes 77% of android devices from working.

peabody124:
@mlyle I'm going to separate the java changes. It looks like the API version bump is needed to support BT LE but I'd like to see if there is a workaround that just makes it fail on that gracefully (I think there should be).

peabody124:
@elafargue you might want to look at this too since you pushed the original PR with a lot of this code.

mlyle:
Should not land before #1759

peabody124:
@mlyle is #1759 going to merge for the next release? there is still a lot on the testing combinations to go.

I'm happy to rebase 1759 onto this if the merge happens in the other order. I'll be doing that either way, I'm sure and had to do it once already because of all the indentation changes.

mlyle:
Yes, #1759 is going to merge for this release; it was opened before the high risk window, has undergone review, and is undergoing test.    I don't expect to get full coverage for the expanded matrix but we will have a good shotgunning of it which is what counts.

If I don't receive a revomini like I expect I may have to take that target out, though.

peabody124:
Oh great. I'll prepare this on top of #1759 then.

flight/targets/pipxtreme/fw/pios_board.c
```diff
@@ -164,43 +323,34 @@ void PIOS_Board_Init(void) {
...
 #endif /* PIOS_INCLUDE_PPM */
-				break;
-			}
-		case HWTAULINK_PPMPORT_DISABLED:
-		default:
-			break;
-	}
+        break;
+    }
+    case HWTAULINK_PPMPORT_SPORT:
```
peabody124: @mlyle you might have a thought about using the shared system to do this part. note that this is a slightly different module that is attached to the SPORT than normal but it uses the same ```PIOS_COM_FRSKY_SPORT``` variable

peabody124:
rebased onto the HAL system

peabody124:
Telemetry verified on basic pipxtreme/oplink/taulink

peabody124:
PPM+Telemetry mode verified on TauLinkModule

flight/targets/pipxtreme/fw/pios_board.c
```diff
@@ -128,32 +164,155 @@ void PIOS_Board_Init(void) {
...
+	    hwTauLink.MaxRfSpeed, NULL, rfm22b_cfg,
+	    hwTauLink.MinChannel, hwTauLink.MaxChannel,
+	    hwTauLink.CoordID, 0);
+
+	if (bdinfo->board_rev == TAULINK_VERSION_MODULE) {
+		// Configure the main serial port function
+		switch (hwTauLink.BTPort) {
+		case HWTAULINK_BTPORT_TELEMETRY:
+		{
+			// Note: if the main port is also on telemetry the bluetooth
```
mlyle: So I think it's a little magical/strange that this inits the bluetooth module for telemetry but not for combridge.

flight/targets/pipxtreme/fw/pios_board.c
```diff
@@ -31,16 +31,57 @@
...
 uintptr_t pios_uavo_settings_fs_id;
 
+#define PIOS_COM_TELEM_RX_BUF_LEN 256
+#define PIOS_COM_TELEM_TX_BUF_LEN 256
+#define PIOS_COM_RFM22B_RF_RX_BUF_LEN 256
+#define PIOS_COM_RFM22B_RF_TX_BUF_LEN 256
+#define PIOS_COM_FRSKYSPORT_TX_BUF_LEN 16
+
+#if defined(PIOS_INCLUDE_USART) && defined(PIOS_INCLUDE_COM)
+static void PIOS_Board_configure_com (const struct pios_usart_cfg *usart_port_cfg, size_t rx_buf_len, size_t tx_buf_len,
```
mlyle: If this has to be here-- and the init code for bluetooth doesn't live in PIOS_HAL.. could at least factor so the version in PIOS_HAL could be called here.

flight/targets/pipxtreme/fw/pios_board.c
mlyle: I think the shared system could do it (we don't care what the target module is.  But-- right now we don't really handle PPM in any way and I'd likely have a different wrapper call for receiver-port type things that have serial etc on them.

peabody124: Ok then I will leave this for now as is.

flight/targets/pipxtreme/fw/pios_board.c
peabody124: when I used combridge it was normally because i wanted to send commands directly to the bt module and configure it that way so I wanted it at the previous (or default) configuration

flight/targets/pipxtreme/fw/pios_board.c
peabody124: Will do

mlyle: Sorry for the conflict that showed up here.  If rebased and this one is touched up I can run my tests on this and most likely merge Tuesday.

peabody124:
@mlyle rebased and updated

mlyle:
Needs #1828 per comment

peabody124:
@mlyle should be good to merge now. Updated to use #1828

mlyle:
Yup.. just trying to get a few moments to try it.  I have a kitchensink build that this is in.

mlyle:
First attempt couldn't get revomini to talk to pipxtreme.  Possible: A) I was doing something dumb, B) this Chinese revomini is defective, C) my previous refactor introduced something defective with revomini in particular, or D) there's something wrong with this changeset.  Tried binding in both directions using the wizard; double checked the UAVOs looked sane.

Going to try sparky2 + pipxtreme next, as that was working a couple days ago on next.

mlyle:
Worked sparky2 + pipxtreme.  Going to retest revo.

mlyle:
No dice on revomini again.  Hmm.

mlyle:
Linkstate on pipxtreme toggles between "Connected" and "Enabled" with revomini plugged in, but resets accumulate and no data transfers.

mlyle:
Holding off on merging just a little.  Evidence is pretty clear it's A, B, C or other prior breakage, but I want to investigate more.

peabody124:
@mlyle did you try revomini+next?

mlyle:
That's on the todo-list, and then revomini+master.

mlyle:
Just for my information while I try and run this down-- do you know of the last time you saw RF telemetry work on revomini?

mlyle:
Works on revomini.  Didn't realize that on F4 targets only.. that if you power via USB all other telem is shut down (e.g. CDC device doesn't need to be open).

PR #1794, opened 2015-08-14
Sparky2: update STP file for correct thickness

peabody124:


peabody124:
Thanks to @m-thread for noticing that the thickness of the Sparky2 3D model
was 2.2mm instead of 1.6.

m-thread:
Tested and this is ok. Another bonus was that the holes in the board was there also :) Great!

Issue #633, opened 2013-06-12
Does MavlinkBridge work on CC?


guilhermito:
Recently there was an increase in stack usage making the available stack for modules much less than it used to be.
The MavLinkBridge for example is now impossible to work on CopterControl since it requires about 600 bytes (probably more now with Vinz recent changes) and the stack available is about 400 bytes.
I don't know enough about the memory management to pinpoint where the stack usage increased but buy reverting to other revisions I think it was around the FreeRTOS upgrade. The reason I was not able to exactly pinpoint the revision will be discussed in another issue.

mlyle:
Need to see if mavlinkbridge works on CC.

SystemsGuy:
Tested on CC3D with mlyle's 20150813 build with no issues.  Even flys :)

SystemsGuy:
![tau labs gcs_2015-08-13_18-18-03-cc3d](https://cloud.githubusercontent.com/assets/2819484/9264019/ee1fc77e-41e7-11e5-919a-e784b4f75571.jpg)

SystemsGuy:
![tau labs gcs_2015-08-13_18-37-05-ss](https://cloud.githubusercontent.com/assets/2819484/9264297/605f9754-41ea-11e5-88ed-179e92bebb76.jpg)

mlyle:
Fixed by #1797.

PR #1797, opened 2015-08-15
Save stack space in mavlink bridge

mlyle:
For issue #633.  @SystemsGuy generously bench-tested, and found it to work but heap was very low.  This gets us some more bytes.  Needs verification that it still works.


peabody124:
how much is free when it is running? typically it is nice to keep a margin of at least 50 bytes so we don't get surprise segfaults

peabody124:
logic looks good, moving variables into local scope is sensible and having explicit defaults is not that useful

mlyle:
When I've done refactors like this, I've slightly increased the amount free.  I didn't measure this.  Static analysis says the task function uses 112 bytes less (from gcc -fstack-usage), so I turned down the task stack by 104 bytes.  There should be 8 more bytes free.

Ultimately I'd like to put together a tool that looks at the entire call graph and data from -fstack-usage and comes up with a worst-case number for a task.  But now the only tool I have in my toolbox is to look at functions that I know are in the worst-case stack usage path (and a task's main function certainly falls in that category) and look at their change.

Note the explicit defaults are still present-- assigning the structure to {} is equivalent to the previous initialization.

mlyle:
Rebased due to merge conflict.

Artifact at http://jar.lyle.org/~mlyle/TauLabs-20150189-4e6ab8ca-install.exe

SystemsGuy:
Bench tested on CC3D with  MinimOSD-Extra_Copter_Pre-release_2.4_r726 , seems to be working fine, flight modes as well.

mavlink-tau labs gcs_2015-08-21_23-15-41

mlyle:
Woot! 480 bytes free.  That's happiness right there.

Thanks @systemsguy!

PR #1798, opened 2015-08-16
gcs: change export to fix build on windows

mlyle:
Corrects Windows build regression introduced in #1739.


mlyle:
Quick-merging when Jenkins is done because 1) trivial, 2) major regression.

PR #1799, opened 2015-08-16
VtolPathFollower: fix a regression that resulted in landing halting

peabody124:
This fixes a regression from #1543 that did not enable the landing mode flag
during the landing phase of RTH. This results in the motors not properly
spooling down and sometimes the multi rotor not hitting the ground.


peabody124:
I'd suggest merging this and doing a hotfix release since this is a nasty regression.

PR #1802, opened 2015-08-17
GCS crash reporting

guilhermito:
This allows the creation and handling of crash reports.
The user is given the option to save or send the report to the TL forum.

This was tested on linux and windows.
The builds for this to work need special attention:
On windows the RELEASE_WITH_SYMBOLS config option must be passed.
On linux the build needs to be a DEBUG build and the "linux_extract_debug_symbols" script (part of this changeset) must be run to extract the symbols to the build/debugsymbols directory and then strip them from the original binaries.
Symbols files don't need to be part of the release packages of course, jenkins can remove and store PDB files from the windows build on a debug compressed file and the same for the debug_symbols directory on linux.
Mac handling should be similar to linux but I don't have any way to test it.


VArcht:
Kind of messy to send them to the forum. Any other options? 

I know that gists can be created anonymously, maybe that would be an option. Need to look into it more.

VArcht:
You can create anonymous gists https://developer.github.com/v3/gists/#authentication

Then find them back https://gist.github.com/search?q=TauLabsGcsCrashReport

mlyle:
I like the gist idea.  At the very least we would need to tell the IRC bot not to send crash report details to IRC, as that'll flood the channel and kill conversation.

guilhermito:
@VArcht  why do you think is messy sending them to the forum?
Anonymous reports is something I really don't like, chasing a bug for hours that someone created just for fun is not really something I fancy.

VArcht:
Kind of like leaving dirty laundry on the floor? Might not look good to have a bunch of errors front stage?

UX wise, best to have the least path of resistance, forum posting forces user to have a forum account, not the worse thing, but may stop some. Also some may worry to post publicly, though it could foster some "community".

I did consider that there is some opportunity for abuse. I think this could be solved with some basic encryption of the error report and maybe a hash for id. Hash can be the created time salted or something. Hmm, I'm  used to closed source, not sure how to hide a salt or key in opensource.

mlyle:
So, I'd like to just mine anonymous-ish error reports and figure out the most frequent failures crossing multiple users and fix those.  I know there's some particular failures I get that I could hunt, but I have a feeling a lot of them are much more common to developers (boards in weird reboot loops, boards not doing telemetry, etc).

I'd also like detailed error reports from users with information on their environment and repro scenarios.

These two things aren't mutually exclusive, but the latter requires some degree of user intervention to be useful.

guilhermito:
> Kind of like leaving dirty laundry on the floor? Might not look good to have a bunch of errors front stage?

Not sure how to read that, I do have a small basket to put my dirty laundry in, not sure if that helps.

> UX wise, best to have the least path of resistance, forum posting forces user to have a forum account, not the worse thing, but may stop some. Also some may worry to post publicly, though it could foster some "community"

The path of less resistance is not sending a report at all. My opinion is that with a project of this nature if a user doesn't even have a forum account his issue with the software is to me non existent, and I wouldn't spend a sec trying to fix it. 

> I did consider that there is some opportunity for abuse. I think this could be solved with some basic encryption of the error report

Not following how encryption would minimize the possibility for abuse, one can still create a bunch of issues by messing with the source and sending those encrypted. 


> I'd also like detailed error reports from users with information on their environment and repro scenarios.

This PR contains code for a form where the user is given the chance to complement the report with a small text description of the issue. Also the google breakpad generated dump contains information about the system where it was generated.

peabody124:
From a pragmatic perspective, while there might be other ways of storing the crash reports - e.g. a database that has the contact and crash information - that are easier to mine, this code exists and works and provides an option for people to pass us useful info (and then they can also link to that from a github issue). However, those changes can always be done as a subsequent change. If the code works and is clean I think merging this is an improvement.

mlyle:
I think at minimum we need to change the IRC bot and the thing that shows posts on the welcome screen if we do it on the forum.  It's bad for the project if the channel gets flooded and the welcome screen is always filled crash reports.

Once I had a very active IRC channel used for developer communications at a company completely die from a CI bot that was too spammy and scrolled off active conversations.  I love the bot linking us to the forum, but I have a feeling we're going to get a lot of crash reports.

guilhermito:
Changed both the crash reporting and autotune results sharing forum thread for a newly created one invisible to both bots and guests.

VArcht:
>Not sure how to read that, I do have a small basket to put my dirty laundry in, not sure if that helps.
>The path of less resistance is not sending a report at all. My opinion is that with a project of this nature if a user doesn't even have a forum account his issue with the software is to me non existent, and I wouldn't spend a sec trying to fix it.

I could post a bunch of links on UX and why otherwise brilliant devs just don't get it but not much fight in me atm. I like TL the way it is, probably best for it not to be too usable or popular.

>Not following how encryption would minimize the possibility for abuse, one can still create a bunch of issues by messing with the source and sending those encrypted.

You would need a private key, but like I said, not sure how/if that could be managed in OSS

As it is this much better than nothing, I wouldn't think to try to block it (even if I could) just wanted to share my ideas/concerns.

mlyle:
I will get to this Monday, most likely.

mlyle:
So on OSX these files stack up: 

`448K	/var/folders/kb/g2m5j1f576l3z7lk3ypyq6hr0000gn/T/taulabsgcs_mpl-allstuff_20150824-efb03a18/B83C7CD2-B485-40CD-8072-570097970FB5.dmp`

400-500k per crash.  not sure this is fatal... but they're in a weird location and odds are the user will never clean them up.

An actual crash outputs this:

```
m_crashReporterWChar: crashreporterappDump file was written to: /var/folders/kb/g2m5j1f576l3z7lk3ypyq6hr0000gn/T/taulabsgcs_mpl-allstuff_20150824-efb03a18/B83C7CD2-B485-40CD-8072-570097970FB5.dmp
Error: Can't launch CrashReporter!
```

guilhermito:
> ... but they're in a weird location and odds are the user will never clean them up.

That is weird. Breakpad's library is initialized with the temporary directory as argument so that should be the place for the reports. 

> An actual crash outputs this:

So it didn't start the reporting app, that could be because it didn't find it or...

I'm unable to try/fix this on OSX because I don't have access to a Mac machine. So the only option will be to disable this for that OS type.

mlyle:
Tried a mingw windows build and tested until I got a crash.  That seemed normal (I know this is not expected to report a crash-- just wanted to look for overall regression).

mlyle:
Tried on Linux.  kill -SEGV'd taulabsgcs.  Got a crash report.

1) Tried to view it, which has resulted in a hanging gedit session.  I believe this is because the crash report is a raw binar y
2) tried to send it to the forum, and got "Could not send the crash report-- Ooops, something went wrong"

Tried twice and it doesn't seem to work.

guilhermito:
> 2) tried to send it to the forum, and got "Could not send the crash report-- Ooops, something went wrong"

There was a problem with the forum's thread permissions. It is fixed and tested now.

guilhermito:
> What constrains it to not doing this during mingw builds? It seemed to build unless I screwed something up, but I don't understand how this works?

This uses google's breakpad library (https://code.google.com/p/google-breakpad/) which doesn't have the necessary hooks for mingw. However we shouldn't really continue to use it for windows since MSVC is now free and completely legal to use with this project. MSVC makes the GCS faster to build and run, and also seems more stable.

mlyle:
No, I meant, why does mingw not use those command line flags?  I don't think -Zi is a legal mingw flag.

I agree we should look at moving to MSVC.  I just don't think it's going to happen for this release, because of Jenkins infrastructure, etc.  It's great that we're introducing the option now and can get experience with it.

I am going to set up a Jenkins and I will get MSVC in place on it for next release, and then we can have a whole release cycle of testers using MSVC builds to look for any possible breakage.

guilhermito:
> No, I meant, why does mingw not use those command line flags? I don't think -Zi is a legal mingw flag.

So Zi tells the compiler to produce pdb files containing the debug information. AFAIK that is a windows only format and it is not produced by non MS compilers.
The debug switch allows debugging information to be produced but then the OPT:REF eliminates functions and data that are never referenced same as with a release build. We could also probably use OPT:ICF to perform COMDAT folding.
All of this is only necessary to generate debug info without bloating the binaries and making the code run slower.
Since a release build already does all of this (besides the debug info generation) there is no point on using these switches with mingw.

mlyle:
I'm sorry, what I'm trying to ask is--

```
win32 {
    # http://gcc.gnu.org/bugzilla/show_bug.cgi?id=52991
    QMAKE_CXXFLAGS += -mno-ms-bitfields
    RELEASE_WITH_SYMBOLS {
        QMAKE_CFLAGS_RELEASE += -Zi
        QMAKE_CXXFLAGS_RELEASE += -Zi
        QMAKE_LFLAGS_RELEASE += /DEBUG /OPT:REF
    }
}
```

I think it should look more like

```
win32 {
    # http://gcc.gnu.org/bugzilla/show_bug.cgi?id=52991
    QMAKE_CXXFLAGS += -mno-ms-bitfields
}

win32-msvc*:CONFIG(release, debug|release) {
    QMAKE_CFLAGS_RELEASE += -Zi
    QMAKE_CXXFLAGS_RELEASE += -Zi
    QMAKE_LFLAGS_RELEASE += /DEBUG /OPT:REF
}
```

Rather than have a RELEASE_WITH_SYMBOLS configuration that can only build with msvc.

mlyle:
Still have "Ooops, something went wrong" on linux.  Also the view a crash report thing is bad because it just opens it in gedit.

guilhermito:
> Still have "Ooops, something went wrong" on linux. Also the view a crash report thing is bad because it just opens it in gedit.

The report is human readable so I don't understand why your system has problems reading it. I have tested on windows using notepad and linux using kate and both show no issues.
About the posting I know the code works because it is the same used on the autotuning sharing. If you are sure the login info you used is ok then it is probably some forum setting I'm missing since I tested it with a admin account. If you can try manually create a forum post in http://forum.taulabs.org/viewtopic.php?f=26&t=830 and attach the dump file.

guilhermito:
> ...Rather than have a RELEASE_WITH_SYMBOLS configuration that can only build with msvc.

That would unnecessarily pollute the build with unneeded pdb files and make the build time longer than a normal release and even longer than a debug build. IMHO debug symbols for release builds should only be generated for alpha or beta releases where that information needs to be extracted and stored by the ci system. During development it makes much more sense to just use a debug build.

mlyle:
OK, when I clicked the "view the file" it tried to open a raw crash dump that started with MDMP.

I get a different error message when I use the wrong credentials-- it tells me I have a bad login name or password.  But with the correct credentials I get the "oops something went wrong".

Is there any chance that maybe my account is too new to create attachments?  My username is 'icee'

guilhermito:
> OK, when I clicked the "view the file" it tried to open a raw crash dump that started with MDMP.

Yes that is the right file. Most of the file is human readable and the intention of making it viewable to the user is showing that no personal information is sent.
To properly analyzing it the minidump_stackwalk tool needs to be used together with the debugging symbols.

> Is there any chance that maybe my account is too new to create attachments? My username is 'icee'

Yes that was why I asked you to try to post an attachment on the thread I referred above.

guilhermito:
> get a different error message when I use the wrong credentials-- it tells me I have a bad login name or password. But with the correct credentials I get the "oops something went wrong".

Are you behind a proxy? Are you able to use the autotune sharing function?

mlyle:
The 10 second timeout sounds likely.  I will check that out.

mlyle:
Please increase the timeout to something like 30 seconds, and if possible disable this on Mac since it seemed like it did a couple of bad things..

Then I'll do a sequence of packaging and test builds tomorrow and merge if at all possible for this release.  At first it'll only be effective on Linux (no Jenkins MSVC build for this version, etc).

The release schedule requires all functional work to nominally be merged by tomorrow, Monday Aug 31.  I think this helicopter support stuff is probably worth waiting a day or two for, but then it'll be time to bundle things up and really try testing systematically.

guilhermito:
> and if possible disable this on Mac since it seemed like it did a couple of bad things..

f2642fb

mlyle:
Hey I'm really sorry this has been so difficult to get merged.

I just tried running it and sending it a SEGV on linux, but it doesn't seem to pop up the crash reporter.

I see the crashreporter lines at startup.

I'm going to try a couple of other things.

mlyle:
I think this was my fault.. doing a new build and trying again.

mlyle:
It looks like the crashreporter bin can't find libUtils.so.  Trying again from a package.

mlyle:
```
mlyle@stuffy:~/TauLabs$ /home/mlyle/TauLabs/build/ground/gcs/bin/crashreporterapp
/home/mlyle/TauLabs/build/ground/gcs/bin/crashreporterapp: error while loading shared libraries: libUtils.so.1: cannot open shared object file: No such file or directory
mlyle@stuffy:~/TauLabs$ ldd /home/mlyle/TauLabs/build/ground/gcs/bin/crashreporterapp
	linux-vdso.so.1 =>  (0x00007fff0fbd0000)
	libUtils.so.1 => not found
	libQt5Widgets.so.5 => /home/mlyle/TauLabs/tools/Qt5.5.0/5.5/gcc_64/lib/libQt5Widgets.so.5 (0x00007f2e560a8000)
	libQt5Gui.so.5 => /home/mlyle/TauLabs/tools/Qt5.5.0/5.5/gcc_64/lib/libQt5Gui.so.5 (0x00007f2e55890000)
	libQt5Core.so.5 => /home/mlyle/TauLabs/tools/Qt5.5.0/5.5/gcc_64/lib/libQt5Core.so.5 (0x00007f2e55148000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f2e54e40000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2e54c28000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2e54860000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2e54640000)
	libgobject-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007f2e543e8000)
	libgthread-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgthread-2.0.so.0 (0x00007f2e541e0000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f2e53fd8000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f2e53cd0000)
	libXext.so.6 => /usr/lib/x86_64-linux-gnu/libXext.so.6 (0x00007f2e53ab8000)
	libX11.so.6 => /usr/lib/x86_64-linux-gnu/libX11.so.6 (0x00007f2e53780000)
	libGL.so.1 => /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1 (0x00007f2e534f0000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2e531e8000)
	libicui18n.so.54 => /home/mlyle/TauLabs/tools/Qt5.5.0/5.5/gcc_64/lib/libicui18n.so.54 (0x00007f2e52d78000)
	libicuuc.so.54 => /home/mlyle/TauLabs/tools/Qt5.5.0/5.5/gcc_64/lib/libicuuc.so.54 (0x00007f2e529c8000)
	libicudata.so.54 => /home/mlyle/TauLabs/tools/Qt5.5.0/5.5/gcc_64/lib/libicudata.so.54 (0x00007f2e50f98000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2e50d90000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2e56928000)
	libffi.so.6 => /usr/lib/x86_64-linux-gnu/libffi.so.6 (0x00007f2e50b88000)
	libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007f2e50948000)
	libxcb.so.1 => /usr/lib/x86_64-linux-gnu/libxcb.so.1 (0x00007f2e50728000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f2e504f8000)
	libglapi.so.0 => /usr/lib/x86_64-linux-gnu/libglapi.so.0 (0x00007f2e502c8000)
	libXdamage.so.1 => /usr/lib/x86_64-linux-gnu/libXdamage.so.1 (0x00007f2e500c0000)
	libXfixes.so.3 => /usr/lib/x86_64-linux-gnu/libXfixes.so.3 (0x00007f2e4feb8000)
	libX11-xcb.so.1 => /usr/lib/x86_64-linux-gnu/libX11-xcb.so.1 (0x00007f2e4fcb0000)
	libxcb-glx.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-glx.so.0 (0x00007f2e4fa98000)
	libxcb-dri2.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-dri2.so.0 (0x00007f2e4f890000)
	libxcb-dri3.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-dri3.so.0 (0x00007f2e4f688000)
	libxcb-present.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-present.so.0 (0x00007f2e4f480000)
	libxcb-sync.so.1 => /usr/lib/x86_64-linux-gnu/libxcb-sync.so.1 (0x00007f2e4f278000)
	libxshmfence.so.1 => /usr/lib/x86_64-linux-gnu/libxshmfence.so.1 (0x00007f2e4f070000)
	libXxf86vm.so.1 => /usr/lib/x86_64-linux-gnu/libXxf86vm.so.1 (0x00007f2e4ee68000)
	libdrm.so.2 => /usr/lib/x86_64-linux-gnu/libdrm.so.2 (0x00007f2e4ec58000)
	libXau.so.6 => /usr/lib/x86_64-linux-gnu/libXau.so.6 (0x00007f2e4ea50000)
	libXdmcp.so.6 => /usr/lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007f2e4e848000)
```

tracernz:
I just test this (Ubuntu 15.04 x64) and it seemed to work just fine although I haven't looked at the [actual dump output](http://forum.taulabs.org/viewtopic.php?p=6927#p6927). If I could make a suggestion it would be to disable the buttons or at least provide some kind of UI feedback once the send button is clicked and it is sending the report to the forums?

guilhermito:
hmm I'm a bit lost, first you say it doesn't find libUtils and then you say you tested it? English is not my first language so that sounds like a contradiction to me.

tracernz:
@guilhermito I think you're confusing me (tracernz/Michael Corcoran) with icee (icee/Michael Lyle)? ;) I don't think he's managed to get it working.

guilhermito:
> @guilhermito I think you're confusing me (tracernz/Michael Corcoran) with icee (icee/Michael Lyle)? ;) I don't think he's managed to get it working.

Ok that does make more sense. We should probably create a rule to limit project members with the same first name.

mlyle:
I am going to try and test again today and see if I can nail down what is going on.  It wasn't working for me before, but it was getting farther than it was last night.

mlyle:
Tested again on this rebased to next.  Nothing.  Next step, retest  dc48647

mlyle:
Worked on dc48647.  trying on raw branch next.

mlyle:
Doesn't work on 46e6081b7

mlyle:
OK, LOL.  I think I figured this out.  git checkout guilhermito/ground_crash_reporting doesn't work.

git checkout -b ground_crash_testing works
git checkout 26ca89b works

The problem is the slash in the git revision name.

Also, the progress bar stays at 0% for a long time and then instantly jumps to 100%, so it's not effective.

PTDreamer:
Seems there is no interest in this. Lets just continue to guess the reported segfaults root cause.

tracernz:
@guilhermito/@PTDreamer (??) There is certainly interest in this and I hope you will consider reopening this PR. It could be very valuable in understanding the single biggest frustration for users.

guilhermito:
> It could be very valuable in understanding the single biggest frustration for users

Yes that was the idea, this code was certainly not for me because I rarely see segfaults and since I only run the GCS in debug mode I'm always capable of looking at what went wrong.
Thank you for testing this, I saw that you were the onlyone capable/willing to upload the dmp.
As to reopening this, ATM I have no incentive to do it, this PR review was a joke.
Also this not being merged means that the autotunning results sharing  is now broken, and it will probably continue to be for the release.

mlyle:
@guilhermito Why did the autotune forum change?  And why was that tied into this unrelated PR?

mlyle:
> Thank you for testing this, I saw that you were the onlyone capable/willing to upload the dmp.

I did spend hours testing it, and did end up being able to upload a dump with the increased filename length once I figured out that it was the slash that was the problem -- http://forum.taulabs.org/viewtopic.php?p=6934#p6934

If you don't want to deal with this anymore, that's fine.  I'll pick up the diff and go from there.

Issue #231, opened 2013-01-28
Automatically generate uavobjects.pro and UAVObjects.inc


kubark42:
These files have large parts that are nothing more than a list of all available UAVOs on the system. GCS will fail to compile if the uavobjects.pro list does not match the list of available UAVOs, so there doesn't seem much reason not to automatically generate the list

The firmware fails a little more gracefully, but it's still a extra work to have to add new UAVOs to the list in each UAVObjects.inc, and as we add more boards this work will grow in size. (A new UAVO already requires modifying eight files if it is to be added to all flight boards.) The downside of automatically adding all UAVOs might be that there cases where we don't want to list them all of them. Are there?

mlyle:
My plan is:

GCS builds with all UAVOs programmatically.

UAVOs are broken into two lists:  

* UAVOs that are present on all boards
* Of other UAVOs, UAVOs that are present on all larger (non-SMALLF1 boards).

Then F1 makefiles include the first list; big makefiles include both lists; and each makefile can add any target-specific ones that apply.

mlyle:
#1759 causes GCS to automatically pick up UAVO list.  Still need to do the flight side.

mlyle:
Merged and done.

PR #1808, opened 2015-08-18
Tree cleanup

mlyle:
* Work towards #231: factor UAVOs common
* Remove 'pathfollower' module
* Remove faultsettings from CC/Naze32.  I kinda like the fault injection module, but it probably belongs on larger targets today.  Either that or we should totally remove it.  (it's currently compiled against nothing)


peabody124:
Love the LOC count :)

BrainFPV:
LGTM :)

mlyle:
> I would like to figure out how to get rid of UAVO_NAV (and similar Makefile flags). In an ideal world, including the appropriate navigation module in the Makefile would be the only requirement necessary to build a working binary.

I think the way to do this is to have module specific makefiles that enabling a module includes.  Then they can either set flags or declare dependencies on UAVOs in there.  They can also declare anything about how they want their code compiled.

> Pardon my ignorance, but why aren't all UAVOs included for all firmware targets? Shouldn't the linker handle stripping out the unused functions, thus removing the need for per-target specification?

So we need to turn them off or on somewhere so that they don't use memory on smaller targets.  So that's why we have a UAVOBJ_Init_... flag.

And as long as we're doing that, we might as well not build the corresponding source files and save a little build time.  The makefile produces both lists (the list of sourcefiles and the list of _Init_ flags) adjacent to each other.

One alternative option would be to compile all the uavos (except uavobjectsinit) into a library before building the flight targets.  The different ARM targets don't really need different compile flags for the uavos-- there's not a lot of prospect for improvement from architecture-specific code. The simulator adds a small bit of complexity here.  I'm not sure it's worth it.

Basically at this point my objective is to just deduplicate and factor out stuff.  There's a lot of things that can be done to potentially clean stuff up, but the duplication is worst (and makes it harder to do those other things).

mlyle:
Also I would like to end run-time registration of UAVOs.  We have a lot of code to build up a datastructure and figure what what uavo is from an id, etc.  We can avoid all that just by generating switch statements at compile time and run this directly in code.

kubark42:
> So we need to turn them off or on somewhere so that they don't use memory on smaller targets.

Grumble, grumble. There's a can of worms here about obsolete targets interfering with future progress, but that's for another day.

> One alternative option would be to compile all the uavos (except uavobjectsinit) into a library before building the flight targets. 

Compiling all UAVOs with ```-fdata-sections -ffunction-sections````and linking the firmware blob with ```-Wl,--gc-sections``` in theory strips out the unused functions. That *should* get the unused UAVObjects, right?

> Also I would like to end run-time registration of UAVOs. We have a lot of code to build up a datastructure and figure what what uavo is from an id, etc. We can avoid all that just by generating switch statements at compile time and run this directly in code.

Good idea!

mlyle:
> Grumble, grumble. There's a can of worms here about obsolete targets interfering with future progress, but that's for another day.

I like CC3D/Naze32.  It's nice to have a cheap widely available target.  Also it's debatable whether we should have hw objects for unsupported board types, etc.

> Compiling all UAVOs with -fdata-sections -ffunction-sections and linking the firmware blob with -Wl,--gc-sections in theory strips out the unused functions. That should get the unused UAVObjects, right?

Yes, but if they're registered they're "used".

There's a "clever" way to do the initialization/registration dynamically using linker scripts, but I don't like being too "clever."

kubark42:
> Yes, but if they're registered they're "used".

Ah, I see. So the autogenerated ```uavobjectsinit.c``` is the culprit here. And the only way to formally keep the UAVObjects out is not include them in the first place. Sigh.

P.S. I kind of doubt it, but I wonder if there's a function to tell ```—gc-sections``` to ignore certain files when deciding if something is used or not.

mlyle:
> P.S. I kind of doubt it, but I wonder if there's a function to tell —gc-sections to ignore certain files when deciding if something is used or not.

If there was, the call in the uavobjectsinit.c would fail to resolve to a symbol.  The linker is not anyone's friend.

PR #1809, opened 2015-08-19
smallf1: eliminate obsolete trimming mode

mlyle:
There used to be a mode to calculate a trim based on a large number of
samples.  However, it's specific to smallf1 and there is no GCS code to
use it.  Removin'.


BrainFPV:
LGTM

Issue #1801, opened 2015-08-17
PicoC disabled on RevoMini


mlyle:
commit dd588113360d28cf35e8a4d554022ff95bbfd822
Author: ernie of infect @ernieift
Date:   Fri Sep 5 07:19:41 2014 +0200

    picoC: remove revomini support
    
    Temporary removed due a compiler issue on make ef_. Will be enabled
    again when fault is removed.

mlyle:
```
 PADDING   [ef|rm  ]  build/ef_revomini/ef_revomini.fw_pre.pad
dd if=/dev/zero count=98304 bs=1 2> /dev/null | LC_CTYPE=c tr '\000' '\377' > /Users/mlyle/taulabs/build/ef_revomini/ef_revomini.fw_pre.pad && [ ${PIPESTATUS[0]} -eq "0" ]
 PADDING   [ef|rm  ]  build/ef_revomini/ef_revomini.fw_post.pad
dd if=/dev/zero count=-17688 bs=1 2> /dev/null | LC_CTYPE=c tr '\000' '\377' > /Users/mlyle/taulabs/build/ef_revomini/ef_revomini.fw_post.pad && [ ${PIPESTATUS[0]} -eq "0" ]
```

mlyle:
```
FW_POST_PAD := $(shell echo $$[$(FW_BANK_SIZE)-$(FW_DESC_SIZE)-$(FW_DESC_SIZE)-$(shell wc -c < $(FW_BIN))])

FW_BANK_SIZE        := 0x00040000  # Should include FW_DESC_SIZE (256kb)
FW_DESC_SIZE        := 0x00000064
-rwxr-xr-x  1 mlyle  staff  279632 Aug 17 02:24 fw_revomini.bin
```

274k vs 256k.  384k on part total.  Bootloader occupies first 32k, next 96k unoccupied I think, then 256k of flash.

Without picoc we are at 216k.   So picoc increases our image size by a startling 58k.

It may be time to claw back some of that 96k.  We're already building with -Os.  Or maybe we give up on picoc on revomini.

ernieift:
It think it is a linker problem not a compiler error. Have a look at this https://github.com/TauLabs/TauLabs/commit/290690f58f1a18e52fc004d287591da655f6eda6 . It should solve the issue on revomini. I don't have it, so i can't test.

tracernz:
[...]

PR #1810, opened 2015-08-19
revomini: adjust flash partitioning, enable picoc

mlyle:
Corrects #1801.  RELNOTE: Requires new bootloader.

Now consistent with Quanton and Sparky2.

Needs test; when my revomini shows up I'll give it a whirl.


peabody124:
So the only issue with this is that it makes it harder for people to jump back and forth between TL and OP/LP. I'm not sure if picoC (for those users) would be worth the increased barrier to entry.

I'm not particularly against this, just laying out the pros/cons.

mlyle:
Does the new bootloader prevent loading old OP images?

It looks like OpenPilot raised it to a 640k limit in 2003: https://github.com/openpilot/OpenPilot/blob/master/flight/targets/boards/revolution/board-info.mk

mlyle:
That should be 2013. :P

peabody124:
Oh, then I'm less sure. Should be tested - maybe they aren't even compatible anymore but I don't think that is the case.

mlyle:
Confirmed to boot and work with OP bootloader.

mlyle:
Confirmed to boot and work with TL bootloader.  Spotted an issue where ccache doesn't know about bootloader bin includes.. need to fix.

mlyle:
That is , to boot and work with -new- TL bootloader.

Issue #1807, opened 2015-08-18
Gyro cutoff frequency change problematic


mlyle:
@Trex4Git Please confirm my reasoning here.

@pug398 I believe this is the issue you've seen.

It looks like the changes to GyroCutoff in PR #1765 have detuned existing Quanton setups.  I believe the old Sparky2 default cutoff was 41Hz.  If my reasoning is correct, the old Quanton cutoff for the Gyro LPF was 55Hz with default settings (@Trex4Git is this true?).  We picked a compromise value of 45Hz.  Unfortunately, this means Quanton has more filter delay than it used to, and existing autotunes have too much PID gain as a result.

On one level, this is good news-- it means that autotune is getting fairly optimal values for the PID loop and a small introduction of additional delay results in an overtuned situation.

I can think of two possible approaches to fixing this.

1. Setting our default Gyro LPF to 55Hz on all platforms.  I think this value is safe from a vibration perspective on most vehicles, and reducing delay is always safe-- bonus phase margin.  Quanton would continue to operate the same and people with Sparkies would pick up some more PID headroom.
2. Document that retuning is necessary.

Any thoughts?

Trex4Git:
I based the default value on 500Hz MPU rate, what seemed to me the default setting for most targets.
Quanton has 666.67Hz so had a higher cutoff with the old implementation (where cutoff changes with MPU rate).

When i remember correctly about my investigation about the cutoff frequency it was with default GyroTau (0.005):
 500Hz:  40.5Hz
 666Hz:  54Hz

So 55Hz should be ok, in cleanflight a lot of people seems to use 60Hz at the moment.

peabody124:
has anyone plopped in debug output to see the calculated gyro_alpha that #1765 produces (compared to the original 0.005), and also to make sure it is nice and stable?

i'm also putting some comments in the original PR that should be fixed with the next fix. That PR was marked WIP so I never had a chance to review it.

peabody124:
Yeah option 1 seems ideal. And nice thing is people can test without a new build by changing the param :)

mlyle:
So old gyroalpha was sqrt(1/e) = 0.607.

Quanton/Colibri/AQ32/FlyingF4 (667Hz):
New value @55Hz, assuming 667Hz update rate is: e^(-2 * pi * 55 * (1 / 667)) =~ **0.595**
[...]

PR #1815, opened 2015-08-21
stabilization: New attempt at gyro filter infrastructure

mlyle:
@Trex4Git can you please review?

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -231,21 +231,19 @@ static void stabilizationTask(void* parameters)
 		// exponential moving averaging (EMA) of dT to reduce jitter; ~200points
 		// to have more or less equivalent noise reduction to a normal N point moving averaging:  alpha = 2 / (N + 1)
 		// run it only at the beginning for the first samples, to reduce CPU load, and the value should converge to a constant value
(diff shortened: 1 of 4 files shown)


Trex4Git:
From a first quick look it loos fine.
Some refactoring am some functional changes.
Just some comments:

- a default value for gyro alpha that should be fine (similar to old gyro tau default valus as you have shown here https://github.com/TauLabs/TauLabs/issues/1807#issuecomment-132288125)
- the refactoring of the flags etc. is a little bit different but looks fine
- i had implemented when lpf is disabled no filtering is done, that maybe saves a little bit processing time. But running the filter  also with gyro alpha = 0 and having no conditional check if it should be run or not , might be only a minimal difference
- you removed completely the testing of cutoff frequency against nyqusit frequency that you had requested earlier. So we got rid of the problem how to deal with th disabled LPF ;)

mlyle:
I made sure that things come very close to 1 in all cases.  A compare and branch is not much cheaper than few multiplies on faster platforms, and the cost of the compare is incurred whether the filter is on or off.

@pug398 says this flies great.  Gonna merge.

Trex4Git:
As said, I also think  there should be only a minimal difference; I just wanted to mention. But nice to see that there is already positive feedback and it is merged again.

Issue #1505, opened 2015-03-07
Add "Cancel" button


Zyamstein:
It will be good to add Cancel button on all steps of ESC Calibration

Today I try to make and realized that props or in the motors and decide to put it out, but I have a lack of space and want to cancel the ESC calibration to safely put them down - but there was no options and I press the Esc key and motors spenn to maximum... Everything else I think every body know what's happens :)

mlyle:
#1818 merged, closing

PR #1818, opened 2015-08-21
gcs: ESC calibration: allow cancellation

mlyle:
Fixes issue #1505.


mlyle:
Confirmed to work on hardware-- revomini

PR #1820, opened 2015-08-22
Loiter control of altitude

### LINKED ISSUE #1444: allow control of throttle in positionhold mode
when set "throttlecontrol = true" ,the alttitude is hold there ,but cant chontrol it. can we separate  the alttitude control from ph mode ,and use the ah mode to control it?

mlyle:
Fixes #1444 .  Defaults to off, requires you to turn on ThrottleControl and LoiterAllowAltControl in VtolPathFollowerSettings.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_control.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_control.c
@@ -177,7 +177,8 @@ int32_t vtol_follower_control_path(const float dT, const PathDesiredData *pathDe
  * Controller to maintain/seek a position and optionally descend.
  * @param[in] dT time since last eval
  * @param[in] hold_pos_ned a position to hold
- * @param[in] landing whether to descend
+ * @param[in] alt_rate if not 0, a requested descent/climb rate that overrides
+ * hold_pos_ned
  * @param[in] update_status does this update path_status, or does somoene else?
  */
 static int32_t vtol_follower_control_impl(const float dT,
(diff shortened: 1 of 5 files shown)


flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -560,12 +576,29 @@ bool vtol_follower_control_loiter(float dT, float *hold_pos, float *att_adj) {
 		cmd.Pitch
 	};
 
+	const float CMD_THRESHOLD = 0.2f;
+	const float CMD_ALT_THRESHOLD = 0.4f;
```
BrainFPV: I think it would make sense to use the same deadbands as for `AltitudeHold`.. can we just use `AltitudeHoldSettings` here? IMO would be best to have the same AH behaviour in normal AH and loiter.

mlyle: That's a good point.  Probably still makes sense to use the navigation configured ascent/descent rates, I think, but using the althold configured deadbands makes sense.

mlyle:
Updated based on feedback to use althold deadband and expo.  Default althold deadband adjusted to 30% to better cope with non-self-centering throttle sticks.

mlyle:
@pug398 indicates the previous rev, 269b064, flies well.  There's minimal change since then (just the althold settings).

mlyle:
Mergin', getting ready for overall navigation release integration tests.

PR #1822, opened 2015-08-23
Use fastheap, take 2

BrainFPV:
New attempt to use the fastheap. Needs to be tested on Colibri (and other F4 targets), which previously crashed. `pios_streamfs` no longer uses the fatheap for the com buffer, which may have been the culprit (see comment by @solidgoldbomb in TauLabs/TauLabs#1691)

c.f. TauLabs/TauLabs#1691 TauLabs/TauLabs#1752

### DIFF
--- a/flight/Modules/FirmwareIAP/firmwareiap.c
+++ b/flight/Modules/FirmwareIAP/firmwareiap.c
@@ -182,7 +182,7 @@ static void FirmwareIAPCallback(UAVObjEvent* ev)
 						/* Note: Cant just wait timeout value, because first time is randomized */
 						reset_count = 0;
 						lastResetSysTime = PIOS_Thread_Systime();
-						UAVObjEvent * ev = PIOS_malloc(sizeof(UAVObjEvent));
+						UAVObjEvent * ev = PIOS_malloc_no_dma(sizeof(UAVObjEvent));
 						memset(ev,0,sizeof(UAVObjEvent));
 						EventPeriodicCallbackCreate(ev, resetTask, 100);
 						iap_state = IAP_STATE_RESETTING;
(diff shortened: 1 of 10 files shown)


mlyle:
Indeed it looks like Colibri has external flash and the pios_jedec_flash_driver uses DMA.  I tried checking for that before reverting but missed the jedec stuff-- sorry about that.

mlyle:
Confirmed to work on revomini.

mlyle:
Confirmed to work on colibri.

lolorc:
it was previously crashing on quanton, does it work on quanton as well ?

mlyle:
I'm not able to try Quanton right now.  But it looks like the crash bug would cross all targets with external flash, and has been confirmed to work on two targets that would crash before.

lolorc:
are you saying this has been merged without being tested on all f4 platforms ?

mlyle:
It's not feasible to verify every PR against every platform, nor is there a reason to suspect the remaining platforms would vary.

The release will get tested against every platform and I'm continually rotating between platforms during work on -next.

mlyle:
@lolorc I'd welcome your feedback from trying -next on quanton; if not I'll be cycling to do tests on Quanton in a few days.  I simply don't have lab bench space or flight platforms to have all the targets active at once, and few people are providing comments that things work on individual targets.

As it is, the previous version was cursorily tested on various platforms but not flight tested; the problem that was introduced was a crash on arm on parts with JEDEC flash.  I don't think it's reasonable to gate every change on flight test on every platform.  If a problem is introduced, we'll bisect and fix it.

lolorc:
i like how you make things moving forward lately, still I don't see the hurry in merging untested stuff that quickly.
just bench-tested it on quanton, I'm able to arm, it doesn't crash.

mlyle:
I'm glad Quanton is good.  I figured since I saw two regressed platforms go green (and wasn't aware that Quanton was bad before) that the changeset could land.  And things are landing a bit quickly because I"m trying to get next to a good point for overall integration testing.

mlyle:
I've now gotten one report of this crashing AQ32 ... sometimes... on arm.  I'm not reverting it again quite yet but I'm concerned.

PR #1824, opened 2015-08-24
build: remove space from end of tool url

mlyle:


mlyle:
This is insufficient.  It attempts to decompress using bzip2.  I will be working on this when I have access to my Windows machine.

peabody124:
Did this work? I've found weirdness with the URL when setting them up on windows.

mlyle:
Yes, this works.  Just confirmed packaging still works today.

I don't think we should go to an error right away and break existing setups.

PR #1825, opened 2015-08-24
make: fix a dumb bug resulting in fail to init on some targets.

mlyle:
About to go to systemic test today of this fused next... Picked up a little more risk and made some more dumb mistakes than I like.


mlyle:
Tested simulator and revomini.  Going to test more today in general... Also going to try and look at Windows tools issues reported by Aviraj on the forums. 

Merging this one once jenkins completes.

Issue #489, opened 2013-05-07
Explicitly handle divide by 0 and NaN


kubark42:
In https://github.com/TauLabs/TauLabs/pull/488#discussion_r4101533, @lilvinz pointed out that we can set a trap for divide by 0. This seems like a good idea to consider, since a divide by 0 error can be catastrophic for real-time control, especially if it leads to a NaN in the system. 

In general, what would appropriate behavior be in the case of divide by 0? While this ideally shouldn't ever occur, algorithmic bugs or erroneous user input are hard to stop in all cases. Should we ignore the illegal operation and hope that it doesn't spread? Or should we reset the MCU, at the risk of destabilizing an otherwise functioning system?

PR #1832, opened 2015-08-25
Nan fixes

kubark42:
Interesting problem regarding NaN and Inf checks. The compile optimization ```--fast-math``` does many useful things, but it also removes the ability to check if a number is a ```Nan```, even if the check is the classic ```x == x```. Thus we need to add this back in, and we do it as a function which has the optimizations turned off.

Souce: http://stackoverflow.com/questions/22931147/stdisinf-does-not-work-with-ffast-math-how-to-check-for-infinity

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -46,10 +46,16 @@
  * simple filter to smooth the altitude.
  */
 
-#include "pios.h"
 #include "openpilot.h"
+#include "pios.h"
+#include "pios_thread.h"
+#include "pios_queue.h"
+#include "misc_math.h"
 #include "physical_constants.h"
+#include "coordinate_conversions.h"
+#include "WorldMagModel.h"
 
+// UAVOs
 #include "accels.h"
 #include "attitudeactual.h"
 #include "attitudesettings.h"
@@ -71,10 +77,6 @@
 #include "stateestimation.h"
 #include "systemalarms.h"
 #include "velocityactual.h"
-#include "coordinate_conversions.h"
-#include "WorldMagModel.h"
-#include "pios_thread.h"
-#include "pios_queue.h"
 
 // Private constants
 #define STACK_SIZE_BYTES 2200
@@ -595,8 +597,9 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 		MagnetometerData mag;
 		MagnetometerGet(&mag);
 
-		// If the mag is producing bad data (NAN) don't use it (normally bad calibration)
-		if  (mag.x == mag.x && mag.y == mag.y && mag.z == mag.z) {
+		// Only use the magnetometer data if it is good, i.e. not NAN. A NAN would
+		// normally only arise due to a bad magnetometer calibration.
+		if  (!(IS_NOT_FINITE(mag.x) || IS_NOT_FINITE(mag.y) || IS_NOT_FINITE(mag.z))) {
 			float bmag = 1.0f;
 			float brot[3];
 			float Rbe[3][3];
@@ -676,9 +679,9 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 	cf_q[2] = cf_q[2] / qmag;
 	cf_q[3] = cf_q[3] / qmag;
 
-	// If quaternion has become inappropriately short or is nan reinit.
+	// If quaternion has become inappropriately short or has become Nan reinit.
 	// THIS SHOULD NEVER ACTUALLY HAPPEN
-	if((fabsf(qmag) < 1.0e-3f) || (qmag != qmag)) {
+	if((fabsf(qmag) < 1.0e-3f) || IS_NOT_FINITE(qmag)) {
 		cf_q[0] = 1;
 		cf_q[1] = 0;
 		cf_q[2] = 0;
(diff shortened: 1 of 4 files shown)


flight/Modules/Attitude/attitude.c
```diff
@@ -1015,7 +1018,7 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 		GPSVelocityGet(&gpsVelData);
 
 	// Discard mag if it has NAN (normally from bad calibration)
-	mag_updated &= (magData.x == magData.x && magData.y == magData.y && magData.z == magData.z);
+	mag_updated &= !(IS_NOT_FINITE(magData.x) || IS_NOT_FINITE(magData.y) || IS_NOT_FINITE(magData.z));
```
mlyle: Yay demorgan.

mlyle:
So.. overall fast-math makes me nervous.  Just about every optimization in there is dangerous-- either to precision or overflow handling.  But turning it off in general scares me too at this late juncture with unknown performance implications.

How did you come to discover this?  Was a situation generating NaNs that you noticed were not caught?

This fix is in a kinda gray area.  If the FPU itself generates NaNs or Infs we'll be able to spot them with this.  But the compiler is free to eat NaNs in other circumstances (e.g. it can const-evaluate things without worrying an input is nan).  So even sprinkling these in the code isn't really a guarantee we're dealing with it appropriately.

OTOH, NANs should never happen and this is mostly a safety net.   on the other other hand-- things like the INS are doing pretty scary, pretty deep math that operation reordering in fast-math is scary for and that can't be bounded / guaranteed to not occasionally numerically blow up.

flight/Modules/Attitude/attitude.c
kubark42: I struggled with that one because it's uglier than before. In retrospect, I think I should have made an ```IS_FINITE``` macro which simply inverts IS_NOT_FINITE.

kubark42:
I discovered it when I was getting NaNs while doing a curve fitting for closed loop rpm-based control. The mapping involved taking the sqrt of a number between 0 to 1, but the actual input to the function could jump to -1 in certain circumstances and I didn't check for sanity before taking the ```sqrt()```.

We know that our physical reality never goes to ```NaN``` or ```inf```, so if we had infinite resources and precision I don't see that it's intrinsically wrong to allow the software to make optimizations with this simplification in mind. Especially since the NaNs can occur anyway due to coding errors or algorithmic limitations.

There doesn't seem to be a way to generally resolve the issue: https://github.com/TauLabs/TauLabs/issues/489. There are a finite-- albeit perhaps large--- number of places where the NaN can occur. Sanitizing them all is within the realm of reason. 

At some point we'll probably burn the risk down on this issue through unit and system testing. It does seem to be a non-academic problem, I saw a NaN today while testing using the existing EKF.

mlyle:
Yah, I don't like the current EKF and we have issues open to reinit on NaN--even though any kind of meaningful convergence before crash is dodgy.  Also, neither the current complementary filter nor the EKF does a very good job of estimating attitude.

The fast-math potentially reordering operations and sacrificing precision on things we care about could be encouraging the EKF to blow up, actually.

There's various techniques that we can use to trade off a little optimality for numerical stability, and I hope to look at them when a bunch of other things are fixed. ;)

Based on this rationale, code review, and empirical verification that this is OK (bench test) merging.

kubark42:
At least one comparison paper showed that the EKF is inferior to the Mahoney filter. (This doesn't surprise me because the Mahoney filter operates on SO(3), whereas the EKF is operating in a linearized space.) In fact, according to their findings, the Mahoney filter was the best amongst those tested on traditional multirotors. There are upgrades to be made to our complementary filter (c.f. my commits from summer of 2012), but the attitude estimate seems to be otherwise very robust for the cost. What do you feel is the best filter for attitude estimation? I can fly in four motion capture rooms, using Vicon and OptiTrack solutions. Happy to compare against ground-truth if a new algo needs testing.

Numerical stability on the EKF can be gained by moving to a square-root filter. See Chui and Chen for their SQSF algorithm.

mlyle:
So, I'm deliberately not digging into the INS yet, because it's a huge undertaking and there's large wins to made bade elsewhere.

Both the INS and the complementary filter do really badly for estimating attitude under accelerated flight.  It's really obvious watching videos of BrainFPV and watching the horizon line bounce around.  Likewise when I was flying Horizon mode on overpowered vehicles it would return to really wacky attitudes at times. Right now the complementary filter does obviously better than INSOutdoor but still not very good-- and the INS ought to be able to do better because of the additional information.  At least the complementary filter is "good enough"-- if you switch to attitude mode it goes to a sane enough attitude to converge and eventually find level given enough altitude and time. :P

I think there are a lot of wins to be had-- I've not looked at the state equations at all for the INS, and there seem to be some low hanging fruit in the complementary filter (de-weighting the accelerometer under high accelerations).

As to the actual filter architecture, I think almost anything can be good with the proper state equations and sensor weightings--- but I seriously favor architectures that we can prove constraints about numerical stability as even a fairly thorough flight test program doesn't really explore very much of the envelope of possible inputs.

I appreciate the mocap offer-- that's super interesting.  Right now I don't think I have enough knowledge of those parts of the system to be dangerous yet, and I'm not good enough with tools to understand flight logs to do a lot of this analysis in a super rigorous way.  I have on my to-do list to encapsulate serial streams in UAVO so we can shuttle differential pseudorange information to the onboard GPS.. and to look at logging of pseudorange data to do RTK after flight.

kubark42:
I see what you mean. It's not so different from the estimation problems I faced with LVL 1, although my solutions there won't work because here the problem is flipped. I had to deal with an initial lateral acceleration, followed by a steady-state high velocity phase where the vehicle is perfectly flat, whereas the classic multirotor is level in steady-state zero velocity phase and then needs to handle constant tilt-- i.e. sensed lateral acceleration-- when cruising.

Regarding the complementary filter, the ```AccelKp``` and ```AccelKi``` are probably due for a retune.  They were done several years ago and sensors and vehicle builds have improved significantly since then. I find that values ~20% smaller reduce drift due to lateral acceleration without compromising gyro bias drift rejection.

You might enjoy Randy Beard's work on this subject: http://www.et.byu.edu/~beard/papers/preprints/LeishmanMacdonaldBeard__.pdf. His results are quite promising.

PR #1833, opened 2015-08-25
Fix internal sorting of XML in .uav files

kubark42:
It is currently difficult to visually compare two .uav settings files. The saved settings don't line up because in XML there is no guarantee that the output file will always have the same node order. This resolves that problem by having Qt sort the nodes by UAVObject name.

Solution inspired by http://stackoverflow.com/questions/14647638/xslt-sorting-how-to-sort-xml-childnodes-inside-a-parent-node-with-an-attribute

### DIFF
--- a/ground/gcs/src/plugins/uavsettingsimportexport/uavsettingsimportexport.pro
+++ b/ground/gcs/src/plugins/uavsettingsimportexport/uavsettingsimportexport.pro
@@ -2,6 +2,7 @@
 TEMPLATE = lib
 QT += xml
 QT += widgets
+QT += xmlpatterns
 TARGET = UAVSettingsImportExport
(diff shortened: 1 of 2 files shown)


mlyle:
XSLT.  Nice.

mlyle:
```
Warning in file:///Users/mlyle/taulabs/build/ground/gcs/bin/Tau Labs GCS.app/Contents/MacOS/Tau Labs GCS, at line 1, column 88: Running an XSL-T 1.0 stylesheet with a 2.0 processor.
```

There's nothing that seems to be version specific in here, so it's probably just a question of bumping the xsl stylesheet version in the document to 2.0.

kubark42:
@mlyle Thanks. Fixed.

mlyle:
Cool, thank you!   Merging.

PR #1834, opened 2015-08-25
UAVObject: Fix for multiinstance metadata object initialization

kubark42:
Inspired by http://git.openpilot.org/cru/OPReview-693. (Unfortunately this link is currently dead since OPNG took down the public git server. I am placing it here because 1) the original work should be credited and 2) I am hopeful that OPNG's git server will become open once again.)

From the original review (which IIRC was started by CorvusCorvax):
"When a UAVO with instance id > 1 is read from flash it will cause the metadata to be set to default values. The metadata is correctly saved to flash but if there are more than one instance it will be reset to default when second instance is loaded."

This has been tested and works as expected.


mlyle:
Verified rationale and tested.  Merging.

PR #1835, opened 2015-08-26
Map: coord in homeitem shadows coord in mappointitem and interferes with setting home

kubark42:
The presence of this variable prevented the map from correctly setting the home location, with regards to calculating the mouse cursor's NED position.

This fix can be verified by examining the NED field, as shown in the screenshot.

screen shot 2015-08-26 at 5 48 31 pm


mlyle:
Adjusted title.  Looks good, will look at a little more.

mlyle:
Verified, read code.  Mergin'.

PR #1836, opened 2015-08-27
revomini: enable combridge in build

mlyle:
Runnin' down the test matrix for the HAL refactor.. finding lots of pre-existing boogs.


mlyle:
Trivial-- merging.

PR #1839, opened 2015-08-27
Remove  timeutils while() clause that always evaluates to false

kubark42:
The time utilities has a value which is a uint, and thus can never be < 0, but is compared to < 0. This causes the unit tests on OSX to fail (but not on the others, which is strange because it's all ARM-GCC.)

This is not the only solution which would resolve the problem, but since it's unlikely that we'll have a GPS reporting negative unix time (zero day for GPS is ten years after Unix zero time) this seemed the best solution.

### DIFF
--- a/flight/Libraries/timeutils.c
+++ b/flight/Libraries/timeutils.c
@@ -57,11 +57,7 @@ void date_from_timestamp(uint32_t timestamp, DateTimeT *date_time)
 
 	days = timestamp / SECS_PER_DAY;
 	rem = timestamp % SECS_PER_DAY;
-	while (rem >= SECS_PER_DAY)
-	{
-		rem -= SECS_PER_DAY;
-		++days;
-	}
+
 	date_time->hour = rem / SECS_PER_HOUR;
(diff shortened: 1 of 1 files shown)


BrainFPV:
Looking at the code, I'm wondering how it can be that `rem >= SECS_PER_DAY`, given that `rem = timestamp % SECS_PER_DAY'. Doesn't this mean that `rem` should be within [0, .SECS_PER_DAY - 1]`?

Note: I only had 0.5 coffees so far, so I may be missing something.

kubark42:
Good point. That second function always evaluates to false as well.

P.S. C's "modulo" operator (```%```) is misleading (and a pet peeve of mine). It can return a negative number, which a true modulus operator can never do. http://stackoverflow.com/questions/4003232/how-to-code-a-modulo-operator-in-c-c-obj-c-that-handles-negative-numbers

mlyle:
With the second while loop removed I'd be glad to merge.

kubark42:
Works for me. Additional code removed.

mlyle:
Thank you! :)

Issue #1838, opened 2015-08-27
New vtol navigation PID defaults


mlyle:
We've learned some stuff about tuning navigation--and are still learning.  We should alter the defaults.

mlyle:
Done in #1871.

PR #1840, opened 2015-08-27
Txpid add path follower tuning

kubark42:
This PR would help with #1838 .

Note, it is not compatible with targets that do not include the VTOLPathFollowerSettings UAVO (and thus Jenkins should fail to build). If the functionality is of interest beyond helping out with #1838, it should be easy to include.

### DIFF
--- a/flight/Modules/TxPID/txpid.c
+++ b/flight/Modules/TxPID/txpid.c
@@ -153,7 +153,7 @@ static void updatePIDs(UAVObjEvent* ev)
 	StabilizationSettingsGet(&stab);
 	AccessoryDesiredData accessory;
 
-	uint8_t needsUpdate = 0;
+	uint8_t stabilizationSettingsNeedsUpdate = 0;
 
 	// Loop through every enabled instance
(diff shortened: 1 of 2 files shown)


mlyle:
I'd like it if we can merge this in general-- so it'd be nice if we could conditionalize the flight code on the presence of the UAVO.

One thing to be aware of-- we're in the later phases of functional work on this release.  https://github.com/TauLabs/TauLabs/wiki/Endomorphism-(September)-release

I love this stuff that's coming in, but I'm going to have to taper off merging it as we've already incurred a lot of risk and we're at the tail end.  So some of this stuff may need to be deferred to the next.  (I'm going to do my best to get in what I can).

kubark42:
Understood and agreed. I've only been pushing bug fixes because, hey, who doesn't love a good bugfix before release?

I labeled this "PR for comment", I don't expect it to merge. It's work I did earlier to resolve the problem you mention in #1838, so my hopes are that it might be again useful.

BrainFPV:
Nice :).

Maybe we could use and `#if defined(UAVO_NAV)` to conditionally compile in the code on targets that support navigation.

mlyle:
Closed in favor of #1857, which is this fixed up to build on all targets.

Issue #1837, opened 2015-08-27
New rateloop/attitudeloop PID defaults


mlyle:
We turned them up some last release; should turn them up a little more for this release as there's been no screaming yet.

mlyle:
Merged/done.

Issue #1434, opened 2015-01-04
Warning about Integral Settings


x0054:
So, I admit it, I didn't properly read the OpenPilot manual about Stabilization page. So, this is mostly my bad, by I would like to add a request to add a warning about setting the Integral setting in BOTH, the Inner and Outer loop to a non zero value. Setting both values to a nonzero makes the multirottor crash in a hurry. Maybe a warning that pops up saying something like:

"Setting Integral value in both the inner and outer loop to a nonzero value will very likely result in a crash. Are you sure you would like to proceed?"

mlyle:
Integral in both loops can be sane, though you usually want a pretty low value in the outer loop.

PR #1842, opened 2015-08-28
stabilization: turn up default outer loop Kp

mlyle:
Rationale:  Rate PI is still lower than almost all autotunes.

This outer loop P will be lower than all autotune output, but increased by 20%.  It should improve user experience.

No complaints of oscillation with default settings yet.

### DIFF
--- a/shared/uavobjectdefinition/stabilizationsettings.xml
+++ b/shared/uavobjectdefinition/stabilizationsettings.xml
@@ -14,9 +14,9 @@
 	<field name="RollRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.002,0.0015,0,0.3" limits="%BE:0:0.01,%BE:0:0.01,, "/>
 	<field name="PitchRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.002,0.0015,0,0.3" limits="%BE:0:0.01,%BE:0:0.01,, "/>
 	<field name="YawRatePID" units="" type="float" elementnames="Kp,Ki,Kd,ILimit" defaultvalue="0.0035,0.0035,0,0.3" limits="%BE:0:0.01,%BE:0:0.01,, "/>
-	<field name="RollPI" units="" type="float" elementnames="Kp,Ki,ILimit" defaultvalue="2,0,50" limits="%BE:0:10,%BE:0:10,"/>
-	<field name="PitchPI" units="" type="float" elementnames="Kp,Ki,ILimit" defaultvalue="2,0,50" limits="%BE:0:10,%BE:0:10,"/>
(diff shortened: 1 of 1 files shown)


mlyle:
Issue #1837

BrainFPV:
Good idea. I find that once the inner loop is properly tuned, `Kp = 4`, `Ki = 2` usually works pretty well for the outer loop. But those values would probably be too high to be used for defaults. I wonder if it would make sense to add some `Ki`. Thoughts?

kubark42:
In some sense, the cascaded PID architecture allows us to think about the outer loop tuning in terms of a geometric command instead of a model-dependent command. This suggests that the range of tuning values across models might be relatively compact. 

I have the equivalent experience to @BrainFPV, it seems that ```Kp=4``` and ```Ki=2``` seems to work quite well.

mlyle:
I appreciate the feedback.  I'm being very conservative on outer loop tuning for now.  It's complicated by the fact that very little of our userbase runs integrator on the outer loop because of mythology (integrator in both loops causes instability and immediate crash eg #1434)

There's a tricky job to do with the defaults-- ideally they will work well for most models to get through autotune.  It's OK if they don't work for ridiculously overpowered or strange things-- presumably people building that stuff will know what to do.  They're much better to be undertuned in each loop than overtuned because lethargic is better than oscillation.

Right now my procedure is pretty simple-- each release I'm looking at the autotune thread and turning the parameter that looks the "lowest" up a bit and then listening for screaming from users that it oscillates with the default settings. Last release I added default I term of 15; this release it seems like outer loop P is a natural thing to turn up a little.  I'n both looking at the autotune threads and talking to end-users about their tunings to do this.  If you guys have suggestions on a more systematic way to do this I'll listen, but right now I'm deliberately underdoing it.

mlyle:
Hey, this is a selfish question. ;)

@kubark42 -- I know you flew fixed wing on Tau.  Do you have a recommended starting point (still prefer undercontrol) for rate/attitude PIDs for a 1.6m elevon-mixed wing?

mlyle:
@peabody124 says he's good with this on IRC.  mergin'.

kubark42:
Seems a reasonable approach. Tuning via git commit, I kind of like it. ;)

=============================

Regarding fixed-wing, as far as the outer loop goes the values are very similar to what works for multirotors (cf my earlier comment about geometric control). On my TechPod (~2.5m wingspan), I have:

```
<field values="0.005,0,0,0.3" name="RollRatePID"/>
<field values="0.006,0,0,0.3" name="PitchRatePID"/>
<field values="0.25,0.0035,0,0.3" name="YawRatePID"/>
<field values="4,2,50" name="RollPI"/>
<field values="3,1.5,50" name="PitchPI"/>
<field values="2,0,50" name="YawPI"/>
```

Those were good for stable autonomous flight.

mlyle:
Thanks for the values--- I'll pick something just slightly lower.

PR #1843, opened 2015-08-28
pfd_qml: try the workaround for upstream QT bug again

mlyle:
So, this is a slight variant of the previous workaround for the int8-in-qml issue.

@pug398 has seen intermittent issues with the previous workaround, where he got errors about the charCodeAt property.  They spontaneously resolved.

At a minimum, this change is slightly cleaner.  It possibly fixes it if it's about what happens when the value is 0.  Opening for evaluation/consideration, but it's a bit of a hail mary.

### DIFF
--- a/ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
+++ b/ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
@@ -1,4 +1,5 @@
 import QtQuick 2.0
+import "convertint8.js" as CInt8
 
 Item {
     id: sceneItem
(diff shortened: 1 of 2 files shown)


mlyle:
Issue #1709

BrainFPV:
So the issue was because `GPSPosition.Satellites` can for some reason be a string of length 0? and `CharCodeAt` can't handle this?

mlyle:
That's a guess, but yes.

In any case it factors things into one place so we can do better.

Basically, QML marshals QT objects into javascript, but it treats char foo[]="String"; and char a=43; the same-- both are converted to javascript strings blindly.

It's weird because the new code works most of the time.  So maybe it's a critical thing when it's initially 0 it results in a zero length string (guess).  I think I should also add a conditional on the existence of the length and charcodeat properties to avoid sending errors.

It's possible pug's problem could have been caused by mismatched objects and this stuff never getting an instance or something similar.  I really don't know.  It works for him now without this PR.

pug398:
At various times I would totally delete the configuration and after the last time it started working. Hasn't stopped since.

mlyle:
@pug398  For clarity, the GCS configuration or the flight configuration?

pug398:
GCS configuration

mlyle:
Plan: check for existence of properties length and charCodeAt; if don't exist return 0.  Smoke test and call this good.

mlyle:
All's good.. Merging.

PR #1846, opened 2015-08-28
GCS: Enable halt for non-USB devices

tracernz:
Currently the GCS uploader widget only shows the halt button when connected via USB. I imagine the original intent of this was to stop boards from being halted unless connected by USB. This check is ineffective though as the board is still able to be halted when connected via PipXtreme/OPLink, since that device is connected via USB. This PR removes the check and always shows the halt button. Additionally the device disconnection is handled gracefully for non-USB connections (previously the connect dropdown and button would be left disabled when resetting via non-USB connection).

Benefits:
* Naze target is able to be put into bootloader mode by GCS (even if the uploader can't speak ST bootloader protocol yet, this is still very helpful),
* Ineffective check removed,
* Reset disconnection handled gracefully on non-USB connections.

Downsides:
* Boards can be halted (put into bootloader mode) by serial connection even though the bootloader cannot use this interface (except Naze).
* No automatic reconnect after reset on non-USB connections.

Future work:
* Add a check on flight side to prevent IAP commands working over radio links.
* Add automatic reconnect after reset for non-USB connections.

### DIFF
--- a/ground/gcs/src/plugins/uploader/uploadergadgetwidget.cpp
+++ b/ground/gcs/src/plugins/uploader/uploadergadgetwidget.cpp
@@ -1171,12 +1171,21 @@ void UploaderGadgetWidget::onResetButtonClick()
         }
         timeout.stop();
     }
-    setUploaderStatus(uploader::BOOTING);
-    conMngr->disconnectDevice();
-    timeout.start(200);
-    loop.exec();
-    conMngr->suspendPolling();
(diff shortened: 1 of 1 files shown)


mlyle:
Please open and link issues for "future work" and I'll merge.  Looks good to me.

tracernz:
Thanks, and done. :-) #1860 #1861

mlyle:
:heart:

PR #1847, opened 2015-08-28
Naze: Hack to make firmware info work correctly

tracernz:
Currently the Makefile just concatenates the 100 byte firmware info blob onto the end of the bin file to make the tlfw. When flashing GCS chops the last 100 bytes off and flashes them separately at the correct location in flash. As the GCS uploader plugin is currently unable to speak ST bootloader protocol this method doesn't work for the Naze target. This PR pads the firmware bin up to the FW info location and then concatenates the info blob. By doing this we can flash the tlfw with standard ST bootloader tools (like stm32flash) and still have the firmware info blob in the correct location.

### DIFF
--- a/flight/targets/naze32/fw/Makefile
+++ b/flight/targets/naze32/fw/Makefile
@@ -516,3 +516,12 @@ LDFLAGS += -T$(LINKERSCRIPTPATH)/link_$(BOARD)_sections.ld
 
 include ./UAVObjects.inc
 include $(MAKE_INC_DIR)/firmware-common.mk
+
+# Hack to place firmware info at correct location in flash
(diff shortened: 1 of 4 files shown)


tracernz:
Note: [Flashing instructions](https://github.com/TauLabs/TauLabs/wiki/OpenNaze-Naze32#flash-firmware-and-erase-settings) are available in the wiki to make this work effectively.

fujin:
cool hax

tracernz:
Thank you. :-P

Note: An alternative to this approach would be to fill the remainder of the flash section using the linker script (FILL()), but I don't like that because it pads the elf, bin, and hex to fix a problem in the tlfw...

mlyle:
LGTM.  Mergin'.

PR #1849, opened 2015-08-28
UAVOGen: Add max UAVO size limit

### LINKED ISSUE #1841: "Large" UAVObject size causes crash on sparky2
Summary:
- If the total size of any UAVObject is greater than 244 bytes, sparky2 goes into a reboot loop

Platform info:
- Core-i7 CPU
- ubuntu 12.04.5 amd64 / gcc-4.8.1
- sparky2 connected via USB (settings wiped)

Software
- branch: next (8f5764990d99192d1abf0cde28904b86a5a54ae6)

Issue Details:
- UAVObjects size is initially 3942
- StabilizationSettings size is initially 222 bytes
- modified shared/uavobjects/stabilizationsettings.xml; added 23 bytes of data via one 16-element uint8 vector and one 7-element uint8 vector
- compilation verifies that UAVObjects size is now 3965, StabilizationSettings is now 245 bytes
- flash firmware; result: reboot loop, with "Telemetry disconnected" each loop

Resolution Details
- reduce 7-element vector by one element; compilation verifies that UAVObjects size is now 3964, StabilizationSettings is now 244 bytes
- flash firmware; result: sparky2 functions as expected (boot, connect to USB telemetry)

Additional Details
- attempted to break ManualControlSettings in the same way
- ManualControlSettings is initially 112 bytes
- added eight 16-element uint8 vectors and one 5-element uint8 vector
- compilation verifies that ManualControlSettings is now 245 bytes
- flash, boot loop
- reduce 5-element vector by one element; size is 244; flash, no issues

I'm guessing this is a result of the message size being stored as a uint8 somewhere, but the message size has some additional length added due to headers, etc, so going above 244 causes an overflow or some other unexpected behavior.

tracernz:
Fixes #1841. I think my overhead math is correct based on what is stored and it certainly lines up with #1841 which I was able to reproduce and confirm this fix with.

### DIFF
--- a/ground/uavobjgenerator/uavobjectparser.cpp
+++ b/ground/uavobjgenerator/uavobjectparser.cpp
@@ -404,6 +404,15 @@ QString UAVObjectParser::parseXML(QString& xml, QString& filename)
         // Calculate size
         calculateSize(info);
 
+        /**
+         * The maximum size of UAVOs is limited by the FlashFS filesystem in the flight code
+         * The flash slot size is 256 bytes which is comprised of the FlashFS header (12 bytes)
+         * and the UAVO. This leaves a maximum of 244 bytes for the UAVO.
+         */
+        if(info->numBytes > 244)
+            return genErrorMsg(filename, QString("total object size(%1 bytes) exceeds maximum limit (244 bytes)")
+                    .arg(QString::number(info->numBytes)), 0, 0);
+
         // Add object
         objInfo.append(info);
 
--- a/ground/uavobjgenerator/uavobjectparser.cpp
+++ b/ground/uavobjgenerator/uavobjectparser.cpp
@@ -404,14 +404,10 @@ QString UAVObjectParser::parseXML(QString& xml, QString& filename)
         // Calculate size
         calculateSize(info);
 
-        /**
-         * The maximum size of UAVOs is limited by the FlashFS filesystem in the flight code
-         * The flash slot size is 256 bytes which is comprised of the FlashFS header (12 bytes)
-         * and the UAVO. This leaves a maximum of 244 bytes for the UAVO.
-         */
-        if(info->numBytes > 244)
-            return genErrorMsg(filename, QString("total object size(%1 bytes) exceeds maximum limit (244 bytes)")
-                    .arg(QString::number(info->numBytes)), 0, 0);
+        // Check size against max allowed
+        if(info->numBytes > UAVO_MAX_SIZE)
+            return genErrorMsg(filename, QString("total object size(%1 bytes) exceeds maximum limit (%2 bytes)")
+                    .arg(QString::number(info->numBytes), QString::number(UAVO_MAX_SIZE)), 0, 0);
 
         // Add object
         objInfo.append(info);
--- a/ground/uavobjgenerator/uavobjectparser.h
+++ b/ground/uavobjgenerator/uavobjectparser.h
@@ -36,6 +36,13 @@
 #include <QDomNode>
 #include <QByteArray>
 
+/**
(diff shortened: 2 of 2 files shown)


BrainFPV:
Looks good to me.

tracernz:
There was a small mistake which I have just corrected. :P

ground/uavobjgenerator/uavobjectparser.cpp
```diff
@@ -404,6 +404,12 @@ QString UAVObjectParser::parseXML(QString& xml, QString& filename)
         // Calculate size
         calculateSize(info);
 
+        // Buffer on flight side is 256 bytes
+        // Additional overhead is 2+4+2+2+1+1=12
```
kubark42: It would be nice to have the overhead bytes explicitly defined here in the comment.

ground/uavobjgenerator/uavobjectparser.cpp
```diff
@@ -404,6 +404,12 @@ QString UAVObjectParser::parseXML(QString& xml, QString& filename)
         // Calculate size
         calculateSize(info);
 
+        // Buffer on flight side is 256 bytes
+        // Additional overhead is 2+4+2+2+1+1=12
+        if(info->numBytes > 244)
```
kubark42: @mlyle would it be possible to replace this magic number with a #define or const?

mlyle: I think the issue may be that this module doesn't have visibility to the uavtalk headers and plumbing it may be unfortunate.

I suggest 

```
#define UAVTALK_OVERHEAD (2+4+2+2+1+1)
#define MAX_UAVOSIZE (UINT8_MAX - UAVTALK_OVERHEAD)
```

Later that define can be factored out to uavtalk and replaced with constants for the specific sizes (object ids, lengths, etc).

kubark42: That looks good (although I never know if ```#define``` or ```const``` is preferable in this situation). If you get around to it, it would be good to comment on where the ```2+4+2+2+1+1``` is coming from, although it's sufficiently clear now what it's related to.

tracernz: @kubark42 As icee mentioned there isn't a common header between the two locations and I can't really see a clean way to do so. I'll take another look at this later today and implement the requested modifications re: overhead. Thanks @icee and @kubark42.

tracernz:
You know I have spent a bit of time looking through the path by which UAVOs are loaded and I think the limit is in the flash filesystem code.
See https://github.com/TauLabs/TauLabs/blob/next/flight/PiOS/Common/pios_flashfs_logfs.c#L873
The slot_size is 256 bytes. I think the size of the slot_header struct is 4 + 4 + 2 + 2 = 12. That gives the max size of a UAVO that can be loaded from flash as 244 bytes.
```
struct slot_header {
	enum slot_state state;
	uint32_t obj_id;
	uint16_t obj_inst_id;
	uint16_t obj_size;
} __attribute__((packed));
```
I've rewritten the comment to reflect this. I'm not sure it's worth adding a hash define for a number only used in a single place in the file.

kubark42:
This code upgrade is a clear improvement on the existing code so if you'd rather not put in the ```#define``` that's fine.

P.S. I personally think it's better never to have magic numbers because comments and code can get separated, plus magic numbers have a way of being reused over time without people noticing.

guilhermito:
LGTM

tracernz:
@kubark42 I do tend to agree with you on magic numbers. I've had a little think and decided on the cleanest way I could think to do it. Do you like 73f00c4?

mlyle:
I like it :)

PR #1853, opened 2015-08-28
UAVOBrowser: Fix crash when using buttons on non-UAVObjects

kubark42:
When a non-UAVObject is selected and any of the upload/download/save/erase buttons are used GCS crashes because the returned value is NULL. It is reasonable for an inexperienced user to click on the buttons without having first clicked on a UAVObject. Thus this crash seems unnecessary.

This removes the assert and adds an explicit return. There is no adverse consequence to directly returning, although it might be desirable to alert the user to the misuse of the buttons, since it is otherwise a silent failure.

The crash can be reproduced by clicking any of the buttons 1) without having selected any items (e.g. just after starting GCS), 2) with having selected one of the two top category ```Data``` or ```Settings``` items, or 3) having selected a category (which requires category view to be turned on).

This PR resolves the crash in all three of the above scenarios.


peabody124:
Confirmed bug and fix. Code looks good. Ok to merge, although we might want to get jenkins verifying things first for good practice.

Issue #1844, opened 2015-08-28
Tracking-- possible reset regression from #1822


mlyle:
#1822 was tested on several F4 platforms and found to be innocent.  There was a report from a user on IRC about intermittent resets on arm on AQ32.

This is likely to be something very subtle, if real.

mlyle:
Reverted for now.

PR #1855, opened 2015-08-28
Revert "Use fastheap, take 2"

mlyle:
Reverts TauLabs/TauLabs#1822

Issue #1844 ; just too much risk.

### DIFF
--- a/flight/Modules/FirmwareIAP/firmwareiap.c
+++ b/flight/Modules/FirmwareIAP/firmwareiap.c
@@ -182,7 +182,7 @@ static void FirmwareIAPCallback(UAVObjEvent* ev)
 						/* Note: Cant just wait timeout value, because first time is randomized */
 						reset_count = 0;
 						lastResetSysTime = PIOS_Thread_Systime();
-						UAVObjEvent * ev = PIOS_malloc_no_dma(sizeof(UAVObjEvent));
+						UAVObjEvent * ev = PIOS_malloc(sizeof(UAVObjEvent));
 						memset(ev,0,sizeof(UAVObjEvent));
 						EventPeriodicCallbackCreate(ev, resetTask, 100);
 						iap_state = IAP_STATE_RESETTING;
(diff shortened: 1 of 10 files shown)


lolorc:
been flying it ok with acroplus on 2 quantons today. however I now have critical issues with AH while it was working ok last week, could it be related ? otherwise I'll try to bisect it tomorrow.

peabody124:
I was going to comment before this merged. Future version should probably track the stack using in the fast help like we do in the regular heap so we know if we are running out.

mlyle:
@lolorc were those crashes/resets?

lolorc:
@mlyle oh that's a relevant question, nope it wasn't crashing/resetting as far as I observed. the quad just flips and crashes 10sec after enabling AH, I already have some logs, i'll try to record some vid tomorrow.

mlyle:
OK, that sounds super bad.  I'm willing to check out logs.

doctorseus:
I guess I am the IRC user who reported that this might be the cause for the AQ32 resets because of (possible) hard faults. (There was no proof for this claim.)
Today I tried to run a fw build for the AQ32 target based on f697631dadb1716208bf8cafea5ea84433915319 (latest at this point) and applied the fast heap patch on top (52caa92cf1ba9d2681daa6aef1e22d3995910ed1). There is no reset on the AQ32 target so far!
I do not have any idea what could have been the cause for the resets at this point.

mlyle:
@doctorseus -- thanks for your report.  Also @pug398 had unexpected resets.

Valuable information would be what previous hash you ran on before without reset.  

I would suggest you fly without fastheap for now.  It's a small win in additional free memory, but it's not going to be in for this release.  And if we have an unrelated reset bug it would be nice to know.  That is, right now I don't really care whether fastheap is OK-- I want to know if we have any regression outside of fastheap.  We have had no evidence of any, but the evidence pointing to fastheap is not unambiguous.

PR #1857, opened 2015-08-28
txpid: allow tuning of vtolpathfollowersettings

mlyle:
Inspired by #1840.  Altered to build on all targets.

### DIFF
--- a/flight/Modules/TxPID/txpid.c
+++ b/flight/Modules/TxPID/txpid.c
@@ -48,7 +48,10 @@
 #include "accessorydesired.h"
 #include "manualcontrolcommand.h"
 #include "stabilizationsettings.h"
+#ifdef UAVOBJ_INIT_vtolpathfollowersettings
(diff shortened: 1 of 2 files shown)


flight/Modules/TxPID/txpid.c
```diff
@@ -48,6 +48,10 @@
 #include "accessorydesired.h"
 #include "manualcontrolcommand.h"
 #include "stabilizationsettings.h"
+#ifdef UAVOBJ_INIT_vtolpathfollowersettings
```
guilhermito: Why is this needed?

mlyle: It's needed on the other lines to not try and do things with objects that don't exist on non-navigation targets.  It could strictly be removed up there on that first one... but then we wouldn't get a compile time error if someone does something with vtolpathfollowersettings outside a similar #ifdef.  (It'd show up at link time instead and be a lot more confusing).

Issue #1848, opened 2015-08-28
Auto-detection of DSMX channel resolution fails for walkera devo10 in 9-channel mode


ekaszubski:
Hardware
- 480-class collective-pitch helicopter, 120 degree swash
- sparky2
- DSMX satellite receiver (SPM9645) bound in slave or master mode, 11-bit channels, connected to sparky2 3-pin receiver header
- Walkera devo10 with deviation firmware in 9-channel DSMX mode

Issue
- auto-detection fails; DSM_RESOLUTION_MASK does not correctly detect 11-bit mode
- sparky2 falls back to 10-bit mode, which fails miserably

Solution
- force resolution to 11-bit mode in dsm_resolution (pios_dsm.c)

Extended solution
- allow user to override DSM resolution via uavobjects
- add DSMResolutionMode to manualcontrolsettings
- options are Autodetect, Force 10-bit, Force 11-bit (default Autodetect)
- I have a branch with this solution implemented and heavily tested; should I create a pull request?

mlyle:
I like the idea of being able to force it-- please send a PR for this release.

I do think the auto-detection we have requires you to bind using sparky-- usually 4 pulses I think (it's actually 5 pulses but one pulse comes from it floating high in the bootloader :/ )

As to the other PRs-- really neat stuff.  I think they need a little more discussion.  We are coming up to a release though ( https://github.com/TauLabs/TauLabs/wiki/Endomorphism-(September)-release ) and have crossed the deadline for high risk PRs to be opened.  We've also incurred more risk than I'd like in general so it will probably be a small amount of time before these ideas can really be reviewed and commented upon.

Issue #1792, opened 2015-08-14
When sanity checks fail we should panic.


mlyle:
Through the code I see a lot of things like:

        if (PIOS_RFM22B_Validate(rfm22b_dev)) {
                return rfm22b_dev->deviceID;
        }

If we have a garbage pointer, we should stop running.  Something has badly gone wrong and we shouldn't continue to execute flight code with undefined results.   Comments?

peabody124:
Panic in the sense of put a high priority task into a tight spin loop and block the rest of the RTOS? I think in a debug compile that is appropriate. For a release compile, we should try and keep the system up as much as possible.

That being said when basic assumptions like this fail, who knows how long we are going to keep running. But planning for essentially segfault is maybe not the best option.

mlyle:
I think we should reset, stop motors, etc.  It's important to fail as fast as possible in a deterministic way instead of trying to go forward.

When I've worked on far less safety critical stuff in the past, this has always been teh

Normal failsafe, we at least know the basic situation-- we've lost the receiver or the stuff from the receiver doesn't make sense anymore.  Here, we've lost the receiver and unknown things have happened to our system state.

If pointers point to other stuff, random things have been written to sections of memory, or power supplies have sagged and we've lost some RAM contents, and any variable we touch could be tainted.  We could be acting on arbitrary things, sending pulses to the wrong place, who knows what..

PR #1858, opened 2015-08-28
Implemented DSM resolution override

ekaszubski:
DSM: Fix issue #1848 
- Added new DSMResolutionMode field to manualcontrolsettings; options are Autodetect, Force 10-bit, Force 11-bit (default Autodetect)
- Implemented resolution mode switch in pios_dsm.c | dsm_resolution()

### DIFF
--- a/flight/PiOS/Common/pios_dsm.c
+++ b/flight/PiOS/Common/pios_dsm.c
@@ -32,6 +32,7 @@
 /* Project Includes */
 #include "pios.h"
 #include "pios_dsm_priv.h"
+#include "manualcontrolsettings.h"
 
 #if defined(PIOS_INCLUDE_DSM)
 
@@ -229,7 +230,24 @@ int PIOS_DSM_UnrollChannels(struct pios_dsm_dev *dsm_dev)
 	// If no stream type has yet been detected, then try to probe for it
 	// this should only happen once per power cycle
 	if (dsm_dev->resolution == DSM_UNKNOWN) {
-		dsm_dev->resolution = PIOS_DSM_DetectResolution(state->received_data);
+
+		// check user settings to determine which resolution mode to use
+		uint8_t dsm_resolution_mode;
+		ManualControlSettingsDSMResolutionModeGet( &dsm_resolution_mode );
+
+		switch( dsm_resolution_mode )
+		{
+		case MANUALCONTROLSETTINGS_DSMRESOLUTIONMODE_AUTODETECT:
+			dsm_dev->resolution = PIOS_DSM_DetectResolution(state->received_data);
+			break;
+		case MANUALCONTROLSETTINGS_DSMRESOLUTIONMODE_FORCE10BIT:
+			dsm_dev->resolution = DSM_10BIT;
+			break;
+		case MANUALCONTROLSETTINGS_DSMRESOLUTIONMODE_FORCE11BIT:
+			dsm_dev->resolution = DSM_11BIT;
+			break;
+		}
+
 	}
 
 	/* Stream type still not detected */
--- a/flight/PiOS/Common/pios_dsm.c
+++ b/flight/PiOS/Common/pios_dsm.c
@@ -32,7 +32,8 @@
 /* Project Includes */
 #include "pios.h"
 #include "pios_dsm_priv.h"
-#include "manualcontrolsettings.h"
+#include <uavobjectmanager.h>
+#include <manualcontrolsettings.h>
 
 #if defined(PIOS_INCLUDE_DSM)
 
--- a/flight/PiOS/Common/pios_dsm.c
+++ b/flight/PiOS/Common/pios_dsm.c
@@ -319,6 +319,11 @@ int32_t PIOS_DSM_Init(uintptr_t *dsm_id,
 	uint8_t num_pulses = 0;
 
 	// check user settings to determine which resolution mode to use
+	// if invalid mode selected, bail
+	if (mode > HWSHARED_DSMXMODE_BIND10PULSES)
+		return -1;
+
+	// set resolution or bind mode depending on user selection
 	switch (mode)
 	{
 	case HWSHARED_DSMXMODE_AUTODETECT:
@@ -340,8 +345,6 @@ int32_t PIOS_DSM_Init(uintptr_t *dsm_id,
 	case HWSHARED_DSMXMODE_BIND10PULSES:
 		num_pulses = 3 + mode - HWSHARED_DSMXMODE_BIND3PULSES;
 		break;
-	default:
-		return -1;
 	}
 
 	/* Bind the receiver if requested */
--- a/flight/PiOS/Common/pios_dsm.c
+++ b/flight/PiOS/Common/pios_dsm.c
@@ -32,8 +32,6 @@
 /* Project Includes */
 #include "pios.h"
 #include "pios_dsm_priv.h"
-#include <uavobjectmanager.h>
-#include <manualcontrolsettings.h>
 
 #if defined(PIOS_INCLUDE_DSM)
 
@@ -104,7 +102,7 @@ static bool PIOS_DSM_Validate(struct pios_dsm_dev *dsm_dev)
 }
 
 /* Try to bind DSMx satellite using specified number of pulses */
-static void PIOS_DSM_Bind(struct pios_dsm_dev *dsm_dev, uint8_t bind)
+static void PIOS_DSM_Bind(struct pios_dsm_dev *dsm_dev, uint8_t num_pulses)
 {
 	const struct pios_dsm_cfg *cfg = dsm_dev->cfg;
 
@@ -116,8 +114,8 @@ static void PIOS_DSM_Bind(struct pios_dsm_dev *dsm_dev, uint8_t bind)
 #endif
 
 	/* just to limit bind pulses */
-	if (bind > 10)
-		bind = 10;
+	if (num_pulses > 10)
+		num_pulses = 10;
 
 	GPIO_Init(cfg->bind.gpio, (GPIO_InitTypeDef*)&cfg->bind.init);
 
@@ -127,7 +125,7 @@ static void PIOS_DSM_Bind(struct pios_dsm_dev *dsm_dev, uint8_t bind)
 	/* on CC works up to 140ms, guess bind window is around 20-140ms after power up */
 	while (((float) PIOS_DELAY_GetRaw() / (float) PIOS_SYSCLK) < 0.02f);
 
-	for (int i = 0; i < bind ; i++) {
+	for (int i = 0; i < num_pulses ; i++) {
 		/* RX line, drive low for 120us */
 		GPIO_ResetBits(cfg->bind.gpio, cfg->bind.init.GPIO_Pin);
 		PIOS_DELAY_WaituS(120);
@@ -231,24 +229,7 @@ int PIOS_DSM_UnrollChannels(struct pios_dsm_dev *dsm_dev)
 	// If no stream type has yet been detected, then try to probe for it
 	// this should only happen once per power cycle
 	if (dsm_dev->resolution == DSM_UNKNOWN) {
-
-		// check user settings to determine which resolution mode to use
-		uint8_t dsm_resolution_mode;
-		ManualControlSettingsDSMResolutionModeGet( &dsm_resolution_mode );
-
-		switch( dsm_resolution_mode )
-		{
-		case MANUALCONTROLSETTINGS_DSMRESOLUTIONMODE_AUTODETECT:
-			dsm_dev->resolution = PIOS_DSM_DetectResolution(state->received_data);
-			break;
-		case MANUALCONTROLSETTINGS_DSMRESOLUTIONMODE_FORCE10BIT:
-			dsm_dev->resolution = DSM_10BIT;
-			break;
-		case MANUALCONTROLSETTINGS_DSMRESOLUTIONMODE_FORCE11BIT:
-			dsm_dev->resolution = DSM_11BIT;
-			break;
-		}
-
+		dsm_dev->resolution = PIOS_DSM_DetectResolution(state->received_data);
 	}
 
 	/* Stream type still not detected */
@@ -320,7 +301,7 @@ int32_t PIOS_DSM_Init(uintptr_t *dsm_id,
 		      const struct pios_dsm_cfg *cfg,
 		      const struct pios_com_driver *driver,
 		      uintptr_t lower_id,
-		      uint8_t bind)
+		      HwSharedDSMxModeOptions mode)
 {
 	PIOS_DEBUG_Assert(dsm_id);
 	PIOS_DEBUG_Assert(cfg);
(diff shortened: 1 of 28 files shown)


mlyle:
LGTM.

mlyle:
Only thought is i'm not positive this belongs in manualcontrolsettings.

ekaszubski:
@mlyle agreed, it doesn't feel quite right, but I wasn't sure where else to put it; manualcontrolsettings is the closest thing to receiversettings I could find. Otherwise we could put it in systemsettings or in hw<board>settings, but the latter would create a bunch of redundant fields in multiple UAVOs and the former still doesn't feel quite right.

mlyle:
@peabody124 thoughts?

peabody124:
Let me look at the code. @ekaszubski can you elaborate in the text of the PR what it does. I know the issue says this, but if the text is in the PR then that will end up in the merge log which is nice.

flight/PiOS/Common/pios_dsm.c
```diff
@@ -229,7 +229,24 @@ int PIOS_DSM_UnrollChannels(struct pios_dsm_dev *dsm_dev)
 	// If no stream type has yet been detected, then try to probe for it
 	// this should only happen once per power cycle
 	if (dsm_dev->resolution == DSM_UNKNOWN) {
-		dsm_dev->resolution = PIOS_DSM_DetectResolution(state->received_data);
+
+		// check user settings to determine which resolution mode to use
+		uint8_t dsm_resolution_mode;
+		ManualControlSettingsDSMResolutionModeGet( &dsm_resolution_mode );
```
peabody124: We try to avoid the design pattern of pios drivers are reaching into UAVObjects. It would be better to move this into the part of the "userspace" code when the PIOS_DSM_Init is called and pass in a mode enum there.

peabody124:
I'm bummed that this is necessary (and would like a log file still to add to units tests) but if the stream cannot be auto detected this is a good fallback option. Also agree with @mlyle in MCS isn't great but also can't think of a better place.

peabody124:
Also I would like to hear from 1 or 2 ppl that this doesn't break existing setups, but I don't believe it will.

An alternative way of representign this that just popped into my mind. In the Hw<BoardName> files there is the DSMBind field for each. It would be better for consistency to keep other DSM Settings in the HwBoard file too. This might be well done with MLyle's new fancy UAVO enum sharing too.

We could have DSMMode --> Autodetect, 10bit, 11bit, bind1, bind2, bind3, ... bind11. Then one field would entirely specify the behavior.

ekaszubski:
@peabody124 Updated PR message. Looking into passing value through init. I have a sparky2 and have been using this patch for a few weeks with no issues.

ekaszubski:
@peabody124 My only issue with passing the value through init is that the PIOS_DSM_Init API has to change for all FCs. If you're okay with that then I'll go that route (I have no issue changing all of the references).

peabody124:
@ekaszubski yeah that's better to change all targets (i'd do one commit per target to keep it clean). They all should support this if it is necessary.

ekaszubski:
Fixed compilation issue with aq32 and colibri: added include for uavobjectmanager.h

guilhermito:
Code looks good to me but I would really like to see a binary dump of the DSM stream to understand why this is needed.

mlyle:
At the time we merged the other autodetect code we knew that if we didn't use the right set of binding pulses to bind the satellite as a slave (as opposed to master) it wouldn't work.  @peabody124 asked it to be taught to fall back and use our previous autodetection algorithm (channel order) for some cases..   commit 408e5cd in #1561 .  But the channel order detection algorithm is a hack.  There may be a better way, but for now allowing the user to override seems sane.

ekaszubski:
@peabody124 so I tried pulling the ManualControlSettingsDSMResolutionModeGet() function out of pios_dsm.c and tested it in a couple other places including PiOS/Common/pios_hal.c and targets/sparky2/fw/pios_board.c and it always crashes; it seems to only work in pios_dsm.c because it doesn't get called until later (once data is being received). Do I need to call some function to initialize all the UAVOs prior to calling any of the Get() functions?

ekaszubski:
@peabody124 Implemented DSMxMode for all relevant targets
- excluded discoveryf4 since it has no DSM machinery in its firmware
- excluded taulink/pipxtreme since it has no DSM machinery in its firmware

flight/PiOS/Common/pios_hal.c
```diff
@@ -313,7 +313,7 @@ void PIOS_HAL_ConfigurePort(HwSharedPortTypesOptions port_type,
 /* TODO: future work to factor most of these away */
 	const struct pios_usart_cfg *usart_dsm_hsum_cfg,
 	const struct pios_dsm_cfg *dsm_cfg,
-	int dsm_bind,
+	uint8_t dsm_mode,
```
mlyle: should be the hwshared enum type

flight/PiOS/Common/pios_hal.c
```diff
@@ -220,12 +220,12 @@ void PIOS_HAL_ConfigureCom(const struct pios_usart_cfg *usart_port_cfg,
...
  * @param[in] dsm_cfg Configuration for DSM on this target
  * @param[in] usart_com_driver The COM driver for this USART
- * @param[in] bind A number of binding pulses to transmit on this port.
+ * @param[in] mode Mode in which to operate DSM driver; encapsulates binding
  */
 static void PIOS_HAL_ConfigureDSM(const struct pios_usart_cfg *usart_dsm_cfg,
                 const struct pios_dsm_cfg *dsm_cfg,
                 const struct pios_com_driver *usart_com_driver,
-		int bind)
+		uint8_t mode)
```
mlyle: should also be the hwshared enum type.  Also I notice this part of the file is already indented wrong and needs a reindent in the future.

flight/PiOS/Common/pios_dsm.c
```diff
@@ -230,6 +232,7 @@ int PIOS_DSM_UnrollChannels(struct pios_dsm_dev *dsm_dev)
 	// this should only happen once per power cycle
 	if (dsm_dev->resolution == DSM_UNKNOWN) {
 		dsm_dev->resolution = PIOS_DSM_DetectResolution(state->received_data);
+
```
mlyle: kill new blank line

flight/targets/sparky2/fw/pios_board.c
```diff
@@ -472,8 +472,8 @@ void PIOS_Board_Init(void) {
 	uint8_t hw_rcvrport;
 	HwSparky2RcvrPortGet(&hw_rcvrport);
 
-	if (bdinfo->board_rev != BRUSHEDSPARKY_V0_2) {
-		hw_DSMxBind = 0; /* Do not try to bind through XOR */
+	if (bdinfo->board_rev != BRUSHEDSPARKY_V0_2 && hw_DSMxMode >= HWSHARED_DSMXMODE_BIND3PULSES) {
```
mlyle: should be the HWSPARKY2 version for consistency with its actual type.

flight/targets/sparky2/fw/pios_board.c
```diff
@@ -426,8 +426,8 @@ void PIOS_Board_Init(void) {
 #endif	/* PIOS_INCLUDE_USB */
 
 	/* Configure IO ports */
-	uint8_t hw_DSMxBind;
-	HwSparky2DSMxBindGet(&hw_DSMxBind);
+	uint8_t hw_DSMxMode;
```
mlyle: each of these should be the target-specific enum type.

ekaszubski:
@mlyle Thanks for the comments; fixing now. Will also look out for more comments :)

flight/PiOS/Common/pios_dsm.c
```diff
@@ -301,7 +304,7 @@ int32_t PIOS_DSM_Init(uintptr_t *dsm_id,
 		      const struct pios_dsm_cfg *cfg,
 		      const struct pios_com_driver *driver,
 		      uintptr_t lower_id,
-		      uint8_t bind)
+		      uint8_t mode)
```
ekaszubski: @mlyle Should this be HwSharedDSMxModeOptions too?

ekaszubski:
@mlyle Made requested changes; extrapolated fixes for sparky2 to other relevant targets.

mlyle:
We should test binding on at least one platform and basic DSM functionality on at least a couple (existing autodetect rigs) before merge.

mlyle:
@ekaszubski have you seen positive results on any platform?  Have you tried binding?

ekaszubski:
@mlyle I've tested all my PRs on sparky2. Haven't tried a bind (my hardware is set up to use the RCVR port which can't bind) but autodetect/10-bit/11-bit modes all function the same way they did prior to the PR. The binding function itself was not modified, so I guess I can use the debug console to verify that the correct value is still being passed to it when executing a bind.

peabody124:
I think I have a sparky 1 somewhere with an orange rx. I can test binding tonight if home at a reasonable time.

flight/PiOS/Common/pios_dsm.c
```diff
@@ -316,9 +316,37 @@ int32_t PIOS_DSM_Init(uintptr_t *dsm_id,
...
+	case HWSHARED_DSMXMODE_BIND4PULSES:
+	case HWSHARED_DSMXMODE_BIND5PULSES:
+	case HWSHARED_DSMXMODE_BIND6PULSES:
+	case HWSHARED_DSMXMODE_BIND7PULSES:
+	case HWSHARED_DSMXMODE_BIND8PULSES:
+	case HWSHARED_DSMXMODE_BIND9PULSES:
+	case HWSHARED_DSMXMODE_BIND10PULSES:
+		num_pulses = 3 + mode - HWSHARED_DSMXMODE_BIND3PULSES;
+		break;
+	default:
```
mlyle: Just an FYI:  I'm ambivalent about default: clauses in switch statements.  They may catch (are not required to catch per C99 standard) when invalid values are passed in.  But they also squelch a compiler warning about not naming all enum values.

flight/targets/sparkybgc/fw/pios_board.c
```diff
@@ -180,7 +180,7 @@ static void PIOS_Board_configure_com (const struct pios_usart_cfg *usart_port_cf
 #ifdef PIOS_INCLUDE_DSM
 static void PIOS_Board_configure_dsm(const struct pios_usart_cfg *pios_usart_dsm_cfg, const struct pios_dsm_cfg *pios_dsm_cfg,
 		const struct pios_com_driver *pios_usart_com_driver,
-		ManualControlSettingsChannelGroupsOptions channelgroup,uint8_t *bind)
+		ManualControlSettingsChannelGroupsOptions channelgroup,HwSparkyBGCDSMxModeOptions *mode)
```
mlyle: Heh I really need to get all the other targets ported to PIOS_HAL.

flight/PiOS/Common/pios_dsm.c
ekaszubski: @mlyle Any thoughts on how to resolve this? Range-check on `mode` maybe?

mlyle: It can be left alone.  Just pointing out the trade off for future consideration

kubark42: Our flags turn compiler warnings into build errors. So either the switch..case needs ```default```, or it needs an exhaustive list of the values in the enum or else the build will fail. (Although I can't comment if this is specific to ARM GCC or is in a C standard).

Personally, I like to see an exhaustive list, because the compiler protects against adding new values to an enum and accidentally forgetting to handle them.

ekaszubski: @kubark42 I have a version with the range check + exhaustive list - default; unless you guys are going to merge this within the next 30 minutes, I'm going to commit that version just so it's done right.

ekaszubski:
Fixed commit message; `DMS` -> `DSM`

flight/PiOS/Common/pios_dsm.c
```diff
@@ -316,9 +316,40 @@ int32_t PIOS_DSM_Init(uintptr_t *dsm_id,
 	/* Bind the configuration to the device instance */
 	dsm_dev->cfg = cfg;
 
+	uint8_t num_pulses = 0;
+
+	// check user settings to determine which resolution mode to use
+	// if invalid mode selected, bail
+	if (mode > HWSHARED_DSMXMODE_BIND10PULSES)
```
kubark42: I'm hoping @mlyle or @solidgoldbomb can chime in here. I see what you're doing and I wouldn't argue with the why: you're covering a case where the variable has potentially received a value not in the enum. In theory, the compiler should prevent this from occurring because you could not assign ```mode = 5;```, but I'm betting you could force it with ```mode = (HwSharedDSMxModeOptions) 5;```. And I'm sure you could force it even more easily with pointer foo.

So it makes a lot of sense to be on the lookout for this type of error, but this is the first time I've seen this range check applied so there's not exactly a long history of handling this. Even then, there is no guarantee that the enum is sequential, and that -1 is not a value. Without the ```default``` case, I don't see a way to *guarantee* goodness here, except via code review. And while I'm not a big fan of getting rid of ```default``` because the compiler catches the more likely error-- not updating all switches to handle a new enum-- even though it it could hide an insidious error, I'd love to hear from some Mike or Stacey.

ekaszubski: @kubark42 also, see #1886

mlyle: Basically all variants of these patterns are bad in one way or another.

1. The case where we have a default prevents the compiler from emitting a message about uncovered case elements.  That warning (that becomes an error) is super nice which is why I'm not a fan of default in general.
2. Enums aren't guaranteed in C to be in range.  They're more like "a collection of more tightly scoped constants."  There's been a long history of using them for flags, which the standard recognizes.  e.g. 
        enum flags { A=1, B=2, C=4, D=8 };

        enum flags test = A | B;

        printf("%d\n", test);
is completely legal code.  All the compiler promises is to pick a underlying type that is big enough to store all the enumeration values, and to give us warnings if we don't cover given values in switch statements.  (It's assumed if we're using the enum in a switch we really mean a true enumeration of possible values).
3. If we hit an uncovered case-- an invalid enum value with missing default, undefined behavior occurs.  So really we need to validate enum ranges somewhere.
4. Do we do this loading from flash / over the telemetry link?  That would be best rather than contaminating our code with lots of checks all over the place.

So my end thought is "meh."  There's flaws with just about anything you can do.  Also, I strongly believe in #1792 .  If we do validation, we should put it in a sufficient number of places to catch all the inputs to the code, and then we should A) not spend a lot of effort checking values afterwards, and B) where we do sprinkle in assertions afterwards, they should be fatal / cause the system to fail fast rather than do undefined things.

ekaszubski: @mlyle For 2), I'm mainly thinking about these enums in the context of generated UAVOs, in which case we have complete control over the starting value and increment of each element and can ensure they're generated in such a way that a range check will always work. I agree with 4).

peabody124:
I can confirm Bind4 worked for me using an OrangeRX satellite and DX7. I was able to parse the data in auto detect and 10 bit mode, and not in 11 bit mode (as expected - or more specifically it picked up a subset of channels with odd values).

mlyle:
@peabody124 thanks so much for the test!

peabody124:
Bind3 and auto detect also worked, interestingly

Issue #1850, opened 2015-08-28
Collective pitch functionality appears broken


ekaszubski:
Hardware
- 480-class collective-pitch helicopter, 120 degree swash
- sparky2
- DSMX satellite receiver (SPM9645)
- Walkera devo10 with deviation firmware in 9-channel DSMX mode

Issue
- My transmitter is set up to output full PWM range [-100,100] for collective, as with all other channels
- On the input menu in GCS, I have the collective neutral set to the zero-stick position, which is approximately half-way between the min and the max PWM
- Vehicle type is set to helicopter, 120 degree CCPM, collective pass-through
- Collective curve in GCS is 1-to-1 (or more specifically, [-1,1])
- For positive collective stick values, the swashplate works as expected
- For negative collective stick values, the swashplate snaps to approximately zero collective

Partial solution 1
- in GCS, under input, set neutral point of collective to min collective channel value
- now, collective works as expected, but when the transmitter is shut off, the swashplate snaps to full negative pitch, which can cause catastrophic results in-flight

Partial solution 2
- based on the MixerCurve function implementation (Actuator/actuator.c), setting the first point in the collective curve to a value <=-1 will cause the value to be passed through as-is
- it appears this must be done manually; not sure what effect the "collective passthrough" option under vehicle setup actually has; maybe it was supposed to do this?
- this works as a pass-through, but causes the remaining points in the collective curve to be ignored

Full solution
- Upon further inspection, it appears that the MixerCurve function (Actuator/actuator.c) expects its input to be on the rage [0,1], which make sense for throttle curves; however, the collective curve *should* be defined on [-1,1], should be configurable, and should default to a position of 0 (or some other configurable value) when the transmitter is not connected
- MixerCurve should be renamed to ThrottleCurve; it should continue to handle input in [0,1]
- A new function, CollectiveCurve, should be created to handle input in [-1,1]
- ThrottleCurve should be replaced with CollectiveCurve for MIXERSETTINGS_CURVE2SOURCE_COLLECTIVE
- I have this solution implemented and extensively tested in a separate branch

tracernz:
Nice catch. I think our heli support is very lightly used (and I haven't yet tried it myself) so it's nice that someone is actually testing it. Feel free to submit a PR with your fix, it sounds good to me. :+1:

mlyle:
Please open a PR on this one. :)

mlyle:
Fixed #1859.

PR #1859, opened 2015-08-28
Fixed mixer curve for collective input in actuator.c

ekaszubski:
Mixer: Fix issue #1850 
- Modified MixerCurve function in Actuator/actuator.c to accept an arbitrary input range
- Added alias functions ThrottleCurve and CollectiveCurve for MixerCurve with input range [0,1] and [-1,1] respectively
- No changes to UAVOs or GCS are necessary since the input range of the curves is never defined

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -248,16 +248,16 @@ static void actuatorTask(void* parameters)
 		AccessoryDesiredData accessory;
 		switch(mixerSettings.Curve2Source) {
 			case MIXERSETTINGS_CURVE2SOURCE_THROTTLE:
-				curve2 = ThrottleCurve(desired.Throttle, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = CollectiveCurve(desired.Throttle, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_ROLL:
-				curve2 = ThrottleCurve(desired.Roll, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = CollectiveCurve(desired.Roll, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_PITCH:
-				curve2 = ThrottleCurve(desired.Pitch, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = CollectiveCurve(desired.Pitch, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_YAW:
-				curve2 = ThrottleCurve(desired.Yaw, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = CollectiveCurve(desired.Yaw, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_COLLECTIVE:
 				ManualControlCommandCollectiveGet(&curve2);
@@ -270,7 +270,7 @@ static void actuatorTask(void* parameters)
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY4:
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY5:
 				if(AccessoryDesiredInstGet(mixerSettings.Curve2Source - MIXERSETTINGS_CURVE2SOURCE_ACCESSORY0,&accessory) == 0)
-					curve2 = ThrottleCurve(accessory.AccessoryVal, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+					curve2 = CollectiveCurve(accessory.AccessoryVal, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				else
 					curve2 = 0;
 				break;
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -78,7 +78,9 @@ static volatile bool mixer_settings_updated;
 static void actuatorTask(void* parameters);
 static float scaleChannel(float value, float max, float min, float neutral);
 static void setFailsafe(const ActuatorSettingsData * actuatorSettings, const MixerSettingsData * mixerSettings);
-static float MixerCurve(const float throttle, const float* curve, uint8_t elements);
+static float MixerCurve(const float input, const float* curve, uint8_t elements, const float input_min, const float input_max);
+static float ThrottleCurve(const float input, const float* curve, uint8_t elements);
+static float CollectiveCurve(const float input, const float* curve, uint8_t elements);
 static bool set_channel(uint8_t mixer_channel, float value, const ActuatorSettingsData * actuatorSettings);
 static void actuator_update_rate_if_changed(const ActuatorSettingsData * actuatorSettings, bool force_update);
 static void MixerSettingsUpdatedCb(UAVObjEvent * ev);
@@ -239,29 +241,28 @@ static void actuatorTask(void* parameters)
 		bool positiveThrottle = desired.Throttle >= 0.00f;
 		bool spinWhileArmed = actuatorSettings.MotorsSpinWhileArmed == ACTUATORSETTINGS_MOTORSSPINWHILEARMED_TRUE;
 
-		float curve1 = MixerCurve(desired.Throttle,mixerSettings.ThrottleCurve1,MIXERSETTINGS_THROTTLECURVE1_NUMELEM);
-		
+		float curve1 = ThrottleCurve(desired.Throttle,mixerSettings.ThrottleCurve1,MIXERSETTINGS_THROTTLECURVE1_NUMELEM);
+
 		//The source for the secondary curve is selectable
 		float curve2 = 0;
 		AccessoryDesiredData accessory;
 		switch(mixerSettings.Curve2Source) {
 			case MIXERSETTINGS_CURVE2SOURCE_THROTTLE:
-				curve2 = MixerCurve(desired.Throttle,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Throttle,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_ROLL:
-				curve2 = MixerCurve(desired.Roll,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Roll,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_PITCH:
-				curve2 = MixerCurve(desired.Pitch,mixerSettings.ThrottleCurve2,
+				curve2 = ThrottleCurve(desired.Pitch,mixerSettings.ThrottleCurve2,
 				MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_YAW:
-				curve2 = MixerCurve(desired.Yaw,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Yaw,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_COLLECTIVE:
 				ManualControlCommandCollectiveGet(&curve2);
-				curve2 = MixerCurve(curve2,mixerSettings.ThrottleCurve2,
-				MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = CollectiveCurve(curve2,mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY0:
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY1:
@@ -270,7 +271,7 @@ static void actuatorTask(void* parameters)
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY4:
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY5:
 				if(AccessoryDesiredInstGet(mixerSettings.Curve2Source - MIXERSETTINGS_CURVE2SOURCE_ACCESSORY0,&accessory) == 0)
-					curve2 = MixerCurve(accessory.AccessoryVal,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+					curve2 = ThrottleCurve(accessory.AccessoryVal,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				else
 					curve2 = 0;
 				break;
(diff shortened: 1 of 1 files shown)


flight/Modules/Actuator/actuator.c
```diff
@@ -444,6 +445,43 @@ static float MixerCurve(const float throttle, const float* curve, uint8_t elemen
 	return curve[idx1] * (1.0f - scale) + curve[idx2] * scale;
 }
 
+/* curve is defined in N intervals connecting N+1 points
```
mlyle: Can this to be factored to be common code and take the range as parameters to avoid duplication of code?

ekaszubski:
@mlyle Updated implementation; MixerCurve is the most generic; accepts arbitrary input range, ThrottleCurve is the same as old mixer curve; calls MixerCurve with input range [0,1], and CollectiveCurve calls MixerCurve with input range [-1,1])

mlyle:
Looks great.  Can you test this implementation please?  With positive test I can merge.

mlyle:
I will bench test.

ekaszubski:
@mlyle Verifying now

ekaszubski:
@mlyle Verified; throttle and collective both work as expected

flight/Modules/Actuator/actuator.c
```diff
@@ -239,29 +241,28 @@ static void actuatorTask(void* parameters)
...
-		
+		float curve1 = ThrottleCurve(desired.Throttle,mixerSettings.ThrottleCurve1,MIXERSETTINGS_THROTTLECURVE1_NUMELEM);
+
 		//The source for the secondary curve is selectable
 		float curve2 = 0;
 		AccessoryDesiredData accessory;
 		switch(mixerSettings.Curve2Source) {
 			case MIXERSETTINGS_CURVE2SOURCE_THROTTLE:
-				curve2 = MixerCurve(desired.Throttle,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Throttle,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
```
kubark42: Fix spacing between arguments.

flight/Modules/Actuator/actuator.c
```diff
@@ -239,29 +241,28 @@ static void actuatorTask(void* parameters)
...
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_ROLL:
-				curve2 = MixerCurve(desired.Roll,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Roll,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
```
kubark42: Fix spacing between arguments.

flight/Modules/Actuator/actuator.c
```diff
@@ -239,29 +241,28 @@ static void actuatorTask(void* parameters)
...
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_PITCH:
-				curve2 = MixerCurve(desired.Pitch,mixerSettings.ThrottleCurve2,
+				curve2 = ThrottleCurve(desired.Pitch,mixerSettings.ThrottleCurve2,
```
kubark42: Fix spacing between arguments.

flight/Modules/Actuator/actuator.c
```diff
@@ -239,29 +241,28 @@ static void actuatorTask(void* parameters)
...
 				MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_YAW:
-				curve2 = MixerCurve(desired.Yaw,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Yaw,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
```
kubark42: Fix spacing between arguments.

flight/Modules/Actuator/actuator.c
```diff
@@ -239,29 +241,28 @@ static void actuatorTask(void* parameters)
...
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_COLLECTIVE:
 				ManualControlCommandCollectiveGet(&curve2);
-				curve2 = MixerCurve(curve2,mixerSettings.ThrottleCurve2,
-				MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = CollectiveCurve(curve2,mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
```
kubark42: Fix spacing between arguments.

flight/Modules/Actuator/actuator.c
```diff
@@ -270,7 +271,7 @@ static void actuatorTask(void* parameters)
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY4:
 			case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY5:
 				if(AccessoryDesiredInstGet(mixerSettings.Curve2Source - MIXERSETTINGS_CURVE2SOURCE_ACCESSORY0,&accessory) == 0)
-					curve2 = MixerCurve(accessory.AccessoryVal,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+					curve2 = ThrottleCurve(accessory.AccessoryVal,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
```
kubark42: Fix spacing between arguments.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +414,56 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
+ * this is convenient for throttle, since the neutral value is used as a failsafe and would thus shut off the motor
+ * however, this makes no sense for collective, since the collective neutral point is not the min point for most vehicles with collective pitch
+ * instead, we need to accept input in [-1,1] so that the neutral point may be set arbitrarily within the typical channel input range, which is [-1,1]
  */
-static float MixerCurve(const float throttle, const float* curve, uint8_t elements)
+static float MixerCurve( float const input, float const * curve, uint8_t elements, const float input_min, const float input_max )
 {
-	float scale = throttle * (float) (elements - 1);
+	// shift our input [min,max] into the typical range [0,1]
+	float scale = ( ( input - input_min ) / ( input_max - input_min ) ) * (float) (elements - 1);
```
kubark42: Out of curiosity, is the compiler smart enough to optimize this when the inputs are constants?

(It doesn't change mergeability, just an idle question.)

Also, I don't know if we have a strong style guide for equations, but I find all the whitespace makes it a little harder to read than something more compact. ```((input - input_min) / (input_max - input_min)) * (float)(elements - 1)``` puts things into logical groups a little better, IMO. But that's just opinion.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +414,56 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
-	scale -= (float)idx1; //remainder
-	if(curve[0] < -1)
-	{
-		return(throttle);
-	}
-	if (idx1 < 0)
+	// save the offset from the starting bin for linear interpolation
+	scale -= (float)idx1;
+	// if the starting bin is below the first bin
+	if( idx1 < 0 )
```
kubark42: I believe the style guide (which adheres to the [Linux kernel style guide](https://www.kernel.org/doc/Documentation/CodingStyle)) would have the ```{``` be on the same line as the ```if()```. And place a space after ```if```, but not around the ```(``` and ```)```.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +414,56 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
 	{
 		idx1 = 0; //clamp to lowest entry in table
 		scale = 0;
 	}
+	// select an ending bin (linear interpolation occurs between starting and ending bins)
 	int idx2 = idx1 + 1;
-	if(idx2 >= elements)
+	// if the ending bin bin is above the last bin
+	if( idx2 >= elements )
```
kubark42: Likewise, per style guide.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +414,56 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
 	{
 		idx2 = elements -1; //clamp to highest entry in table
-		if(idx1 >= elements)
+		// if the starting bin is above the last bin
+		if( idx1 >= elements )
```
kubark42: And one last style change here. (Realizing totally that this is an inherited problem since when we were at OP we didn't adhere to a style guide.)

kubark42:
Style questions aside, looks good. Hats off to @ekaszubski for providing excellent comments. They make the mathematical intention very easy to follow.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +414,56 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
 	return(result);
 }
 
-
 /**
- *Interpolate a throttle curve. Throttle input should be in the range 0 to 1.
- *Output is in the range 0 to 1.
+ * Interpolate a mixer curve. Input is assumed to be in the range [0,1] unless otherwise specified.
+ * Output range is defined by the curve vector
+ * curve is defined in N intervals connecting N+1 points
```
tracernz: Maybe use Doxygen comments for the parameters? e.g.
```
 * @param[in] curve Curve is defined in N intervals connecting N+1 points
```
I know only some of the code is currently documented in this way but it would be nice to have more so we can generate nice code documentation. :)

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +414,56 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
+ * throttle curve assumes input is [0,1]
+ * this means that the throttle channel neutral value is nearly the same as its min value
+ * this is convenient for throttle, since the neutral value is used as a failsafe and would thus shut off the motor
+ * however, this makes no sense for collective, since the collective neutral point is not the min point for most vehicles with collective pitch
+ * instead, we need to accept input in [-1,1] so that the neutral point may be set arbitrarily within the typical channel input range, which is [-1,1]
  */
-static float MixerCurve(const float throttle, const float* curve, uint8_t elements)
+static float MixerCurve( float const input, float const * curve, uint8_t elements, const float input_min, const float input_max )
```
tracernz: Watch out for spacing. We don't use space inside the brackets, see https://www.kernel.org/doc/Documentation/CodingStyle. There are a few places in this diff that need to be fixed up. :)

tracernz:
Ahhh, I had this open in a tab for a while before I got to it, sorry for duplicating your style comments @kubark42 . Overall this looks very good once the minor style issues are fixed up, well done. :D

flight/Modules/Actuator/actuator.c
tracernz: Agreed, I like spaces only around the operators. :+1:

mlyle: Yes, the compiler is usually smart enough.  It will fix it up if inlined, or if it notices a given variant set of args is often provided within a module, it'll generate a .constprop.n symbol that has performed constant propagation.

flight/Modules/Actuator/actuator.c
kubark42: Yes, good point @tracernz. And, again, kudos to @ekaszubski for excellent and clear documentation.

ekaszubski:
Oh man, I can deal with all the other style changes, but I really hate having the opening curly brace on the same line as the "if"; IMO it makes the code so much harder to read. I'll make the changes, but I'm dying a little inside each time I move those curly braces :/

ekaszubski:
Changed formatting and added doxygen comments as requested. Changed "elements" to "num_points" in all mixer functions for clarity.

tracernz:
Thanks! :)

FWIW: I find the mixed use of newlines before braces for functions, ifs etc. quite odd but I think there is a justification for it in that kernel style guide document.

kubark42:
> Oh man, I can deal with all the other style changes, but I really hate having the opening curly brace on the same line as the "if"; IMO it makes the code so much harder to read. I'll make the changes, but I'm dying a little inside each time I move those curly braces :/

Totally, totally agreed. And it makes it harder to comment ```if...else``` statements as well. It might have made sense back when screens where monochrome, but now... :P Still, until we agree to do a different style we're stuck with it. 

Now, if we were to go through all the code with astyle and fix all the formatting boogs, it would be an excellent opportunity to discuss change style guides as well.  ;)

mlyle:
I've always liked K&R style so I vote that way, though I've gone either way on the function-line one.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +413,80 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
+
+/**
+ * Interpolate a collective curve
+ *
+ * we need to accept input in [-1,1] so that the neutral point may be set arbitrarily within the typical channel input range, which is [-1,1]
+ *
+ * @param input The input value, in [-1,1]
+ * @param curve Array of points in the curve
+ * @param num_points Number of points in the curve
+ * @return the output value, in [0,1]
```
mlyle: Is this correct for collective, or should it be an output range of [-1,1]?

ekaszubski:
@mlyle You're right, it should be [-1,1]; fixing now :)

flight/Modules/Actuator/actuator.c
```diff
@@ -239,29 +241,27 @@ static void actuatorTask(void* parameters)
...
+				curve2 = ThrottleCurve(desired.Throttle, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_ROLL:
-				curve2 = MixerCurve(desired.Roll,mixerSettings.ThrottleCurve2,MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Roll, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
 				break;
 			case MIXERSETTINGS_CURVE2SOURCE_PITCH:
-				curve2 = MixerCurve(desired.Pitch,mixerSettings.ThrottleCurve2,
-				MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
+				curve2 = ThrottleCurve(desired.Pitch, mixerSettings.ThrottleCurve2, MIXERSETTINGS_THROTTLECURVE2_NUMELEM);
```
mlyle: Should roll/pitch/yaw be of the "collectivecurve" type because they are range -1 to 1?

peabody124:
FWIW I would love to see the MixerCurve method made into a general interpolation method (e.g. ```n_point_interp_on_domain``` and then add it to the misc_math units tests: https://github.com/TauLabs/TauLabs/blob/next/flight/tests/misc_math/unittest.cpp

ekaszubski:
@mlyle Switched to CollectiveCurve as suggested

ekaszubski:
@peabody124 I think this PR should be merged as-is (pending any other small changes) and then I can open a separate issue for moving the MixerCurve function into the math libs and adding some unit tests

mlyle:
I agree the factoring out should be separate.

mlyle:
It appears there's an issue.

https://gist.github.com/mlyle/ddbd975671d5a9f4fe02

The first few lines of output are:

```
-1.100 +0.300 +0.280
-1.090 +0.300 +0.282
-1.080 +0.300 +0.284
```

I'd expect the collectivecurve output to be +0.300 not +0.280.

ekaszubski:
@mlyle Good catch; issue was with clamping negative values. Fixed in latest commit.

peabody124:
@ekaszubski I'll defer to @mlyle ultimately. However I never like deferring important things to separate PRs as it decreases the likelihood of it being done. In the case of the actuator code it is so critical that that do the right thing I'd just be much happier having unit tests as part of it.

I haven't reviewed the interp part of this code though either, so someone needs to read that very carefully for correctness and corner cases.

Ultimately I'm definitely not wanting to discourage you! It's greater to have a productive coder interested in helicopters and you are doing great work.

ekaszubski:
@peabody124 Just to clarify, this code is identical to the old MixerCurve code, except I scale the inputs into [0,1] from [min,max]. So as long as that scaling code works correctly, the rest of it will behave exactly the same as the old MixerCurve. Even the bug with clamping to the lower bound was inherited from the old MixerCurve code.

MixerCurve clamping bug in next (d0744a442c3736f3bc6de9c5248a15b7d5a357f2):
```
Curve : [0.3, 0.4, 0.5, 0.6, 1.0]
Input | Output | Expected
-0.060 +0.276 +0.300
-0.050 +0.280 +0.300
-0.040 +0.284 +0.300
-0.030 +0.288 +0.300
-0.020 +0.292 +0.300
-0.010 +0.296 +0.300
+0.000 +0.300 +0.300
+0.010 +0.304 +0.304
+0.020 +0.308 +0.308
+0.030 +0.312 +0.312
+0.040 +0.316 +0.316
+0.050 +0.320 +0.320
+0.060 +0.324 +0.324
```

mlyle:
My personal preference on things like this is that the moving-to-library and refactor happen in separate PRs so that the diff is clear on the code change during refactor.  It can't always work out that way.  Else it requires careful work during CR to line up the old vs. new implementations.

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +413,80 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
-static float MixerCurve(const float throttle, const float* curve, uint8_t elements)
+static float MixerCurve(float const input, float const * curve, uint8_t num_points, const float input_min, const float input_max)
 {
-	float scale = throttle * (float) (elements - 1);
+	// shift our input [min,max] into the typical range [0,1]
+	float scale = ( (input - input_min) / (input_max - input_min) ) * (float) (num_points - 1);
+	// select a starting bin via truncation
 	int idx1 = scale;
-	scale -= (float)idx1; //remainder
-	if(curve[0] < -1)
```
mlyle: Does anyone know what this line was for?

flight/Modules/Actuator/actuator.c
```diff
@@ -413,37 +413,80 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
...
+ * @return the output value, in [0,1]
  */
-static float MixerCurve(const float throttle, const float* curve, uint8_t elements)
+static float MixerCurve(float const input, float const * curve, uint8_t num_points, const float input_min, const float input_max)
 {
-	float scale = throttle * (float) (elements - 1);
+	// shift our input [min,max] into the typical range [0,1]
+	float scale = ( (input - input_min) / (input_max - input_min) ) * (float) (num_points - 1);
+	// select a starting bin via truncation
 	int idx1 = scale;
```
mlyle: So the real issue is the round-towards-zero here.  Can we use floorf() and remove the scale check below?

mlyle:
Confirmed OK with current test vectors.

Made a couple of code comments.  First is I'm not sure why there was that check to just pass through throttle before and I'd like to give people a chance to chime in.  (Do you have any idea @ekaszubski ?) Second is something I slightly prefer but it's completely at @ekaszubski 's discretion.

ekaszubski:
@mlyle It looks like it allows a direct pass-through of the input value if the first point on the curve is <-1; my best guess is that it's a hack for collective pass-through or something similar

flight/Modules/Actuator/actuator.c
ekaszubski: @mlyle It looks like it allows a direct pass-through of the input value if the first point on the curve is <-1; my best guess is that it's a hack for collective pass-through or something similar

mlyle: Yah, it basically completely disables the curve.  I wonder if anyone is using it for anything.

flight/Modules/Actuator/actuator.c
ekaszubski: @mlyle I think the issue is that if scale is ever negative, then input is below input_min and needs to be clamped. I'm confused as to how `floorf` would help; what about replacing
`float scale = ( (input - input_min) / (input_max - input_min) ) * (float) (num_points - 1);`
with
`float scale = fmax( (input - input_min) / (input_max - input_min), 0.0f) * (float) (num_points - 1);`
and then removing the scale check?

mlyle: If scale is ever negative,    int idx1 = floorf(scale);   will produce -1, and then it'll flal into that next if.

ekaszubski: @mlyle OK that makes sense. BTW I tested the `fmax` version and it passes the test vector. Let me know which you think is better and I'll commit the change.

ekaszubski: @mlyle I'm just gonna commit the `fmax` version for now; I think it's cleaner overall and allows us to remove the first `if` block entirely

mlyle: It would have to be fmaxf.  I think I prefer the floor version but up to you.

mlyle:
Looks good through test vector.

Issue #1793, opened 2015-08-14
Mac OS X packaging on 10.10.x broken after move to 5.5 


mlyle:
Principal problem is we use otool to find things in our development path and that no longer works due to use of @rpath by toolchain.

    QT_DIR=$(otool -L "${APP}/Contents/MacOS/Tau Labs GCS"  | sed -n -e 's/\/QtCore\.framework.*//p' | sed -n -E 's:^.::p')

mlyle:
#1863 fixed it and merged.

PR #1863, opened 2015-08-29
make/package: correct OSX packaging

mlyle:
For various QT bugs and changes in OSX deployment between versions.

Issues #1442 / #1793


mlyle:
Confirmed packaging, welcome screen, and PFD work.  Will do a little more verification.

mlyle:
Did a little more verification.  Mergin'.

Issue #1862, opened 2015-08-29
Stack overflow on FlyingF4


ducminhkt:
I just updated firmware build on 27/8/15 (367e90b).  I have a problem when config my quad. The system health show Stack Overflow error when I turn on TX. See images below
![systemhealth](https://cloud.githubusercontent.com/assets/13902828/9559673/9dd32eb4-4e21-11e5-8101-3cfa4f1c5bb0.PNG)

Stack remaining: Before turn on TX
![1](https://cloud.githubusercontent.com/assets/13902828/9559647/2a537fe8-4e21-11e5-915f-3a8f45173d65.PNG)
Stack remaining: After turn on TX
![2](https://cloud.githubusercontent.com/assets/13902828/9559686/c072125a-4e21-11e5-978c-914767cb8240.PNG)
System Alarm
![systemalarm](https://cloud.githubusercontent.com/assets/13902828/9559690/dc6d9ae2-4e21-11e5-918d-80ade0c457ef.PNG)
System Stat
![systemstat](https://cloud.githubusercontent.com/assets/13902828/9559693/e8dc1e84-4e21-11e5-89f8-d31cf8ecf801.PNG)
My hardware
![hardwaretab](https://cloud.githubusercontent.com/assets/13902828/9559699/f90096d2-4e21-11e5-9b00-963de806f0f8.PNG)
Input config
![inputtab](https://cloud.githubusercontent.com/assets/13902828/9559704/04d286fa-4e22-11e5-862e-a5ffdaa9222b.PNG)
Task running
![taskrunning](https://cloud.githubusercontent.com/assets/13902828/9559708/0cf094d0-4e22-11e5-80e5-252421a20f54.PNG)

In Task remaining, you can see the value at Manual Control come to zero (sometimes < 60) when I turn on TX. If I boot up my Quad when turn off TX, Stack Overflow is OK, but after that I turn on TX, it Critical. 
Before update this firmware, I use version 744c79e  (8 days ago), Stack Overflow is OK, no problem.
Please show me where's my mistake, or because firmware bug ?

mlyle:
merges in this range:

    Merge pull request #1797 from mlyle/mpl-mavlinkstackdiet
    Merge pull request #1808 from mlyle/mpl-makededup
    Merge pull request #1809 from mlyle/mpl-deleteoldtrim
    Merge pull request #1814 from mlyle/mpl-removeoplinksettings
[...]

PR #1866, opened 2015-08-30
misc_math: adjust to inline nan check

mlyle:
This dramatically decreases stack usage in several tasks.

The tasks still probably need to be turned up in size.

### DIFF
--- a/flight/Libraries/math/misc_math.h
+++ b/flight/Libraries/math/misc_math.h
@@ -64,11 +64,22 @@ void vector2_rotate(const float *original, float *out, float angle);
 float cubic_deadband(float in, float w, float b, float m, float r);
 void cubic_deadband_setup(float w, float b, float *m, float *r);
 
+/* NOTE: Interesting problem regarding NaN and Inf checks. The compile optimization --fast-math
+ * does many useful things, but it also removes the ability to check if a number is a Nan, even
+ * if the check is the classic `x == x`. Thus we need to add this back in, and we do it as a
+ * function which has the optimizations turned off.
+ *
+ * Souce: http://stackoverflow.com/questions/22931147/stdisinf-does-not-work-with-ffast-math-how-to-check-for-infinity
+ */
 #if defined (__clang__) // Clang can't turn off specific optimizations, so turn them all off. This is currently only useful when compiling the simulator with Clang
-bool IS_NOT_FINITE(float x) __attribute__((optnone()));
+static inline bool IS_NOT_FINITE(float x) __attribute__((optnone()));
 #else
-bool IS_NOT_FINITE(float x) __attribute__((optimize("no-finite-math-only")));
+static inline bool IS_NOT_FINITE(float x) __attribute__((optimize("no-finite-math-only")));
 #endif
+
+static inline bool IS_NOT_FINITE(float x) {
+	return (!isfinite(x));
+}
 
 #endif /* MISC_MATH_H */
 
--- a/flight/Libraries/math/misc_math.h
+++ b/flight/Libraries/math/misc_math.h
@@ -31,6 +31,7 @@
 #ifndef MISC_MATH_H
 #define MISC_MATH_H
 
+#include <math.h>
 #include "stdint.h"
 #include "stdbool.h"
 
--- a/flight/Libraries/math/misc_math.c
+++ b/flight/Libraries/math/misc_math.c
@@ -256,17 +256,6 @@ void cubic_deadband_setup(float w, float b, float *m, float *r)
 }
 
 
-/* NOTE: Interesting problem regarding NaN and Inf checks. The compile optimization --fast-math
- * does many useful things, but it also removes the ability to check if a number is a Nan, even
- * if the check is the classic `x == x`. Thus we need to add this back in, and we do it as a
- * function which has the optimizations turned off.
- *
- * Souce: http://stackoverflow.com/questions/22931147/stdisinf-does-not-work-with-ffast-math-how-to-check-for-infinity
- */
-bool IS_NOT_FINITE(float x) {
-	return (!isfinite(x));
-}
-
 /**
  * @}
  * @}
--- a/flight/Modules/AltitudeHold/altitudehold.c
+++ b/flight/Modules/AltitudeHold/altitudehold.c
@@ -59,7 +59,7 @@
 
 // Private constants
 #define MAX_QUEUE_SIZE 4
-#define STACK_SIZE_BYTES 600
+#define STACK_SIZE_BYTES 648
 #define TASK_PRIORITY PIOS_THREAD_PRIO_LOW
 
 // Private variables


mlyle:
@kubark42 @peabody124 @pug398
relates to #1862, althold resets

Believed introduced in #1832.

Working theory is call to external module (e.g. not visible to optimizer) greatly increased the amount of temporaries that the compiler needed to store and thus increased stack utilization.

mlyle:
Pushed to branch pr-1866 on main repo so we get a jenkins build for test/verification.

flight/Libraries/math/misc_math.h
```diff
@@ -64,12 +64,23 @@ void vector2_rotate(const float *original, float *out, float angle);
...
+ */
 #if defined (__clang__) // Clang can't turn off specific optimizations, so turn them all off. This is currently only useful when compiling the simulator with Clang
-bool IS_NOT_FINITE(float x) __attribute__((optnone()));
+static inline bool IS_NOT_FINITE(float x) __attribute__((optnone()));
 #else
-bool IS_NOT_FINITE(float x) __attribute__((optimize("no-finite-math-only")));
+static inline bool IS_NOT_FINITE(float x) __attribute__((optimize("no-finite-math-only")));
 #endif
 
+static inline bool IS_NOT_FINITE(float x) {
```
peabody124: why move the function definition into a header file?

peabody124:
Could we get a snapshot of stack usage before and after?

flight/Libraries/math/misc_math.h
mlyle: So that it is possible for it to be inlined and for the compiler to be able to reason about its side effects and can freely reorder it.

My theory of this is that additional temporaries had to be preserved on the stack because this became a cross-module call.

mlyle:
Before, altitudehold, something negative.  After, 24 bytes on sparky2.

Before, manualcontrol, 104 bytes; after about 200.

mlyle:
Had to add math.h include because of unit tests.

mlyle:
http://jenkins.taulabs.org/artifacts/taulabs_pr-1866_20150830_004508_cf73d2b185_win32.exe is a windows exec for this.

mlyle:
@pug398 says this does not fix things.  I need to retest with a clean build.  Interestink.

mlyle:
@pug398 says it frees stack but still crashes.  Now trying cranking up the altitudehold stack by 48 bytes.

mlyle:
@tracernz says before arm, 100 bytes task stack in AH with this; 68 bytes without.

tracernz:
I have increased the AltitudeHold stack by 48 bytes as per @mlyle and it no longer crashes. Interestingly the UAVO tells me there is 52 bytes free on the AH stack.

mlyle:
I want my 4 bytes back!@#

mlyle:
Still kinda a mystery how the original PR affects stack usage in this particular task.  Oh well :frowning:

lolorc:
F34R, so you're implying this weird mystery could also affect other tasks ?
if it's introduced by #1832 why not reverting it until it's correctly fixed and fully tested ?

kubark42:
> Still kinda a mystery how the original PR affects stack usage in this particular task.

Yeah, this is strange. Neither ```AtitudeHold``` nor ```ManualControl``` uses ```IS_NOT_FINITE()```. It's only used in three spots throughout the codebase: https://github.com/TauLabs/TauLabs/search?utf8=%E2%9C%93&q=IS_NOT_FINITE. So presumably if this were a local problem we should see it only in the three modules that use it, and it if were a global problem we should see it in all modules which include ```misc_math.h```.

Is it to early to ring the "something's fishy" alarm bell?

P.S. Sometime in the future it might be worth thinking about how integration testing could help generate automated before and after reports for RAM usage.

tracernz:
It doesn't really seem right. We should only see the stack usage go up for the Sensors task and Attitude task..
>P.S. Sometime in the future it might be worth thinking about how integration testing could help generate automated before and after reports for RAM usage.

Definitely, we were discussing it on IRC earlier. It is possible for GCC to spit out the maximum stack usage per function using the -fstack-usage switch. We would just need to figure out how to match this up with callgraph data and we might have something usable.

pug398:
quanton has stopped resetting on the bench so I will try the sparky1 on the bench and if it goes well move on to sparky2 and then flight test them. quanton reports mancntrl stack is 204 and AH is 52 bytes which was what is was before.

mlyle:
@lolorc this is a correct fix.  It undoes (most of) the stack usage regression in #1866.  It also turns up the altitude hold stack to have margin. #1866 is a valuable change that had an unexpected side effect on unrelated thread stacks.

mlyle:
Merged based on successful flight and bench tests by @mlyle, @tracernz, and @pug398.

kubark42:
I'm not terribly happy about this mystery. Can those who saw the problem document the crap out of it here so we've got a good paper trail? This is analogous to memory overflow bugs, where unrelated code fails because it's variables are being overwritten. 

Nothing makes sense to me yet about how unused code can impact these modules. Is it possible the ```__attribute``` isn't as well bounded as we think it is?

mlyle:
@kubark42 I plan to look at object code during the fairly idle weeks we have coming up in the schedule.  We have revision control, so it's pretty easy to get the two revs and read code and retest.

kubark42:
Heck, you can probably just remove the ```__attribute``` part and the compiler should go back to optimizing the NaN check out of existence.

Note, one thing we haven't discussed is that this is actually more than a simple NaN check. It also looks at infinity, and perhaps other things. I would suggest that all these checks are useful and valid, esp. with the filter since it can go to ```Inf``` as easiliy as ```NaN```. But that might be the source of the stack growht.

mlyle:
Whatever the cause of the stack growth it's nonintuitive and warrants investigation.  The ARM GCC has in the past been wonky, so it's worth understanding this particular wonkiness.

tracernz:
I compared the generated assembly before @kubark42's fix, after his fix, and after this fix by @mlyle. 
This has caused GCC to add some additional floating point ops (and in some cases push additional fp registers to stack)  in every file which includes misc_math.h, whether IS_NOT_FINITE is used or not. The generated code for altitudehold.c is the same with @kubark42's original fix and @mlyle's updated fix. 

So apparently disabling that optimisation on a single function has caused GCC to make different assumptions about FP ops in general. That is why we are seeing increased stack usage in modules which don't use IS_NOT_FINITE (but do include misc_math.h).

As a quick test to confirm I did this is misc_math.h:
```c
#ifndef NO_FINITE_STUFF

#if defined (__clang__) // Clang can't turn off specific optimizations, so turn them all off. This is currently only useful when compiling the simulator with Clang
static inline bool IS_NOT_FINITE(float x) __attribute__((optnone()));
#else
static inline bool IS_NOT_FINITE(float x) __attribute__((optimize("no-finite-math-only")));
#endif

static inline bool IS_NOT_FINITE(float x) {
	return (!isfinite(x));
}

#endif /* NO_FINITE_STUFF*/
```
And in altitudehold.c:
```c
#define NO_FINITE_STUFF
#include "misc_math.h"
```
This generated identical code to original (.i.e before @kubark42's fix). With this in mind perhaps we might add a` USE_FINITE_FLOAT_CHECK` switch, wrap the `IS_NOT_FINITE` code in it, and only `#define USE_FINITE_FLOAT_CHECK` when used?

mlyle:
Other options--

make IS_NOT_FINITE into a true macro that produces a block with the proper
__attribute__.

Implement IS_NOT_FINITE in assembly in the header.

PR #1867, opened 2015-08-30
bu: update f1 and f3 targets to remove quotes

mlyle:


mlyle:
This one is bizarre because it builds fine for jenkins and some windows environments.  The quotes should really be there.  But removing them is necessary for some build environments to work.

jihlein:
Just successfully completed a "make package" build on Win7 Professional with this change.  LGTM.

mlyle:
@Jinlein thanks for the report and the verification.

PR #1868, opened 2015-08-30
GCS: Detect TL official releases

tracernz:
Our release tag pattern is yyyymmdd, but the GCS looks for RELEASE* to detect official builds. Official builds show a little certificate instead of an exclamation mark in the firmware info box on the uploader. 

I have modified the GCS plugin to match our tag format. Unfortunately our last couple of releases don't seem to be tagged correctly/tagged after the builds were done so I compiled a [firmware to test this from the 20150713 tag](http://git.bertie.nz:81/stuffs/fw_sparky2.tlfw). It is important that the release be tagged and THEN built for this to work correctly.

### DIFF
--- a/ground/gcs/src/plugins/uavobjectutil/uavobjectutilmanager.cpp
+++ b/ground/gcs/src/plugins/uavobjectutil/uavobjectutilmanager.cpp
@@ -35,6 +35,7 @@
 #include <QDebug>
 #include <QEventLoop>
 #include <QTimer>
+#include <QRegularExpression>
 #include <objectpersistence.h>
 
 #include "firmwareiapobj.h"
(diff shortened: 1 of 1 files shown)


mlyle:
Nice!  Hopefully I can coerce jenkins to build after the tag.

guilhermito:
This is based on the premise that only releases will be tagged and not alfa or beta testing phases. It would be best to create an issue and discuss the tagging syntax to use in the future.

mlyle:
Right now all we've tagged is releases, and this has been broken for all of them.

I'm not sure if a testing build should show the certificate or exclamation mark-- it's not a custom build.

We can always improve this in the future.. i'd like it if the next release would have the certificate rather than be broken this way.. and we're at the end of the functional work window for this release so if it's coming in it's coming in now...

guilhermito:
> We can always improve this in the future

Agreed. Merging.

mlyle:
Thanks for the merge.

Can I suggest that if we don't want the tag, we just put a prefix BETA-201508... in front?  Then it won't match the regex.

mlyle:
Er don't want the certificate.

tracernz:
Agreed that it may need revised if we change the tagging rules, but these are the current rules. We can easily change it if we change the rules.

Is it feasible to configure Jenkins to only pull tags on the main branch?

PR #1870, opened 2015-08-30
GCS: Implemented helicopter swash channel index shift for non-zero indices

### LINKED ISSUE #1869: GCS: Swashplate leveling broken; swash servo channels off by 1 for CPHELI
Applies to:
- Vehicle type: helicopter
- Swash type: any

Issue:
- When running swashplate leveling plugin, only N-1 swash servos are configurable
- Furthermore, the spinbox for channel N actually controls channel N+1, and the spinbox for the lowest channel does nothing
- For example, if the selected servos are channels 6, 5, 4, the spinbox for channel 6 does nothing, the box for 5 controls servo channel 6, and the box for 4 controls servo channel 5
- Additionally, upon cancelling the swashplate leveling plugin, the output panel is locked down

Explanation:
- The channel spinboxes all start with the "None" item; thus, if a channel 1 is selected, the spinbox index is 1, so channel N maps to index N in [1,N]
- However, this index is used later as the servo channel index, which is expected to be in \[0,N-1\] (since channel 1 is index 0, etc)

Solution:
- When interpreting the spinbox indices as servo channels, if the spinbox selection is not "None" (i.e. spinbox index > 0), then the saved servo channel should be spinbox index - 1

ekaszubski:
GCS: Fix #1869 
- All heli swash servo channels will be off by 1 if spinbox indices are used directly
- Simply shift all non-zero indices back by 1


ekaszubski:
Fixed extra channels (potentially throttle channel) from being set when not used

mlyle:
Looks great.  Per discussion on IRC, mergin'.

PR #1871, opened 2015-08-31
vtol_follower: New tuning.

mlyle:
Based on @pug398 flight test experience.  Also increase target RTH speed
from 2.5 m/s to 3.0 m/s.  Many platforms will not hit the target vel
because of the weak (previously nonexistant) integral.

Previous was 3.0/0/0/0.  @pug398 saw good behavior at 2.5/0.5/0.0/5.0.  New
settings are velkP=2.6, velkI=0.4, velkD=0, velILimit=5deg.  (doing
slightly less than recommended out of conservatism).


mlyle:
Issue #1838.

PR #1874, opened 2015-08-31
Stabilization: Clarified PID enum elements in stabilization.c

### LINKED ISSUE #1873: Stabilization: Clarify PID enum names in stabilization.c
- enum for PIDs has no elements for PID groups (no `PID_RATE`, only `PID_RATE_ROLL`, etc)
- common practice in the code is to refer to `pids[PID_RATE_ROLL + i]` which can be confusing
- enum should be changed so that the first element in a PID group is the name of that group, e.g.:
```
enum {
  PID_RATE,
  PID_RATE_ROLL = PID_RATE,
  PID_RATE_PITCH,
  PID_RATE_YAW,
  PID_ATT,
  PID_ATT_ROLL = PID_ATT,
  PID_ATT_PITCH,
  PID_ATT_YAW,
  ...
};
```

- Then PIDs can be referenced via `pids[PID_RATE + i]`

ekaszubski:
Fixed #1873

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -84,16 +84,20 @@
 #define HORIZON_MODE_MAX_BLEND               0.85f
 
 enum {
-	PID_RATE_ROLL,   // Rate controller settings
+	PID_GROUP_RATE,   // Rate controller settings
+	PID_RATE_ROLL = PID_GROUP_RATE,
 	PID_RATE_PITCH,
 	PID_RATE_YAW,
-	PID_ATT_ROLL,    // Attitude controller settings
+	PID_GROUP_ATT,    // Attitude controller settings
+	PID_ATT_ROLL = PID_GROUP_ATT,
 	PID_ATT_PITCH,
 	PID_ATT_YAW,
-	PID_VBAR_ROLL,   // Virtual flybar settings
+	PID_GROUP_VBAR,   // Virtual flybar settings
+	PID_VBAR_ROLL = PID_GROUP_VBAR,
 	PID_VBAR_PITCH,
 	PID_VBAR_YAW,
-	PID_MWR_ROLL,   // Virtual flybar settings
+	PID_GROUP_MWR,   // Multiwii rate settings
+	PID_MWR_ROLL = PID_GROUP_MWR,
 	PID_MWR_PITCH,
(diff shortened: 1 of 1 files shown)


flight/Modules/Stabilization/stabilization.c
```diff
@@ -84,16 +84,20 @@
 #define HORIZON_MODE_MAX_BLEND               0.85f
 
 enum {
-	PID_RATE_ROLL,   // Rate controller settings
+	PID_RATE,   // Rate controller settings
```
kubark42: It took me a second to see where you were going with this, but only because the name ```PID_RATE``` is pretty broad. I don't know if there's a convention for naming this grouping something like ```_BASE```, ```_GROUP```, ```_BLOCK```, etc...

kubark42:
I quite like the effort. It's a good idea to help increase comprehension of ```pids[foo + i]```, which is challenging to understand for those who don't understand the finer point of C and memory handling.

flight/Modules/Stabilization/stabilization.c
ekaszubski: @kubark42 That's a good point... how about `PID_GROUP_RATE`? Otherwise I'll probably go with `PID_RATE_GROUP`

ekaszubski:
@kubark42 How does it look now?

mlyle:
I like it / think it's a win. It replaces one unintuitive thing with another, though ;)  (Duplicate enum values).

kubark42:
There's probably no perfect solution because the underlying memory approach is unintuitive. And this is a more perfect solution than in current ```next```, so I echo @mlyle's comment that it's a win.

I'm good to commit, but since this is a style question I'd like to give it a day to hear from others before pressing the merge button.

mlyle:
I've tagged it for the next release.  Today's ostensibly the last day of functional work on the September release... and there's no direct benefit from this / it's just cleanup.

Just for context, we're not going to actually finish today.  I want to keep the window open for a couple of days so the heli stuff has a small chance of getting in.

ekaszubski:
@mlyle Sounds good

mlyle:
Conflicts need fixing, can you please?

ekaszubski:
@mlyle Working on it now

mlyle:
Cool :)

mlyle:
Thanks for this :)

PR #1883, opened 2015-09-01
Actuator: Fix magic numbers relating to number of actuators

kubark42:
Previously, several sections of code depended on magic numbers. These code segments were replaced with strict dependencies on the UAVOs.

Note: If we want the children to be able to overwrite this value, then a faster and easier solution is simply to change static ```const quint32 CHANNEL_NUMELEM = 10;``` to use the value in ```ActuatorCommand::CHANNEL_NUMELEM```. I don't particularly see the reason for it, though.

### DIFF
--- a/ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
+++ b/ground/gcs/src/plugins/config/cfg_vehicletypes/configccpmwidget.cpp
@@ -38,8 +38,8 @@
 #include <math.h>
 #include <QMessageBox>
 
-#include "mixersettings.h"
 #include "actuatorcommand.h"
+#include "mixersettings.h"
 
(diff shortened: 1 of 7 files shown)


kubark42:
This one can probably merge, unless there's a feeling it's not okay to move away from the ```const int```, in which case I can close it.

mlyle:
Yup, this is another in the rate-limited-plus-I-wanna-read-it-more-slowly set.

kubark42:
I think the question simply boils down to whether we want to be able to subclass the number of actuators based on flight models. I'm not seeing any reason to do this, and every reason not to. If, however, the consensus is that we do want to be able to subclass then the magic number can be eliminated by replacing the ```const int``` with an ```int``` initialized in the constructor.

PR #1884, opened 2015-09-01
PID Library: Remove unnecessary scaling

kubark42:
The justification for the scaling does not seem to be valid. The original commit, a09642675cf66650f1cccaf7567878d2f514822b, says "Tweak the integral calculation so it is scaled in ms internally.  This avoids a loss of  precision on the accumulation." However, no additional support for the claim is made in the code or commit, and is seemingly contrary to the theory of floating point precision.

If there is some justification in IEEE-754, it's not immediately obvious. If dynamic range is a problem, a cleaner solution is to add more bits. It's dangerous to expect that the all compilers will play nice and make the multiplication and division by 1000 a worthwhile operation.

Note to reviewers: This needs coverage on several targets, I've only flight-tested on LVL 1. It would also be helpful to double-check that I haven't missed any divide/multiply operations hidden elsewhere in the code. (Bloody magic numbers!)

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_control.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_control.c
@@ -396,7 +396,7 @@ static int32_t vtol_follower_control_accel(float dT)
 static float vtol_follower_control_altitude(float downCommand) {
 	AltitudeHoldStateData altitudeHoldState;
 	altitudeHoldState.VelocityDesired = downCommand;
-	altitudeHoldState.Integral = vtol_pids[DOWN_VELOCITY].iAccumulator / 1000.0f;
+	altitudeHoldState.Integral = vtol_pids[DOWN_VELOCITY].iAccumulator;
 	altitudeHoldState.AngleGain = 1.0f;
 
 	if (altitudeHoldSettings.AttitudeComp > 0) {
--- a/flight/Libraries/math/pid.c
+++ b/flight/Libraries/math/pid.c
@@ -53,9 +53,8 @@ float pid_apply(struct pid *pid, const float err, float dT)
 		// If Ki is zero, do not change the integrator. We do not reset to zero
 		// because sometimes the accumulator term is set externally
 	} else {
-		// Scale up accumulator by 1000 while computing to avoid losing precision
-		pid->iAccumulator += err * (pid->i * dT * 1000.0f);
-		pid->iAccumulator = bound_sym(pid->iAccumulator, pid->iLim * 1000.0f);
+		pid->iAccumulator += err * (pid->i * dT);
+		pid->iAccumulator = bound_sym(pid->iAccumulator, pid->iLim);
 	}
 
 	// Calculate DT1 term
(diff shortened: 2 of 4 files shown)


mlyle:
This has bugged me every time i've seen it.  Nice.

peabody124:
I agree it never made sense, but admit to being involved in creating it. This stretches my memory, but IIRC we did see a difference on F1 back then which is why it got in. I'm happy to see it go - perhaps it was a bug in the math library on arm back then - but it should be tested on an F1 target ideally before merging.

mlyle:
Whenever I saw this I assumed there was some history of fixed point code.

The main thing the *1000 does is it gets you additional range before you underflow.   But the smallest normal number is something like 1 * 10^-38 ... shifting the range to 1 * 10^-41 shouldn't really be a win.

Note that subnormal numbers often do really strange things, but probably not on ARM cortex microcontrollers.  I once saw a bit of code that had a 500x difference in runtime based on input values even though it ran the same number of iterations no matter the input and had no conditionals.  The reason?  Subnormal numbers, which were processed by an exception handler rather than the FPU on x86.

kubark42:
Triple ack. It was clearly for a legacy reason, so I spent a while looking for supporting code, commits, or reviews that would shine some light on it. Unfortunately the original commit is way back in time and so I came up short. 

If subnormals are/were the problem, then a 1000x multiplier seems too convenient of a solution: there's a good chance we'd be better off with a precisely chosen value (and then we could get rid of the pesky magic numbers). However, I don't see how we would get down into those ranges, we only have two (possibly three) multiplicands, and all of them are pretty much guaranteed to be >10e-10.

@peabody124 thanks for pointing out that if it's anywhere, it's on F1.

peabody124:
Ok, LOL. for no reason I found the commit where I did that:

https://github.com/ekaszubski/TauLabs/commit/a09642675cf66650f1cccaf7567878d2f514822b

Unfortunately the message doesn't back it up :(  I do recall having a fair amount of skepticism about it. I think there was concern that on the F1 with a pretty Ki, error, and dT we might run out of resolution in the mantissa.

dT ~ 1e-3
Ki ~ 1e-5 (we use much higher now)
err ~ 1e-2

doesn't seem too bad. I think there was also discussion that the change in error might become so small compared to the accumulated error it gets lost in the trailing digits, but scaling the whole representation by 1e3 wouldn't change that either.

flight/Modules/VtolPathFollower/vtol_follower_control.c
```diff
@@ -396,7 +396,7 @@ static int32_t vtol_follower_control_accel(float dT)
 static float vtol_follower_control_altitude(float downCommand) {
 	AltitudeHoldStateData altitudeHoldState;
 	altitudeHoldState.VelocityDesired = downCommand;
-	altitudeHoldState.Integral = vtol_pids[DOWN_VELOCITY].iAccumulator / 1000.0f;
```
mlyle: I'm glad you thought of these.  Hopefully they're the only points we micromanage integrals.

tracernz:
I have look at this code before and wondered what the story was. Nice to know I'm not alone. :P

mlyle:

PR #1885, opened 2015-09-01
Notify: Port to Qt5 and reenable

kubark42:
Remove Phonon and replace with QMediaPlayer. No (intentional) functional changes. (NOTE: on_toggled_checkEnableSound() was removed, as it was not doing useful work.)

Inspired by OpenPilot commit: a2ba33f28a76cd4e2e7e88200408c1d7cf842cb9

=========================
As an aside, it would be useful to set up defaults for sound. I suspect that this module hasn't been used because without default sounds most people didn't know it was there. Default sounds seem to work for ArduPilot, so there is precedence for which ones are useful to the community.

### DIFF
--- a/ground/gcs/src/plugins/notify/notify.pro
+++ b/ground/gcs/src/plugins/notify/notify.pro
@@ -1,5 +1,5 @@
 TEMPLATE = lib 
-QT += widgets
+QT += multimedia
 TARGET = NotifyPlugin 
  
 include(../../taulabsgcsplugin.pri) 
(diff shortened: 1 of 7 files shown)


mlyle:
So I tested this and it works.  I'm going to merge with caveats.

The caveats are that the options screen sometimes does very odd things.  I had it crash a couple of times, have empty rows that did nothing, and sometimes refuse to allow selection of UAVOs.  Also, the UAVOs are out of order so it's very hard to select things.

That being said, it is really nice, turned off by default, and getting it in for further work shouldn't be gated on reaching perfection.

mlyle:
PS almost all anomalous test results were on windows.  On other platforms it did much better.

PR #1887, opened 2015-09-01
UAVObjectGenerator: Added NUMOPTS #define for options fields

### LINKED ISSUE #1886: UAVObjectGenerator: no length info for options field
- Generated UAVOs have no \#define or enum element for the length of `options` fields (this info *is* generated for `elementnames`)
- Options fields are often used in case statements, but the preference is for an exhaustive list with no default, which implies a range check on the options field
- Without any length information, this range check is conditional on the last enum element; if another option is added to the end of the options list (and thus the enum), any existing range checks become invalid

ekaszubski:
Fixes #1886 
- Field `MyCoolUAVO->MyCoolField` with `options="Option1, Option2, Option3"` will now have `#define MYCOOLUAVO_MYCOOLFIELD_NUMOPTS=3` in `uavobject-synthetics/flight/mycooluavo.h`
- Also note that this should (and does) happen even if there's only one option

### DIFF
--- a/ground/uavobjgenerator/generators/flight/uavobjectgeneratorflight.cpp
+++ b/ground/uavobjgenerator/generators/flight/uavobjectgeneratorflight.cpp
@@ -184,17 +184,43 @@ bool UAVObjectGeneratorFlight::process_object(ObjectInfo* info)
                           .arg( info->name )
                           .arg( info->fields[n]->name ) );
 
-            // find topmost parent and get the length of its options field
+            // find topmost parent to get the length of its options field
             FieldInfo_s * topmost_parent = info->fields[n];
             while (topmost_parent->parent) {
                 topmost_parent = topmost_parent->parent;
             }
 
-            enums.append(QString("/* Max value of any options in topmost parent of field %1 */\r\n").arg(info->fields[n]->name));
-            enums.append( QString("#define %1_%2_MAXOPTVAL %3\r\n")
+            enums.append(QString("/* Max value of any option in topmost parent %2 of field %1 */\r\n").arg(info->fields[n]->name).arg(topmost_parent->name));
+            enums.append( QString("#define %1_%2_GLOBAL_MAXOPTVAL %3\r\n")
                           .arg( info->name.toUpper() )
                           .arg( info->fields[n]->name.toUpper() )
                           .arg( topmost_parent->options.length() - 1 ) );
+
+            // find largest value in this option vector
+            int max_optval = 0;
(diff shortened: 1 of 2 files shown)


mlyle:
This is very good.

It should probably be combined with code to check the range when we convert a binary blob to a uavo, preventing an invalid value from ever getting into the system-- both over the telemetry link and when we read objects from flash.  Then we can stop range checking in most of our code (or make most of the range checks an assert).

mlyle:
However, this looks incorrect as it doesn't seem to be corrected for inherited uavos.  The correct thing to use here is the highest index number of all options (which because of value inheritance may not be the number of options).

Because of different child ordering this can't even be form_enum_name of the last option.. sorry.

ekaszubski:
@mlyle Hmm tricky. I'll look into it.

ekaszubski:
@mlyle Implementation updated

mlyle:
A suggestion of a hack:

add to child enums a value      DONTSWITCHONCHILDENUMS = 255

Then if someone does a switch on a child enum, without the default clause, they'll get a compile error for the uncovered case.  Then for sure all we need to do is ensure that things are in range.  All switches will either A) cover all values within range, or B) have a default clause.

mlyle:
@ekaszubski --- thanks for this.  

I'd prefer for this to come in with enforcement of ranges on ingress and the suggested hack above.. but if you're not eager to do that I'll merge as is.  Your call.

ekaszubski:
@mlyle Sounds good, I'll implement the check as suggested

ekaszubski:
this@mlyle Implementation updated; is that something along the lines of what you were thinking?

Generated enums end up looking like this (newlines added for readability):
```
typedef enum { 
HWSPARKY2_USB_HIDPORT_USBTELEMETRY=HWSHARED_USB_HIDPORT_USBTELEMETRY, 
HWSPARKY2_USB_HIDPORT_RCTRANSMITTER=HWSHARED_USB_HIDPORT_RCTRANSMITTER, 
HWSPARKY2_USB_HIDPORT_DISABLED=HWSHARED_USB_HIDPORT_DISABLED, 
HWSPARKY2_USB_HIDPORT_DONTSWITCHONCHILDENUMS=255 
}  __attribute__((packed)) HwSparky2USB_HIDPortOptions;
```

ekaszubski:
To clarify, generated root enums are not affected, only generated child enums.

mlyle:
Neat!

You or I can implement the rest-- the input validation-- in a subsequent pull.

PR #1888, opened 2015-09-01
Math: Added linear_interpolate to misc math

### LINKED ISSUE #1880: Actuator: Transplant MixerCurve from actuator.c into math libs; add unit tests
- Pending #1859 
- Consider renaming to something like linear_interpolate or linear_interpolate_curve
- Determine unit test conditions

ekaszubski:
Fixes #1880 
- Added unit tests in misc_math/unittest.cpp
- Updated references in actuator.c

### DIFF
--- a/flight/Libraries/math/misc_math.c
+++ b/flight/Libraries/math/misc_math.c
@@ -255,6 +255,43 @@ void cubic_deadband_setup(float w, float b, float *m, float *r)
 	*r = *m * powf(w, 3) + b * w;
 }
 
+/**
+ * Perform a linear interpolation over N points
+ *
+ * output range is defined by the curve vector
+ * curve is defined in N intervals connecting N+1 points
+ *
+ * @param[in] input the input value, in [input_min,input_max]
+ * @param[in] curve the array of points in the curve
+ * @param[in] num_points the number of points in the curve
+ * @param[in] input_min the lower range of the input values
+ * @param[in] input_max the upper range of the input values
+ * @return the output value, in [0,1]
+ */
+float linear_interpolate(float const input, float const * curve, uint8_t num_points, const float input_min, const float input_max)
+{
+	// shift our input [min,max] into the typical range [0,1]
+	float scale = fmaxf( (input - input_min) / (input_max - input_min), 0.0f) * (float) (num_points - 1);
+	// select a starting bin via truncation
+	int idx1 = scale;
+	// save the offset from the starting bin for linear interpolation
+	scale -= (float)idx1;
+	// select an ending bin (linear interpolation occurs between starting and ending bins)
+	int idx2 = idx1 + 1;
+	// if the ending bin bin is above the last bin
+	if (idx2 >= num_points) {
+		//clamp to highest entry in table
+		idx2 = num_points -1;
+		// if the starting bin is above the last bin
+		if (idx1 >= num_points) {
+			// we no longer do interpolation; instead, we just select the max point on the curve
+			return curve[num_points-1];
+		}
+	}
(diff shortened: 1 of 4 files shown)


peabody124:
Love it!

flight/tests/misc_math/unittest.cpp
```diff
@@ -237,3 +237,63 @@ TEST_F(CircularModulusDeg, SweepError) {
...
+
+  // 21 points in range
+  for (size_t i = 0; i <= 20; ++i) {
+    float const input = i * 0.05f;
+    EXPECT_EQ(input, linear_interpolate(input, curve, curve_numpts, range_min, range_max));
+  }
+
+  // 10 points below min range
+  for (size_t i = 1; i <= 10; ++i) {
+    float const input = range_min - i * 0.1f;
```
kubark42: This isn't really an issue because you do it in the above tests, but it doesn't explicitly test the behavior at the boundary. If you do another commit, it might be good to test that ```range_max``` maps to ```range_max```, in case the first unit test gets changed.

kubark42:
@ekaszubski Cool stuff, thanks.

Should there also be a test for non-monotonic curves? I don't see it makes a big difference, since the curve is anyway monotonic inside a bin.

flight/Libraries/math/misc_math.c
```diff
@@ -255,6 +255,43 @@ void cubic_deadband_setup(float w, float b, float *m, float *r)
 	*r = *m * powf(w, 3) + b * w;
 }
 
+/**
+ * Perform a linear interpolation over N points
```
kubark42: My apologies if I'm being overly pedantic, but I wonder if we could do better for the naming here. It's only because that since its mapping is constrained to [0,-1], it's not a full linear interpolation.

ekaszubski: @kubark42 the mapping is actually constrained to [input_min, input_max] and the number of points is adjustable, so I believe it is a full linear interpolation

ekaszubski:
@kubark42 Agreed; I started working on a cubic test at least for CollectiveCurve but it's not completed yet

flight/tests/misc_math/unittest.cpp
ekaszubski: @kubark42 By "not explicitly" do you mean there isn't a named TEST_F for testing the boundaries? Because they are covered in the 21-point test (first and last point are the min and max bounds).

flight/Libraries/math/misc_math.c
kubark42: The output range is constrained to [0,1], right? So I couldn't provide a set of ```x``` values with corresponding ```y``` values, and then get the ```y_interpolated``` vs ```x_intermediate``` values.

ekaszubski: @kubark42 There are no constraints on the inputs or outputs; you can set the input range to -50,50 and set the curve points to [-5,-2.5,-,2.5,5] or whatever you want; the min/max output will be defined by the curve points. The [0,1] thing is only used to select the curve point index to use for interpolation, but the point at that index has no constraints on its range.

mlyle:
Not for action, just a remark:  It might be nice to have this https://en.wikipedia.org/wiki/Monotone_cubic_interpolation eventually

mlyle:

mlyle:
Argh picked up conflicts on this one.

ekaszubski:
@mlyle Will fix this evening

mlyle:
Cool.  Sorry for the conflicts, etc.. this is the downside of closing the merge window for awhile.

ekaszubski:
@mlyle It's all good :)

mlyle:
And... merged

Issue #1879, opened 2015-08-31
Detailed analysis of NaN check stack usage needed


mlyle:
Relating to #1866 / #1832.

mlyle:
So @tracernz investigated some: https://github.com/TauLabs/TauLabs/pull/1866#issuecomment-136607919

The sum of it is, that he saw increased stack usage and additional FP instructions across all files that include the misc_math header, like a lot of fast-math is being disabled.

There's a lot of potential approaches that can be taken to fix this.

* First, determine whether the attribute is needed.  That is, see whether isfinite() becomes a no-op with -ffast-math.  If not, we just use it.
* If it is, then we can do this:  implement IS_NOT_FINITE in a separate c file with fast-math disabled.  Do not put the attribute in the public header.
* Alternatively, there are approaches we could experiment with-- making it a true macro that only comes into source files that use IS_NOT_FINITE.  Implement it as raw ARM assembly that does what we want.  etc.  Option 2 is better, probably.

kubark42:
According to ```--fast-math```, ```isfinite()``` "should" become a no-op. Although the compiler "shouldn't" add FP instructions randomly, so who knows what is going on. I look at this and wonder if ```__attributes()``` is somehow not being bounded to only the ```IS_NOT_FINITE()``` function.

* In ```misc_math.h``` we can turn off optimization options with ```#pragma``` directives, instead of using ```__attributes```. This is another way to arrive at the same results for actually testing ```NaN```.
* In the Makefile, we can turn expand ```--fast-math``` into its subcomponents, leaving out ```--finite-math-only```
* Speaking of macros, we could do a bitwise comparison to look for ```NaN```. Not the cleanest solution, but it will work.

mlyle:
```
head with 1866 and 1832 reverted vs. head:
-altitudeHoldTask	192	static
+altitudeHoldTask	208	static
-AttitudeTask	200	static
+AttitudeTask	208	static
-cubic_deadband	0	static
+cubic_deadband	24	static
-expo3	0	static
+expo3	24	static
+IS_NOT_FINITE	8	static
-pathfollowerTask	248	static
+pathfollowerTask	256	static
-pid_apply	16	static
+pid_apply	24	static
-pid_apply_setpoint	24	static
+pid_apply_setpoint	32	static
-stabilizationTask	288	static
+stabilizationTask	312	static
-updateAttitudeComplementary	192	static
+updateAttitudeComplementary	200	static
[...]

PR #1889, opened 2015-09-01
misc_math: remove optimizer attribute on IS_NOT_FINITE

mlyle:
Discussion in #1879 .  This is an attempt at the minimal change.  It does come with some long term risk if the compiler chain gets smarter.


mlyle:
@peabody124 @tracernz @kubark42 any comment?

tracernz:
I think this is the best course for this release. We should continue to look into the issue (fast-math) as a whole for the next release.

kubark42:
Agreed. Although it makes my head hurt that GCC optimized away our other NaN checks, but not this one. So many references on the 'net say that this should be optimized out of existence.

PR #1892, opened 2015-09-02
GCS Explicit UAVObjects List

peabody124:
This makes the list of uavobjects explicit again. This was changed at the change
got wrapped into the HAL refactor so I think wasn't fully reviewed.

Historically we auto-populated the list of GCS Uavobjects (like we have now). However
this creates issues when changing between branches because objects build files will
persist in build/uavobjects-sync/gcs and trigger compilation errors. With the previous
(explicit) approach we had been using for ~5 years I don't think these type of things
were an issue.

Ultimately, this auto populating approach would be better if the entire build process
were improved. This would require a better build from shares/uavobjects to the 
build/uavobjects-sync as well as making sure GCS picks up this appropriately 
(which for an explicit list, it does on the files that changed).

Of note, when auto-populating the list, even ```gcs_clean``` is not sufficient to get
a good compile when changing branches. You must delete that particular directory.


peabody124:
@mlyle I know you aren't in love with making this explicit again. I just saw @tracernz also having similar issues on next. I really think it is better for now to keep thinks explicit while the implicit list is unreliable. The burden of adding to the list when creating a new uavobject isn't too bad and ultimately takes less time than repeatedly deleting the build directory.

tracernz:
Yeah, this is a bit of a pain. A quick Google suggests that adding the synthetics directory to both INCLUDEPATH and DEPENDPATH should make QMake generate these dependencies correctly but that didn't seem to work for me.

mlyle:
Seriously, you guys know I disagree on this and just merged it to next while we're after the functional freeze point?

PR #1893, opened 2015-09-02
Revert "GCS Explicit UAVObjects List"

mlyle:
Reverts TauLabs/TauLabs#1892

Was merged with no chance for discussion... past the functional merge deadline... when it was said in the very description it'd be contentious.

I'd like the chance to at least discuss what to do about this if it's critical.  Basically, the uavobjectgenerator can atomically replace the set of uavos in the case of change/hash mismatch.  It's about 20 lines of code, would fix additional build problems, and not require this ickiness.


mlyle:
For future reference, the schedule document is at https://github.com/TauLabs/TauLabs/wiki/Endomorphism-(September)-release

Right now we are in a window where only critical bugfixes should merge to get ready for a snapshot build.  Currently there's a pipxtreme regression (subtle, some telemetry objects not making it over the link) that's being troubleshooted.

mluessi:
+1 on what @mlyle said. Wouldn't the proper way to fix this be to modify the uavobjectgenerator so it deletes generated files for which no corresponding XML file exist? Given that the XML files are managed by git, this would make sure that we never have old files in `build/uavobject-synthetics`.

PTDreamer:
> For future reference, the schedule document is at https://github.com/TauLabs/TauLabs/wiki/Endomorphism-(September)-release

As part of the Tau Labs board I don't remember having agreed with that. I do however remember about a rule about not merging your own PR's which you seem to like doing. I won't go into the "revert the revert" (I'll let someone else do it) game but I'm about to take your merging privileges.

> Right now we are in a window where only critical bugfixes

If this isn't a bug what is it?

> Wouldn't the proper way to fix this be to modify the uavobjectgenerator so it deletes generated files for which no corresponding XML file exist?

You seem to not really understand the root cause of this bug.

mluessi:
@PTDreamer the way I see it, @mlyle has pretty much assumed the role as the release manager at the moment. IMO, we should respect that and if the TL board disagrees, they should assign someone from the board to manage the release.

> You seem to not really understand the root cause of this bug.

Apparently not. Can you give a an explanation so we can open an issue and get it fixed?

mlyle:
#1895 is open.  If it doesn't correctly encompass things it can be updated.

kubark42:
I appreciate that there are some underlying issues and those will need to be addressed. But currently a majority of us voted confidence in @mlyle, and to reverse that would take another board decision.

#1892 was merged within three hours of opening, which did not give other developers time to respond. Under that circumstance, this PR correctly reverts #1892, with no prejudice about re-opening and fixing the original problem. I agree with @peabody124 that the issue needs to be addressed, and I'm hoping soon we'll see good descriptions of the problem so we can fix it or understand why we can't.

PTDreamer:
> But currently a majority of us voted confidence in @mlyle,

Sorry? We agreed in giving him merging rights, yes, I voted yes if you remember. But what was the majority that said it was ok to merge own PRs?

PTDreamer:
> I appreciate that there are some underlying issues and those will need to be addressed. But currently a majority of us voted confidence in @mlyle, and to reverse that would take another board decision.

Funny thing is that reviewing the board's voting about @mlyle I noticed that you weren't even part of it.

mlyle:
Yup, I've merged my own PRs.  And I've screwed it up a fair bit, too.  I think I've done some good things, too.

If you guys don't like the way I'm running things you're well within your rights to take my commit bit away.

The way I see it, things weren't getting done, and I figured it was easier to ask forgiveness than permission.  The fact is, committers other than myself weren't very active, and there wasn't anyone else stepping up to run a release.  I put the Endo schedule up for comment multiple times on the #taulabs-dev channel, and offered several ideas for process changes that didn't get much interest.

peabody124:
Let's not escalate an argument over this minor change.

- we should stick to the policy of not merging our own PRs. This has been a long standing thing that had demonstrably helped code quality. In the case where something is minor and we want to merge it ourself historically what you can do is post "I will merge this in 2 days if there are no objections" which gives people sufficient time to address is. part of having merge access to the main repo is abiding by these policies.
- I'll open another PR with this change. It's pretty irrelevant on the merge window since it has no functional change and we've been merging more aggressive changes recently, but simultaneously there is minimal urgency too it

mlyle:
So I know there's been a lot of aggressive merging lately.. that's pretty typical for the end of a functional work window (ideally the "aggressive" happens a few days before the end but it doesn't always work out that way and the NaN bug really bit us).  The idea for right now is that all the merging should be minimal,  contemplative, and for release blockers.  I'm willing to consider this as a blocker, and it's small and minimal.

Issue #1895, opened 2015-09-02
Make dependencies broken with uavos in GCS


mlyle:
https://github.com/mlyle/TauLabs/commit/bb083e39cd1f8d0385919f618fc45975b43d99b2 / #1759 
#1893 #1892 

IMO, a correct fix is this:

1. uavobjectgenerator in build is adjusted to always create all outputs; it's not a significant difference in execution time and is part of the fix to allow parallel package builds.
2. uavobjectgenerator generates output to directory uavobject-synthetics-TMP where TMP is a randomly generated string.
3. After it is done and the uavo hash has been generated, uavobjectgenerator tests if uavobject-synthetics-uavohash exists.
  * If so, it just erases uavobject-synthetics-TMP
  * If not, it renames uavobject-synthetics-TMP to uavobject-synthetics-uavohash
4. uavobjectgenerator forcibly links uavobject-synthetics-githash to uavobject-synthetics

mlyle:
Note re: above.  the name of the uavobjects-synthetics dir (now symlink) should be renamed so that this scheme can co-exist well with prior schemes.

PR #1898, opened 2015-09-02
GCS: make the list of uavobjects explicit

peabody124:
This makes the list of uavobjects explicit again. This was changed at the change
got wrapped into the HAL refactor so I think wasn't fully reviewed.

Historically we auto-populated the list of GCS Uavobjects (like we have now). However
this creates issues when changing between branches because objects build files will
persist in build/uavobjects-sync/gcs and trigger compilation errors. With the previous
(explicit) approach we had been using for ~5 years I don't think these type of things
were an issue.

Ultimately, this auto populating approach would be better if the entire build process
were improved. This would require a better build from shares/uavobjects to the 
build/uavobjects-sync as well as making sure GCS picks up this appropriately 
(which for an explicit list, it does on the files that changed).

Of note, when auto-populating the list, even gcs_clean is not sufficient to get
a good compile when changing branches. You must delete that particular directory.


tracernz:
So I'm not sure I fully understand the issue. Does deleting the UAVO synthetics fix it?

peabody124:
This is happy to be deferred until after the release, although it does not introduce any functional changes.

I will comment that fa4a42c34f461c69faa3f3c5cab79b8a91c50c92 was buried in the HAL changes and was really unrelated to that, so the original change (to make the list implicit) wasn't really reviewed.

#1895 discusses a more ultimate improvement that allows implicit listing, which is the ultimate fix. However, until that is implemented my preference would be to go back to the way we've had it for 3-4 years which worked quite well.

peabody124:
@tracernz yes, when changing branches you can delete that directory then normally won't even need a clean build. it's just annoying and for people that don't know that counterintuitive hiccup they have to delete the whole build directory and get a long build time.

mlyle:
I'm happy to write #1895 quickly for next release in the next few days.

mlyle:
So there's a minimal subset of #1895 which we could merge for this release, and it comes with benefits in crossing versions.

It would be just to unlink the uavobjects-synthetics unconditionally at the beginning of the build.  It would slightly increase build times because we'd have to rebuild objects that aren't strictly necessary, but we can speed up the build again by finishing out #1895.  Thoughts?

kubark42:
Suggested temporary solution: why not have the Makefile do a clean on the ```build/uavobjects-sync/gcs``` directory?

Regarding the merits of the topic at hand, the old way led to some pretty arcane compiler errors as well. You had to know that the source of the GCS linker error was the presence of an xml file in the ```shared/uavobjectsdefinition``` directory that wasn't letter-for-letter mirrored in ```uavobjects.pro```. This was doubly confounding when adding a new UAVO for flight-side only, since it would still cause a GCS build error. I'm not saying the new compiler error is any less arcane, just that the old failure was pretty non-obvious and it required significantly more knowledge to resolve. 

This new way offers a path forward to resolving what has been an unnecessary "gotcha" for the last 3-4 years. And if no one can get it right after a few days, then let's merge this PR. If neither approach has a clearer compiler failure then I'd prefer the devil I know to the one I don't. 

P.S. As far as mixing bb083e39cd1f8d0385919f618fc45975b43d99b2 with the rest of the #1759, it's not the only example of PRs that could have been more atomic. I would suggest that both authors and reviewers make more efforts to keep PRs maximally atomic. It has many upsides, and I'm not aware of any downsides.

mlyle:
The old way can lead to using stale data under some circumstances.

One more relatively innocuous example:  if you remove a uavobject, but still accidentally include it one place.. your incremental build will succeed.  But then when you check it in, anyone who does a clean build afterwards will see breakage.

peabody124:
jenkins will prevent you making those kind of mistakes, right? You mentioned having breakage with the old system, but honestly I've never seen any weirdness when jumping branches I would attribute to having an explicit list of objects versus it including everything in the build directory.

That's not to say I'm opposed to a better fix, but would prefer this while one is pending. I saw your #1899 but haven't had the chance to evaluate it.

mlyle:
Jenkins will prevent it from getting into the code, but it'll still waste a lot of developer time.  Just like the switching-revs-without-clean with this issues does.  Also CI is the last line of defense.

PR #1899, opened 2015-09-03
makefile: do not use stale generated uavo.

mlyle:
Partial implementation of #1895.  Fully correct, but comes with a small
build time penalty.  (Real #1895 will not).

When there are no changes and the cache is hot, on my machine:

make all_flight takes 7.0 seconds with this
make all_flight takes 3.3 seconds with current next.


mlyle:
* Eliminates separate targets to make different types of uavobject synthetics.  Do 'em all in one go.
* Always cleans up prior synthetics when synthetics are needed for a build.

Needs careful test across platforms, especially if for endomorphism.
Test needs to include packaging.

mlyle:
This is the 5 minute alternative to #1898.

mlyle:
Pushed branch tl1899 to Tau repository so Jenkins will try packaging

PTDreamer:
Does this work if the GCS is built with QtCreator?

mlyle:
I'll look at building with qtcreator too, as I've never done that.

mlyle:
Jenkins artifacts OK: http://jenkins.taulabs.org/artifacts/taulabs_pr1899_20150903_173516_36ae76a6eb_win32.exe http://jenkins.taulabs.org/artifacts/taulabs_pr1899_20150903_173516_36ae76a6eb_linux_amd64.tar.xz

mlyle:
CLEAN_BUILD=NO make package artifacts OK on OSX

mlyle:
Both without and with this change, building in qt creator fails because of broken paths in the qt project.
```
cd uavobjects/ && ( test -e Makefile || /Users/mlyle/taulabs/tools/Qt5.5.0/5.5/clang_64/bin/qmake /Users/mlyle/taulabs/ground/uavobjects/uavobjects.pro -spec macx-clang CONFIG+=debug CONFIG+=x86_64 CONFIG+=declarative_debug CONFIG+=qml_debug -o Makefile ) && /Applications/Xcode.app/Contents/Developer/usr/bin/make -f Makefile 
mkdir -p -p ../../uavobject-synthetics 
cd ../../uavobject-synthetics && ../ground/uavobjgenerator/uavobjgenerator -gcs -flight -matlab ../../shared/uavobjectdefinition ../.. && cd ../ground/gcs && /Users/mlyle/taulabs/tools/Qt5.5.0/5.5/clang_64/bin/qmake ../../../ground/gcs/gcs.pro -spec /Users/mlyle/taulabs/tools/Qt5.5.0/5.5/clang_64/mkspecs/macx-clang CONFIG+=debug -r 
/bin/sh: ../ground/uavobjgenerator/uavobjgenerator: No such file or directory
```

I believe ../ground/uavobjgenerator should be ../uavobjgenerator/uavobjgenerator.  Am I not running this property? I tried building 'ground.pro'.

kubark42:
@mlyle You are likely not configuring the project correctly. For whatever reason, the Qt Creator project requires one minor modification before compiling. Go to the project pane and change the build directory to ```/Users/foo/bar/TauLabs/build/ground/ ```.
screen shot 2015-09-03 at 6 45 52 pm

mlyle:
@kubark42: thanks!

mlyle:
CLEAN_BUILD=NO make package works on windows

mlyle:
Works building in QT Creator.  However I think a change should be made to be consistent with the toplevel make, so I'm going to verify that next.

mlyle:
@PTDreamer: Thanks for noticing/remembering that there's another implementation of the uavobjects creation here.  Altered to be consistent with the way the toplevel makefile does it, though I don't think it's strictly necessary.   Confirmed to work also with this change in qt creator.

mlyle:
Confirmed to work with make -j8 all ; CLEAN_BUILD=NO make package on linux.

kubark42:
This compiles successfully, but there was no specific test methodology for failure so I can't guarantee it fixes the original problem raised by @peabody124. Would be good to have a test case...

mlyle:
I believe the test procedure is:

* check out master
* build
* check out newer code
* build without clean.

I'll run those test cases now.

mlyle:
Checking out master, building, and then parallel building next results in:

```
make[1]: *** No rule to make target `/Users/mlyle/taulabs/flight/PiOS/STM32F4xx/pios_dsm.c', needed by `/Users/mlyle/taulabs/build/fw_flyingf4/pios_dsm.o'.  Stop.
```

From stale dependency files being left around (due to another bug in the build system).  pios_dsm was moved, so the old dependency files mess things up.

Bypassing this, we get the error from the issue:

```
/Users/mlyle/taulabs/build/uavobject-synthetics/gcs/fixedwingpathfollowersettingscc.cpp:49:24: error: 
      no matching constructor for initialization of 'UAVObjectField'
  ...UAVObjectField(QString("VerticalVelMax"), QString("m/s"), UAVObjectField::FLOAT32, VerticalVelMaxElemNames, QStringList(), QString("")));
     ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/mlyle/taulabs/ground/gcs/src/plugins/uavobjects/uavobjectfield.h:55:5: note: 
      candidate constructor not viable: no known conversion from 'QString' to
      'const QList<int>' for 6th argument
    UAVObjectField(const QString& name, const QString& units, FieldType ty...
    ^
/Users/mlyle/taulabs/ground/gcs/src/plugins/uavobjects/uavobjectfield.h:54:5: note: 
      candidate constructor not viable: no known conversion from 'QStringList' to
      'quint32' (aka 'unsigned int') for 4th argument
    UAVObjectField(const QString& name, const QString& units, FieldType ty...
    ^
```

So repro scenario confirmed (though there's at least one other pre-existing build issue that prevents jumping between master and next with incremental builds).

mlyle:
Found another build issue that prevents jumping between master and next-- incorrect dependencies in gcs that cause objects built with the wrong standard library to be included:

```
linking ../../../bin/Tau Labs GCS.app/Contents/Plugins/TauLabs/libGpsDisplayGadget_debug.dylib
Undefined symbols for architecture x86_64:
  "std::ios_base::Init::Init()", referenced from:
      ___cxx_global_var_init in nmeaparser.o
  "std::ios_base::Init::~Init()", referenced from:
      ___cxx_global_var_init in nmeaparser.o
ld: symbol(s) not found for architecture x86_64
```

mlyle:
When the DSM and the objects-built-with-wrong-stdlib problems are bypassed, a GCS binary can be built using this code.   I really suspect it because of all the other loud dependency issues-- there may be quiet ones that result in a subtly broken binary.

However, none of the breakage is because of UAVO with this pull request.  So, a win, I guess?  Same outcome as #1898.

mlyle:
So, here's my thoughts:

I prefer this to #1898.  It's likely to be more compatible with future code, and a step along the way.

I prefer doing nothing, just yet, though.  Neither of these is enough to let you do a build of the previous release, move to this release, and build.

tracernz:
It's not a simple problem we can quickly patch in time for this release unfortunately as really, our whole build system breaks when switching branches, especially as the number of commits between them gets large. Patching the specific UAVO problems still doesn't allow jumping between even current master and next for example (both flight code and gcs are still broken by other dependency issues). Perhaps for this release just add a uavobjects_clean target which allows working around the UAVO synthetics problem?

kubark42:
I prefer this to #1898 as well. However, since it addresses @peabody124's immediate workflow bug, I think we should consider it for immediate merge.

One thing that isn't clear to me is whether there's a tradeoff beyond compilation time. This seems to affect matlab and flight Makefile code as well, but I can't quite tell why since the original problem was limited in scope to just compiling the GCS.

mlyle:
@kubark42 -- it's because to do this we blow away the uavobjects directory and create anew.  And we generate all the synthetics in one pass, because it doesn't take more than the time to do one.

Strictly speaking a slightly more contained fix is possible, but this is on the road to #1895.

Anyways, I don't really really care.  We should merge #1898 or #1899 soon, or leave it alone through the release.

mlyle:
No real consensus around merge.  Deferring to pyreneus for real fix.

kubark42:
Now that hold on ```next``` has been released, what do we think about either merging this or merging #1898?

My vote is for this one, but we should do it soon because it fixes a known problem.

mlyle:
OK, my plan is this:  I'll focus on completing this today the full "right way", and having something together for review.

mlyle:
This is nominally complete, but I am fried from thinking about it and need to review it all and test it thoroughly later.  Not worthy of anyone else's time. :P

ground/uavobjgenerator/generators/gcs/uavobjectgeneratorgcs.cpp
```diff
@@ -42,8 +42,9 @@ bool UAVObjectGeneratorGCS::generate(UAVObjectParser* parser,QString templatepat
     gcsCodeTemplate = readFile( gcsCodePath.absoluteFilePath("uavobjecttemplate.cpp") );
     gcsIncludeTemplate = readFile( gcsCodePath.absoluteFilePath("uavobjecttemplate.h") );
     QString gcsInitTemplate = readFile( gcsCodePath.absoluteFilePath("uavobjectsinittemplate.cpp") );
+    QString gcsVerTemplate = readFile( gcsCodePath.absoluteFilePath("uavogcsvertemplate.h") );
```
kubark42: I don't mean to be a stickler, but it would be much easier to read if the full word "version" were used throughout. I worry that ```uavogcsvertemplate``` is hard enough to read that it will give people pause as they try to digest it. Whereas, on the flight side we call if ```flightVersionTemplate```.

ground/uavobjgenerator/uavobjectparser.cpp
```diff
@@ -169,6 +177,24 @@ void UAVObjectParser::calculateAllIds()
     foreach (ObjectInfo *item, objInfo) {
         calculateID(item);
     }
+
+    /* Sort the object info list by ID, now that they're all defined */
+    std::sort(objInfo.begin(), objInfo.end(), [](ObjectInfo *o1, ObjectInfo *o2) {
+            return o2->id < o1->id;
+            });
+
+    /* This is a 64 bit hash to have a bit more protection against collisions
```
kubark42: 64 bits, nice. ;)

ground/uavobjgenerator/uavobjgenerator.pro
```diff
@@ -5,14 +5,14 @@ QT += xml
 QT -= gui
 
 macx {
-    QMAKE_CFLAGS_X86_64 += -mmacosx-version-min=10.7
-    QMAKE_CXXFLAGS_X86_64 = $$QMAKE_CFLAGS_X86_64
+    QMAKE_MACOSX_DEPLOYMENT_TARGET=10.9
```
kubark42: I appreciate this change, but is it perhaps out of scope? The commit message for 6a703c9a91598e99165f3a75a8ee3e286c7bc28d doesn't describe why this change is necessary or what it's intended to do.

flight/UAVObjects/inc/uavobjecttemplate.h
```diff
@@ -39,6 +39,8 @@
 #define $(NAMEUC)_H
 
 #include "pios_queue.h"
+#include "uavover.h"
```
kubark42: Not to beat a dead horse, but this could be parsed like "UAV over", when we want people to understand "UAVO Version".

ground/uavobjgenerator/uavobjgenerator.pro
mlyle: It's consistent with the other GCS makefiles; I use a lambda to sort which is a C++11 feature.  C++11 in general was brought in with PR #1738.  

(the deployment target change is necessary to have a valid C++11 standard library).

mlyle:
Made the requested ver->version changes, fairly blindly.

I still have not tested this on all targets in various conditions and still have not read through it myself carefully-- still WIP.  Sorry this has been a bit slower than we'd like-- I've not had much time.

mlyle:
Argh.  This is not OK on windows-- the link semantics in qt and mingw are different.  I need a little more time.

mlyle:
OK, this works nicely on Windows, Mac, and Linux.  Ready for final review and merge.

mlyle:
(also pushed to mpl1899 on main repo to confirm Jenkins builds it successfully across platforms)

mlyle:
OK, Jenkins built this, after a long time.  I think this is ready to merge.  I'm not 100% confident in this, but I'm about as confident as I can get without wider test/routine use.

kubark42:
Testing now...

kubark42:
Hmm... I couldn't make the system fail by checking out ```master``` (Endomorphism), then building a local software version which has several changed UAVO. Returning to ```master``` and rebuilding again doesn't cause a failure. Even going to ```next``` after all this and rebuilding doesn't cause a build failure.

Note, the only time there was a ```make clean``` was before starting the above process.

So either I'm testing wrong, or there isn't a problem on my computer. Thoughts?

mlyle:
We haven't removed or added any UAVOs since Endomorphism, which was the thing that caused the fault condition.

This isn't really expected to make things unconditionally safe with past revs, but rather unconditionally safe going forward.

kubark42:
I have added several UAVOs in my side branch. I didn't have any troubles going back and forth between endomorphism and this (forward) branch.

kubark42:
I have added several UAVOs in my side branch. I didn't have any troubles going back and forth between endomorphism and this (forward) branch.

@peabody124 Can you test on your computer? I can't at all imagine that this problem magically resolved itself, but it doesn't seem to be causing a problem on my platform. If I can't reproduce it, it might be useful to understand why there's a platform dependency.

kubark42:
Still waiting to hear if this fixes the problem. It would also be good to get deeper understanding on why I don't see this problem (OS X10.10.4) vs. others who do.

mlyle:
So I've seen the problem before.  I will work on getting a repro scenario together so this can merge.

mlyle:
Closing in favor of #1985

PR #1900, opened 2015-09-03
Use breakpad for GCS crash dump reports

### LINKED ISSUE #1217: Use a crash report system for the GCS
Google's breakpad (https://code.google.com/p/google-breakpad/) looks like a good choice, and there are some examples on how to include it on a Qt multi-platform aplication.

mlyle:
Rescued from #1802.  Thanks to @guilhermito for the code in this important area.  Resolves #1217.

Additional changes:
- Adjusted to tolerate / and other unexpected characters in revisions well
- Simplified revision handling in GCS / eliminate stringification/token pasting.  This needs further work.

Things to work on:
- Release builds not made with MSVC so this does not work on Windows.  Need to evaluate feasibility of moving to MSVC Windows builds in Pyreneus
- Currently does not work on Mac OS X.
- Dialog that invites user to look at crash dump opens a text editor of a raw binary dump.
- Need to carefully look at packages to make sure there is no regression.
- On Linux, still occasionally blows up with missing libUtils.so.1 for unknown reason:

```
mlyle@stuffy:~/TauLabs$ ./build/ground/gcs/bin/taulabsgcs.bin 
m_crashReporterChar:  /home/mlyle/TauLabs/build/ground/gcs/bin/crashreporterapp
"gtk+"
Starting the Udev client
*** addAction(): Unknown group:  "Bind Wizard"
UavTalkRelay SERVER listening on  QHostAddress("0.0.0.0") 2000
qt.network.ssl: QSslSocket: cannot resolve SSLv2_client_method
qt.network.ssl: QSslSocket: cannot resolve SSLv2_server_method
Loaded stylesheet: "/home/mlyle/TauLabs/build/ground/gcs/share/taulabs/stylesheets" "default"
"UAVGadgetInstanceManager::createOptionsPages - failed to create options page for configuration TelemetrySchedulerGadget:default, configuration will be removed."
Board not found
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
Using Telemetry parser
/home/mlyle/TauLabs/build/ground/gcs/bin/crashreporterapp: error while loading shared libraries: libUtils.so.1: cannot open shared object file: No such file or directory
Segmentation fault (core dumped)
```

But on other builds it doesn't.  Don't know what the critical/deciding factor is.  Need to determine root cause and then how to fix up library resolution path appropriately.


mlyle:
This is rebased to next, merges cleanly, and does the right thing most of the time on Linux.  I can't explain some of the prior anomalies, but they don't seem to happen anymore and only prevented the crash reporting from working.

I've made fixes to version handling, tolerating "invalid" characters in branch names, and removed the option for the user to view the crash dump file as it resulted in a hang.

mlyle:
(So it's ready for final review and merge IMO)

kubark42:
So this only works on Linux right now, correct? I'm doing a last review of code and comments from #1802.

ground/gcs/src/app/main.cpp
```diff
@@ -258,6 +272,18 @@ int main(int argc, char **argv)
 
     SharedTools::QtSingleApplication app((QLatin1String(appNameC)), argc, argv);
 
+#ifdef USE_CRASHREPORTING
+    QString dirName(GCS_REVISION_PRETTY);
+    dirName = dirName.replace("%@%", "_");
+
+    // Limit to boring characters
```
kubark42: When possible, it's good to specifically spell out what Regexp is looking for. It's a finicky language that many people are unfamiliar with.

Can the comment explain why we need to limit these characters?

mlyle:
Comment added.  Yes, this only works on Linux, or if built with MSVC which no one except @guilhermito has infrastructure to do currently.

kubark42:
@mlyle Okay, based on your and @PTDreamer's results it's clear that it works on Linux. 

Thanks for adding a regexp comment. I don't see it, though. :( Did you (force) push the change? Or did github hiccup?

mlyle:
Why, no, I didn't push the change :P  Sorry

kubark42:
Haha, thanks! I'll merge as soon as Jenkins is happy. (Really, I can't possibly imagine a change to a comment could break Jenkins, but once upon a time @lilvinz made the convincing point that the CI should always give the go-ahead, no ifs-ands-or-buts.)

kubark42:
LGTM, merging.

---

Soapbox P.S.: It's better if commits like c9da6451bed286574f056cfe1ce0942f0e3f12ab go into a separate PR. They are standalone from the breakpad, and affect a wider target (e.g. Matlab) than just the crash reporter. Plus, they're just downright useful, no matter what the state of the larger PR is.

I feel everyone's pain when it comes to having to wait for lots of little PRs to merge before the big PR can go in, but since they're atomically tiny they can be reviewed and merged quickly. And in principle, if they get held up in review because they're controversial, then the larger PR would get held up just the same.

PR #1901, opened 2015-09-03
pipxtreme: force usb rx/tx buf to 256

### LINKED ISSUE #1897: Objects not getting updated over radio telemetry
From @pug398:

> I compared 20150819 f6593076eb (broken telem) against the master 2015817 b5c70e3599 (working) because I thought it would more pinpoint the problem but if peabody could have custom changes then maybe not. I will back up to next 20150812 6b2f4f4de5 as new starting point. If that doesn't work then I will go back further. Windows build of course :)

> Testing is straight forward. Apply firmware to fc with working radio telemetry and make some vpf changes and save. Disconnect and restart gcs which populates (vpf) with defaults. Connect via radio telemetry and on linkup uavo should transfer and update settings to match the ones on fc. Non working does not update the settings via radio telem. usb works.

> I was testing with sparky2 using both taulink and oplink mini with matching pipxteme. A test with quanton/xbee would eliminate a couple more variables and will try to tonight.

> Sucks to be at work when there is way more important things like testing :)

mlyle:
Unknown why this is necessary-- exceeds the 65 of other targets.

Presumably the radiocombridge module doesn't like buffering stuff itself.

Confirmed to work and resolve #1897 between sparky2 & pipxtreme.


mlyle:
Pushed as pr1901 to tl repo to get jenkins to build artifacts.

mlyle:
Artifact at http://jenkins.taulabs.org/artifacts/taulabs_pr1901_20150903_040932_e2863eee8f_win32.exe

Props to @pug398 for finding this and all other issues ever :D

PTDreamer:
This makes no sense to me, it seems more like a workaround that an actual fix. The USB data frame is comprised of a byte for the report ID, a byte for the data lenght, and 62 bytes of data, that is far from the 256 bytes used by this code.

kubark42:
I wonder if there's a solution that would keep more of the code parameters in the target's C code. A thorough examination of the C wouldn't reveal that the pipxtreme buffer size is different. It only becomes apparent by looking through the build chain.

peabody124:
I agree with @kubark42 the style of this fix isn't ideal. It might be ok as a hotfix for the breakage, but should be improved on.  What about a second method that configures the port and accepts the buffer sizes as a parameter. The original method can call into that with the default sizes.

mlyle:
@PTDreamer it's a fix for the actual regression-- pipxtreme stopped working because the USB buffer size was turned down.  I agree we should know why pipxtreme uses larger buffer sizes than the other targets-- someone made a choice to do this at some point in the past.  Was this documented anywhere?

@peabody124 I think we should be getting away from custom buffer sizes and this is legacy.  Because we currently have different buffer sizes for every port "type" this would get unwieldy very quickly and work against the goal of slimming down the target code to the bare essential.

mlyle:
The issue is the use of PIOS_COM_SendBufferNonBlocking in radiocombridge, which in turn says

```
        if (len > fifoBuf_getFree(&com_dev->tx)) {
#if defined(PIOS_INCLUDE_FREERTOS) || defined(PIOS_INCLUDE_CHIBIOS)
                PIOS_Mutex_Unlock(com_dev->sendbuffer_mtx);
#endif /* PIOS_INCLUDE_FREERTOS */
                /* Buffer cannot accept all requested bytes (retry) */
                return -2;
        }

```

mlyle:
As an aside, it seems like we copy data an -awful- lot in these telemetry paths.

mlyle:
@PTDreamer: Notice the size of this isn't for the USB data frame-- that's at lower levels.  It's for the actual COM "FIFO".

PTDreamer:
> @PTDreamer it's a fix for the actual regression-- pipxtreme stopped working because the USB buffer size was turned down. I agree we should know why pipxtreme uses larger buffer sizes than the other targets-- someone made a choice to do this at some point in the past. Was this documented anywhere?

I'm sorry I was under the impression that the buffer size had been smaller since the beginning.

pug398:
PR1901 fixed the issue for me. Settings populate now as well as others like INSState objects.

mlyle:
@pug398, thanks!

tracernz:
Well this patch matches [the pre-HAL refactor buffer size](https://github.com/mlyle/TauLabs/blob/f42376dedb96de28d9e225aba51ebb3615927eea/flight/targets/pipxtreme/fw/pios_board.c#L40) for this target and is confirmed to fix the identified problem. It looks like we've also identified the reason why a larger buffer is required and raised an issue to rework the code and eliminate the need for this fix after the Endomorphism release. I think this is a clean enough solution until the code is reworked to use a smaller buffer? LGTM

mlyle:
Merging, based on @tracernz and @pug398 comments.

peabody124:
This merge was a big aggressive. As @kubark42 pointed out the code style isn't great and burying constants as a CDEF override in the makefile is a practice we should avoid. It does fix an annoying breakage from HAL though and reverting that isn't desirable/feasible. 

I thought of a better approach to fixing this and will create a separate issue to track it.

PR #1906, opened 2015-09-04
PFD: Add magnetic compass indicator. 

kubark42:
This has been very nice for the last 18 months. It shows the calculated magnetic heading on the PFD  (corrected for magnetic inclination). It provides a true "compass" reading and we've used it for a variety of uses, such as determining that the heading initialization is wrong when the board is inclined.

Ideally, the red line should always be in the middle of the box on the compass ticker, although it will bounce around due to error.
screen shot 2015-09-03 at 8 44 41 pm

### DIFF
--- a/ground/gcs/share/taulabs/pfd/default/Compass.qml
+++ b/ground/gcs/share/taulabs/pfd/default/Compass.qml
@@ -44,7 +44,7 @@ Item {
             elementName: "homewaypoint-bearing"
             sceneSize: sceneItem.sceneSize
 
-            visible: HomeLocation.Set != 0
+            visible: HomeLocation.Set !== 0
 
             property real bearing_D : Math.atan2(-PositionActual.East, -PositionActual.North)*180/Math.PI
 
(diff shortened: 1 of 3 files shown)


mlyle:
Nice! I've really wanted this.

ground/gcs/share/taulabs/pfd/default/Compass.qml
```diff
@@ -44,7 +44,7 @@ Item {
             elementName: "homewaypoint-bearing"
             sceneSize: sceneItem.sceneSize
 
-            visible: HomeLocation.Set != 0
+            visible: HomeLocation.Set !== 0
```
mlyle: I really love Javascript.  One of the statements in this comment is true.

mlyle:

ground/gcs/share/taulabs/pfd/default/Compass.qml
kubark42: >... One of the statements in this comment is true...

^^^
I'm guessing that of the two statements in your comment, this is the true one? ;)

mlyle: Yes. :)

PR #1907, opened 2015-09-04
Settings: Workaround for sporadic corruption of settings files.

kubark42:
GCS is known to crash frequently on certain platforms, and some of those crashes cause user settings file corruption. This provides a workaround by loading a copy of settings into local memory and then closing the QSettings object. Upon close, the new settings are copied back onto the old settings file in an (almost) atomic operation. This minimizes the window when settings can be corrupted by undefined behavior during a crash.

Note: This has been in operation for a year. It's been a wonderful fix!

### DIFF
--- a/ground/gcs/src/app/main.cpp
+++ b/ground/gcs/src/app/main.cpp
@@ -263,12 +263,17 @@ int main(int argc, char **argv)
     // Must be done before any QSettings class is created
     QSettings::setPath(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
             QCoreApplication::applicationDirPath()+QLatin1String(SHARE_PATH));
-    // keep this in sync with the MainWindow ctor in coreplugin/mainwindow.cpp
-    QSettings settings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
-                                 QLatin1String("TauLabs"), QLatin1String("TauLabs_config"));
 
-    overrideSettings(settings, argc, argv);
-    locale = settings.value("General/OverrideLanguage", locale).toString();
+    // Scope this so that we are guaranteed that the QSettings file is closed immediately. This
+    // prevents corruption.
+    {
+        // keep this in sync with the MainWindow ctor in coreplugin/mainwindow.cpp
+        QSettings settings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
+                           QLatin1String("TauLabs"), QLatin1String("TauLabs_config"));
+
+        overrideSettings(settings, argc, argv);
+        locale = settings.value("General/OverrideLanguage", locale).toString();
+    }
 
     QTranslator translator;
     QTranslator qtTranslator;
--- a/ground/gcs/src/plugins/coreplugin/mainwindow.cpp
+++ b/ground/gcs/src/plugins/coreplugin/mainwindow.cpp
@@ -133,8 +133,6 @@ MainWindow::MainWindow() :
     // keep this in sync with main() in app/main.cpp
     m_settings = new QSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
                              QLatin1String("TauLabs"), QLatin1String("TauLabs_config.autosave"), this);
-    m_globalSettings = new QSettings(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
-                                 QLatin1String("TauLabs"), QLatin1String("TauLabs_config.autosave"), this);
     m_settingsDatabase = new SettingsDatabase(QFileInfo(m_settings->fileName()).path(),
                                             QLatin1String("TauLabs_config"),
                                             this);
--- a/ground/gcs/src/plugins/coreplugin/mainwindow.cpp
+++ b/ground/gcs/src/plugins/coreplugin/mainwindow.cpp
@@ -103,14 +103,6 @@ MainWindow::MainWindow() :
     m_uniqueIDManager(new UniqueIDManager()),
     m_globalContext(QList<int>() << Constants::C_GLOBAL_ID),
     m_additionalContexts(m_globalContext),
-    // keep this in sync with main() in app/main.cpp
-    m_settings(new QSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
-                             QLatin1String("TauLabs"), QLatin1String("TauLabs_config"), this)),
-    m_globalSettings(new QSettings(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
-                                 QLatin1String("TauLabs"), QLatin1String("TauLabs_config"), this)),
-    m_settingsDatabase(new SettingsDatabase(QFileInfo(m_settings->fileName()).path(),
-                                            QLatin1String("TauLabs_config"),
-                                            this)),
     m_dontSaveSettings(false),
     m_actionManager(new ActionManagerPrivate(this)),
     m_variableManager(new VariableManager(this)),
@@ -138,6 +130,31 @@ MainWindow::MainWindow() :
 #endif /* Q_OS_MAC */
     m_toggleFullScreenAction(0)
 {
+    // keep this in sync with main() in app/main.cpp
+    m_settings = new QSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
+                             QLatin1String("TauLabs"), QLatin1String("TauLabs_config.autosave"), this);
+    m_globalSettings = new QSettings(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
+                                 QLatin1String("TauLabs"), QLatin1String("TauLabs_config.autosave"), this);
+    m_settingsDatabase = new SettingsDatabase(QFileInfo(m_settings->fileName()).path(),
+                                            QLatin1String("TauLabs_config"),
+                                            this);
+
+    // Copy original settings file to working settings. Do this in scope so that
+    // we are guaranteed that the originalSettings file is closed. This prevents corruption
+    // since the QSettings being used are copies of the original file.
+    {
+        QSettings originalSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
+                                   QLatin1String("TauLabs"), QLatin1String("TauLabs_config"), this);
+
+        // There is no copy constructor for QSettings, so we have to do it manually
+        m_settings->clear();
+        QStringList keys = originalSettings.allKeys();
+        for( QStringList::iterator i = keys.begin(); i != keys.end(); i++ )
+        {
+            m_settings->setValue( *i, originalSettings.value( *i ) );
+        }
+    }
+
     setWindowTitle(QLatin1String(Core::Constants::GCS_NAME));
     if (!Utils::HostOsInfo::isMacHost())
         QApplication::setWindowIcon(QIcon(Core::Constants::ICON_TAULABS));
@@ -235,6 +252,23 @@ MainWindow::~MainWindow()
     m_generalSettings = 0;
     delete m_workspaceSettings;
     m_workspaceSettings = 0;
+
+    // Copy working settings back to original settings file. Do this in scope so that
+    // we are guaranteed that the originalSettings file is closed. This minimizes the risk
+    // of corruption since the QSettings are saved (almost) atomically.
+    {
+        QSettings originalSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
+                                   QLatin1String("TauLabs"), QLatin1String("TauLabs_config"), this);
+
+        // There is no copy constructor for QSettings, so we have to do it manually
+        originalSettings.clear();
+        QStringList keys = m_settings->allKeys();
+        for( QStringList::iterator i = keys.begin(); i != keys.end(); i++ )
+        {
+            originalSettings.setValue( *i, m_settings->value( *i ) );
+        }
+    }
+
     delete m_settings;
     m_settings = 0;
     delete m_uniqueIDManager;
@@ -1302,8 +1336,17 @@ void MainWindow::saveSettings(IConfigurablePlugin* plugin, QSettings* qs)
 
 void MainWindow::deleteSettings()
 {
+    // Clear the in-memory settings
     m_settings->clear();
     m_settings->sync();
+
+    // Clear the on-disk settings
+    QSettings originalSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
+                               QLatin1String("TauLabs"), QLatin1String("TauLabs_config"), this);
+    originalSettings.clear();
+    originalSettings.sync();
+
+    // Don't save the settings when exiting
     m_dontSaveSettings = true;
 }
 


mlyle:
Shoot, this is such a persistent annoyance and support problem I'd like to consider this for endomorphism.

mlyle:
(Not saying this is in, just for discussion-- would need heavy test and sequencing).

kubark42:
I appreciate the thought, and it has been an absolute boon to development, but since it's very much a workaround I'd like to see it get wide consideration before merge. It's never let me down, but my coverage is very limited.

peabody124:
Yeah this is a big user annoyance and I think would be a great fix to get into this release

ground/gcs/src/app/main.cpp
```diff
@@ -263,12 +263,17 @@ int main(int argc, char **argv)
     // Must be done before any QSettings class is created
     QSettings::setPath(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
             QCoreApplication::applicationDirPath()+QLatin1String(SHARE_PATH));
-    // keep this in sync with the MainWindow ctor in coreplugin/mainwindow.cpp
-    QSettings settings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
-                                 QLatin1String("TauLabs"), QLatin1String("TauLabs_config"));
 
-    overrideSettings(settings, argc, argv);
-    locale = settings.value("General/OverrideLanguage", locale).toString();
+    // Scope this so that we are guaranteed that the QSettings file is closed immediately. This
```
peabody124: Huh. this really does that? there  isn't a command like flush to make it happen immediately / more explicitly?

peabody124: also just silly question since i'm trying to understand. when is this code called? i can understand the saving to the settings having in the application destructor. this one confuses me a bit.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -138,6 +130,31 @@ MainWindow::MainWindow() :
 #endif /* Q_OS_MAC */
     m_toggleFullScreenAction(0)
 {
+    // keep this in sync with main() in app/main.cpp
+    m_settings = new QSettings(XmlConfig::XmlSettingsFormat, QSettings::UserScope,
```
peabody124: so I understand, this file on disk is never truly used? it is just required in order to use the QSettings object? and if settings are changed on the fly, are they written to this file?not that it is bad to have the autosaved settings hanging around.

ground/gcs/src/app/main.cpp
mlyle: Destructors can be called when objects go out of scope.  They're not guaranteed to be done promptly.. an explicit new/delete might be better, but it makes things more verbose.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
```diff
@@ -138,6 +130,31 @@ MainWindow::MainWindow() :
...
+                             QLatin1String("TauLabs"), QLatin1String("TauLabs_config.autosave"), this);
+    m_globalSettings = new QSettings(XmlConfig::XmlSettingsFormat, QSettings::SystemScope,
+                                 QLatin1String("TauLabs"), QLatin1String("TauLabs_config.autosave"), this);
```
PTDreamer: AFAIK this is never used, nor does the settingsDatabase class functionality.

ground/gcs/src/app/main.cpp
PTDreamer: You can do the same with the deleteLater method which schedules destruction as soon as possible on the event loop.

mlyle: I don't think we want to do that.  The event loop is where a lot of the crashes happen.

PTDreamer:
I can't say I like the solution. I have a branch somewhere that allows the user to use portable settings and while working on that I came to the conclusion that part of this code isn't even used.
Also the scoping done here is not guaranteed to work as intended and is not as obvious as using the deleteLater() slot inherited from QObject

ground/gcs/src/app/main.cpp
PTDreamer: lol

mlyle:
I'm a bit nervous about it... both with the timing for this release, and  the fact that it's a bit of a hack.

OTOH, someone shows up asking me about a question that's traced to corrupted settings about once a week, and it also happens to me every few days.  It's much worse than an ordinary crash because it's usually confusing at first and takes time to get GCS set back up than normal.

Even pretty skillful users like @pug398 and developers end up thinking there's a product regression when the PFD disappears or another symptom of corrupted settings show up.  So it's a pretty severe defect.

PTDreamer:
The corrupted settings files I have seen are produced after a segfault and while I agree we should make the settings file imune to that the truth is that the root cause is still unknown and it will probably continue to be for at least a period of two more release cycles.

PTDreamer:
Also this doesn't deal with the QSettings object that all the plugins use which uses the same file.

mlyle:
> The corrupted settings files I have seen are produced after a segfault and while I agree we should make the settings file imune to that the truth is that the root cause is still unknown

Sure, but I think there's a lot of different ways that GCS can crash.  We should do both: we should fix crashes as we find them, and we should contain crashes so they don't do worse damage.

> and it will probably continue to be for at least a period of two more release cycles.

If you're referring to the crash reporter not getting in-- I can give you plenty of backtraces if that's helpful.  I was gathering them for awhile and I see plenty of crashes-- usually about 3-4 per day.

> Also this doesn't deal with the QSettings object that all the plugins use which uses the same file.

Well, I've not vetted this one in detail myself (yet), so I can't speak as to the details.

ground/gcs/src/app/main.cpp
kubark42: > Huh. this really does that? there isn't a command like flush to make it happen immediately / more explicitly?

The closest command is ```sync()```. According to the docs:

> Writes any unsaved changes to permanent storage, and reloads any settings that have been changed in the meantime by another application.

>This function is called automatically from QSettings's destructor and by the event loop at regular intervals, so you normally don't need to call it yourself

Destroying it immediately prevents the helpful event loop from helpfully corrupting settings. The assumption is that it is this event loop, and not the destructor, that causes corruption when GCS segfaults.

Scoping forces the destructor to be called-- understood it's not guaranteed to be immediate-- and also removes the variable from local scope. Perhaps this might be seen as silly, but IMO it protects against future accidental use of the class. If some wayward soul decided to reuse ```settings``` outside the scope, it would cause a compiler failure instead of a runtime failure. This could also be accomplished with a function, but I'm not sure that's any cleaner.

ground/gcs/src/plugins/coreplugin/mainwindow.cpp
kubark42: > not that it is bad to have the autosaved settings hanging around.

This. When I wrote this code, I had no expectation that it would work so robustly. So I kept around an autosave because recreating all my settings was a giant PITA as well as impediment to workflow. In retrospect, since implementing this change I haven't had a single settings corruption and so making the autosave probably isn't necessary.

PTDreamer:
> If you're referring to the crash reporter not getting in-- I can give you plenty of backtraces if that's helpful. I was gathering them for awhile and I see plenty of crashes-- usually about 3-4 per day.

No backtraces are not useful at all and that only proves you missed that PR's point completely.

PTDreamer:
For the record, I'm against merging this.

kubark42:
> I'm a bit nervous about it... both with the timing for this release, and the fact that it's a bit of a hack.

Yup. The value in this PR is that it demonstrates a working solution to a problem that has existed for years. I consider the code an academic exercise in not-losing-my-shit-anymore and I'm pleased with the results.

Improve away!

mlyle:
> No backtraces are not useful at all and that only proves you missed that PR's point completely.

Hostile much?  Hey, I get that a backtrace in the event loop that it's trying to deallocate something 3 widgets deep isn't very useful.  And I know what crash reporting frameworks are for.  But solving the null dereferences that regularly happen to me 2-3 times per day don't seem like they require a whole lot more context.  We can fight among ourselves or we can work to make things better.

> For the record, I'm against merging this.

OK, so far we have Kenn who brought the code, and peabody and I seem to agree this is an important issue... Do you have another proposed solution?  Note that I've seen more than 5 distinct GCS crashes so it doesn't seem likely we'll address them all for this release.

I agree this needs careful review and verification.  On the other hand, Kenn has a year of mileage on it and anecdotal experience of making this problem go away.  It seems like it's at least a good starting point.

mlyle:
So I studied this some-- I'm a far less qualified language lawyer for C++ than I am for C-- and it's actually guaranteed to do things promptly by recent versions of the C++ standard, including C++11 which we now build with.

The standard states:

> 2 On exit from a scope (however accomplished), objects with automatic storage duration that have been constructed in that scope are destroyed in the reverse order of their construction.  Transfer out of a loop, out of a block, or back past an initialized variable with automatic storage duration involves the destruction of objects with automatic storage duration that are in scope at the point transferred from but not at the point transferred to.

Which is pretty unambiguous that the destruction must be done promptly in a defined order.

mlyle:
I like this for merge.  With the suggested change, I have a bit of a lengthy test sequence to run on it, that I can get to Sunday.

mlyle:
#1918 is the new issue.

peabody124:
@PTDreamer regarding not wanting this to merge, can you elaborate? 

At least on OS X I have a fair number of crashes while exiting that corrupt the settings. If @kubark42 says this improves that, it seems like a good step. Is there something about the approach you don't like, or the fact that it is required is the problem?

peabody124:
P.S. that isn't disputing a better solution might exist. The scoping isn't the most obvious approach IMO either, but if it works and for the (consistent with the code) reason that it writes the data when the object goes out of scope. You are saying to just call deleteLater on the object? that would seem like cleaner code but might need more testing if kenn has tested this for a year.

mlyle:
I definitely see some event loop crashes on shutdown, so having this wait on the event queue could be bad.

PTDreamer:
> At least on OS X I have a fair number of crashes while exiting that corrupt the settings. If @kubark42 says this improves that, it seems like a good step. Is there something about the approach you don't like, or the fact that it is required is the problem?

I don't like what I can't understand. And this PR is full of misconceptions as far as I can tell.

The eventloop is referred several times as the culprit of the problems. That makes no sense since with the exception of main() everything runs on the event loop.

> Destroying it immediately triggers sync() and prevents the helpful event loop from helpfully corrupting settings at regular intervals.
?

> I don't think we want to do that. The event loop is where a lot of the crashes happen.
?

Throughout the comments there is also the believe that the settings file remains open during the settings object life, this is also wrong. The file is opened on the call to QConfFileSettingsPrivate::syncConfFile(int confFileNo)  and closed on this function exit.

This code also touches unused objects and classes and instead of cleaning them it just re-shuffles them.
A mysql database file is being created and never used. Again this was already the case but the code that does it has also been shuffled around making it look like newly added code which masks its staleness.

Again there is no technical explanation why this is any better than the previous code (it does create a new file on disk, is this a feature?).
Having code that is supported only by technical fallacy, misconception and a particular success story on a particular OS or machine is just wrong.

kubark42:
>  With the suggested change, I have a bit of a lengthy test sequence to run on it

Suggested change is implemented.

mlyle:
> The eventloop is referred several times as the culprit of the problems. That makes no sense since with the exception of main() everything runs on the event loop.

We have tons of threads.  Not everything runs on the event loop.  E.g. lots of RawHID crashes are not relating to the event loop in any way.

It's not that the event loop is the culprit of problems.  It's that deferring deleting this object to a future time in the event loop, when there could be something already queued there that's going to blow up GCS, seems unwise.  One frequent crash we have is shutdown-time relating to the teardown of widgets and looks like double-free; it seems to result in this failure.  Deferring the deletion increases the window for things to go wrong.  

During shutdown, lots of things trigger update of the QSettings object, causing it to be saved many times.  It is very easy to lose the race and be halfway through writing the file when something else bad happens.

PTDreamer:
> We have tons of threads. Not everything runs on the event loop. E.g. lots of RawHID crashes are not relating to the event loop in any way.

Again you don't know what you are talking about, the threads used are created with QThread which has its own eventloop.

I won't continue justifying my opinion, writing in english requires to much effort and I have had similar discussions (e.g. object destruction and QObject parenting with peabody124) which can be avoided reading some documentation before commenting based on preconceived ideas.
I'm probably the current developer with more LOC written on the GCS codebase but hey...

peabody124:
I think this is probably an OSX-specific problem. In that case there _is_ a weird exception to how the eventloops happen. I can't recall the details, but there is an OSX system event handler that we have to register against for the HID calls. That required some pretty odd code in terms of making sure variables/objects are running on particular threads.

As to whether a file is persistently opened, that might not be the case, but on OSX there is definitely a frequently observed problem that a crash occurs while exiting will cause a partial configuration file to be written to disk and then subsequent user frustration... 

@PTDreamer Going back to the PR code itself though rather than general architecture, using deleteLater might work but doesn't the scoping approach work too to have the same result? More critically without knowing _why_ the crashes are doing this to the configuration file, doesn't atomically dump the working configuration file into the permanent file quickly seem like a reasonable approach? 
Obviously the better fix is resolving the crashes but this seems to fix a pretty user-annoying issue.

PTDreamer:
m

> doesn't atomically dump the working configuration file into the permanent file quickly seem like a reasonable approach?

Yes it does! The problem is that this PR does much more than that, it is shuffling things around that have nothing to do with that. It is taking variables initialization out of the initializer list and placing them in the constructor, it is scoping variables just because (against Qt's source design style), etc... and while touching them without any technical justification it is reusing stale and unused code. Also it creates a temp file (for the working settings) on disk that is never cleaned.

mlyle:
It may be more common on osx but definitely happens on windows..  Pug and others have hit it.

mlyle:
Yes, this is getting silly.  There are plenty of threads operating asynchronously.  Some are processing other events in separate queues and some are running conventional code and using conventional synchronization mechanisms.

I would be in favor of doing both-- lets take the fix (upon testing that it improves things).  Sunday I am planning on adding an abort to see if I can get a repo and then seeing if this improves things.  Ultimately we should factor this more cleanly, remove the superfluous code, and fix the crashes.  I, you, or someone else can work on these things for next release.

pug398:
Get "taulabs.exe has stopped working" with error in qt5 module on about 40% of GCS exiting (Windows). Since I am exiting it has not been big deal. When it corrupts the settings it is annoying. Have had to redo a few times since it showed up. That is the extent of my knowledge.

mlyle:
This diff allows you to reliably repro config loss on crash:

```
diff --git a/ground/gcs/src/plugins/uavobjectbrowser/uavobjectbrowserconfiguration.cpp b/ground/gcs/src/plugins/uavobjectbrowser/uavobjectbrowserconfiguration.cpp
index fb7ede6..361ade7 100644
--- a/ground/gcs/src/plugins/uavobjectbrowser/uavobjectbrowserconfiguration.cpp
+++ b/ground/gcs/src/plugins/uavobjectbrowser/uavobjectbrowserconfiguration.cpp
@@ -84,6 +84,7 @@ void UAVObjectBrowserConfiguration::saveConfig(QSettings* qSettings) const {
     qSettings->setValue("manuallyChangedColor", m_manuallyChangedColor);
     qSettings->setValue("notPresentOnHwColor", m_notPresentOnHwColor);
     qSettings->setValue("recentlyUpdatedTimeout", m_recentlyUpdatedTimeout);
+    abort();^M
     qSettings->setValue("onlyHighlightChangedValues", m_onlyHighlightChangedValues);
     qSettings->setValue("CategorizedView", m_useCategorizedView);
     qSettings->setValue("ScientificView", m_useScientificView);
```

I am going to test with this patch now and see if the problem goes away.

Uh... am I stupid if I didn't backup my configuration before running the code that I thought might reliably corrupt config? :D

mlyle:
OK, confirmed that on next this corrupted config 3/3.

mlyle:
With this PR, it corrupted config 0/3. This PR appears to be effective in preventing config corruption from crashes during shutdown.  The normal cases still need verification (later tonight or tomorrow).

PTDreamer:
> With this PR, it corrupted config 0/3. This PR appears to be effective in preventing config corruption from crashes during shutdown. The normal cases still need verification (later tonight or tomorrow).

Not surprisingly. And completely unrelated to my comments. As such my vote for not merging this is maintained.

mlyle:
I hear you that it's not clean, but it solves a huge source of developer and user pain.

We could merge this and then revert it immediately after release, if you like, while bringing in a better fix... then users have a better experience and your concerns about code cleanliness are met.

This is a very frequent problem.  I have hit it on Linux and Mac OS X.  Pug has hit it on windows.  Sometimes users come asking for help with it; probably some people just give up.  Sometimes developers get stuck thinking they've regressed something when it's just the config being corrupted instead.

mlyle:
> Again there is no technical explanation why this is any better than the previous code (it does create a new file on disk, is this a feature?).

There's a very simple explanation.  It does not rewrite the main configuration file about a hundred times on shutdown during a window when crashes frequently occur.

mlyle:
Confirmed to properly persist configuration in a variety of scenarios on OSX and Linux.

mlyle:
@PTDreamer-- sorry, but this makes a common source of user pain much better, is confirmed to fix the problem, and has been verified correct.  I'm happy to revert and replace with something cleaner after release.

Issue #919, opened 2013-08-30
Refactor airspeed module


kubark42:
- Create a sensor queue for airspeed data. Each driver will apply it's own hardware specific scaling there
- Sensors can be initialized in pios_board.c with #ifdef flags. This will cause them to register their sensor queue. Most of them will likely need to create their own tasks.
- Process the queued data in sensors.c. Convert to calibrated airspeed and true airspeed as necessary, and push that into the AirspeedActual UAVO.

Note, we might want to make AirspeedActual a multiple-instance UAVO since we can have multiple simultaneous airspeed sources.

PR #1908, opened 2015-09-04
PX4Flow driver

kubark42:
Add driver support for the PixHawk PX4Flow optical flow module.

Note: This PR also renames ```SonarAltitude``` to ```RangeFinderDistance```. This is a more correct description, as we also use laser rangefinders for the purpose of measuring the drone's altitude (and distance to objects). I apologize for not making a more atomic PR. If it's a problem, I can extract the ```SonarAltitude``` renaming to another PR.

Note: This PR only adds support to Quanton. This has also been tested on our internal boards and BrainFPV, but those aren't Tau Labs targets.

P.S. The module works pretty well, although we have some carpet that totally defeats it.
P.S.S. We need to run the bus at 100kHz to make sure the I2C bus has enough time to rise. Realistically, the bus needs to run at >=200kHz if we were sampling at the maximum rate of ```2.5ms```.


DTFUHF:
Nice. Very nice.

mlyle:
Man, like my entire todo list is getting crossed off.  Is a driver for the lidar-lite coming next? ;)

One thing we need to think about is how to handle sensor fusion where we're measuring different but related values-- altitude AGL (rangefinder) vs. altitude MSL (GPS) vs. pressure altitude (baro).

kubark42:
Force pushed a change to remove the last ```sonaraltitude``` from firmware-common.mk.

flight/PiOS/Common/pios_px4flow.c
```diff
@@ -0,0 +1,286 @@
...
+#include "pios_px4flow_priv.h"
+
+#include "pios_semaphore.h"
+#include "pios_thread.h"
+#include "pios_queue.h"
+
+/* Private constants */
+#define PX4FLOW_TASK_PRIORITY        PIOS_THREAD_PRIO_HIGH
+#define PX4FLOW_TASK_STACK_BYTES     512
+#define PX4FLOW_SAMPLE_PERIOD_MS     5
```
kubark42: Note that while this works out to a 200Hz sample rate, the actual fastest rate that the PX4Flow can run is at 400Hz. Unfortunately, I don't believe there's a way to get the sample rate to ```2.5```ms, since ```PIOS_Thread_Sleep_Until()``` takes ```ms``` as arguments.

mlyle:
So do you have any code that does something with this-- feeds it into stabilization or the INS velocity estimate?  Or are you mostly using this to measure performance of internal platforms?

kubark42:
I have  a branch where this augments the INS, but it won't be ready for a PR for quite some time yet. It's challenging to robustly fuse the absolute velocity (from GPS) and the relative velocity (from OF).

kubark42:
I was using this code for position hold over a rug on Monday. Are there any thoughts on improving this PR, or can it merge as-is?

mlyle:
I've been meaning to read this and give it thorough code review... but haven't gotten to it yet.  sorry.  I should have more time for Tau soon.

flight/PiOS/Common/pios_px4flow.c
```diff
@@ -0,0 +1,286 @@
...
+ * @addtogroup PIOS_PX4FLOW PX4Flow Functions
+ * @brief Deals with the hardware interface to the PixHawk optical flow sensor
+ * @{
+ * @file       pios_px4flow.c
+ * @author     Kenn Sebesta, Copyright (C) 2014
+ * @brief      PX4Flow optical flow sensor
+ *
+ ******************************************************************************
+ */
+
```
mlyle: Should have an appropriate license header, e.g. these lines

```
 * @see        The GNU Public License (GPL) Version 3
 *
 *****************************************************************************/
/*
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful, but
 * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
 * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
 * for more details.
 *
 * You should have received a copy of the GNU General Public License along
 * with this program; if not, write to the Free Software Foundation, Inc.,
 */
```

flight/PiOS/Common/pios_px4flow.c
mlyle: The actual multitasking tick rate we have is 1KHz.. 400Hz isn't a factor of this, and turning it up too much further probably isn't good.  One could do 250Hz or 333Hz though...

flight/PiOS/Common/pios_px4flow.c
```diff
@@ -0,0 +1,286 @@
...
+	/* Optical flow structs */
+	struct I2C_Frame
+	{
+		uint16_t frame_count; // counts created I2C frames [#frames]
+		int16_t pixel_flow_x_sum_px10; // latest x flow measurement in pixels*10 [pixels]
+		int16_t pixel_flow_y_sum_px10; // latest y flow measurement in pixels*10 [pixels]
+		int16_t flow_comp_x_m1000; // x velocity*1000 [meters/sec]
+		int16_t flow_comp_y_m1000; // y velocity*1000 [meters/sec]
+		int16_t qual; // Optical flow solution quality [0: bad, 255: maximum quality]
```
mlyle: This has a much larger range than 0 to 255; is the preceding byte used for anything?

kubark42: I don't believe so. This structure is defined by the PX4Flow people, so they might have wanted to reserve the last bit for something more. Or maybe just keep this nicely aligned on 2-byte boundaries.

flight/PiOS/Common/pios_px4flow.c
kubark42: Yup, yup! Thanks for catching this.

flight/PiOS/Common/pios_px4flow.c
kubark42: Would you recommend 333Hz? Internally, optical flow sampling rate should be as fast as humanly possible. However, the external digestion of the OF data can happen at a slower rate.

flight/PiOS/Common/pios_px4flow.c
mlyle: Maybe we should put a reserved byte there, so that if they use the top byte for something we don't get invalid quality numbers.

kubark42: The struct is copied from https://pixhawk.org/modules/px4flow#i2c.

mlyle: Yah, I understand that is the source, but it seems like we don't want to react to an invalid quality that comes over the wire, either through bounds checking or changing the structure so that the adjacent byte is ignored.

kubark42: So you're suggesting that we could use this field as an error check? If the upper byte is non-zero, we throw out the entire packet?

That makes a lot of sense to me, but it depends on PX4Flow intentionally writing 0 there. I suspect/hope they do, since otherwise their description of the field is off...

mlyle: Either way.  Right now we're ignoring the second byte by virtue of the assignment to uint8, but if we ever do any math/quality scaling/discarding of frames with poor quality we could be surprised.  So I think it makes sense to either change that to a uint8_t in the structure definition and then put a reserved byte after it, or to do if ((quality < 0) || (quality > UINT8_MAX)) { /* reject the frame or bound the quality */ }

There's arguments in favor of either approach, but silently ignoring the byte for now in this way leaves a trap for whoever hacks on it in the future.

kubark42: Perfectly reasonable, I'll implement it.

peabody124:
I think this is a hugely useful (and especially fun) thing. My only hesitation is that it adds so much code into core code that likely won't be used most of the time. Do you think it would be possible to implement as an independent module? Sorry if there is an obvious reason why not as I only skimmed the code.

kubark42:
Historically we've been wanting to move all sensors out of modules and into queues. See for instance #919, which I believe was the outcome of a discussion you and I had on IRC.

Not relevant here, but I think that optical flow will become a basic staple of robust robotic control. Wow, it makes a difference.

flight/PiOS/Common/pios_px4flow.c
kubark42: Done, see https://github.com/TauLabs/TauLabs/pull/1908/files#diff-e6d12b87cf512fcdfcab20894f68339bR243.

kubark42:
What's the current thinking on this? Merge or no-merge?

mlyle:
I have a commit bit again-- haven't seen anyone speak against this so will merge :)

mlyle:
Merging isn't working from my phone so will do later.

mlyle:
Once the compile error is fixed I'll merge

kubark42:
Build error fixed.

peabody124:
I'm not against the flow sensor using queues. Definitely right thing to do. Just with all the compile switches I think it could be cleaner as a standalone module

mlyle:
I'm not sure which way is better; I could argue either way.  And better is the enemy of good-- we can always refactor.

kubark42:
> Just with all the compile switches I think it could be cleaner as a standalone module

I think ideally we would have other sensors which would fill these same queues. It wouldn't have to be just the PX4Flow.

We could also just refactor this to include the checks even if OF and RF support isn't present. That's less efficient, but it adheres more to the notion of the sensors module.

We can also refactor sensors to make the module specifically IMU sensors, and then push everyone else into their own modules. I really love this abstraction, because it makes it much cleaner to add multiple redundant sensors (6 IMUs? Why not?!!), but it will definitely have a CPU hit.

PR #1890, opened 2015-09-02
Fix fw naze32 tlfw target detection

fujin:
* Make it such that similarly named boards (e.g. @jihlein's naze32pro support) are able to build without the weird tlfw hax.

PR #1909, opened 2015-09-04
[RFC] Make: Add option to pad firmware description in TLFW

tracernz:
The motivation for this is that my previous workaround for the naze breaks @jihlein's out-of-tree naze32pro build as it is not specific enough (matches all targets with naze32 in the name). This is a much cleaner general solution which can be easily used for other targets in future. In particular it is useful for others doing out-of-tree branches for boards with CP2102 instead of USB. The side effect when building a .tlfw is that a `$(TARGET).padded.bin` will be created in the build directory for every target, but will be identical to the regular `$(TARGET).bin` unless the padded option is enabled.

Replaces #1890 .

### DIFF
--- a/flight/targets/naze32/board-info/board-info.mk
+++ b/flight/targets/naze32/board-info/board-info.mk
@@ -3,6 +3,8 @@ BOARD_REVISION      := 0x01
 BOOTLOADER_VERSION  := 0x80
 HW_TYPE             := 0x01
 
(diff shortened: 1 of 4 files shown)


fujin:
Cool this should work for my copied dodo/spracingf3 target too (cp2102) :dancer: nice work @tracernz

mlyle:
I am marking PRs to merge in the first pass after a release branch is created.  They are not for merge now.

PR #1914, opened 2015-09-04
Mixer maximum value abstraction

kubark42:
Currently the maximum mixer value is assumed to be 127. This abstracts out the magic number so other values could be used. It also fixes a brittle dependency on the UAVObject autogenerated structure, which was previously an impediment to changing the maximum value type.

The GCS changes are pretty superficial, but the firmware changes are a little more in-depth and require more reviewer attention. (To be honest, I'm surprised that the firmware never broke before because of the dependency on the autogenerated MixerSettings struct.)

Note that there is currently no way to deduce the maximum mixer value from any UAVObject. Thus, it needs to be manually set in vehicleconfig.h and actuator.c. Also note that in one the maximum was ```127``` and in the other ```128```. I have elected not to change this at this time.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -399,11 +399,11 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
 	const Mixer_t * mixers = (Mixer_t *)&mixerSettings->Mixer1Type; //pointer to array of mixers in UAVObjects
 	const Mixer_t * mixer = &mixers[index];
 
-	float result = (((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE1] / 128.0f) * curve1) +
-		       (((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE2] / 128.0f) * curve2) +
-		       (((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_ROLL] / 128.0f) * desired->Roll) +
-		       (((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_PITCH] / 128.0f) * desired->Pitch) +
-		       (((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_YAW] / 128.0f) * desired->Yaw);
+	float result = (((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE1] * curve1) +
+	                ((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE2] * curve2) +
+	                ((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_ROLL] * desired->Roll) +
+	                ((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_PITCH] * desired->Pitch) +
+	                ((float)mixer->matrix[MIXERSETTINGS_MIXER1VECTOR_YAW] * desired->Yaw)) / 128.0f;
 
 	if((mixer->type == MIXERSETTINGS_MIXER1TYPE_MOTOR) && (result < 0.0f))
 	{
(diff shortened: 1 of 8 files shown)


flight/Modules/Actuator/actuator.c
```diff
@@ -61,6 +61,7 @@
 #define TASK_PRIORITY PIOS_THREAD_PRIO_HIGHEST
 #define FAILSAFE_TIMEOUT_MS 100
 #define MAX_MIX_ACTUATORS ACTUATORCOMMAND_CHANNEL_NUMELEM
+#define MULTIROTOR_MIXER_UPPER_BOUND 128
```
mlyle: So I notice this in conjunction with the type promotion in the uavo.. What is the intention--- do we want to allow mixer values over 100%?  Or is it just that 128 was out of range before?

I think I'd prefer 127 and keeping the old range, because it wouldn't break peoples' existing mixers.  That said, if there's a reason for over-unity mixers we should consider it.  (While it would seem superficially equivalent to increasing rate PIDs to clip sooner, I have some potential mixer changes I'm evaluating which would make the two cases different).

kubark42: It turns out that 7 bits of resolution is too little for many things. This lays the groundwork for increasing resolution without having to change too much code. I plan to open up a PR in the future where we can discuss what's the best balance of memory vs. resolution.

The type promotion was included by accident. I use the extra resolution a lot, but it's not necessary for this PR so I'm removing it.

FWIW, 128 is the original value. Changing it to 127 would probably not break anything, because it's such a minor difference.

kubark42:
Rebased to remove UAVO type promotion to ```int16```.

flight/Modules/Actuator/actuator.c
mlyle: OK.  The original value makes me a little nervous because 128 doesn't fit in an int8.

mlyle: (Not on the flight side-- on the GCS side.  Presumably appropriate saturation is in place so that we don't accidentally get the opposite mixer value than we'd like, but I'll verify that.)

mluessi: Is this only used for multirotors? Using 

    #define MIXER_RANGE 127

would make it more consistent with the GCS code.

peabody124:
A cool approach for things like this might be a way of exposing associated constants with uavobjects that works in gcs and flight. Not for this pr though

kubark42:
> A cool approach for things like this might be a way of exposing associated constants with uavobjects that works in gcs and flight. Not for this pr though

I'm thinking an additional attribute field that would define a max and a min. That should be easy to accomplish. (+1 for doing this in another PR, though)

kubark42:
We should decide what to do about the 127 vs. 128 question. That seems to be the only unresolved issue before merge, right?

FWIW, I vote to leave it the same as before. I would like this PR to be a pure refactoring and have no impact on flight.

mlyle:
Let's set it to the same as before.  There's not a big difference either way.

I need to read this one with a bit more paranoia than most since it's in actuator, so I've deferred merging it.  Just to let you know how I'm doing stuff-- I merged a bunch of things and I'll probably be merging 1-2 things per day.  If we keep our total merge rate at 4/day or less, there's time for things to shake out (e.g. make sure the merged results still build from Jenkins, and there's no effects from combining code) and we'll still clear out the pull backlog relatively quickly.

kubark42:
No problem. I've been using this code for the better part of a year, so I know it *flies*, but I love having a second eye to see if there are unexpected side effects. And I'm not terribly happy with having to introduce that ```switch..case()``` structure, but it was the only thing I could think of to keep the original mixer code structure and yet break the dependency on the structure of automagically generated UAVO code.

mlyle:
Benchtests OK / I see actuators do seemingly the right thing.

mlyle:
Woops, but we have conflict now before it can merge.

kubark42:
@mlyle Fixed. It was a conflict in ground/gcs/src/plugins/config/cfg_vehicletypes/vehicleconfig.h due to #1883.

mlyle:
Once unit tests are happy again I'll merge this.  (Waiting on Jenkins to do an artifact build of the unit test change)

kubark42:
@mlyle Repeat of https://github.com/TauLabs/TauLabs/pull/1962#issuecomment-143080259. ;)

PR #1916, opened 2015-09-04
Fix AxisLock rate limitation

kubark42:
AxisLock was saturating to the wrong value. The takeaway is that we can now push yaw rates well past 720deg/s (although it's mind numbing to fly a drone spinning at that speed).

The bug was an errant bounding by ```MaximumRate``` where there should have been bounding by ```ManualRate```.

### DIFF
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -425,16 +425,19 @@ static void stabilizationTask(void* parameters)
 
 					if(fabs(stabDesiredAxis[i]) > max_axislock_rate) {
 						// While getting strong commands act like rate mode
-						rateDesiredAxis[i] = stabDesiredAxis[i];
+						rateDesiredAxis[i] = bound_sym(stabDesiredAxis[i], settings.ManualRate[i]);
+
+						// Reset accumulator
 						axis_lock_accum[i] = 0;
 					} else {
-						// For weaker commands or no command simply attitude lock (almost) on no gyro change
+						// For weaker commands or no command simply lock (almost) on no gyro change
 						axis_lock_accum[i] += (stabDesiredAxis[i] - gyro_filtered[i]) * dT;
 						axis_lock_accum[i] = bound_sym(axis_lock_accum[i], max_axis_lock);
-						rateDesiredAxis[i] = pid_apply(&pids[PID_ATT_ROLL + i], axis_lock_accum[i], dT);
-					}
 
-					rateDesiredAxis[i] = bound_sym(rateDesiredAxis[i], settings.MaximumRate[i]);
+						// Compute the inner loop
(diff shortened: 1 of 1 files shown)


flight/Modules/Stabilization/stabilization.c
```diff
@@ -423,18 +423,21 @@ static void stabilizationTask(void* parameters)
 					if (reinit)
 						pids[PID_RATE_ROLL + i].iAccumulator = 0;
 
-					if(fabs(stabDesiredAxis[i]) > max_axislock_rate) {
+					if (fabsf(stabDesiredAxis[i]) > max_axislock_rate) {
```
peabody124: good catch :)

flight/Modules/Stabilization/stabilization.c
```diff
@@ -423,18 +423,21 @@ static void stabilizationTask(void* parameters)
...
-						// For weaker commands or no command simply attitude lock (almost) on no gyro change
+						// For weaker commands or no command simply lock (almost) on no gyro change
 						axis_lock_accum[i] += (stabDesiredAxis[i] - gyro_filtered[i]) * dT;
 						axis_lock_accum[i] = bound_sym(axis_lock_accum[i], max_axis_lock);
-						rateDesiredAxis[i] = pid_apply(&pids[PID_ATT_ROLL + i], axis_lock_accum[i], dT);
-					}
 
-					rateDesiredAxis[i] = bound_sym(rateDesiredAxis[i], settings.MaximumRate[i]);
+						// Compute the inner loop
+						float tmpRateDesired = pid_apply(&pids[PID_ATT_ROLL + i], axis_lock_accum[i], dT);
```
peabody124: why define tmp variable here rather than just assign rateDesiredAxis[i] on this line and then bound it on the next line? not wrong but more consistent style with code above.

peabody124:
Looks good although personally I'd drop that temp variable.

peabody124:
@mlyle this might want to be tagged for endo release, since it is a bug fix

flight/Modules/Stabilization/stabilization.c
kubark42: As a general rule, I avoid functions that are called as arguments to functions. IMO this has two advantages: 1) it makes each step of the process clearer, 2) it decouples the theoretical control math from the real-world system constraints, facilitating changing each part without accidentally influencing the other.

flight/Modules/Stabilization/stabilization.c
kubark42: Strangely there are several of these throughout our code. I had thought that there would be warnings-- and thus because of ```-Werror``` build failures-- when promoting to a double as an argument to a function, but it turns out that the compiler catches very few of these double promotions.

mlyle:
What the heck, let's take it.

Issue #1913, opened 2015-09-04
Jenkins: Linux amd64 builds, GCS don't start up


LinuxJohannes:
Latest Master is working fine. But all latest NEXT on Jenkins don't work.

Output from the command line:
jh@linux-cjtf:~/TL/taulabs_next_20150904_013719_f30055fd4e_linux_amd64/gcs/bin> ./taulabsgcs
"fusion"
Starting the Udev client
*** addAction(): Unknown group:  "Bind Wizard"
UavTalkRelay SERVER listening on  QHostAddress("0.0.0.0") 2000
Loaded stylesheet: "/home/jh/TL/taulabs_next_20150904_013719_f30055fd4e_linux_amd64/gcs/share/taulabs/stylesheets" "default"
"UAVGadgetInstanceManager::createOptionsPages - failed to create options page for configuration TelemetrySchedulerGadget:default, configuration will be removed."
void PfdQmlGadgetWidget::setQmlFile(QString) "/home/jh/TL/taulabs_next_20150904_013719_f30055fd4e_linux_amd64/gcs/share/taulabs/pfd/default/Pfd.qml"
Cannot open file 'pfd.svg', because: No such file or directory
file:///home/jh/TL/taulabs_next_20150904_013719_f30055fd4e_linux_amd64/gcs/share/taulabs/pfd/default/SvgElementImage.qml:18: TypeError: Cannot read property 'height' of undefined
file:///home/jh/TL/taulabs_next_20150904_013719_f30055fd4e_linux_amd64/gcs/share/taulabs/pfd/default/SvgElementImage.qml:17: TypeError: Cannot read property 'width' of undefined
GLC_Context::GLC_Context
GLC_Context::chooseContext
QXcbIntegration: Cannot create platform OpenGL context, neither GLX nor EGL are enabled
enable to create context  0x2715900
ModelView: aircraft file not found.
Board not found
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
Using Telemetry parser
Abgebrochen
jh@linux-cjtf:~/TL/taulabs_next_20150904_013719_f30055fd4e_linux_amd64/gcs/bin>

mlyle:
What graphics card do you have?  Does glxgears work with hardware acceleration?

mlyle:
Initially put into endomorphism release for triage.

LinuxJohannes:
[...]

PR #1920, opened 2015-09-07
Make:Package: Copy xcbglintegrations to qtplugins directory on linux

tracernz:
Required for Qt 5.5, as per #1913. Seems Qt changed from [compile time detection](http://code.qt.io/cgit/qt/qtbase.git/tree/src/plugins/platforms/xcb/qxcbintegration.cpp?h=v5.4.1#n249) to [runtime detection](http://code.qt.io/cgit/qt/qtbase.git/tree/src/plugins/platforms/xcb/qxcbintegration.cpp?h=v5.5.0#n214) of GL in 5.5 and it depends on this plugin being present. Mostly solved by @mlyle.

### DIFF
--- a/package/Makefile.linux
+++ b/package/Makefile.linux
@@ -40,7 +40,7 @@ ALL_DEB_FILES += $(foreach f, $(DEB_MACHINE_FILES), $(DEB_BUILD_DIR)/$(f))
 
 DEB_PACKAGE_NAME	:= taulabs_$(VERSION_FULL)_$(DEB_PLATFORM)
 
-QT_PLUGINS_USED		:=imageformats platforms sqldrivers
+QT_PLUGINS_USED		:= imageformats platforms sqldrivers xcbglintegrations
 
 standalone: gcs
 	@echo $@ starting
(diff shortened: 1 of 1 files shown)


mlyle:
Mostly solved by @mlyle my ass, you did all the work. :)

Merging to see if it affects Jenkins artifacts.  I don't really expect it to.

kubark42:
@mlyle could we have a bit more time to comment before merging? I realize this doesn't affect much, but 6 hours is an awfully narrow window, esp. on a holiday weekend.

@tracernz I have no objections to the changes, but the git commit message could be improved. From the sound of your and @mlyle's discussion, this was a pernicious problem to run to ground. The git commit message is a great place to document your efforts, and ensure 1) that no one accidentally breaks important code and 2) help other project maintainers years down the road judge code for continued pertinence.

For example, tracking down https://github.com/TauLabs/TauLabs/pull/1884 was much harder for me because there was no breadcrumb trail. And even though no one has been able to find theoretical justification for the original scaling, as @peabody124 says they might have been necessary in order to work around an ARM FPU bug. So we're somewhat in the dark about whether merging #1884 is a good idea, and we'll only find out if we see breakage after the change is pushed to a wider audience. This might have been avoided if the original justification were present in the commit history. (At the very least, it would have given us a test case.)

In general, and here I'm getting on my soapbox, a git commit message is like good journalism, and it should answer the Who/What/Where/Why/When. Fortunately for us, the "Who" and "When" are already answered automatically. The rest are left to the author, but they are equally critical.

Read Section 14 of https://www.kernel.org/doc/Documentation/SubmittingPatches for inspiration. In particular, 

> The canonical patch subject line is:
>    Subject: [PATCH 001/123] subsystem: summary phrase
>
>The canonical patch message body contains the following:
>  - A "from" line specifying the patch author (only needed if the person
>    sending the patch is not the author).
>  - An empty line.
>  - The body of the explanation, line wrapped at 75 columns, which will
>    be copied to the permanent changelog to describe this patch.

Here are two commits which provide good examples of how the commit messages should look:
https://github.com/peabody124/TauLabs/commit/8367348b5e9f7ee5d860870c6f538f8a76b9a353
https://github.com/cyphyworks/TauLabs/commit/32c3f4f03111a1fe4a5ad0a3e0adde12d5548a2b

mlyle:
@kubark42 sorry about that.  I probably should have copied it to another branch so Jenkins would build it.  I don't hold much hope that this is going to fix things, but might allow me to manually create good packages at least.

mlyle:
And no packages are spitting out from Jenkins for whatever reason.. since this is tied to the Linux build it couldn't have broken packaging for Windows, but no output.

tracernz:
@kubark42 Thanks for your suggestion, I should have added more detail to that message. Hopefully for this particular commit the PR (which will be linked to the commit when viewing on Github) will give the necessary detail to anyone looking in the future.

PR #1921, opened 2015-09-08
PIOS HAL: make RFM22B optional

kubark42:
Fix by @mluessi.

We couldn't compile a new target using the HAL because it was looking for modules we weren't compiling. This resolves that problem.


mlyle:
Nice.  Yup, I expect there will need to be lots of small changes like this as we broaden the set of platforms.

mlyle:

PR #1924, opened 2015-09-09
UAVSettingsImportExport: Alphabetize attributes in generated XML document

### LINKED ISSUE #1922: Stop shuffling of attributes in XML files generated by the GCS
Every time GCS restarts, it arbitrarily rearranges the ordering of XML attributes. This potentially interferes with the progress in https://github.com/TauLabs/TauLabs/pull/1833. Fortunately only trivial things like ```names``` and ```values``` will invert, so it's still easy to visually compare files, but this inversion still interferes with diffs.

E.g.:
```
<object id="0x3557036A" name="ActuatorSettings">
    <field values="0.991900027,0.421400011" name="MotorInputOutputCurveFit" />
    <field values="2,0,50,0,0,50" name="TimerUpdateFreq"/>
    <field values="232,232,232,232,232,232,1000,2000,2000,2000" name="ChannelMax" />
    <field values="136,136,136,137,137,137,1000,1000,1000,1000" name="ChannelNeutral" />
    <field values="125,125,125,125,125,125,1000,1000,1000,1000" name="ChannelMin" />
    <field values="12MHz,12MHz,1MHz,1MHz,1MHz,1MHz" name="TimerPwmResolution" />
    <field values="PWM,PWM,PWM,PWM,PWM,PWM,PWM,PWM,PWM,PWM" name="ChannelType" />
    <field values="TRUE" name="MotorsSpinWhileArmed" />
</object>
```

vs.

```
<object name="ActuatorSettings" id="0x3557036A">
    <field name="MotorInputOutputCurveFit" values="0.991900027,0.421400011"/>
    <field name="TimerUpdateFreq" values="0,0,50,0,0,50"/>
    <field name="ChannelMax" values="232,232,232,232,232,232,1000,2000,2000,2000"/>
    <field name="ChannelNeutral" values="136,136,136,137,137,137,1000,1000,1000,1000"/>
    <field name="ChannelMin" values="125,125,125,125,125,125,1000,1000,1000,1000"/>
    <field name="TimerPwmResolution" values="12MHz,12MHz,1MHz,1MHz,1MHz,1MHz"/>
    <field name="ChannelType" values="PWM,PWM,PWM,PWM,PWM,PWM,PWM,PWM,PWM,PWM"/>
    <field name="MotorsSpinWhileArmed" values="TRUE"/>
</object>
```

It turns out that in XML attribute order is *never* important, and XSLT cannot change this. There are two "easy" solutions as I see it:

1. Change the randomizing seed for Q_HASH. This, IMHO, is *bad*, for any number of security reasons
2. Figure out how to convert the file to Canonical XML. http://lists.w3.org/Archives/Public/public-qt-comments/2003Nov/0029.html hints that XSLT should be able to do this, but the info there is poor and not enough to go very far on.

kubark42:
The XML spec does not care about the ordering of attributes, it considers the order "unimportant". This patch attempts to use XSLT to sort these attributes nonetheless.

This works on all tested computers so far. I'd love to hear feedback from others in case it breaks anywhere. 

**Note**: According to the original code source, http://stackoverflow.com/questions/1429991/using-xsl-to-sort-attributes, this is not guaranteed to work every time. In fact, because XML doesn't care about attribute ordering, it can be said that this is never guaranteed to work. However, in practice it works, and it reasonable to hope that if it works across all implementations of Qt that it will not likely break in Qt's near future.

Hopefully this tides things over until Qt implements a Canonical XML parser.

Fixes #1922.

P.S. A million thanks to ankh and can-of-bees on freenode #xml for all their help in running this to ground.

### DIFF
--- a/ground/gcs/src/plugins/uavsettingsimportexport/uavsettingsimportexportfactory.cpp
+++ b/ground/gcs/src/plugins/uavsettingsimportexport/uavsettingsimportexportfactory.cpp
@@ -338,35 +338,74 @@ QString UAVSettingsImportExportFactory::createXMLDocument(const enum storedData 
         }
     }
 
-    // This sorts the XML <object> children's <name=".."> attribute by alphabetical order. This
-    // is particularly helpful when comparing *.uav files to each other.
+    // This sorts the XML <object> children's <name=".."> attribute by alphabetical order,
+    // as well as sorting all attributes by alphabetical order (see note below for
+    // special circumstances around sorting attributes. This is particularly helpful when
+    // comparing *.uav files to each other.
     QString preliminaryXMLDoc = doc.toString(4);
+    QString alphabetizedAttributesXMLDoc;
     QString alphabetizedXMLDoc;
 
-    QString xmlAlpheticalSorter(" \
-        <xsl:stylesheet version=\"2.0\" xmlns:xsl=\"http://www.w3.org/1999/XSL/Transform\"> \
-          <xsl:output method=\"xml\" indent=\"yes\" omit-xml-declaration=\"no\"/> \
-          <xsl:strip-space elements=\"*\"/> \
-          \
-          <xsl:template match=\"@* | node()\"> \
-            <xsl:copy> \
-              <xsl:apply-templates select=\"@* | node()\"/> \
-            </xsl:copy> \
-          </xsl:template> \
-          \
-          <xsl:template match=\"settings\"> \
-            <xsl:copy> \
-              <xsl:apply-templates select=\"@*\" /> \
-              <xsl:apply-templates select=\"object\"> \
-                <xsl:sort select=\"@name\" data-type=\"text\"/> \
-              </xsl:apply-templates> \
-            </xsl:copy> \
-          </xsl:template> \
-        </xsl:stylesheet> \
-    ");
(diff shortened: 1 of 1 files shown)


mlyle:
Oh man, a can of bees.  That doesn't sound good ;)

liamquin:
XML cares about attributes - it's the ordering of attributes within a tag that's not considered significant.

kubark42:
@liamquin I updated the PR text to make that explicit. Thanks again for all the #xml help on this.

swolfe0105:
I tested this with the Jenkins binary on Ubuntu. It appears to be working fine.

mlyle:
@swolfe0105 I don't think this would be present in any Jenkins binaries-- PRs aren't published/built by default and this isn't in-tree yet.

kubark42:
> I don't think this would be present in any Jenkins binaries-- PRs aren't published/built by default and this isn't in-tree yet.

PRs are built by default, but only for Linux. Check out http://jenkins.taulabs.org:8080/job/taulabs-pr-linux-amd64/3394/ (gotten to by clicking the little green checkmark showing the build passed)

mlyle:
Ah, I forgot about those artifacts.  Thanks.

mlyle:

PR #1925, opened 2015-09-10
pios_mpu*: Correct orientation math / prevent overflow

### LINKED ISSUE #1923: Problem with axis reverse code in MPU6K driver
There is a problem within the code that invert mpu6000 axis to adapt them to board convention.
it is done as 
axis = - (int16_t)original value;
this create a problems as INT16_MIN != - INT16_MAX
`(int16_t)0x8000 = -32767`

`(int16_t)0x8001 = -32768`
when inverted they turns into:
` - (int16_t)0x8000 =  32767`

` - (int16_t)0x8001 = -32768`

mlyle:
Signed wraparound has undefined behavior in C, and here it was worse,
where the obvious thing flipped sign for maximum values in one
direction.  Used @trex4git's suggestion of a floating point multiply to
ensure sign promotion.  Hat tip to @AlessioMorale for pointing out this
severe issue.

Fixes #1923.

### DIFF
--- a/flight/PiOS/Common/pios_mpu6000.c
+++ b/flight/PiOS/Common/pios_mpu6000.c
@@ -605,61 +605,61 @@ static void PIOS_MPU6000_Task(void *parameters)
 			accel_data.x = (int16_t)(mpu6000_rec_buf[IDX_ACCEL_YOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_YOUT_L]);
 			gyro_data.y  = (int16_t)(mpu6000_rec_buf[IDX_GYRO_XOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_XOUT_L]);
 			gyro_data.x  = (int16_t)(mpu6000_rec_buf[IDX_GYRO_YOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_YOUT_L]);
-			gyro_data.z  = - (int16_t)(mpu6000_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_ZOUT_L]);
-			accel_data.z = - (int16_t)(mpu6000_rec_buf[IDX_ACCEL_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_ZOUT_L]);
+			gyro_data.z  = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_ZOUT_L]);
+			accel_data.z = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_ACCEL_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_ZOUT_L]);
 			break;
 		case PIOS_MPU60X0_TOP_90DEG:
-			accel_data.y = - (int16_t)(mpu6000_rec_buf[IDX_ACCEL_YOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_YOUT_L]);
+			accel_data.y = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_ACCEL_YOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_YOUT_L]);
 			accel_data.x = (int16_t)(mpu6000_rec_buf[IDX_ACCEL_XOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_XOUT_L]);
-			gyro_data.y  = - (int16_t)(mpu6000_rec_buf[IDX_GYRO_YOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_YOUT_L]);
+			gyro_data.y  = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_GYRO_YOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_YOUT_L]);
 			gyro_data.x  = (int16_t)(mpu6000_rec_buf[IDX_GYRO_XOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_XOUT_L]);
-			gyro_data.z  = - (int16_t)(mpu6000_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_ZOUT_L]);
-			accel_data.z = - (int16_t)(mpu6000_rec_buf[IDX_ACCEL_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_ZOUT_L]);
+			gyro_data.z  = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_ZOUT_L]);
+			accel_data.z = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_ACCEL_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_ZOUT_L]);
 			break;
 		case PIOS_MPU60X0_TOP_180DEG:
-			accel_data.y = - (int16_t)(mpu6000_rec_buf[IDX_ACCEL_XOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_XOUT_L]);
-			accel_data.x = - (int16_t)(mpu6000_rec_buf[IDX_ACCEL_YOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_YOUT_L]);
-			gyro_data.y  = - (int16_t)(mpu6000_rec_buf[IDX_GYRO_XOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_XOUT_L]);
-			gyro_data.x  = - (int16_t)(mpu6000_rec_buf[IDX_GYRO_YOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_YOUT_L]);
-			gyro_data.z  = - (int16_t)(mpu6000_rec_buf[IDX_GYRO_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_GYRO_ZOUT_L]);
-			accel_data.z = - (int16_t)(mpu6000_rec_buf[IDX_ACCEL_ZOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_ZOUT_L]);
+			accel_data.y = -1.0f * (int16_t)(mpu6000_rec_buf[IDX_ACCEL_XOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_XOUT_L]);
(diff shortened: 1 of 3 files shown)


mlyle:
Target: merge this Friday.  It's a high severity issue, but we don't want to be hasty and make anything worse.

phobos-:
What about pios_mpu9250_spi?

mlyle:
@phobos- It appears to do all the rotations after things are already floats.

Trex4Git:
My quick suggestion 
```
axis = - 1.0f * (int16_t)original value;
``` 
seems to already fix it.
But this also affects the down-sampling PR, especially the Small F1 behavior.
I could implement it in there the same way, but as already using macros a little bit cleaner might be:
```
axis = -  (float)((int16_t)original value);
```

Or more concrete: https://github.com/Trex4Git/TauLabs/blob/flight_gyro_downsampling/flight/PiOS/Common/pios_mpu6000.c#L58-L66
e.g. instead of:
```
 #define RAW_ACCEL_X  ((int16_t)(mpu6000_rec_buf[IDX_ACCEL_XOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_XOUT_L]))
```
and than using the "-1.0f *" fix I would prefer to do it like that:
```
 #define RAW_ACCEL_X  (float)(((int16_t)(mpu6000_rec_buf[IDX_ACCEL_XOUT_H] << 8 | mpu6000_rec_buf[IDX_ACCEL_XOUT_L])))
```

mlyle:
@Trex4Git Yes-- I was concerned this would clash with your PR.

I think your new RAW_ACCEL_X is fine, but you probably want another set of parens.

Can I suggest something like (not verified)...

```
#define PACK_SENSOR_VALUE16(offa, offb) ((float)((int16_t)(mpu6000_rec_buf[offa] << 8 | mpu6000_rec_buf[offb])))
#define RAW_ACCEL_X PACK_SENSOR_VALUE16(IDX_ACCEL_XOUT_H, IDX_ACCEL_XOUT_L)
```

Trex4Git:
I think i had not used a cascaded macro before, but if that's work, in principal this looks fine to me.

BrainFPV:
LGTM

peabody124:
So can we come.up with a solution that avoids fpu. On f1 this is a burden, hence the code that shuffled it before. The code from Libre pilot that has a 1lsb bias is probably a better approach

kubark42:
@mlyle Should this fix be propagated more widely? Just poking around, I see the same issue in another driver: https://github.com/TauLabs/TauLabs/blob/f81a193f00f62c601761c8e653d829b3bca49d6f/flight/PiOS/Common/pios_lsm303.c#L66. It's not a priority, because no target hardware is using this, but we could open an issue after this PR is done.

@peabody124 https://github.com/TauLabs/TauLabs/search?utf8=%E2%9C%93&q=pios_sensor_accel_data suggests that there will be a floating point promotion no matter what. Are you concerned that the FPU operation to negate the variable will cost significantly more after the promotion than before? It might require looking at the actual assembly, since toggling the sign on an IEEE-754 float is equivalent to toggling a single bit. So this might still be equally efficient.

======================

P.S.: I personally don't like the fixed point arithmetic solution. It makes for more awkward, i.e. less maintainable, code, with the only benefit of supporting hardware designed 5+ years ago. I realize this is a somewhat controversial point, so I don't intend to start a discussion on it here and will go with the flow of whatever we decide.

Trex4Git:
For the L3DG20 (Gyro) https://github.com/TauLabs/TauLabs/blob/next/flight/PiOS/Common/pios_l3gd20.c#L68-L73 it seems to be the same as for the LSM303 (ACC + MAG) , and might also need to be fixed.

I am also not sure if it is better to get worth data on all targets, for maybe a small performance improvement on some targets.

peabody124:
If all these drivers are using a 16 bit signed representation, it might be worthwhile making a library common code that does the rotation. Take in an xyz raw value and return a safely rotated/scaled floating point then.

If all the drivers have this issue, I'd prefer to see a clean fix. As much as it superficially appears critical, we have been flying this rotation code for years and I bet even when you are pulling saturating g forces noise make this a relatively minor practical problem...

peabody124:
P.s. @kubark42, true we end up floating quickly. Perhaps I'm being a bit silly trying to stick with fixed point.

kubark42:
As pointed out by @mluessi, https://github.com/TauLabs/TauLabs/blob/97c0a39290036a7601ce2ac86a8d03c2453192ec/flight/PiOS/Common/pios_mpu9250_spi.c#L656 is another solution to this problem.

Trex4Git:
There additional float variables are needed, as the the raw data is at first stored and converted to float in a temp buffer, and then, when the orientation stuff is done, the data is already float.

In the down-sampling PR i do it also like that, because i need a data buffer there (so on not SmallF1 targets that use MPU6000 this problem is also already fixed in that PR). 
But this implementation is not used on SmallF1 targets, so for them saving  the usage of additional variables might be an advantage, and i don't see a disadvantage in doing the sensor data orientation and float conversion in 1 step, when there is no other need for additional buffers.

kubark42:
@Trex4Git makes sense.

Thanks to @mlyle for implementing this. There seems to be a consensus forming around this solution. If there are no objections, I'll go ahead and merge it once someone confirms this has been bench or flight tested. (@AlessioMorale says "you can test with a bare board, by choosing the smallest range for gyro and turning fast using the usb wire".)

I agree with @peabody124 that it's unlikely it's caused many problems, but @AlessioMorale reported that this was originally discovered "that suddenly after we fixed some flight performance issues, sometimes it started spinning like crazy on hard rolls." so this bug has been observed in flight.

mlyle:
@peabody124 I agree these should end up factored common, but that's something that should happen after the release.

As far as the.. "this probably isn't critical" thing... Here's one of three spin-down-at-high-rate crashes I have had since setting roll rates to +/-1440deg/sec. https://www.youtube.com/watch?v=1VEJV3YKGXI  I had been blaming hardware or construction failures, but since it's happened three times I was starting to think the software needed to be audited.  This issue is a probable cause.

mlyle:
So I'm not done testing yet.

The good news is that this does not seem to cause regression on the targets I've looked at (Quanton, Colibri, CC3D).

The bad news is I do not thing I've been able to cause saturation by twirling-on-USB-cable.  I can get bad attitude estimates but I assume they're gain error.

I want to test a little bit more, and look at the output listing.  Third party review / test / "ok for merge" appreciated so that when I have satisfied myself I can merge and set about making a snapshot build (abandoning 8515a25c / e0d3d513).

kubark42:
Strange that you've been unable to get gyro saturation while twirling the cable. 500deg/s should be pretty easy. I frequently get that myself when doing these kind of twirling tests. Are you unable to cause the original bug, and thus demonstrate that this fixes it?

We should have a second dev merge when possible. There are several benefits to this, but for me the biggest is that it increases ownership of the code and also keeps members up to date with changes (since presumably the merger has reviewed and signed off on the code).

mlyle:
Oh, I should turn down the maximum rate down from 2000 deg/sec.  Duh.  That might help. :)

> Are you unable to cause the original bug, and thus demonstrate that this fixes it?

Yes, that's the current status.  I will have another chance to play with this sometime tomorrow.

> We should have a second dev merge when possible. 

I think there might be times when it makes sense for the originator to merge things themselves-- e.g. when getting a snapshot together or when the originator has some last things to confirm themselves.  Can't another dev saying "Yes, I would merge this" suffice?

We have a "PR for merge" tag that could be used for this purpose.

kubark42:
> Oh, I should turn down the maximum rate down from 2000 deg/sec. Duh. That might help. :)

Double plus ::))

> Can't another dev saying "Yes, I would merge this" suffice?

Historically, this has only been used when no other devs are active. Not saying that the past defines the future, but the least number of feathers will be ruffled when this principle is kept. 

I can't explain why, but I suspect that reviewers who press the merge button themselves are more attentive. Maybe for the same reason that the "undo send" function in GMail is so powerful. It's probably easier to pay attention to the real than the possible. 

Anyway, I'm digressing because it's 3AM and I need sleep.

mlyle:
Tested on sparky1 with the 500deg/sec limit.  Appears to solve the problem.  I couldn't get inversion every twirl but saw it a few times without the fix and none with the fix.

listing on sparky1 looks like:

```
 1084 062a 9DF80520             ldrb    r2, [sp, #5]    @ zero_extendqisi2
 1085 062e 9DF80430             ldrb    r3, [sp, #4]    @ zero_extendqisi2
 1086 0632 43EA0223             orr     r3, r3, r2, lsl #8
 1087 0636 1BB2                 sxth    r3, r3
 1088 0638 07EE903A             fmsr    s15, r3 @ int
```

On coptercontrol looks like:

```
 346 00b8 9DF81C30              ldrb    r3, [sp, #28]   @ zero_extendqisi2
 347 00bc 9DF81B00              ldrb    r0, [sp, #27]   @ zero_extendqisi2
 348 00c0 43EA0020              orr     r0, r3, r0, lsl #8
 349 00c4 00B2                  sxth    r0, r0
 350 00c6 FFF7FEFF              bl      __aeabi_i2f
 351                    .LVL29:
 352 00ca 00F10040              add     r0, r0, #-2147483648
```

Compared to the baseline of

```
 386 0108 9DF81A30              ldrb    r3, [sp, #26]   @ zero_extendqisi2
 387 010c 9DF81900              ldrb    r0, [sp, #25]   @ zero_extendqisi2
 388 0110 43EA0020              orr     r0, r3, r0, lsl #8
 389 0114 00B2                  sxth    r0, r0
 390 0116 4042                  negs    r0, r0
 391 0118 FFF7FEFF              bl      __aeabi_i2f
```

So it looks equivalently efficient.

mlyle:
Still, I need careful audit/inspection of the code because I only tested one platform and this was search and replace.

I have read it once, and I'm about to read it again.

mlyle:
Re-skimmed and it looks good.

I am in favor of, after this release, removing all this driver orientation code and teaching attitude/stabilization to add an offset to the board rotations based on hardware rotations.  We're already doing a rotation operation there, so.. *shrug*

That way as we add higher frequency stuff, filtering, etc, can be done before we normalize coordinate systems.

peabody124:
I'm less inclined to that. There is a lot to be said for pios_sensor providing nice normalized data. We used to have that set of mixtures in the sensor module and it just glot clunky, which is why abstracted it this way.

However, I'm all for having a shared library called from the drivers that uses some code with unit tests in place, though. If it can use a common enum type from hwshared all the

tracernz:
The axis orientation is sensor dependent so it makes sense to do rotation in the sensor driver. We should be able to refactor the Invensense drivers to share more of the code between them though.

mlyle:
It just seems silly to me that we might rotate the sensor by 90 degrees here and then end up rotating it back.  Ah well.

Can someone else please review/test/merge?  This is the sole issue blocking a RC build.

kubark42:
@mlyle, LGTM.

kubark42:
As a closing thought, I don't have much confidence that our rotations are correct in general, except for the orientations that are used by active boards. It's unclear what the convention should be for defining these angles. @mluessi and I discussed it and internally we decided to use the EE convention[*], where 0 degrees is defined by pin #1, and the rotation is always + in the clockwise direction when looking at the sensor from its top, no matter which side of the board the sensor mounted on.

I'll echo everyone's suggestion that this area of the code could be cleaned up and abstracted out. Ideally, it should be enough to look at the data sheet and define the three sensor axes of the sensor, with the rest of the macro math being done automatically.

[*]This is different from the mathematical convention, where there is one world reference frame and everything inside it is transformed relative to the origin, no matter which side of the board it's on.

Issue #1926, opened 2015-09-10
UAVO: New HwShared parent causes field misalignment


kubark42:
Not quite sure where this is happening (ground? firmware? both?), so pardon me for an incomplete bug report.

Symptom:

I have a hardware UAVO which has a UART3 field with only five options. The fourth option is I2C. 

```
<field name="Uart3" units="function" type="enum" elements="1" parent="HwShared.PortTypes"  defaultvalue="Disabled">
	<options>
		<option>Disabled</option>
		<option>Telemetry</option>
		<option>GPS</option>
		<option>I2C</option>
		<option>DSM</option>
	</options>
</field>
```
When selecting I2C via the UAVO browser and saving, upon rebooting the UAVO browser populates the name "DSM". Inside HwShared.xml, DSM is the fourth option. When moving I2C to the third place in the list

```
<field name="Uart3" units="function" type="enum" elements="1" parent="HwShared.PortTypes"  defaultvalue="Disabled">
	<options>
		<option>Disabled</option>
		<option>Telemetry</option>
		<option>I2C</option>
		<option>GPS</option>
		<option>DSM</option>
	</options>
</field>
```

And performing the same save/reboot operation, the UAVO browser populates the name with GPS (the third option in HwShared.xml).

I don't know if this is only a UAVO browser problem, or if it is a bonafide firmware misconfiguration. I'll look into it later when I get a chance, but for the moment I opted to remove all options from the UART, thus using the default from HwShard and ensuring compatibility. Note, this is not a good workaround, as many of the options aren't supported by my board.

Only tested on in-house hardware, don't know if this is repeatable on officially supported targets. I suspect it is, though.

mlyle:
Argh.  Thanks for the report.  I originally suspected some of the combobox indexing code would have to change in the uavo browser to support parent fields, but I did some test and it seemed OK in practice.  I will come up with the minimal fix ASAP.

mlyle:
I am having trouble reproducing this.  Right now I'm playing with sparky2 which has a pretty limited set of options on RcvrPort.  Both the uavobject browser and hardware pane seem to do the right thing.

kubark42:
Hmm... Okay, I'll dive more into it when I get the chance. It might mean that I implemented the HAL changes incorrectly, or that there's something else like #1921 which is most easily seen on a new target.

[...]

PR #1927, opened 2015-09-10
UAVOBrowser: Fix crash which occurred from an out of range combobox index

kubark42:
Came across this when investigating https://github.com/TauLabs/TauLabs/issues/1926.


mlyle:
Does this fix the crash or just make it easier to hunt down?  Won't the Q_ASSERT() just crash the process?

kubark42:
The Q_ASSERT will only be evaluated on debug builds. On a release it should only print the qDebug() warning.

kubark42:
P.S. Whoops, there was supposed to be a ```return``` there. I forgot to copy it in from the other branch. Fixed. This now resolves the crash.

Issue #1865, opened 2015-08-29
Implement and test fastheap support on F4


BrainFPV:
Opening this issue so it doesn't get forgotten. There have been several attempts to add fast heap support, but it keeps crashing on some targets.

Original PR: TauLabs/TauLabs#1691
Take 2 PR: TauLabs/TauLabs#1822 reverted in TauLabs/TauLabs#1855

The idea is to first identify what brings most memory savings and then enable fat heap for a small number of buffers.

mlyle:
Note we're not positive fastheap was guilty this time around.

tracernz:
Can be closed. Fixed in #1939.

PR #1928, opened 2015-09-10
Logging module improvements

mluessi:
Adds several improvements to the logging module:
- The logging module now respects the `logging_period` set in the UAVO meta data, so anything can be logged without having to modify the code.
- Ability to format the flash file system, which helps when logging things fast (as no erasing of pages is required).

The `logging_period` of all UAVOs have been set to sane defaults (can be changed on request). For settings objects, the `logging_period = 1` is used, which means they are logged whenever they are changed (which should happen infrequently when armed)

The greater flexibility comes at some cost. With the default settings in the PR, the module uses about 5% of the CPU (STM32F4, 168MHz) and 4kB of heap memory. Note that most of the memory should be allocatable in fast heap (see #1865). The current logging module uses about 1% of the CPU and almost no heap.

Things that could be improved:
- CPU utilization could be reduced. One way could be to use DMA with a callback in the flash fs driver, so the module doesn't have to wait while writing flash.
- Provide a way for the user to easily configure logging rates of all UAVOs. E.g. by extending the telemetry scheduler.

@kubark42 and I have been using this module over the last few months and it has proven to be very useful for development purposes. Suggestions for improvements and comments in general are welcome :).


mlyle:
jenkins: add to whitelist

mlyle:
jenkins: test this please

mlyle:
```
09:58:52 /var/lib/jenkins/workspace/taulabs-pr-linux-amd64/build/fw_colibri/logging.o: In function `writeHeader':
09:58:52 /var/lib/jenkins/workspace/taulabs-pr-linux-amd64/flight/Modules/Logging/logging.c:526: undefined reference to `date_from_timestamp'
```

mlyle:
This looks super awesome :)

mluessi:
@mlyle thanks :). Working on rebasing / changing the commit history so it merges cleanly into next.

kubark42:
@mlyle Thanks! Martin did a bang-up job on this and the performance has been excellent. We actually ran it a little faster in house, but the performance hit of running the loop at 1kHz isn't really sensible for most people.

For this PR, we hardcoded the loop speed at 100Hz, but @mluessi and I have discussed making that user configurable. We slowed it down principally because I discovered that we could do batch logging of high-speed signals (like sensor data) via dedicated UAVOs, and that seems to work well. However, doing a batch UAVO requires extensive knowledge of the build system, whereas a regular user could in theory want to log a single UAVO at a higher rate of speed than 100Hz.

Also, if reviewers don't feel that they can give specific attention to each proposed UAVO logging rate change, we'd be happy to split those problem children out. It's crucially important that we choose good default rates because they're a total PITA to change one-by-one in the UAVO broswer and settings export files **don't** save metadata.

mluessi:
@mlyle Fixed it so it compiles again on sparky2 and colibri.

I tried rebasing this branch so the reverts at the start are not necessary. However, this turned out to be difficult / messy. Basically, the reverts are necessary because the logging module in next and this one evolved in parallel, so the changes to logging don't merge cleanly into the current logging module.

Trex4Git:
Had not looked at the sources yet.
So this is more a wish list for logging in general. I don't know if its possible with this PR:
* logging of raw gyro data at full speed would be nice. 
 * But that is with my down-sampling PR 8kHz on some targets. So getting all the data for later analyzing, to see how the filters have to be setup  can be nice but is not essential. I think that might not be possible with this PR.

* logging the data of the stabilization module at full rate would be nice, like the blackbox logging in cleanflight seems to work. So logging the gyro & acc data, the RC input, the calculated P,I,D terms every time the stabilization control loop runs. 
 * With the down-sampling PR this runs e.g. at 2kHz with 50-55% CPU load. This would be helpful for tuning the PIDs and analyzing the behavior of the UAV. I think this is more important than the raw gyro data logging.

mluessi:
@Trex4Git look at the comments by @kubark42 above regarding logging faster. For gyros and accels, logging multiple values using a special UAVO would make most sense due to the large header overhead.

shared/uavobjectdefinition/watchdogstatus.xml
```diff
@@ -6,6 +6,6 @@
         <access gcs="readwrite" flight="readwrite"/>
         <telemetrygcs acked="false" updatemode="manual" period="0"/>
         <telemetryflight acked="false" updatemode="periodic" period="5000"/>
-        <logging updatemode="periodic" period="5000"/>
```
kubark42: I think this one actually shouldn't be changed.

phobos-:
I agree with @Trex4Git on the second part, I think adding PID calculated terms to logs would be really useful for tuning/improving the autotune algorithm. Logging Gyro data + RC command (this can be done at 50hz) + PID terms + motor ouptuts would be enough to get the perfect tune. Right now we don't have complete information as far as I know. I've tried to retrieve it using PicoC scripting but failed.

kubark42:
@phobos-, @Trex4Git, ideally, this logging function would continue to be optimized until it can keep up at these high rates. That will probably take a bit of deep thought, though, whereas the batch processing works for analysis needs. 

I can push our batch UAVOs for a PR discussion, we use them for exactly what you're describing. It's very nice to have **every** single measurement sample.

peabody124:
Wow. You can pull that off with internal flash without too much overhead? Awesome!

kubark42:
@peabody124 We had some problems in the beginning with data hiccups, but Martin's tried a few different approaches and found something that didn't stress the system too much. Preformatting can be important when running at higher speeds, since the time to erase an individual page can be very high relative to the incoming data rate.

lilvinz:
IIRC execution of cpu instructions stalls while writing to internal flash which could be pretty undesirable during flight. This may also depend on the internal flash type (F1 and F3 != F4).

mluessi:
@lilvinz usually the external flash is used for logging. The on-board flash on the STM32 is too small to be useful for this purpose.

flight/Modules/Logging/logging.c
```diff
@@ -381,25 +353,135 @@ static void loggingTask(void *parameters)
 				read_sector = loggingData.FileSectorNum;
 			}
 			LoggingStatsSet(&loggingData);
-
+			// fall-through to default case
+		default:
```
kubark42: Not for this PR, but we should probably finish logging whatever's in the queue when the logging disarms.

kubark42:
@mluessi This doesn't merge cleanly anymore because of the change to actuatorsettings.xml.

mlyle:
So sorry, I still have not done any more than cursorily skim this, and I don't think I'll probably get to this today.

It has conflicts but I don't wanna bug someone to do a rebase before review has happened.

One thing I think we should do is avoid dynamic linked lists in ram from compile-time known data, which this looks like it has.  I think we can tolerate that for now for the benefits, but we really need to claw that ram back.

kubark42:
@mlyle appreciate it. The merge conflict is because of the UAVO changes, I can fix it if timing becomes an issue.

Could you explain your hesitancy about the dynamic linked list? I'm not quite yet following how else this would look since the elements of the dynamic list are only known after the board has booted and read the UAVO metadata.

mlyle:
The contents of the UAVO metadata are known at compile time, right?  So there's no fundamental reason why this couldn't be formed into an ordered array that resides in flash; or just direct code that operates on the correct elements/sleeps for the correct time (e.g. by the uavobjectgenerator).  

I know not all files are on all targets, but it could work like uavobjectsinit.c does.

kubark42:
The metadata can be updated and saved. So the user can turn off the logging of individual UAVOs by setting ```loggingUpdatePeriod=0```.

mlyle:
Ah, I see.  That's unfortunate.  Now I'm starting to see where all our resources go.

flight/PiOS/Common/pios_streamfs.c
```diff
@@ -213,10 +213,22 @@ static int32_t streamfs_new_sector(struct streamfs_state *streamfs)
 	streamfs->active_file_arena_offset = 0;
 	streamfs->active_file_segment++;
 
-	if (streamfs_erase_arena(streamfs, streamfs->active_file_arena) != 0) {
+	// Test whether the sector has already been erased by checking the footer
```
peabody124: Can you explain this change since it's the only functional change in streamfs?

mluessi: The purpose is to speed-up writing on a freshly formated file system. Instead of checking the whole page, it checks only the footer and erases the page if the footer isn't all `FF`.

flight/Modules/Logging/logging.c
```diff
@@ -381,25 +353,135 @@ static void loggingTask(void *parameters)
...
+	if (loggingData.Operation != LOGGINGSTATS_OPERATION_LOGGING){
+		// We are not logging, so all events are discarded
+		return;
+	}
+	LL_FOREACH(log_info, info) {
+		if (info->obj == ev->obj){
+			now = PIOS_Thread_Systime();
+			if (info->next_log <= now){
+				// compute the time of the next logging operation
+				while (info->next_log < now) {
```
mlyle: This makes me a little nervous since it's not constant time.  I think it should divide, multiply, and add.  If someone sets navigation things to be logged frequently, and only enables a nav mode 10 minutes into flight, this is going to have to iterate a lot.

Maybe like.

```
/* How much do we need to advance the due date so the log would be newly due in this interval. */
info->next_log += ((now - info->next_log) / info->logging_period) * info->logging_period;
/* And advance into the next interval */
info->next_log += info->logging_period;
```

flight/Modules/Logging/logging.c
```diff
@@ -232,107 +219,92 @@ static void loggingTask(void *parameters)
...
-			PIOS_STREAMFS_Close(streamfs_id);
-			read_open = false;
-		}
-
-		if (loggingData.Operation == LOGGINGSTATS_OPERATION_LOGGING && !write_open) {
-			if (PIOS_STREAMFS_OpenWrite(streamfs_id) != 0) {
-				loggingData.Operation = LOGGINGSTATS_OPERATION_ERROR;
-			} else {
-				write_open = true;
+		switch (loggingData.Operation) {
```
mlyle: This control structure is pre-existing and maybe should survive this PR, but I'm not a fan of it.  Using a FSM in a task is basically a fancy way to write goto statements around the task function.  I understand it also exists so that a UAVO operation can control logging, but also allows a UAVO operation to send us to wrong states, etc.

flight/Modules/Logging/logging.c
```diff
@@ -232,107 +219,92 @@ static void loggingTask(void *parameters)
...
+			loggingData.Operation = LOGGINGSTATS_OPERATION_LOGGING;
+			LoggingStatsSet(&loggingData);
+			break;
+		case LOGGINGSTATS_OPERATION_LOGGING:
+			{
+				// Sleep between writing
+				PIOS_Thread_Sleep_Until(&now, LOGGING_PERIOD_MS);
+
+				// Log the objects with private queues
+				for (int i=0; i<PRIVATE_QUEUE_SIZE; i++){
```
mlyle: I just do not understand the use of multiple queues here.  Surely a single deep queue is better?

By the time something is put into a queue, we've decided that it is due to be logged.  Presumably it's equally bad-- perhaps worse-- to drop an infrequently logged sample as one of the high rate ones-- this isn't a prioritization mechanism to say "high rate things usually win," is it?  In either case we're not attaining the desired class of service.

Having to scan many queues adds overhead here and makes us slower to log the objects in them.  But the worse thing is it eliminates effective depth-- we could have a deeper queue if we didn't have many queues like the above, and it means we wouldn't start to drop objects until all of the queue depth is gone.  We also waste memory on queue structures that could instead be devoted to queue depth.

The one benefit I see is that it possibly lowers synchronization overhead, but if that's the core reason there's better ways to get that, too.

mlyle: P.S. if the desire is prioritization/choosing what things get dropped there are cheaper ways to attain better guarantees on that than this delivers.

flight/Modules/Logging/logging.c
```diff
@@ -381,25 +353,135 @@ static void loggingTask(void *parameters)
...
+static void obj_updated_callback(UAVObjEvent * ev)
+{
+	struct UAVOLogInfo *info;
+	uint32_t now;
+
+	if (loggingData.Operation != LOGGINGSTATS_OPERATION_LOGGING){
+		// We are not logging, so all events are discarded
+		return;
+	}
+	LL_FOREACH(log_info, info) {
```
mlyle: So things are sorted by logging interval... but this is unfortunate because it has nothing to do with the interval at which our callback will get called.  Unless I'm misreading this, this has paradoxical effects-- I could lower the interval at which gyros are logged, which bumps it to the end of the list, which in turn could greatly increase CPU utilization.

I think the interval/next time to log probably belong at the callback layer.  The benefits there are that there's no need to search an additional list at all and it's possible to entirely avoid these issues and remove the need for the list.  The downside is that it increases the memory use for all other callbacks even if logging is not used.  There's ways around this though-- for example the event mask could be updated to have a bit of HAS_MAX_INTERVAL where we allocate an additional trailer structure containing the last time and the interval.

mlyle: Actually delaying callbacks in general can be really bad, right, and jam up the event system?  So this is actually potentially a safety of flight issue.

shared/uavobjectdefinition/inssettings.xml
```diff
@@ -18,6 +18,6 @@
 		<access gcs="readwrite" flight="readwrite"/>
 		<telemetrygcs acked="true" updatemode="onchange" period="0"/>
 		<telemetryflight acked="true" updatemode="onchange" period="0"/>
-		<logging updatemode="manual" period="0"/>
+		<logging updatemode="manual" period="1"/>
```
mlyle: The use of 1 as a substitute for "always" is unfortunate here, because it advances things up to the front of the linked list where they compete with actual frequent events.

flight/Modules/Logging/logging.c
mlyle: One last comment and then time for zzzzzz.  If it's strongly desired to keep things at this layer, one "correct" mechanism is to move things to the front of the list when they're activated.  Then the list is (roughly) sorted by activation frequency instead of logging rate.

flight/Modules/Logging/logging.c
mlyle: (with the use of '1' for settings, if the flight controller's been up an hour this loop is going to have to count to 3.6 million which might take a quarter second on fast F4's)

mluessi: @mlyle very good point. As far as I can tell the code you proposed is correct.

flight/Modules/Logging/logging.c
mluessi: True, the fact that we can send a wrong state is not very nice. Not sure at the moment how it could be done more elegantly.

flight/Modules/Logging/logging.c
mluessi: Yes, the idea of having separate queues is to provide some fairness between objects. If you have a single queue and the user requests something like `Gyros` to be logged at a very high frequency, the queue can overflow if the write speed to flash is not sufficient and samples will be dropped.

Any suggestions on how to improve this?

flight/Modules/Logging/logging.c
mluessi: Hmm.. you are right, the scenario where a high frequency object has a low logging rate will indeed increase the CPU utilization. Which isn't "ideal".

Your suggestion to implement this in the uavobjectmanager callback layer makes a lot of sense. I have actually been thinking about this as well. We could have a `UAVObjConnectThrottledCallback` function that allows passing a minimum interval at which the callback is called. This could be useful for things like telemetry as well. The downside would be that we would need to extend `struct ObjectEventEntry` to include the interval which would increase the memory requirements by 4 bytes (assuming the interval is `uint32`), even for targets that don't use logging.

mlyle: 8 bytes, because you need to store when it's due, too.

But you could shorten the existing mask field by one bit, and then add a bit flag about whether there's an interval/those 8 bytes are present, and adjust the allocation appropriately.  I like this better because like you say there's a lot of other subsystems that could use it.

pseudo-code
```
struct ObjectEventEntryWithTimeout {
    struct ObjectEventEntry oee;
    uint32_t interval, due;
};

...
struct ObjectEventEntryWithTimeout *timeoutInfo;

if (flag) {
    timeoutInfo = (struct ObjectEventEntryWithTimeout *) &objectEventEntry;

   /* implement timeout.. */
}
```

flight/Modules/Logging/logging.c
mlyle: Personally I'd drop it and move to the uavo layer, because it drops most of the resource utilization and makes everything constant time.  Then other modules could benefit from it, too.

The main thing I would think of is a little yucky because our current API makes it require another copy.  Basically, add a volatile variable for how many of a given object type are allowed to be in the queue.  Atomically decrement that (do we have a header with atomics?).  If the result is less than 0, re-increment it and don't log.   Otherwise, enqueue the object + the pointer to the info structure concatenated.  When dequeing, atomically increment that field.

flight/Modules/Logging/logging.c
mluessi: I think you could get away with 4 bytes, i.e., only store when it is due. The interval is already in the meta data and we could grab it from there.

shared/uavobjectdefinition/inssettings.xml
mluessi: Yes, I agree, it's not very elegant. An option could be to use "0" for log always, and "-1" for no logging.

flight/Modules/Logging/logging.c
kubark42: @mlyle I'm having a little trouble following. Are you suggesting that this part of logging be integrated into the UAVO layer itself?

We had experimented with several approaches to queues, and empirically this is the best one we found. It has the lowest memory impact and ensures the least amount of jitter. Still, @mluessi and I have both felt this is a weak area of the logging module, and while it is fit for purpose we love the idea of approaching this from an even more efficient direction.

mlyle: Sorry-- the way I said that was a little convoluted.  My message refers to two different approaches.  In one, the linked list is removed in favor of using the UAVO-callback-layer to keep an optional ratelimit/throttle interval, and a single deep queue.  All of the searches involved and memory utilization of the list go away.  There is no effort to prioritize or be "fair" between different types of objects.   (I am not sure the existing fairness mechanism accomplishes much, and it sacrifices a lot of memory and queue depth to exist.  It also does expensive searches inside the callback/event path if any high rate object is ever set to be logged at a low rate).

On the other hand, if it's decided there must be a fairness mechanism for this to work effectively (I am skeptical), there's not so much of a win from factoring the interval into the UAVO layer.  In that case, what I would recommend is having a single queue along with a volatile variable that is accessed using atomics per source.  The volatile variable specifies how many things from that source are allowed to be in the queue.

In either case, queue depth could be shared between many possible senders and the receiver does not need to iterate/check many queues.  (It could actually block on the queue instead of sleeping an interval, too, which would improve logging performance).

I like factoring the interval to the UAVO layer because it looks like many subsystems could benefit, ultimately-- telemetry, logging, the various telemetry bridges... and because it really could make this module much more efficient in memory, CPU used, and log rate.

kubark42: That's a lot clearer, thanks. FWIW, there was no overt attempt at establishing a fairness mechanism. I see what you're getting at that it is a defacto fairness mechanism, but the assumption is that ALL queues are emptied. Otherwise, we would be getting queue overflow warnings.

I'll let @mluessi respond to this in more depth, but when we had a single deep queue the results were poor. There was a lot of jitter, and it even led to objects being dropped and long blanks in memory where high-rate objects were not logged. In short, the logs were effectively unusable.

As we look for ways to reduce memory and CPU impact, keep in mind that the total module cost is low, on the order of 1-2%. So improving efficiency will produce a measurable, but perhaps not noticeable, improvement. The bigger impact will come from reducing timing jitter. We can see a standard deviation of 6ms on objects being logged every 30ms.

mlyle: > I'll let @mluessi respond to this in more depth, but when we had a single deep queue the results were poor. There was a lot of jitter, and it even led to objects being dropped and long blanks in memory where high-rate objects were not logged. In short, the logs were effectively unusable.

That puzzles me and I suspect there was some other change involved.  I don't know of any way that more queues would be more efficient, given that they are all being generated by one task and drained by one other task.

> As we look for ways to reduce memory and CPU impact, keep in mind that the total module cost is low, on the order of 1-2%. So improving efficiency will produce a measurable, but perhaps not noticeable, improvement.

This is stuff that runs in event system callbacks and effectively is at very high priority.  It should be efficient.  Some of the issues with this approach are severe-- if you turn down logging of gyros, you get to scan the entire list for every gyro update while the attitude and stabilization threads are possibly waiting.

shared/uavobjectdefinition/inssettings.xml
kubark42: I'd rather go for 65535, if only because the resolution is sufficiently limited that we shouldn't sacrifice over 32 seconds of resolution in order to get a single negative value.

flight/Modules/Logging/logging.c
mluessi: The fairness is currently implemented is that we iterate over the private queues several times and write one entry from each queue each time. So if the write speed to flash is not fast enough, the private queues will start dropping elements with the UAVOs that are being logged more frequently dropping more. E.g., if the flash is only fast enough to write 10 UAVOs of each registered type per second, UAVOs with lower logging intervals will be logged at approximately 10 UAVOs / second.

The long memory blanks that @kubark42 are actually time blanks, which occur when a flash page is erased (takes about 50ms). This is still happening unless the file system is freshly formatted (this is the reason the format option was added).

I agree with @mlyle that implementing the throttling in the UAVO callback layer would be the cleanest solution that could also be used elsewhere (e.g. in the telemetry module, which is currently very complex). The only concern is the increased memory consumption, i.e., I'm not sure if the F1 targets could handle it if it is implemented in the way that every callback has an interval associated with it. Maybe we could use a separate callback list for throttled callbacks to avoid this problem.

In general, a single queue would be cleaner. Especially if we implement throttled callbacks in the UAVO layer, as this would mean the logging module wouldn't need a linked list. This would also make it possible to determine the timestamp at the time an object is added to the queue, not at the time of writing to flash. If the queue is sufficiently deep, it will act as a buffer during periods when flash pages are erased and there will be no gaps in the timestamps.

mlyle: > The only concern is the increased memory consumption, i.e., I'm not sure if the F1 targets could handle it if it is implemented in the way that every callback has an interval associated with it. 

You don't like my proposed solution to that in https://github.com/TauLabs/TauLabs/pull/1928#discussion_r41418120  ?  If you do that, there's no need for parallel data structures or additional lookups and only callbacks with interval pay a 4 or 8 byte penalty on size.

I can mock it up / do an initial implementation if unclear.

mlyle: Let me actually take a couple minutes to write that.

mluessi: Oh.. good call on using the mask to store the flag :). I didn't look at your code snipped carefully enough.

mluessi: Feel free to take a stab at adding throttled callbacks. I can then later rebase this PR to use this feature.

mlyle: This is just a little proof of concept of what the implementation would look like.

https://github.com/TauLabs/TauLabs/compare/TauLabs:next...mlyle:mpl-uavothrottle?expand=1

mluessi: Beautiful :) If you want, I can merge that into this PR and finish it once I find some time to do so.

mlyle: Sounds great-- thank you!

mluessi:
PS: Started reworking this on top of latest `next` (#1995 introduced a lot of conflicts). Should be ready soon.

mluessi:
While re-working the logging module code, I realized that the code was based on wrong assumptions about how the PIOS queues work. Basically, the logging module was using queues as buffers to store UAVOs that had been updated and needed to be logged. However, the queues don't actually store the UAVOs themselves, just the events that they have been updated. So when the UAVOs are retrieved from the queues and logged, the UAVO values at the time they are retrieved from the queues are used instead of the values at the time they were added, which is not what I intended to do.

It seems like to make the logging module work, we need to add a buffered logging file system with a method that can be called from the UAVO update callback that writes the data to a ring buffer in RAM from where the logging fs takes the data and writes it to flash. Like this it can handle slow downs that occur when erasing pages without blocking other processes.

Thoughts?

peabody124:
@mluessi if you look at the overo logging module you'll see that is exactly what it does. It has a 1KB ram buffer (coincidentally the DMA size) and takes low latency updates from the UAVO queue to store that instaneous value of the UAVO to the write buffer.

https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/OveroSync/overosync.c

mluessi:
@peabody124 ..the overo sync module is actually where I got the idea to use the queue from. It doesn't seem to use a ram buffer, i.e., it just takes the events from the queue and sends the corresponding UAVOs using `UAVTalkSendObjectTimestamped` (with the same problem that I described above).

mluessi:
Superseded by #2031. Closing this.

PR #1930, opened 2015-09-11
Attitude: Use logical AND instead of bitwise AND

kubark42:
Not a big deal. It always evaluated to true when it needed to. But it's nice to remove this kind of typo anyway.


mluessi:
Good catch. LGTM

PR #1932, opened 2015-09-11
Watchdog: Update watchdog UAVO only when it has changed.

kubark42:
The watchdog kind of spams the UAVO backend. This slows it down considerably, at the cost of a few bytes of memory.

### DIFF
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -347,12 +347,16 @@ static void configurationUpdatedCb(UAVObjEvent * ev)
  * Called periodically to update the WDG statistics
  */
 #if defined(WDG_STATS_DIAGNOSTICS)
+static WatchdogStatusData watchdogStatus;
 static void updateWDGstats() 
 {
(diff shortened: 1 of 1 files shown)


mlyle:
Does this noticeably impact GCS performance?  Just how spammy was it?

(I am concerned about GCS performance-- sometimes having the object browser open slows down my entire system-- seen on both Mac and Windows.  But it isn't repeatable/isn't all of the time.  Haven't had a chance to hunt this down yet).

kubark42:
It was very spammy internally, i.e. on the firmware side it was constantly rewriting the UAVO when there was no need to. 

Over telemetry, it was bounded by the telemetry limits. I came across the problem with the logging set to 1ms, where it was *very* spammy.

(I see the same freezing as well. Up to 10 seconds on particularly bad occasions. Never tracked it down to what was the guilty widget, though.)

BrainFPV:
It was also spamming the new logging module #1928, so this fix is very nice.

mlyle:

Issue #908, opened 2013-08-26
Initial gyro calibration is bad


kubark42:
I have a repeatable bad initial gyro calibration using next (058b06984a34a2c8d6c9c4a130e170e79c7b9ca9) on Revolution. The gyros wind up having between 1 and 3 deg/s of error. This seems small, but is in fact considerable and pulls the attitude estimate many tens of degrees out of alignment. Fortunately zeroing the gyros on arming solves the problem.

This might be linked to the problems-- and crashes-- I have had due to bad attitude estimates in the past.

PR #1933, opened 2015-09-11
State estimation: Fix complementary filter heading initialization

kubark42:
The initialization math required that the vehicle be level. This limitation was non-obvious, and probably why the error was made.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -448,12 +448,38 @@ static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool 
 			MagnetometerGet(&magData);
 		}
 
-		float RPY[3];
+		/* Initialization is tricky. We're using two reference vectors, one which
+		 * is the Earth's gravitational field and the other the Earth's magnetic field
+		 * While we know that gravity is always almost perfectly straight down, and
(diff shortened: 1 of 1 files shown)


mlyle:
Is this root cause for #908 / #917?

kubark42:
I doubt it's the root cause, but it might have been the initial clang that got the system ringing. The issue in #917 was that the convergence routine can drive the system unstable, even though the system is demonstrated to mathematically converge under (almost) all conditions. That indicates that a more adaptive/less aggressive convergence scheme would converge foar a wider range of bad initial values.

P.S. I discovered this thanks to the visualization in #1906.

mlyle:
I'm not in favor for this for endomorphism.  Is this OK?

peabody124:
It also might be worth making this a library method that creates the initial quaternion based on mag/acceleration abd use that from the ins on it code as we repeat essential the same thing.

I haven't reviewed this but i prefer it to turning off gyro bias estimation during init.

kubark42:
> I'm not in favor for this for endomorphism. Is this OK?

Fine by me, I've been using the fix for a year now. ;)

@peabody124 Good idea for making this a library. Although we shouldn't confuse this theoretical bug with the gyro bias initialization instability.

mlyle:
OK, I've read this and it pretty obviously is correct.

Issue #1936, opened 2015-09-13
gcs: "test outputs" on output pane broken when servo reversed


mlyle:
Moving the "neutral" right decreases the interval shown in the slider, and should be expected to decrease the duty cycle of the PWM.  It actually increases it.

<img width="900" alt="whoopsie" src="https://cloud.githubusercontent.com/assets/90903/9835257/6fdff886-5996-11e5-9360-fc76b03cc947.png">

mlyle:
Fixed by #1937

PR #1937, opened 2015-09-13
gcs: no longer re-invert motor test output

mlyle:
Was resulting in incorrect actuator commands in test with reversed
servos.

Issue #1936.  Bench-tested, but would be nice to have someone confirm I'm sane.

### DIFF
--- a/ground/gcs/src/plugins/config/outputchannelform.cpp
+++ b/ground/gcs/src/plugins/config/outputchannelform.cpp
@@ -330,9 +330,6 @@ void OutputChannelForm::sendChannelTest(int value)
     if (!ob)
         return;
 
-    if (ui.actuatorMin->value() > ui.actuatorMax->value())
-            value = ui.actuatorMin->value() - value + ui.actuatorMax->value();	// the channel is reversed
-
     if (ui.actuatorLink->checkState() && parent())
     {	// the channel is linked to other channels
         QList<OutputChannelForm*> outputChannelForms = parent()->findChildren<OutputChannelForm*>();


kubark42:
Looks good in theory. Historically, these sliders-- and their reverse states-- have given us a lot more trouble than you'd think would be possible. 

Since this only affects the test outputs, there's no way this could scramble the logic for reversing channels. Still, the fact that this bug has (re?)appeared makes me cautious. It would be good to see if the reverse button still works, including setting the neutral point to the right value (I believe in the past it didn't and this was one of the things we had to fix).

If no one tests today, I'll be able to give it a look tomorrow.

P.S. It would be good to have both input and output configuration sliders behave the same. IIRC, they're built on different code and move differently.

mlyle:
I'm neutral on this for endomorphism.  If nothing else is found majorly wrong with the RC, I'd say we should defer.  If this vets OK and there are other fixes coming in, maybe it should merge.

kubark42:
Bench tested and resolves the problem. The ActuatorCommand UAVO now exactly mirrors what is written in the text bubble.

Since this is confirmed bug with an extremely limited scope that doesn't impact flight code I'd vote to merge now.

P.S. One thing I noticed when playing with the reversed channels is that the linked channels link the absolute output values, not the relative values in the same sense. In other words, flipped channels' sliders will move backwards (from right to left) and channels that don't have the same range will update their values within that limited range and then saturate at the ends.

There's another bug in when you reverse a channel that's linked to another channel: it moves the neutral point of the linked channels.

mlyle:
Thanks for the test!

I'd prefer to ship the exact source code that's been tested by the user community if the RC tests well, even though this looks completely safe.  I'm leaving it tagged into endomorphism because this should merge if the RC is unshippable for any reason.

kubark42:
That sounds reasonable and I'm fine either way. But isn't the existence of this bug a proof that at least for one user the release **doesn't** test well?

peabody124:
Heh. Good point.

I'm fine either way. Regardless @mlyle you want to make a release-candidate branch so we can return to business as usual on next?

mlyle:
> That sounds reasonable and I'm fine either way. But isn't the existence of this bug a proof that at least for one user the release doesn't test well?

It's just a question of criticality.  Every week we hold things open for minor bugs is another week that users don't get the major fixes we've already proven.  What I have for minimum ship criteria is this:

1. The overall quality of the release is on par with the previous release.
2. No critical regressions are present from the time of the previous release.
3. No known safety-critical problems that a user is reasonably likely to
encounter are known.
4. The release snapshots have seen test across a wide variety of flight
modes, targets, and vehicles.

Overall I think we've fixed a lot of things and are great about point 1-- the GCS may be slightly crashier from the move to 5.5 but may not be-- it's subjective.  Point 2 is looking good.  Point 3, we have the OpenLRS and reset-spinup thing, but they're both not-super-actionable and I think we address them with release notes.  The fourth is where we come up short.  I have testing on it, @pug398 has flown the Windows artifact... and hopefully we hear some from the users from the post to the forum.

>  Regardless @mlyle you want to make a release-candidate branch so we can return to business as usual on next?

That's scheduled for 9-19 on the schedule, and we're running about a week behind because of Jenkins and the critical fixes at the end..  I'm amicable to picking up time and shipping/branching early if things look good but we don't know enough to do that yet IMO.  The way I look at it now, if the thing I just posted to the forum survives well through the weekend, it makes sense to assume we have a reasonable chance of shipping it or something very close to it and to get a release branch together.

mlyle:
@peabody124, @kubark42 ... I'm not super-available yet / resting a little and having fun testing the release.  But there's nothing to prevent us from getting a bunch of PR's pre-reviewed and ready for commit.  Then the moment we branch we can land 3-4 per day to clear the backlog without having too undue of a rate of change.

peabody124:
Ok tested on a brushedsparky with motors running. Does what is expected (although reversing those output channels was amusing/counterintuitive). Since we just merged 1944 @mlyle suggested merging this. Doing so.

PR #1938, opened 2015-09-14
vtol_follower: Move return to home vel to uavo

mlyle:
Not much to say-- make this configurable.  @pug398 asked for the ability to set the value higher, and I don't see the harm in making it a knob-- it's no more dangerous than the other knobs.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -53,11 +53,11 @@
 
 #include "pathdesired.h"
 #include "positionactual.h"
+#include "vtolpathfollowersettings.h"
 #include "vtolpathfollowerstatus.h"
 
 // Various navigation constants
 const static float RTH_MIN_ALTITUDE = 15.f;  //!< Hover at least 15 m above home */
-const static float RTH_VELOCITY     = 3.0f;  //!< Return home at 3.0 m/s */
 const static float RTH_ALT_ERROR    = 1.0f;  //!< The altitude to come within for RTH */
 const static float DT               = 0.05f; // TODO: make the self monitored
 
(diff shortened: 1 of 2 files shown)


mlyle:
Windows artifact at http://jar.lyle.org/~mlyle/TauLabs-20150914-3abcfbcd-install.exe

flight/Modules/VtolPathFollower/vtol_follower_fsm.c
```diff
@@ -671,8 +671,11 @@ static void go_enable_fly_home()
 		vtol_fsm_path_desired.End[2] = -RTH_MIN_ALTITUDE;
 	}
 
-	vtol_fsm_path_desired.StartingVelocity = RTH_VELOCITY;
-	vtol_fsm_path_desired.EndingVelocity = RTH_VELOCITY;
+	float rth_vel = 3.0f;
```
kubark42: IIRC , historically we've not assigned a default value when the UAVO variable is declared, in order to save flash. Since the variable is immediately overwritten by ```VtolPathFollowerSettingsReturnToHomeVelGet()```, it would seem safe to do the same here.

mlyle: That makes me really nervous.  Auditing the Get /  UAVObjGetInstanceDataField appears to show paths where these functions can fail, in which case rth_vel is left set to an undefined value.  Granted this is unexpected on the path follower settings object.

In accordance with #1792, we could alter those paths to fault/reset instead, and it'd start to be safe..

kubark42: Fair enough. Could add a comment-- either here or in the git commit message-- as to why this is set to ```3.0f```?

mlyle: I am going to audit and see how we handle this systematically.  I have a feeling we come up short.  I'm kinda inclined to follow the convention and not assign it, but open a separate issue/PR where we make these things fault for now on.

mlyle: Added a comment here for now.

mlyle:
This one is ready for merge, I believe.

mlyle:
Cool, thank you :)

pug398:
Yes thank you :)

PR #1939, opened 2015-09-14
Use Fastheap, take 3.

mlyle:
This is a revert of a revert, hence the odd history.  It reintroduces @brainfpv 's code that moves various allocations to fastheap.   The issue that describes this in #1865.

On [merge 1](https://github.com/TauLabs/TauLabs/pull/1691), there was a defect causing crashes on arm on some targets.  It was reverted.

On [merge 2](https://github.com/TauLabs/TauLabs/pull/1822), it was probably spuriously blamed for some crashes which related to the NaN changes.  But we are not sure. 

With careful vetting, this can come in as one piece.  But since it's been problematic we need to be especially cautious.  If there are any problems this can come in smaller chunks.

### DIFF
--- a/flight/Modules/FirmwareIAP/firmwareiap.c
+++ b/flight/Modules/FirmwareIAP/firmwareiap.c
@@ -182,7 +182,7 @@ static void FirmwareIAPCallback(UAVObjEvent* ev)
 						/* Note: Cant just wait timeout value, because first time is randomized */
 						reset_count = 0;
 						lastResetSysTime = PIOS_Thread_Systime();
-						UAVObjEvent * ev = PIOS_malloc(sizeof(UAVObjEvent));
+						UAVObjEvent * ev = PIOS_malloc_no_dma(sizeof(UAVObjEvent));
 						memset(ev,0,sizeof(UAVObjEvent));
 						EventPeriodicCallbackCreate(ev, resetTask, 100);
 						iap_state = IAP_STATE_RESETTING;
--- a/flight/PiOS/Common/pios_adc.c
+++ b/flight/PiOS/Common/pios_adc.c
@@ -81,7 +81,7 @@ static struct pios_adc_dev *PIOS_ADC_Allocate(void)
 {
 	struct pios_adc_dev *adc_dev;
 
-	adc_dev = (struct pios_adc_dev *)PIOS_malloc(sizeof(*adc_dev));
+	adc_dev = (struct pios_adc_dev *)PIOS_malloc_no_dma(sizeof(*adc_dev));
 	if (!adc_dev)
 		return (NULL );
 
--- a/flight/PiOS/Common/pios_flashfs_logfs.c
+++ b/flight/PiOS/Common/pios_flashfs_logfs.c
@@ -466,7 +466,7 @@ static struct logfs_state *PIOS_FLASHFS_Logfs_alloc(void)
 {
 	struct logfs_state *logfs;
 
-	logfs = (struct logfs_state *)PIOS_malloc(sizeof(*logfs));
+	logfs = (struct logfs_state *)PIOS_malloc_no_dma(sizeof(*logfs));
 	if (!logfs) return (NULL);
 
 	logfs->magic = PIOS_FLASHFS_LOGFS_DEV_MAGIC;
--- a/flight/PiOS/Common/pios_mutex.c
+++ b/flight/PiOS/Common/pios_mutex.c
@@ -52,7 +52,7 @@
  */
 struct pios_mutex *PIOS_Mutex_Create(void)
 {
-	struct pios_mutex *mtx = PIOS_malloc(sizeof(struct pios_mutex));
+	struct pios_mutex *mtx = PIOS_malloc_no_dma(sizeof(struct pios_mutex));
 
 	if (mtx == NULL)
 		return NULL;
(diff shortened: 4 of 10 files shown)


kubark42:
@mlyle Sounds good. Thanks for bringing this back!

P.S. I'm wondering if it would make sense for us to make a table with the list of issues and the devs/users who had them so we can confirm they're flying okay before merging.

peabody124:
Yeah it might be good to get some devs flying this for a week or two on whatever branches they are testing to avoid a regression again.

kubark42:
Have the original bug reporters are tested and confirmed that this doesn't crash anymore?

mlyle:
So I've tested the original scenario / crash from take 1 (which has since been corrected before take 2) It arms and flies fine.

When we had the issues with the NaN stack overruns we merged out take 2.  There were no issues that in retrospect that couldn't be explained by the NaN change.

kubark42:
Considering the back and forth on the fastheap, is it reasonable to wait a day or two more until an AQ32 guy can test? I'm willing to merge based on reported feedback, but it would help confidence if we could widen the test a little bit.

jihlein:
I'll try to give it a shot this weekend. Got a bit of a family emergency going on, I'll need a small diversion at some point.

mlyle:
@jihlein I'm sorry to hear that, but certainly appreciate any help you can give us :)

jihlein:
Okay, took this PR and added it to my "myNext" branch, then built the AQ32 firmware.  Loaded it onto a board in a 450 size quad.  Set it up with DSM satellite RX, GPS, ext mag, sync PWM at 666 Hz, batt monitor, altitude hold, path follower, and geo fence.  Flight modes acro, level, stab 1, altitude hold, position hold, and RTH.

I replaced RTH with autotune and did a tune in the backyard.  Tune completed successfully, gains feel good.  Put RTH back in place.  Flew acro, level, and stab 1 (horizon, horizon, rate) in the backyard, all seems good.

Going to try to head to the park later today and try altitude and position hold. Anythings else I should look for?

So far, so good on AQ32 hardware.

Does this need F3 testing?  I can build and load on the Naze32Pro if necessary.

mlyle:
@jihlein mostly concerned with F4 on this.  Thanks for the test.  If you get a chance, can you try flying althold?

jihlein:
Didn't have time to tune it, but flew both altitude and position hold with no odd behavior.  Altitude hold worked pretty well with stick centered, trying to adjust up or down got a little flaky, but nothing a little tuning won't cure.  Same for position hold, held position pretty well but loiter got a little silly looking.

peabody124:
Do we have a way of monitoring the stack space on this heap? I think having that is really important for this to not trip us up on the way...

kubark42:
@jihlein Thanks for the test report. 

I'm feeling pretty comfortable with our understanding of this change, and it seems the test results are good. @peabody124 would you like us to hold off on merging until we can monitor the memory usage on the heap, or is that something that should come in on a different PR?

peabody124:
@kubarak42 I think it is critical for this feature. It is non critical and we've had numerous crashes related to it. Knowing we aren't over allocating this memory is pretty key IMO and will help catch insidious crashes that could creep in.

mlyle:
@peabody124 We've had one issue on take one.  It didn't relate to resource overspend, but instead using the fastheap (non-DMA capable) for what is a DMA buffer on some targets.  Then I falsely blamed this for crashes that ended up being the NaN issue.

I'm not sure what you mean by the "stack space on this heap".  Do you mean the remaining heap space?  Or do you mean stack remaining on tasks allocated onto this heap?  Because we don't currently allocate task stacks onto fastheap in this diff.

mluessi:
@peabody124 it shouldn't be possible to over allocate memory in fastheap; if it is full it will fall-back to allocate it in the normal heap. However, I agree that it would be good to be able to monitor the free space in fastheap. It should be fairly easy to add `FastHeapRemaining` to `SystemStats`. That should do the job.

kubark42:
@mluessi Interesting. What happens if the allocation crosses a boundary?

mluessi:
@kubark42 it doesn't cross the boundary (memory has to be a contiguous). It just allocates the requested block in the normal heap instead and if there isn't enough space there either, the allocation fails.

kubark42:
@peabody124 so since according to @mluessi the chip's memory handler handles memory exhaustion gracefully, do you have any further objections? I agree it will be useful to monitor free fast heap space, but once allocations there fail they will start allocating on the normal heap, where we currently have good knowledge of free RAM.

mluessi:
@kubark42 it's not the chip's memory handler, it's our allocation routine. See:

https://github.com/TauLabs/TauLabs/blob/next/flight/PiOS/Common/pios_heap.c#L154

kubark42:
@peabody124 one last poke before merging. It seems that your concerns have been met by @mluessi's code, but it'd be great to hear that from both sides.

If there are no further objections, I plan to merge this tomorrow.

mlyle:
So I've been tied up this week, and behind/inactive.  I had a milestone of 9-29 for "backlog of PRs mostly merged, and smoke test completed".  How about we merge this one, and any others that are ready, and then plan to stop for 4-5 days.  We can ask @pug398 whether he can fly it, and I'll try and get some flight in over the weekend.   That'll give us confidence that the tree is in a good state and we have a good place to move forward from.  Sound reasonable?

kubark42:
Sounds reasonable, merging.

kubark42:
Yay, no problems reported after almost two weeks! Looks like we got it this time.

Issue #1881, opened 2015-08-31
Common functions should be moved to math header so they can get inlined


mlyle:
And so the compiler can realize they're pure functions.

Main ones I'm thinking of offhand are bound_sym / bound_min_max.  Maybe expo3?

mlyle:
(Also the compiler will be able to eliminate the code entirely in some circumstances).

kubark42:
http://embeddedgurus.com/barr-code/2011/03/do-inline-function-bodies-belong-in-c-header-files/#comment-23132 provides an interesting idea for resolving the messiness of putting functions into header files.

PR #1945, opened 2015-09-18
misc_math: correct vector clipping function.

mlyle:
When we exceed the limit, clip to the limit instead of 1.

### DIFF
--- a/flight/Libraries/math/misc_math.c
+++ b/flight/Libraries/math/misc_math.c
@@ -189,11 +189,11 @@ float vector3_distances(const float *actual,
 void vector2_clip(float *vels, float limit)
 {
 	float mag = vectorn_magnitude(vels, 2);    // only horiz component
-	float scale = mag / limit;
 
-	if (scale > 1) {
-		vels[0] /= scale;
-		vels[1] /= scale;
+	if (mag > limit && mag != 0) {
+		float scale = limit / mag;
+		vels[0] *= scale;
+		vels[1] *= scale;
 	}
 }
 
--- a/flight/Libraries/math/misc_math.c
+++ b/flight/Libraries/math/misc_math.c
@@ -192,8 +192,8 @@ void vector2_clip(float *vels, float limit)
 	float scale = mag / limit;
 
 	if (scale > 1) {
-		vels[0] /= mag;
-		vels[1] /= mag;
+		vels[0] /= scale;
+		vels[1] /= scale;
 	}
 }
 
--- a/flight/tests/misc_math/unittest.cpp
+++ b/flight/tests/misc_math/unittest.cpp
@@ -259,7 +259,8 @@ TEST_F(Vector2Clip, TestScale) {
     float test_vec_null[2] = {0, 0};
     float test_vec_within[2] = {limit/2, limit/2};
     float test_vec_edge_numerically_stable[2] = {limit, 0};
-    float test_vec_edge_numerically_unstable[2] = {sqrt(2)/2*limit, sqrt(2)/2*limit};
+    float test_vec_edge_numerically_unstable[2] =
+        {(float) (sqrt(2)/2*limit), (float) (sqrt(2)/2*limit)};
     float test_vec_outside[2] = {limit, limit};
 
     // Test for 0 vector


mlyle:
Since this is a correctness fix on something used in envelope limitation (though usually wrong in the safe direction) marking safety/blocker for release.  Comments appreciated.  Big whoops on my part here.

mlyle:
Thanks to @pug398 for the log that found this-- he's turned up RTH velocity to match/exceed HorizontalVelMax.  I am not sure this could/would be encountered with default settings, which is a mitigating factor.

flight/Libraries/math/misc_math.c
```diff
@@ -192,8 +192,8 @@ void vector2_clip(float *vels, float limit)
 	float scale = mag / limit;
 
 	if (scale > 1) {
-		vels[0] /= mag;
-		vels[1] /= mag;
+		vels[0] /= scale;
+		vels[1] /= scale;
```
kubark42: Wouldn't this whole function be better-- i.e. more efficient-- as

```
float mag = vectorn_magnitude(vels, 2);    // only horiz component

if (mag > limit) {
	vels[0] *= (limit / mag);
	vels[1] *= (limit / mag);
}
```

kubark42:
Good job finding this. 

As an aside, should all functions in ```misc_math.h``` have unit tests? That might have caught this earlier on.

mlyle:
Probably, yes.  We can require it as a criteria for merges of new func to misc_math for now on.

If this merges today, I will try to get together an rc tonight (since more people test over the weekend).

The good news is very little has been seen wrong with rc1.  The bad news is it has only seen moderate amounts of test from what I can tell.

I'll be flying cc3d and doing some acro today for some coverage of base functionality.

flight/Libraries/math/misc_math.c
mlyle: Maybe.  Fast-math lets the compiler do the inverse calculation.  In the worst case it is only a few cycles on something with a fpu, and maybe not even that if loads are executed at the same time or the compiler is feeling smart.

Moving a lot of this stuff as inline into the header should be a big win though--- because the compiler is free to reorder code over larger boundaries and eliminate larger common subexpressions.

peabody124:
Yeah good catch. I agree with @kubarak42. I think whenever we find a math bug too we should try and add a unit test showing the error and then fix it. Its good practice and helps block regressions.

Doesn't misc math have some unit tests already?

flight/Libraries/math/misc_math.c
kubark42: Sounds reasonable for the FPU.

I don't know if you caught that I changed the conditional so that we eliminate any math until after the comparison fails. That seems like it should be a win no matter what.

This is probably the most efficient (before any optimization):

```
float mag = vectorn_magnitude(vels, 2);    // only horiz component

if (mag > limit) {
	float scale = limit / mag;
	vels[0] *= scale;
	vels[1] *= scale;
}
```

kubark42: We also have a divide by 0 error when ```vels = {0,0}``` and ```limit <=0```. I suggest

```
float mag = vectorn_magnitude(vels, 2);    // only horiz component

if (mag > limit && mag != 0) {
	float scale = limit / mag;
	vels[0] *= scale;
	vels[1] *= scale;
}
```

flight/tests/misc_math/unittest.cpp
```diff
@@ -237,3 +237,54 @@ TEST_F(CircularModulusDeg, SweepError) {
...
+
+  virtual void TearDown() {
+  }
+};
+
+TEST_F(Vector2Clip, TestScale) {
+  float eps = 0.000001f;
+
+  // Choose a non-integer limit
+  for (int i=0; i<3; i++) {
```
kubark42: So it's open to debate as to whether this should be iterative or not. I wrote it for compactness, but an argument can be made that when unit testing no forensics should ever be performed in order to understand the test and that verbosity is valued above all things. So we could split this into three tests, ```TEST_F(Vector2Clip, LessThanUnity)```, ```TEST_F(Vector2Clip, Unity)```, and ```TEST_F(Vector2Clip, GreaterThanUnity)``` (and, eventually, ```TEST_F(Vector2Clip, InvalidInputs)```) if that's for the better.

mlyle:
```
12:29:24 unittest.cpp: In member function ‘virtual void Vector2Clip_TestScale_Test::TestBody()’:
12:29:24 unittest.cpp:262:61: error: narrowing conversion of ‘(((double)limit) * 7.0710678118654757e-1)’ from ‘double’ to ‘float’ inside { } is ill-formed in C++11 [-Werror=narrowing]
12:29:24      float test_vec_edge_numerically_unstable[2] = {sqrt(2)/2*limit, sqrt(2)/2*limit};
12:29:24                                                              ^
12:29:24 unittest.cpp:262:78: error: narrowing conversion of ‘(((double)limit) * 7.0710678118654757e-1)’ from ‘double’ to ‘float’ inside { } is ill-formed in C++11 [-Werror=narrowing]
12:29:24      float test_vec_edge_numerically_unstable[2] = {sqrt(2)/2*limit, sqrt(2)/2*limit};
```

Is the failure.  will fix.

kubark42:
Huh. That's why it's good to do unit testing on many platforms!

mlyle:
I'm fine implementing this however for expedience's sake-- I'll take your requested implementation.

For reference,  here's the listing without:

```
 458 0012 80EE080A              fdivs   s0, s0, s16
 461 0016 BDEC028B              fldmfdd sp!, {d8}
 468 001a F7EE007A              fconsts s15, #112
 469 001e B4EEE70A              fcmpes  s0, s15
 470 0022 F1EE10FA              fmstat
 472 0026 C1BF                  itttt   gt
 473 0028 D4ED007A              fldsgt  s15, [r4]
 474 002c C7EE807A              fdivsgt s15, s15, s0
 475 0030 C4ED007A              fstsgt  s15, [r4]
 477 0034 D4ED017A              fldsgt  s15, [r4, #4]
 478 0038 C4BF                  itt     gt
 479 003a 87EE800A              fdivsgt s0, s15, s0
 481 003e 84ED010A              fstsgt  s0, [r4, #4]
 483 0042 10BD                  pop     {r4, pc}
```

vs. the new is:

```
 458 0012 B4EEC80A              fcmpes  s0, s16
 459 0016 F1EE10FA              fmstat
 460 001a 12DD                  ble     .L44
 462 001c B5EE400A              fcmpzs  s0
 463 0020 F1EE10FA              fmstat
 464 0024 0DD0                  beq     .L44
 467 0026 88EE008A              fdivs   s16, s16, s0
 470 002a D4ED007A              flds    s15, [r4]
 472 002e 94ED010A              flds    s0, [r4, #4]
 475 0032 67EE887A              fmuls   s15, s15, s16
 477 0036 20EE088A              fmuls   s16, s0, s16
 480 003a C4ED007A              fsts    s15, [r4]
 482 003e 84ED018A              fsts    s16, [r4, #4]
 483                    .L44:
 486 0042 BDEC028B              fldmfdd sp!, {d8}
 491 0046 10BD                  pop     {r4, pc}
```

Both are correct for mag=limit=0  (this is also true with vanilla C; fast-math with its poorly specified behavior makes it questionable whether the source is correct on the original version).  Worst case performance is 3 cycles slower, typical un-clipped performance is 6-7 cycles better.  Whether this is a win or not is up for debate.  I'm kinda indifferent on the topic of improving typical performance with conditionals-- the way I see it, there's 3 possible outcomes:

1. The performance difference doesn't really matter.
2. The performance difference barely matters, so it provides a slightly higher class of service for lower priority tasks.
3. The performance difference is significant, so in addition to correctness challenges with the conditional we need to be sure that the performance impact of taking the conditional doesn't screw us up with jitter or overall CPU utilization.

kubark42:
> Worst case performance is 3 cycles slower, typical un-clipped performance is 6-7 cycles better.

It's totally not important, but if someone has time I'd love to see a discussion as to what could drive the process to cost more cycles. I'm not terribly familiar with assembly. 

Or are the additional three cycles taking the additional overload from the 0 check into account? Then I don't feel so bad, since that check really needs to be in the original version as well.

mlyle:
The compiler doesn't know the likelihood of any of these conditionals being true or false, so it tries to make both cases efficient (note that we can hint conditionals as likely or not).

The original version uses floating point conditional operations-- operations that only happen based on a flag, otherwise are nops.

The new version uses two compares and two branches, because there's not an easy way to get the results of both conditionals into one flag and that tips the balance towards having separate branches.

As to the 0 check, 

```
 458 0012 80EE080A              fdivs   s0, s0, s16
 461 0016 BDEC028B              fldmfdd sp!, {d8}
 468 001a F7EE007A              fconsts s15, #112
 469 001e B4EEE70A              fcmpes  s0, s15
```

fcmpes doesn't set the GT flag if either operand is a NaN.

kubark42:
Wow, this is neat stuff. Your comment about hinting which conditionals are likely or not got me reading http://igoro.com/archive/fast-and-slow-if-statements-branch-prediction-in-modern-processors/, which does a great job of explaining how the pipeline can work. Thanks.

mlyle:
No problem-- though Cortex-M3/M4 is very simple compared to most microarchitectures.

So this is why moving things to headers and inlining them is helpful, by the way-- for instance integer math and loads and stores are often "free" between floating point instructions.  So if things are in a header:

1. The compiler can be sure if it's a "pure function" / what side effects it has.  This allows the compiler to reorder it relative to other functions.
2. The compiler can evaluate constants across the function boundary and possibly eliminate portions of the inlined function.
3. The compiler can interleave other instructions in the middle of it to maximize use of the hardware resources.

This is why I have #1881 open-- it can be a really big difference.

The compiler is able to use profile information to decide what branch directions are "likely"-- whether it's overwhelmingly likely to go one way or there's about even odds.  Of course, we don't have profile information.  There's also `__builtin_expect`-- https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html   which is often wrapped into `#define LIKELY(x) (__builtin_expect((x), 1)` and `#define UNLIKELY(x) (__builtin_expect((x), 0)`.   So you can do `if (UNLIKELY(mag>limit)) {` ...  Which is probably a convention we should adopt.

flight/tests/misc_math/unittest.cpp
mlyle: Doesn't matter :)  It's not -that- bad to figure out if/when it breaks.

mlyle:
So can this merge?  Unsure who should really push the button.  ;) it's my PR but kubark's code.

kubark42:
> Unsure who should really push the button. ;)  it's my PR but kubark's code.

@mlyle Hah, so true! IMO, since I wrote the code, I'm most likely to suffer from "my baby is beautiful" syndrome. So for me, that means I'm the only one who can't merge. So if you wanna merge, be my guest. ;)

peabody124:
Looks good and nice analysis of performance @mlyle . @kubark42 thanks for getting the unit test in there! I can merge.

pug398:
Very nice job. RTH speed is plenty high now at default and also the surge of velocity at the end of return path seems to have been eliminated.

PR #1946, opened 2015-09-18
Unit test: Add test for vector2_clip()

kubark42:
This tests for the 0 vector, vectors which are within limits, vectors at the limits, and vectors longer than the limits. It tests for limits less than 1, equal to 1, and greater than 1.

Open to debate about whether the test should loop through and increase the limit at each point, or be broken into three separate tests. I kind of prefer the former because it reduces code duplication.

Test fails on next, but passes https://github.com/TauLabs/TauLabs/pull/1945/files.

Note that it does not test the case of ```limit<=0```.


mlyle:
Merged into #1945.  Thank you very much for taking the time to do this.

kubark42:
Pleasure!

PR #1948, opened 2015-09-20
vtol_follower_fsm: update to use waypoint handshaking

mlyle:
This is all stopgap still.. before real handshaking code happens to
control waypoint sequencing.  Another bug discovered by @pug398.  It
requires, I believe, multiple switches to RTH and getting unlucky.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -424,7 +424,9 @@ static int32_t do_hold()
 }
 
 //! The configured path desired. Uses the @ref PathDesired structure
-static PathDesiredData vtol_fsm_path_desired;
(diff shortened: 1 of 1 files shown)


mlyle:
(The consequence of this is that we end up flying back in endpoint mode, which is not awesome... and very fast).

Another thing noticed perusing these logs is that we saturate attitudes and velocities in funny ways.  We can end up flying a different direction because we saturate axes individually rather than cumulatively.

pug398:
PR-1948 appears to correct the arc in the return path and maintains the default 3 m/s RTH velocity now.

mlyle:
This should merge ASAP and then I'll create a release branch (presuming that we'll ship these bits in 4-5 days) and can do the initial pass of pyreneus merges.

mlyle:
Endomorphism release branch created with this merged.  Any comments on this fix welcome.

PR #1952, opened 2015-09-21
flight: Remove unused components.

mlyle:
This includes:  dosfs/SD card support,  headers for wav playing and video processing.  Nowhere close to being done, just grabbed a few of the more obvious things


jihlein:
Would like to see if we could use SD card on AQ32 in place of SPI eeprom.  It has an SD card connector, does not have SPI eeprom.

mlyle:
I agree we should bring SD card back, but this used old libraries and old hooks and was not compiled into target/rotting.

DTFUHF:
Ditto, SD card on a flight controller is definitely in the cards, is it not?

mlyle:
@DTFUHF -- I don't know.  It's something that both OP and TL moved to avoid because of problematic mechanicals, dust, etc.  And it's unlikely we can log to SD at full rates with the peripherals available.

In any case, we didn't have support for the F4 SD controller, etc.. and this wasn't compiled into any targets.

mlyle:
This formerly included code to do error correction which I thought was unused, but was removed in error.

Trex4Git:
So an SD card connected with SPI (on a seperate SPI, not shared with sensors) is not able to log at full rates of high speed control loop?

mlyle:
@Trex4Git this is what a SD card gets for 1, 100, and 16k bytes using fatfs (the old dosfs is unmaintained) on LPC's 4 bit wide "MCI".  http://elm-chan.org/fsw/ff/img/rwtest2.png

So, at one fourth the speed (since the semantics are write, flush, wait for flush to complete, write again) it looks like you could maybe get 120 kilobytes/second which would be enough... though when I was measuring this a couple of years ago I remember seeing considerable jitter where the SD card would "lock up" for 100ms to do some bulk erase operation or whatever.

tracernz:
It is enough if you encode the logged data in a similar fashion to @thenickdude's implementation for CF (using keyframes with full data and a predictor function and encoding the difference from the predictor using only the minimum number of bits required in between keyframes).

kubark42:
> It is enough if you encode the logged data in a similar fashion to @thenickdude's implementation for CF (using keyframes with full data and a predictor function and encoding the difference from the predictor using only the minimum number of bits required in between keyframes).

Have a link to the PR where this was proposed? That sounds like a fascinating way to get a lot more out of the logging system. Our logs can get very large (because we're recording every IMU, baro, mag, and GPS sample) and it's such a PITA to download them off the HID.

kubark42:
> @DTFUHF -- I don't know. It's something that both OP and TL moved to avoid because of problematic mechanicals, dust, etc.

In hindsight, this decision was probably a mistake and I suspect set the OP/TL ecosystem back significantly in logging and doing useful analytics. It's challening to automatically download logfiles because of their size and the slow and unpredictable speed over the USB HID. Plus, an SD card with GB of space would have been useful vs. the tiny MB of space we get out of the flash chips. (Even using the largest memory chip we could find for the footprint, we can't log an entire flight's worth of data before running out of space. And then because the logfile takes 10+ minutes to download and frequently fails to download successfully, we don't download unless we absolutely have to, which is a shame because we don't have a very good database of "good" flights.)

IMO, the benefits of having a robust and easy to use logging system outweigh the occasional mechanical problem. For the autopilot projects and boards which used SD cards vibration and dust haven't become significant issues. Oh, well, Harry Hindsight is 20/20!

kubark42:
@mlyle This looks complete. I agree that the SD card code is so bit-rotted that it doesn't make much sense to keep it around. 

(Lemme know if the PR is ready to merge and I'll push the green button.)

Trex4Git:
Does that mean also for future SD card support is not welcome, or is the current implementation not working and need improvments ? 

As e.g. AQ32 seems to have a µSD card slot.

kubark42:
@Trex4Git I would love to see SD card support. I suspect that the current implementation would need some a code refresh to bring it back up to date. I can't recall ever seeing anyone use it, so I suspect it must be at least 4 years bit-rotted.

This shouldn't discourage anyone, though! If someone wants to take it on, it would be a valuable contribution.

mlyle:
I think it's probably worth starting over for SD card support, to use the fatfs module instead of dosfs, and to use DMA / probably restrict this to F3 and F4 controllers.  I don't know, is AQ32 hooked up to the SD lines or SPI?

@kubark42-- thanks-- I am OK with this merging.

jihlein:
AQ32 SD Card is hooked up to a dedicated SPI interface.

tracernz:
@kubark42 This is the original Cleanflight PR for the blackbox logging feature: Cleanflight/Cleanflight#227

kubark42:
@tracernz Thanks!

@mlyle Merging.

thenickdude:
I've now written up Cleanflight's Blackbox log specification here, I hope it will be of use:

https://github.com/cleanflight/cleanflight/blob/master/docs/development/Blackbox%20Internals.md

If you want an example log file to look at, here's a good one:

http://www.rcgroups.com/forums/showthread.php?t=2386267&page=7#post32683275

tracernz:
@thenickdude I just saw that on IRC and came here to post it. Thanks very much! :D

PR #1955, opened 2015-09-22
AltitudeHold: Scale altitude hold by 10, and control in dm/s

kubark42:
This provides more granularity to autonomous altitude change rates. There didn't seem to be much reason to have a rate resolution of [m] when our altitude precision is on the order of [cm].

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -1072,8 +1072,14 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
 		altitudeHoldDesired.Altitude = -current_down;
 		altitudeHoldDesired.ClimbRate = 0;
 	} else {
-		uint8_t altitude_hold_expo, altitude_hold_maxrate, altitude_hold_deadband;
-		AltitudeHoldSettingsMaxRateGet(&altitude_hold_maxrate);
+		uint8_t altitude_hold_expo;
+		uint8_t altitude_hold_maxrate10;
+		uint8_t altitude_hold_deadband;
+		AltitudeHoldSettingsMaxRateGet(&altitude_hold_maxrate10);
+
+		// Scale altitude hold rate
+		float altitude_hold_maxrate = altitude_hold_maxrate10 * 0.1f;
+
 		AltitudeHoldSettingsExpoGet(&altitude_hold_expo);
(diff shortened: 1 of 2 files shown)


mlyle:
Argh, just yuck.  Probably a good change but I'm not a huge fan of fixed point in the uavos, but it's probably the best possible change.

The name of maxrate should change though.  I don't think we use unit names in computing uavo hashes.

kubark42:
> The name of maxrate should change though. I don't think we use unit names in computing uavo hashes.

I really don't like this. The field is appropriately named, and renaming it **forever** so that a handful of users don't have a problem **now** seems like it's going to get us in trouble in the long run.[1]

That being said, I understand where the problem is coming from and don't have an immediate answer. In most cases, the default will get loaded automatically since users who haven't the UAVO will get the defaults on boot, but that's not **all** cases, which is what we really need to handle when doing a scale change like this.[2]

But I really, really, really want to avoid the principle of changing a variable name just to kick the UAVO.

[1] Of course, if someone proposes a better name, then by all means!
[2] This reminds us the importance to examine proposed UAVOs carefully, since historically it frequently has been traumatic to change them.

mlyle:
Yes I agree it's unfortunate.  We can add units to uavo hashes but that breaks all configurations.  It also doesn't handle the import case, so it doesn't really eliminate the problem.

MaxAltRate? :P

kubark42:
```MaxClimbRate```? That makes space for a ```MaxDescentRate```, which should definitely be different from the max climb, since a 4m/s descent almost guarantees entering into a ring vortex state.

mlyle:
Yup, a MaxDescentRate was in my plans, too.  ;)  we'll see who gets to it first.

kubark42:
@mlyle Fixed. Ready to merge, I think.

> Yup, a MaxDescentRate was in my plans, too. ;) we'll see who gets to it first.
https://github.com/cyphyworks/TauLabs/tree/altitude_hold_descent_rate :D

mlyle:
Alrighty.  Probably going to slow down the merge rate a little because I feel like we've accepted plenty of risk in the past few days.

kubark42:
> Probably going to slow down the merge rate a little because I feel like
we've accepted plenty of risk in the past few days.

No problem. I've been doing my best only to submit stuff that's been
running for months so as to minimize risk. I really appreciate how these
PRs have been tied up so quickly.

-- 
Sent from my Banana Phone.

PR #1956, opened 2015-09-22
Uavobrowser search

kubark42:
This is a cool feature I was asked to develop. It provides a search field for the UAVO browser.

Works very intuitively, all text matching the string is found. This includes partial strings, even ones which are embedded in the middle of a larger string. So it doesn't just search UAVO names, but looks at the children fields as well.

<ESC> button works as expected, i.e. clearing the text. 

Ideally, it would be nicer if the CLEAR button were replaced by a small 'X', like in search fields in cell phones, but that's a nice improvement for another day.

screen shot 2015-09-22 at 10 42 58 am
screen shot 2015-09-22 at 10 43 15 am

In use for more than a year, no bugs found or anomalies to report.


mlyle:
Nice!

PR #1962, opened 2015-09-24
misc_math: unit test: add tolerance to floating retvals

### LINKED ISSUE #1960: linear_interpolate unit tests fail on linux-i686
Probably a floating point precision thing.  All the values "match" on output.

mlyle:
Attempt to fix #1960.  Somewhat in the blind because I don't have linux-i686.  But I have pushed the branch to the Tau repository too, so we'll see if Jenkins spits out an i686 artifact.

@ekaszubski FYI.

### DIFF
--- a/flight/tests/misc_math/unittest.cpp
+++ b/flight/tests/misc_math/unittest.cpp
@@ -305,23 +305,24 @@ TEST_F(LinearInterpolate, ThrottleCurve1to1) {
   float const range_max = 1.0f;
   uint8_t const curve_numpts = 5;
   float const curve[5] = { 0.0f, 0.25f, 0.5f, 0.75f, 1.0f };
+  float eps = 0.000001f;
 
   // 21 points in range
   for (size_t i = 0; i <= 20; ++i) {
     float const input = i * 0.05f;
-    EXPECT_EQ(input, linear_interpolate(input, curve, curve_numpts, range_min, range_max));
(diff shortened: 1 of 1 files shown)


kubark42:
Looks reasonable. There's a fair amount of FPU action going on, so being at worst within 4 significant digits across the entire range is pretty good indication that the coverage is complete.

flight/tests/misc_math/unittest.cpp
```diff
@@ -305,23 +305,24 @@ TEST_F(LinearInterpolate, ThrottleCurve1to1) {
   float const range_max = 1.0f;
   uint8_t const curve_numpts = 5;
   float const curve[5] = { 0.0f, 0.25f, 0.5f, 0.75f, 1.0f };
+  float eps = 0.000001f;
```
kubark42: This could probably be several orders of magnitude more strict, but 1e-6 is fine.

mlyle: Single precision floating point is 6 sigfigs of precision.  Arguably we should have one less zero.

kubark42: Several means ```1```, [right](https://xkcd.com/1070/)? <grin>

mlyle:
And this spat out of jenkins happily, so please merge?

kubark42:
Yay, Jenkins!

PR #1963, opened 2015-09-25
Actuator: Curve fit motor channels

kubark42:
This provides a power curve fitting for the motor channels. All motors are assumed to be identical. This is a poor assumption, but it works in most common scenarios.

* The fitting maps the linear [-1,1] range to a nonlinear [-1,1].
* It uses an equation of the type y = a*x^b.

Been flying with this for about 5 months, it makes a pretty significant difference, especially when using ESCs in closed loop mode where the setpoint is the RPM.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -142,7 +142,7 @@ int32_t ActuatorInitialize()
 
 	return 0;
 }
-MODULE_INITCALL(ActuatorInitialize, ActuatorStart)
+MODULE_INITCALL(ActuatorInitialize, ActuatorStart);
 
 /**
  * @brief Main Actuator module task
@@ -340,12 +340,23 @@ static void actuatorTask(void* parameters)
 				continue;
 			}
 
-			if((mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) || (mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_SERVO))
+			if((mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) || (mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_SERVO)) {
 				status[ct] = ProcessMixer(ct, curve1, curve2, &mixerSettings, &desired, dT);
-			else
+
+				// If a motor type, additional work to be done
+				if(mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) {
+					if (status[ct] > 0) {
+						// Apply curve fitting, mapping the input to the propeller output.
+						status[ct] = actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_A] *
+						      powf(status[ct], actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_B]);
+					} else {
+						 // Idle throttle
+						status[ct] = 0.0f;
+					}
+				}
+			} else {
 				status[ct] = -1;
-
-
+			}
 
 			// Motors have additional protection for when to be on
 			if(mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) {
@@ -454,11 +465,6 @@ float ProcessMixer(const int index, const float curve1, const float curve2,
 	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_ROLL] * desired->Roll) +
 	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_PITCH] * desired->Pitch) +
 	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_YAW] * desired->Yaw)) * (1.0f / MULTIROTOR_MIXER_UPPER_BOUND);
-
-	if((MixerEntry_t[index].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) && (result < 0.0f))
-	{
-			result = 0.0f; //idle throttle
-	}
 
 	return(result);
 }
--- a/shared/uavobjectdefinition/actuatorsettings.xml
+++ b/shared/uavobjectdefinition/actuatorsettings.xml
@@ -8,6 +8,10 @@
         <field name="ChannelMin" units="us" type="uint16" elements="10" defaultvalue="1000"/>
         <field name="ChannelType" units="" type="enum" elements="10" options="PWM,PWM Alarm,Arming LED,Info LED" defaultvalue="PWM"/>
         <field name="MotorsSpinWhileArmed" units="" type="enum" elements="1" options="FALSE,TRUE" defaultvalue="FALSE"/>
+
+        <!-- Actuator mapping of input in [-1,1] to output on [-1,1], using power equation of type a*x^b-->
+        <field name="MotorInputOutputCurveFit" units="-" type="float" elementnames="A,B" defaultvalue="1,1"/>
+
         <access gcs="readwrite" flight="readwrite"/>
         <telemetrygcs acked="true" updatemode="onchange" period="0"/>
         <telemetryflight acked="true" updatemode="onchange" period="0"/>


mlyle:
Is the A necessary?  x^b is range-preserving.

mlyle:
I assume you guys end up using a power like 1.5 or something, assuming you can discuss.

flight/Modules/Actuator/actuator.c
```diff
@@ -340,12 +340,23 @@ static void actuatorTask(void* parameters)
...
 				status[ct] = ProcessMixer(ct, curve1, curve2, &mixerSettings, &desired, dT);
-			else
-				status[ct] = -1;
-
 
+				// If a motor type, additional work to be done
+				if(mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) {
+					if (status[ct] > 0) {
+						// Apply curve fitting, mapping the input to the propeller output.
+						status[ct] = actuatorSettings.MotorInputOutputCurveFit[ACTUATORSETTINGS_MOTORINPUTOUTPUTCURVEFIT_A] *
```
mlyle: This should be adjusted to do the right thing for negative numbers, for future bidir support.

kubark42:
> Is the A necessary? x^b is range-preserving.

It helps, but it's not necessary. 

To be honest, it tends to be a value close to 1, and it is a fair question to look at what happens at full throttle. I tend to be a little greedy and go after the best precision when possible. I'd rather lose a tiny bit of accuracy on the limits and have maximum accuracy for the normal control band.

> I assume you guys end up using a power like 1.5 or something, assuming you can discuss.
On the motor/prop/ESC combo I have here my curve fitting is ```A=0.9919```, ```B=0.4412```.

flight/Modules/Actuator/actuator.c
kubark42: This is where I discovered the ```NaN``` problem. ;)

If it's okay, I'd like to leave the bidirectional support for later. It will require a different set of curve fitting constants for bidirectional support and I don't know how I would test that efficiently. Currently ```ProcessMixer()``` saturates the motors' lower value at 0, so in theory we should never be able to have a negative value at this point. But I put the check in anyway to ensure that we don't accidentally pass a negative number no matter what happens upstream.

P.S. There's a conversation here about the fact that motors are treated as things that can only run forward, and I think we agree that it's a shame that this assumption is sort of baked into our process.

mlyle:
oh yah I was thinking the inverse.  so that's interesting, a 2.26 power law.  most of what the guys are measuring on rotorbench for actual PWM time is in the 1.2-1.7 power range.

A makes me nervous because it introduces new clipping, and we have vehicle types (gemini) that spend time in the clipped regime.  (I have some plans in this area, along with a silly branch in my tree called "ghettomixer" which hacks around this.)

kubark42:
> ...it introduces new clipping...

If the A and B values are left at 1.0 then this is a pass-through. That would avoid any clipping on the Gemini (at the expense of not being able to curve-fit).

> most of what the guys are measuring on rotorbench for actual PWM time is in the 1.2-1.7 power range.

I'm not familiar with their work, but based on your comment I wonder if they're measuring the same thing. In all propellers we've tested, from 2" to 18" relationship between RPM and thrust is close to quadratic, whereas 1.2 sounds like it's a long way away.

P.S. I think you figured this out, but for the record the constants are found by doing an RPM vs. thrust sweep, and normalizing both RPM and thrust onto [0,1], and then finding the curve that fits thrust to RPM.

mlyle:
So with typical ESCs (not RPM commands), and complementary output / "damped light" / etc you you about n^2.  Without complementary output from the ESC it's much more linear.

Lots of overpowered setups are basically sigmoid as you start to run into prop flex and and other saturation effects.

dys1306

e.g. this is a Dys 1306 3100KV driven at 3S by SimonK ESC firmware.  (thrust vs. microseconds input PWM)

flight/Modules/Actuator/actuator.c
mlyle: I was thinking you'd just invert the input and output.. -powf(-x, y) .. granted the prop may not respond exactly the same in reverse depending on geometry of things around it and whether it's a symmetric foil, but it ought to be close.. and beats coming up with some nasty series expansion

> P.S. There's a conversation here about the fact that motors are treated as things that can only run forward, and I think we agree that it's a shame that this assumption is sort of baked into our process.

People are asking me to prioritize fixing that, btw-- make Tau Labs fully 3D and make ground vehicles actually work.  I'm a nav nerd and kinda don't really care about that stuff but I get persuaded to work on acro stuff. ;)  Though I may be working on a vehicle soon that may need negative thrust in some flight regimes for stability.

mlyle: I understand completely if you don't want to take that on for this PR though.

kubark42: Generally, we've had a strong principle about not adding code that doesn't provide complete functionality. But handling the negative case is logical so I can go either way, and am fine with whatever the reviewer(s) decide(s).

> prop may not respond exactly the same in reverse depending on geometry of things around it and whether it's a symmetric foil, but it ought to be close

I expect it to be **very**, **very** different. We've never formally tested our props running backwards, but theoretically there should be a tremendous difference on non-symetrical props. 

Assuming my above assertion is correct, one interesting thing of note is that we will have ```A_reversed << 1``` in the case of reverse props, assuming that ```status[idx]=1``` represents maximum forward thrust, and thus ```status[idx]=-1``` should represent an equal number of Newtons of reverse thrust.

(Not sure if the above sentence is clear, if it's not I'll take another stab at explaining it.)

kubark42:
> Lots of overpowered setups are basically sigmoid as you start to run into prop flex and and other saturation effects.

Oooh, good point. We never have overpowered systems here because agility is not what we're focused on.

> this is a Dys 1306 3100KV, 4045 plastic prop driven at 3S by SimonK ESC firmware. (thrust vs. microseconds input PWM)

That looks a lot like the graphs I see when PWM maps to constant DC/Power, instead of constant RPM. In that case, my B constant tends to be closer to 1, but still far enough away from 1 that it makes a noticeable difference. 

^^^ With the caveat that we never go sigmoid because of overpowering the props.

flight/Modules/Actuator/actuator.c
mlyle: Yah.  I understand what you're saying.  And those are nice rational units to come into actuator.  And this makes a lot more sense when the controlled value is RPM.   When it's PWM (which basically all of our users are running) / some coarse idea of actuator effort on an ESC I'm not sure how true this holds.  Most people doing 3D are using (nearly) symmetric airfoils-- the main difference are prop texture and what side of the airstream the arm is on.     It's also problematic because then negative throttle needs a different curve for the 3D case, because you'll almost certainly not get as much negative thrust as positive.  And there's going to be something icky around 0 throttle no matter what-- both linearity from command to output but also strange time-domain effects.

Basically, I think the difference between A_reversed and just doing it one sided is something like: https://www.desmos.com/calculator/dcd8rumbqm

All in all, it's a difficult set of tradeoffs.

What we've been discussing on IRC for when we have smart ESCs (like perhaps ESC32 integration) is to keep something much like the current actuator effort on the output,  but to run inner angular acceleration and throttle loops based on ESC power or RPM feedback from an inverted mixer matrix.  The higher frequency components of that data could also be (weakly) blended into the rate estimate to further improve the bandwidth of the rate and attitude loops.

So it's an interesting question.  Obviously predistorting/linearize the command is helpful in either case.

mlyle:
Another question is whether we should de-emphasize TPA in favor of this.

kubark42:
> Another question is whether we should de-emphasize TPA in favor of this.

I think your instinct is correct. If the actuators were ideal, then TPA would not be necessary, and in theory this helps increase the linearity of the actuators.

However, I wonder how well a user can determine the curve fit. We have a highly precise and well calibrated test stand, so we're quite confident of our measurements.

Not to mention the fact that these measurements go out the window as soon as you're not measuring static thrust. I have a hunch that controlling in terms of pure power is better than RPM, since power is roughly proportionate to momentum imparted to the air molecules, whereas RPM varies based on a slew of factors, such as climate, geography, flight envelope, etc...

mlyle:
So I've thought about this.  I'm going to merge it, but plan on trying to replace it soon.  I think the conversation here makes it really apparent that mixers really need to be arbitrary functions.

There's other things, like tricopters should really be altering tail thrust based on tangent yaw command, etc.  Which you can't even fake now because there's no way to take an absolute value...

kubark42:
I would be all for arbitrary mixer/actuator mapping functions. This is what you need to almost anything that's not a multirotor.

I wonder if arbitrary math like this could be done through PicoC? Ideally, changing mixer algorithms wouldn't require the user recompiling the firmware.

mlyle:
PicoC is overkill and doesn't fit on all platforms.

We just need a small set of possible functions and a way to combine them.  A simple expression syntax.  I'll think about it some more.  Maybe a little RPN stack thingy.

kubark42:
Noted. I'm assuming that someone who is playing around with new algorithms will be doing this on capable hardware. My thought is that no matter what set of non-linear functions we make available, there will be another combination of state and control which will be useful which we haven't foreseen. I'm thinking of highly-unstable high-efficiency flying wings.

But, jah, having a small set of polynomials will probably be more than sufficient for most people.

PR #1964, opened 2015-09-25
AttitudeHold: Change default deadband

kubark42:
The deadband was very wide at 30%. This reduces that by half. It seems a more sensible default.

Flying with this for the better part of a year. Works fine with Taranis. I think the deadband is still wide enough that even with lower-quality transmitters it should be good.


mlyle:
I realized I regressed this before and have been meaning to fix it..  The deadband in the code is defined poorly (and differently from loiter).  A deadband of 30 is actually a deadband of 60%.. that is, 30% of the whole scale above and 30% of the whole scale below the midpoint.  My plan was a /2 in the code, which would came to the same value and match loiter.  Thoughts?

kubark42:
Sounds like a better plan. In an ideal world they should all be getting their configuration from the same place, but that's probably for the future. 

Should I close this?

mlyle:
Sure.  Opened #1967 as an alternative.

kubark42:
Closed in favor of #1967.

PR #1965, opened 2015-09-25
AltitudeHold: Add support for a separate maximum descent rate

kubark42:
Previously, maximum climb and descent rates were saved in the same setting. This leads to problems where the maximum climb rate can safely be far higher than the maximum descent rate. This PR simply separates the two and sets a reasonably safe rate for maximum descent.

Tested and confirmed to work.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -1074,11 +1074,14 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
 	} else {
 		uint8_t altitude_hold_expo;
 		uint8_t altitude_hold_maxclimbrate10;
+		uint8_t altitude_hold_maxdescentrate10;
 		uint8_t altitude_hold_deadband;
 		AltitudeHoldSettingsMaxClimbRateGet(&altitude_hold_maxclimbrate10);
+		AltitudeHoldSettingsMaxDescentRateGet(&altitude_hold_maxdescentrate10);
 
 		// Scale altitude hold rate
 		float altitude_hold_maxclimbrate = altitude_hold_maxclimbrate10 * 0.1f;
+		float altitude_hold_maxdescentrate = altitude_hold_maxdescentrate10 * 0.1f;
 
 		AltitudeHoldSettingsExpoGet(&altitude_hold_expo);
 		AltitudeHoldSettingsDeadbandGet(&altitude_hold_deadband);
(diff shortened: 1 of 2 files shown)


flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1090,9 +1093,9 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
 		if (cmd->Throttle > DEADBAND_HIGH) {
 			climb_rate = expo3((cmd->Throttle - DEADBAND_HIGH) / (1.0f - DEADBAND_HIGH), altitude_hold_expo) *
 		                         altitude_hold_maxclimbrate;
-		} else if (cmd->Throttle < DEADBAND_LOW && altitude_hold_maxclimbrate > MIN_CLIMB_RATE) {
+		} else if (cmd->Throttle < DEADBAND_LOW && altitude_hold_maxdescentrate > MIN_CLIMB_RATE) {
```
mlyle: I may be being thick headed.  What's the reason for the second half of this conditional?

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1090,9 +1093,9 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
 		if (cmd->Throttle > DEADBAND_HIGH) {
 			climb_rate = expo3((cmd->Throttle - DEADBAND_HIGH) / (1.0f - DEADBAND_HIGH), altitude_hold_expo) *
```
mlyle: Argh. I've stared at code too much but this existing code looks totally wrong to me.

mlyle: Yah, amended.  This is fine.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1090,9 +1093,9 @@ static void altitude_hold_desired(ManualControlCommandData * cmd, bool flightMod
...
 		                         altitude_hold_maxclimbrate;
-		} else if (cmd->Throttle < DEADBAND_LOW && altitude_hold_maxclimbrate > MIN_CLIMB_RATE) {
+		} else if (cmd->Throttle < DEADBAND_LOW && altitude_hold_maxdescentrate > MIN_CLIMB_RATE) {
 			climb_rate = ((cmd->Throttle < 0) ? DEADBAND_LOW : DEADBAND_LOW - cmd->Throttle) / DEADBAND_LOW;
```
mlyle: So say we're at 30% deadband.  DEADBAND_LOW will (now) be .35.   If throttle is less than 0, we will descend at 35% of our maximum descent rate?

mlyle: I missed the parens.  I'm really getting dumb from reading code and gonna stop.

flight/Modules/ManualControl/transmitter_control.c
kubark42: I don't know. The [commit introducing it](https://github.com/cyphyworks/TauLabs/commit/3670c0697a4139cdeae64912499b74bc504db1b3) doesn't have any information about what it's trying to accomplish. There's no justification in the relevant code either, so the best we can do is make educated guesses. :P

My suspicion is that it's related to this comment in it's PR, #1119:
> This also has another implicit change. When using this there is no longer the song and dance to enter vario mode as there was previously. It seemed like not many people liked that. If you do not want to have a vario mode then climb rate can be set to zero and it will never enter that mode.

I think that "climb rate" refers to the UAVO field formerly known as "MaxRate".

flight/Modules/ManualControl/transmitter_control.c
kubark42: FWIW, I think this section could be cleaned up. I'm anti reusing variables, unless for things like loop iterators or clear performance gains.

flight/Modules/ManualControl/transmitter_control.c
mlyle: OK, multiplying by 0 should have the same effect so I'm in favor of killin' it.

kubark42: How about I open an issue to do some minor refactoring overall? It seems like we've both seen a few things we think could be better, but fixing all those things extends the scope of this PR beyond the bare minimum.

kubark42:
@mlyle Do you think this PR is ready for merge before the weekend? Or did you want me to do the refactoring beforehand?

mlyle:
OK I read this while sane.  It looks correct.

PR #1967, opened 2015-09-25
transmitter_control: Divide altitude hold deadband in half

mlyle:
Per discussion in #1964.  Now a deadband of 30% means 30% of stick
deflection, not 60%.


kubark42:
Looks good to me. I'm assuming that the compiler is smart enough to figure out that x/2*.01 = x*.005, so this approach incurs no additional runtime computation.

mlyle:
Yes, something like that will usually get const evaluated.  Sometimes without fast-math it won't, when the compiler can't prove it'll have the exact same results.

kubark42:
LGTM.

PR #1968, opened 2015-09-25
transmitter_control: Adjust deadbands to preserve range

mlyle:


flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1201,13 +1201,19 @@ bool validInputRange(int16_t min, int16_t max, uint16_t value, uint16_t offset)
...
-		if (*value > 0.0f)
+	} else {
+		if (*value > 0.0f) {
 			*value -= deadband;
-		else
+		} else {
 			*value += deadband;
+		}
+
+		*value /= 1 - deadband;
```
kubark42: I know technically it's not necessary, but please put ```()``` around the right side. I can just see this getting someone into copy-pasta trouble at a later date.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -1201,13 +1201,19 @@ bool validInputRange(int16_t min, int16_t max, uint16_t value, uint16_t offset)
  */
 static void applyDeadband(float *value, float deadband)
 {
-	if (fabsf(*value) < deadband)
+	if (deadband < 0.00001f) return;
```
kubark42: I think what you're doing here is choosing the smallest number such that we still have precision over a 32-bit float. Is that correct? If so, could the magic number become a ```#define``` instead?

kubark42:
@mlyle Looks good, thanks for doing this.

mlyle:
Yah, we should have a VERY_SMALL_FLOAT() check-- that isn't quite the smallest value in range, it should be a few bits bigger.  I can do that and scrub a lot of the existing uses in a subsequent pull..

flight/Modules/ManualControl/transmitter_control.c
kubark42: Also, we need to make sure that ```deadband``` is strictly ```>0``` in order to avoid any possible ```NaN``` contamination. If we don't want to do this check every cycle, this should be easy to do at line https://github.com/mlyle/TauLabs/blob/mpl-deadbandrange/flight/Modules/ManualControl/transmitter_control.c#L169.

mlyle: You mean <1?

The runtime checking is an interesting question in general.  THere's plenty of places where we don't do it.  And a lot of times it's not clear what's best to do.  And the strictly "correct" range doesn't really mean it's safe-- if someone sets a deadband of .99, they have almost as big of a problem as if they set it to 1.

kubark42: Yes, ```<1```, sorry! 

You raise excellent points. I agree that the strict range doesn't guarantee safety, although it guarantees numerical stability. I like to think we're pretty good about catching divide by 0 errors. It's such an important problem that there's [a hardware register for it](https://github.com/TauLabs/TauLabs/pull/488#discussion_r4101533).

If it's easiest, we could kick this can down the road and open an issue to decide what is the best value.

kubark42: @mlyle, are you okay to put the parentheses around the righthand side of L1215? And would you like to add a trap for division by zero, or push that off to another PR?

mlyle: Alright alright I'll stop slacking ;)

PR #1974, opened 2015-10-02
SysAlarmsMessaging plugin: Remove crash

kubark42:
Previously, GCS would crash when the SystemAlarms UAVO and the SysAlarmsMessaging plugin were out of sync. Obviously, it is undesirable for GCS to crash, esp. without an explanation of why.

Now GCS will not crash, but it throws an error at start up that it MUST be recompiled.

See also #873 and #933.

### DIFF
--- a/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
+++ b/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
@@ -33,6 +33,7 @@
 #include "systemalarms.h"
 #include <QtPlugin>
 #include <QMainWindow>
+#include <QErrorMessage>
 
 #include <QDebug>
 
@@ -49,7 +50,21 @@ void SysAlarmsMessagingPlugin::extensionsInitialized()
     ExtensionSystem::PluginManager *pm = ExtensionSystem::PluginManager::instance();
     UAVObjectManager *objManager = pm->getObject<UAVObjectManager>();
     SystemAlarms* obj = dynamic_cast<SystemAlarms*>(objManager->getObject(QString("SystemAlarms")));
-    connect(obj, SIGNAL(objectUpdated(UAVObject*)), this, SLOT(updateAlarms(UAVObject*)));
+
+    // Test for a condition which occurs when the GCS has not been recompiled from clean, but the SystemAlarms UAVO has changed.
+    // Either
+    //   1: Recompile from clean
+    //   2: Change this file (sysalarmsmessagingplugin.cpp) and recompile
+    if(obj->getObjID() != SystemAlarms::OBJID) {
+        QString staleSystemAlarmsWarning = QString(tr("The SystemAlarms UAVO has changed. GCS MUST be recompiled from `clean`."));
+        QErrorMessage err;
+        err.showMessage(staleSystemAlarmsWarning);
+        err.exec();
+
+        return;
+    } else {
+        connect(obj, SIGNAL(objectUpdated(UAVObject*)), this, SLOT(updateAlarms(UAVObject*)));
+    }
 
 }
 
@@ -79,6 +94,7 @@ bool SysAlarmsMessagingPlugin::initialize(const QStringList &arguments, QString 
     }
     TelemetryManager* telMngr = pm->getObject<TelemetryManager>();
     connect(telMngr, SIGNAL(disconnected()), this, SLOT(onAutopilotDisconnect()));
+
     return true;
 }
 
@@ -89,8 +105,6 @@ bool SysAlarmsMessagingPlugin::initialize(const QStringList &arguments, QString 
  */
 void SysAlarmsMessagingPlugin::updateAlarms(UAVObject* systemAlarm)
 {
-    Q_ASSERT(systemAlarm->getObjID() == SystemAlarms::OBJID);
-
     foreach (UAVObjectField *field, systemAlarm->getFields()) {
         for (uint i = 0; i < field->getNumElements(); ++i) {
             QString element = field->getElementNames()[i];


peabody124:
The code looks sensible but can you give steps to reproduce the crash? I change systemalarms.xml and recompile and don't see a crash. I have seen it in the past so know it is real though.

ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
```diff
@@ -89,8 +105,6 @@ bool SysAlarmsMessagingPlugin::initialize(const QStringList &arguments, QString
  */
 void SysAlarmsMessagingPlugin::updateAlarms(UAVObject* systemAlarm)
 {
-    Q_ASSERT(systemAlarm->getObjID() == SystemAlarms::OBJID);
```
peabody124: We should still likely return here, right? If this code is executed it is with a mismatched object so essentially all of the values are irrelevant?

kubark42: Good question. I made the change

```
   // Check that the UAVO library is in sync with the plugin. If not, return.
if (systemAlarm->getObjID() != SystemAlarms::OBJID) {
    return;
}
```

but I'm not actually sure if we should specifically handle that case here. The function will terminate normally without having done anything, which is what we expect of it. The problem of having UAVOs out of sync with plugins is universal, and we don't test in other plugins so I'm inclined to think this test would be out of place.

Still, the code is here and ready to go. I can merge it in at reviewer request.

kubark42:
> The code looks sensible but can you give steps to reproduce the crash? I change systemalarms.xml and recompile and don't see a crash. I have seen it in the past so know it is real though.

That was the subject of several questions in #933. To be honest, I don't recall the exact scenario for getting here, either. It happens a lot when improving alarms, though, which is why I got frustrated with the crashes and found this solution-- hopefully more elegant than #933.

kubark42:
Can this get merged?

mlyle:
I'm not really sure what it does or what the mismatch is :/

I would favor solutions that add the proper make dependencies rather than throw a runtime error.

peabody124:
@mlyle I'm not entirely sure how to reproduce it but the bug is legit. I can sketch out the problem:

- the SysAlarm plugin links at compile time to the alarms UAVO
- sometimes you can change the alarms UAVO and not trigger a recompile of the SysAlarms
- the Q_ASSERT then triggers an alarm because the runtime UAVO_ID does not match the compile time one

This isn't a bug that affects clean builds but only when you are jumping between branches.

I think the fix should be merged although agree the ultimate build problem should be resolved. This futzes on the problem of mixing static compile time object definitions and runtime introspection

mlyle:
I looked at the stuff that qmake spits out for the sysalarmsmessaging plugin makefile.  It looks like it correctly makes .obj/.../sysalarmsmessagingplugin.o depend upon                 /Users/mlyle/taulabs/build/uavobject-synthetics/gcs/systemalarms.h

It also seems like moc_systemalarms.cpp properly depends on systemalarms.h.

I wonder if this could relate to an upstream QT bug that is fixed.

PTDreamer:
Shouldn't merge

peabody124:
@PTDreamer You mean because it is a hack workaround for a compile system bug?

PTDreamer:
> @PTDreamer You mean because it is a hack workaround for a compile system bug?
>  Show all checks

I can only speak for what I have seen and know. I have never encountered the bug described here, not saying that it doesn't exist with a particular OS/QtVersion combo (both you and kenz are on OSX). 
So for me this will not fix anything but will bloat an already bloated codebase.
Also if this really fixes something it is something that only affects self built code since CI builds are made from clean. With this in mind a failed assertion is IMO better than a message box.

kubark42:
@PTDreamer, IIRC you said back in 2012 that this is due to the UAVO being a plugin and not a library. This isn't a fix or a workaround for that problem-- this PR turns a crash into a warning, and has no further ambitions.

My philosophy:

1. GCS should never crash, not unless the situation is absolutely unrecoverable and to continue is worse than to crash. As this is in a non-critical pathway, and the consequences of the plugin not working are marginally important at best, then we can categorically state that a crash is the wrong thing to do.
2. We  have efficient checks in place to catch this problem and we perfectly understand the solution. Therefore we have a responsibility to communicate that to the developer. Ambiguous stack dumps are an impediment to code progress.

I'm open to any solution that meets the above two criteria, so if you have a PR on the horizon that would be a better fix for the problem I'm happy to wait. In the meantime, it looks like most people feel this is a net positive to the codebase.

mlyle:
I don't see how it can happen with the present build dependencies if it occurs how I understand it (.h mismatch with linked .cpp mismatch).  I fiddled around trying to reproduce and couldn't.  I don't know if qmake has improved recently or I just don't understand the problem.  I can't recommend a merge without a repro scenario :/

PTDreamer:
> @PTDreamer, IIRC you said back in 2012 that this is due to the UAVO being a plugin and not a library. This isn't a fix or a workaround for that problem-- this PR turns a crash into a warning, and has no further ambitions.

I don't recall having said that, if I did I was an idiot which is a good thing since it implies I'm not anymore :)
A plugin is a C++ library which complies to a particular interface.
The problem with our code, and something I have been against (and in fact opposing to the merging of several PRs), is the use of UAVO classes static members which since they create their own copy on the module being linked (compile time) defeat the shared library purpose and are a source of hard to spot runtime errors.
The interfacing with UAVOs should always be done trough it's base classes (ie the ones that in fact define the interface). This off course bloats the code, makes it harder to read, makes it uglier, and requires better handling of non existing fields but it has the big advantage of flexibility and independence of compile time generated binaries. You can do things like changing the UAVO set during runtime by unloading the present UAVO lib and loading another.

Ex:
ManualControlCommand::GetInstance(objManagerPointer); Wrong, uses compile time defined UAVO ID across modules denying the the used of multi version UAVO shared library.
UAVObjectManagerPointer->getObject("ManualControlCommand"); Correct, returns UAVO regardless of its version
UAVObjectManagerPointer->getObject("ManualControlCommand")->getField("Roll) Correct uses only interface members which will always be present even if roll doesn't exist (requires checking of returned value).
ManualControCommand *obj = UAVObjectManagerPointer->getObject("ManualControlCommand");
obj->getRoll() Wrong, uses child class member which is not guaranteed to exist and can lead to a segfault.

kubark42:
I see two solutions:

1. If we can't reproduce it and it's suspected that it's an old build error, then let's reopen #933 and merge it to remove the crash code. 
2. If we can't reproduce it but we suspect that the build error is still out there, then let's merge this, since the principles of not crashing GCS apply.

Sound good?

----
@PTDreamer, thank you for the insight. As I recall, I think the conclusion was that an approach which leads to compile-time failure is very compelling. Ultimately we-- I think it was mostly @peabody124 and perhaps @solidgoldbomb?-- have opted for the version which prevents bugs which arise from string typos. Is there not a way to programatically fetch UAVO without having to use a string?

PTDreamer:
> @PTDreamer, thank you for the insight. As I recall, I think the conclusion was that an approach which leads to compile-time failure is very compelling. Ultimately we-- I think it was mostly @peabody124 and perhaps @solidgoldbomb?-- have opted for the version which prevents bugs which arise from string typos. Is there not a way to programatically fetch UAVO without having to use a string?

Wanting to integrate runtime checks on code pulled from a shared library on another module is a very strange thing to do.
Using the static defined string (on each UAVO file) instead of the string is a way to avoid typos and should not lead to any problems.

kubark42:
We'd need to upgrade the generator for this. We have ```ManualControlCommand::NAME```, so ```UAVObjectManagerPointer->getObject(ManualControlCommand::NAME);``` would work, but there doesn't seem to be any support for something like ```UAVObjectManagerPointer->getObject(ManualControlCommand::NAME)->getField(ManualControlCommand::Roll);```.

However, since the majority of our usage is ```ManualControlCommand::GetInstance(objManagerPointer); obj->getRoll()```, it would be easy to at least change our standards to make the first half ```UAVObjectManagerPointer->getObject(ManualControlCommand::NAME);``` and then continue to handle the possible segfault by checking that the returned obj is valid before using it.

peabody124:
Ok, I looked into this some more. At least once #1985 is merged, this bug should not occur so I don't think this fix is necessary (likely because 1985 cleaned up the build dependencies)

To test this, I deleted the build directory, did a build, then I modify systemalarms.xml and add a new field (which changes the UAVO ID). When I recompile, the system alarms plugin picks up the changes:

```
cd sysalarmsmessaging/ && ( test -e Makefile || TauLabs/tools/Qt5.5.0/5.5/clang_64/bin/qmake /Users/jcotton81/Programming/TauLabs/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessaging.pro -spec macx-g++ 'CONFIG+=debug silent' -o Makefile ) && /Applications/Xcode.app/Contents/Developer/usr/bin/make -f Makefile
make[4]: Entering directory `TauLabs/build/ground/gcs/src/plugins/sysalarmsmessaging'
compiling TauLabs/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
moc TauLabs/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.h
compiling .moc/debug-shared/moc_sysalarmsmessagingplugin.cpp
rm -f libSysAlarmsMessaging_debug.dylib
linking ../../../bin/Tau Labs GCS.app/Contents/Plugins/TauLabs/libSysAlarmsMessaging_debug.dylib
mv -f libSysAlarmsMessaging_debug.dylib ../../../bin/Tau\ Labs\ GCS.app/Contents/Plugins/TauLabs/
```

Running the GCS also doesn't trigger a crash.

[compile_log.txt](https://github.com/TauLabs/TauLabs/files/17947/compile_log.txt)

PTDreamer:
> We'd need to upgrade the generator for this. We have ManualControlCommand::NAME, so UAVObjectManagerPointer->getObject(ManualControlCommand::NAME); would work, but there doesn't seem to be any support for something like UAVObjectManagerPointer->getObject(ManualControlCommand::NAME)->getField(ManualControlCommand::Roll);.
> 
> However, since the majority of our usage is ManualControlCommand::GetInstance(objManagerPointer); obj->getRoll(), it would be easy to at least change our standards to make the first half UAVObjectManagerPointer->getObject(ManualControlCommand::NAME); and then continue to handle the possible segfault by checking that the returned obj is valid before using it.

Even that sort of approach using static values should be avoided since it only relates to the current set of UAVO being compiled.
Imagine the situation where you have a field named "timeBetweenCrashesSec"  which is then superseded by "timeBetweenCrashesMin". With the non static approach you can deal with both UAVO versions (***). 
    int scale;
    if(UAVObjectManagerPointer->getObject("crashHandlingObject")->getField("timeBetweenCrashesSec"))
      scale = 1;
    else if (UAVObjectManagerPointer->getObject("crashHandlingObject")->getField("timeBetweenCrashesMin"))
      scale = 60;
    else
      scale = .-1;

This would be impossible with statics since crashHandlingObject::timeBetweenCrashesSec would not exist after the refactoring.

*** off course this would only work if we would dynamically load the UAVO lib according to the version present on the hw the same way it is done on the android app.

peabody124:
Yea I think switching to only using dynamic access is something that would need a larger discussion. On the one hand, it could allow better compatibility with different sets of UAVO bundles (although would we really write the code to handle all these cases?).

The down side is that you now get runtime errors for typos instead of compilation errors. Personally, I much prefer the static access approach for reducing bugs.

mlyle:
> The down side is that you now get runtime errors for typos instead of compilation errors. Personally, I much prefer the static access approach for reducing bugs.

I agree with this.  I like that if I remove a field and there's somewhere in the GCS that uses it that I'm not aware of, I get a compilation failure.

Right now we don't use the plugins as independent plugins.  We should probably do one thing or another-- move towards the GCS plugins really being discrete plugins that are very loosely coupled, or collapse the GCS architecture to be flatter with a lot less runtime binding.

tracernz:
I think @PTDreamer's idea to allow the UAVOs to be compiled into a library that can be switched out at runtime has a lot of merit. The downside is of course that all the plugins would need to gracefully handle the non-existence of UAVOs or UAVO fields at runtime. Personally I think that is a good thing as it moves us away from the UAVO versions problem. The lack of compiler warnings for mistyped names can't really be easily solved if we want to get away from that version dependence I think, other than to print to the debug log at runtime when a field is missing.

kubark42:
I wasn't able to reproduce it anymore, although the reason I did this code back in March was because I was definitely having the problem back then. Is the general consensus that the build problems that led to this issue have been resolved?

peabody124:
My recollection of causing it was changing the systemalarms UAVO and not doing a clean build. From the build log / test I did above I believe it is now resolved (and hopefully catches the broader class of errors). Perhaps we can close this PR?

If people want to have a broader discussion about static/dynamic access and making GCS UAVO-verion insensitive I think that 100% warrants its own github issue for discussion. I know @PTDreamer referenced doing some work on swapping out the UAVO plugin itself.

kubark42:
> The lack of compiler warnings for mistyped names can't really be easily solved if we want to get away from that version dependence I think, other than to print to the debug log at runtime when a field is missing.

The two goals seem compatible: @mlyle, @peabody124, and I all want the build to fail when a variable name is changed. @PTDreamer and @tracernz would like to be able to swap plugins at runtime. Maybe the problem could be solved by having #defines instead of const chars, so that the text the compiler sees is a string, but the precompiler will fail if a variable is left unchanged after a naming refactor.

kubark42:
Closing this PR in favor of #933.

PR #1975, opened 2015-10-02
Alarms reset root cause

kubark42:
I implemented this back in March. So far, this hasn't had much value to me, but it might to others. I'm using it because I suspect that one day it will have incredible value, and I want the log data to reflect this.

Tested on F4 target. No guarantees on other targets, and might have unintended consequences such as your mother-in-law complaining to your children about how bad a parent you are, or barnacles growing on the bottom of your shoes.

### DIFF
--- a/flight/PiOS/STM32F10x/pios_reset.c
+++ b/flight/PiOS/STM32F10x/pios_reset.c
@@ -1,0 +1,48 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_RESET Reset functions
+ * @brief Hardware functions to deal with the reset register
+ * @{
+ *
+ * @file       pios_reset.c
+ * @author     Kenn Sebesta, Copyright (C) 2015.
+ * @brief      Reset information
+ *
+ ******************************************************************************/
+
+
+/* Project Includes */
+#include "pios.h"
+
+#include "pios_reset.h"
+
+void PIOS_RESET_Clear()
+{
+    RCC_ClearFlag();
+}
+
+int16_t PIOS_RESET_GetResetReason (void)
+{
+	uint8_t reboot_reason = PIOS_RESET_FLAG_UNDEFINED;
+	
+	// ****************************************************** //
+	// The order of these checks is important. DO NOT CHANGE. //
+	// ****************************************************** //
+	if(RCC_GetFlagStatus(RCC_FLAG_PORRST) == SET) {         // Check #1
+		reboot_reason = PIOS_RESET_FLAG_POWERON;
+	} else if(RCC_GetFlagStatus(RCC_FLAG_SFTRST) == SET) {  // Check #2
+		reboot_reason = PIOS_RESET_FLAG_SOFTWARE;
+	} else if(RCC_GetFlagStatus(RCC_FLAG_IWDGRST) == SET) { // Check #3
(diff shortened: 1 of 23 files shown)


mlyle:
The pios_reset.c file needs to get added to all million makefiles right?

kubark42:
> The pios_reset.c file needs to get added to all million makefiles right?

Yep. If people think this PR is a useful feature I'm happy to make that change.

mlyle:
I think it's useful.  Unfortunately it's probably not much use for the one thing I think of offhand-- we have people show up that say "I was flying and then my quad dropped out of the sky," and it's never clear whether it's a failsafe or FC reset.  If we had a history of the last 5 resets in flash that would be pretty cool.

Or maybe we need a way that certain things get logged even if logging is disabled so that logging doesn't need to be explicitly turned on to get something useful.

kubark42:
Glad you think it's useful! I'll fix it for all targets.

We could certainly have a small page of data (512 bytes?) reserved for this sort of essential info. I like the direction you're thinking.

kubark42:
Heh. Fixed. In my defense, when I wrote this the Naze32 target didn't exist yet. ;)

flight/PiOS/STM32F10x/pios_reset.c
```diff
@@ -0,0 +1,48 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_RESET Reset functions
+ * @brief Hardware functions to deal with the reset register
+ * @{
+ *
+ * @file       pios_reset.c
+ * @author     Kenn Sebesta, Copyright (C) 2015.
```
peabody124: what are your thoughts on updating this to also include taulabs like other files? not sure if there are implications of this are but generally we haven't used individual names.

flight/PiOS/STM32F4xx/pios_reset.c
```diff
@@ -0,0 +1,50 @@
...
+}
+
+int16_t PIOS_RESET_GetResetReason (void)
+{
+	uint8_t reboot_reason = PIOS_RESET_FLAG_UNDEFINED;
+	
+	// ****************************************************** //
+	// The order of these checks is important. DO NOT CHANGE. //
+	// ****************************************************** //
+	if(RCC_GetFlagStatus(RCC_FLAG_PORRST) == SET) {         // Check #1
```
mluessi: A switch statement would be more elegant (and efficient)

flight/PiOS/STM32F10x/pios_reset.c
kubark42: All these need to be updated. They're written with internal copyrights so that the git history makes it clear that code has been deliberately open-sourced. Thanks for catching this, I'll try to do better in the future.

flight/PiOS/STM32F4xx/pios_reset.c
kubark42: How would that work in the ```switch()..case```? ```RCC_GetFlagStatus``` has to be called with an argument so I will have to iterate over each possible reset reason.

I could write my own function to grab the RCC flag directly, but that's code duplication and targets the specific processor instead of using the CMSIS abstraction, right?

mluessi: True.. I probably didn't have enough coffee before looking at your code ;)

PR #1976, opened 2015-10-02
Alarms: Determine which stack overflowed

kubark42:
The board will raise an error when a module's stack overflows. This writes the module's name to UAVO so the faulty modules can be easily diagnosed.

It has been amazingly useful to me, even though I know there are other ways to determine which stack overflowed. It's just so much easier to read the module name than to have to guess which module at 0 bytes in the TaskInfo UAVO is the problem child.

However, due to the increased UAVO packet size, I think there's a reasonable argument not to merge this.

Comments?

### DIFF
--- a/flight/Modules/System/systemmod.c
+++ b/flight/Modules/System/systemmod.c
@@ -80,6 +80,7 @@ static uint32_t idleCounterClear;
 static struct pios_thread *systemTaskHandle;
 static struct pios_queue *objectPersistenceQueue;
 static bool stackOverflow;
+static uint8_t stackOverflowedName[SYSTEMALARMS_STACKOVERFLOWCODE_NUMELEM];
 
 // Private functions
 static void objectUpdatedCb(UAVObjEvent * ev);
@@ -537,6 +538,7 @@ static void updateSystemAlarms()
 	// Check for stack overflow
 	if (stackOverflow) {
 		AlarmsSet(SYSTEMALARMS_ALARM_STACKOVERFLOW, SYSTEMALARMS_ALARM_CRITICAL);
+		SystemAlarmsStackOverflowCodeSet(stackOverflowedName);
 	} else {
 		AlarmsClear(SYSTEMALARMS_ALARM_STACKOVERFLOW);
 	}
@@ -609,6 +611,9 @@ void vApplicationIdleHook(void)
 void vApplicationStackOverflowHook(uintptr_t pxTask, signed char * pcTaskName)
 {
 	stackOverflow = true;
+	for (int i=0; i< SYSTEMALARMS_STACKOVERFLOWCODE_NUMELEM; i++) {
+		stackOverflowedName[i] = pcTaskName[i];
+	}
 #if DEBUG_STACK_OVERFLOW
 	static volatile bool wait_here = true;
 	while(wait_here);
--- a/shared/uavobjectdefinition/systemalarms.xml
+++ b/shared/uavobjectdefinition/systemalarms.xml
@@ -68,6 +68,9 @@
 			</options>
 		</field>
 
+		<field name="StackOverflowCode" units="" type="uint8" elements="8"/>
+
+
 		<access gcs="readwrite" flight="readwrite"/>
 		<telemetrygcs acked="true" updatemode="onchange" period="0"/>
 		<telemetryflight acked="true" updatemode="onchange" period="0"/>


mlyle:
So I've not read any of this code, so I'm just making things up... and I know this is pre-existing code I'm complaining about.  But we don't have an MMU, and our stacks and heap are jumbled.  Can't we only detect overflows if we run over some canary at the end, which means we've almost certainly scribbled on other objects and are going to do undefined things from now on? 

Now presumably the canary should not end up back in place, but it seems like to me the code to clear the alarm shouldn't exist there. 

Also, in the long run it would be nice to be able to report free form errors via a generalized mechanism, and not have to spend bytes on buffers for all the different places (config errors, this, etc) where we might want more detailed talkback than just an enum.  (or we come up with a better way of defining the enums).

kubark42:
```stackOverflow``` is set to ```false``` on boot, and then has no mechanism to return to ```false```. So the canary stays put.

However, it does indeed seem inviting that there is a way to clear the error. It is necessary to at some point change the ```SystemAlarms.StackOverflow``` field from ```Uninitialized``` to ```OK```. But this could be done before going into the main ```while(1){}``` loop as well.

Agreed about the free-form errors. It might be nice to have a generalized string UAVO which simply passes back messages that don't warrant escalation to a full UAVO. On the other hand, it seems somewhat a shame to be optimizing for several bytes in 2015. This system was designed over 5 years ago, when memory, CPU, and bandwidth were all much harder to come by.

**Comment by mlyle** [comment]
SRAM's still expensive and we're not getting tons more of it ever.  And putting string table on an embedded device with no internal UI is kinda strange, too.  So the old tradeoffs still kinda make sense.

**Comment by kubark42** [comment]
Agreed that it's kinda strange, and we can do better. It might even make sense for the ```PIOS_Thread_Create()``` function to use an enum table, such as the list of tasks in the TaskInfo object. Then this "ZOMG, stack overflow!!!" field could likewise be an enum, instead of a string.

Regarding SRAM, Atmega's S70 seems to be a much more capable chip, at a similar price point to the STM32F4s. http://www.atmel.com/products/microcontrollers/arm/sam-s.aspx. Love the doubling of the SRAM!

File: flight/Modules/System/systemmod.c [hunk 1]
```diff
@@ -609,6 +611,9 @@ void vApplicationIdleHook(void)
 void vApplicationStackOverflowHook(uintptr_t pxTask, signed char * pcTaskName)
 {
 	stackOverflow = true;
+	for (int i=0; i< 8; i++) {
```
peabody124: for safety I'd like to see NELEMENTS used here instead of hardcoded 8.

flight/Modules/System/systemmod.c
```diff
@@ -80,6 +80,7 @@ static uint32_t idleCounterClear;
 static struct pios_thread *systemTaskHandle;
 static struct pios_queue *objectPersistenceQueue;
 static bool stackOverflow;
+static uint8_t stackOverflowedName[8];
```
peabody124: similarly can we use the constant that is SYSTEMALARMS_STACKOVERFLOWCODE_SIZE instead of 8

kubark42:
I'm doubting the value of this, the stack overflow hook only works on FreeRTOS.

flight/Modules/System/systemmod.c
kubark42: Both fixed via rebase.

tracernz:
Stack checking is done in Chibi by enabling [CH_DBG_ENABLE_STACK_CHECK](http://chibios.sourceforge.net/html/group__config.html#gab93d9ee904f15d4f2c26ef2a1394a1d7) which does the right thing and halts the system in a tight loop when stack overflow is detected (since you can't really continue at this point, there is an excellent chance that at the very least the state struct of the next task has been overrun).

kubark42:
@tracernz So this PR wouldn't have any long-term value, since if the stack overflow halts the system.

FWIW, I don't like that behavior. A halted system is a lot harder to diagnose in this context than a flaky system. There's no guarantee that things will behave well when the stack overflows, but it's guaranteed it won't when it enters a death spiral because the FC locks up. :p

mlyle:
So I disagree.  I think we should stop what we're doing and wedge or reset (fixed wing you'd have a chance to save things after a reset, maybe).

Undefined behavior scares me.  It could potentially do anything, like wedge actuator or get manualcontrol stuck without failsafe depending on what it runs into on what other stack.

Maybe we should save a little bit of flash for a crash message so we can log what goes wrong when an runtime exception like stack overflow bites us.

mlyle:
(it's difficult to even reason about what it will do when stuff is scribbled on, because it can choose to take paths and do things that aren't even in the code, because of optimizer magic and reaching states that "can never happen".  So, like take neither branch of an if statement, etc)

kubark42:
> So I disagree. I think we should stop what we're doing and wedge or reset (fixed wing you'd have a chance to save things after a reset, maybe).
>
> Undefined behavior scares me. It could potentially do anything, like wedge actuator or get manualcontrol stuck without failsafe depending on what it runs into on what other stack.

I think we agree here. Rebooting is fine, you have a chance to recover. What I don't like is the system halting. A halted FC leads to undefined drone behavior, as in "I can't define which part of the ground the drone is going to hit." ;)

kubark42:
P.S. If this only can target FreeRTOS, then I'm still not seeing the value of this PR since moving forward the RTOS of choice will be ChibiOS. 

In summary, does anyone think it should merge?

tracernz:
Agreed that this PR isn't as useful anymore. I like the idea of saving a little flash to write some crash data. I also like the idea of attempting to soft reset rather than just locking in a tight loop. I wonder if we actually have stack checking enabled for Chibi though, might need to enable it too.

peabody124:
Should we close this then? Seems like without ChiBios it won't apply to many targets going forward.

kubark42:
I'm in favor of closing it, but I wanted to get consensus since @mlyle had initially thought it would be useful.

dustin:
I filed #2044 last night as I had what appears to be a FC reset that cost me a frame and a motor (and some props and some nylon bolts and a few hours as I transplant the working bits).

If something can provide a clue as to what goes wrong in these situations, it seems worth all of that.

PR #1979, opened 2015-10-04
AndroidGCS: Support for TauLinkModule

peabody124:
This adds BT LE support to AndroidGCS which allows usings things like the HM-10.


peabody124:
I'd still like to walk down the minimum SDK version if possible.

peabody124:
@lilvinz if you have a chance can you run ```make android_sdk_update``` on this branch

lilvinz:
jenkins: test this please

peabody124:
Thanks @lilvinz !!

@PTDreamer or @kubark42 if either of you have a chance, it would be great if you can glance at this. It adds support for BT LE to the android code. It fails gracefully so doesn't bump the version API required.

androidgcs/google-play-services_lib/project.properties
```diff
@@ -11,5 +11,5 @@
 #proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
 
 # Project target.
-target=Google Inc.:Google APIs:14
+target=Google Inc.:Google APIs:19
```
PTDreamer: Shouldn't this match minSdkVersion?

peabody124: No, it is actually intentional. You have to compile with a more recent apI to have the BTLE library calls available. However when running on hardware that has an API between 14 and 18, then it will detect this at run time and throw a warning and not make the calls.

Issue #1804, opened 2015-08-18
HAL Refactor/#1759 remnant: Python support for inherited UAVO


mlyle:

tracernz:
Fixed in #1981?

mlyle:
Yup!

PR #1981, opened 2015-10-08
Python API improvements

mluessi:
Several improvements to the Python API:
- It is now possible to directly instantiate UAVOs in Python with the correct default values from the xml file. E.g.: `sparky = uavo.UAVO_HwSparky2()`
- ENUMs are included in the Python UAVOs (including correct parsing of shared definitions) and are e.g. used to produce more informative representations of the objects . E.g.:

```
    print sparky
    UAVO_HwSparky2(name=UAVO_HwSparky2, time=0, uavo_id=609948424, CoordID=0,
    RcvrPort=Disabled(0), MainPort=Disabled(0), FlexiPort=Disabled(0), USB_HIDPort=USBTelemetry(0), 
    USB_VCPPort=Disabled(4), DSMxMode=Autodetect(0), Radio=Disabled(0), MaxRfSpeed=64000(3), 
    MaxRfPower=125(1), MinChannel=0, MaxChannel=250, GyroRange=2000(3),  AccelRange=8G(2),
    MPU9250Rate=500(3), MPU9250GyroLPF=184(0), MPU9250AccelLPF=184(1), VTX_Ch=1(0),
    Magnetometer=Internal(0), ExtMagOrientation=Top0degCW(0))
```
- Python scripts using the githash to get the UAVO definitions can now be run from any directory (previously, one had to be in the TL root directory)
- Fixed a bug that sometimes caused the TCP telemetry to fail due to a "Resource temporarily unavailable" error (not sure why this error occurred, it shouldn't be possible due to the use of `select`).

### DIFF
--- a/python/taulabs/telemetry.py
+++ b/python/taulabs/telemetry.py
@@ -30,8 +30,8 @@ class TelemetryBase():
     def __init__(self, githash=None, service_in_iter=True,
             iter_blocks=True, use_walltime=True, do_handshaking=False,
             gcs_timestamps=False, name=None):
+
         """Instantiates a telemetry instance.  Called only by derived classes.
-        
          - githash: revision control id of the UAVO's used to communicate.
              if unspecified, we use the version in this source tree.
          - service_in_iter: whether the actual iterator should service the
@@ -82,7 +82,7 @@ class TelemetryBase():
 
     def as_numpy_array(self, match_class, filter_cond=None):
         """ Transforms all received instances of a given object to a numpy array.
-        
+
         match_class: the UAVO_* class you'd like to match.
         """
 
@@ -90,7 +90,7 @@ class TelemetryBase():
 
         # Find the subset of this list that is of the requested class
         filtered_list = filter(lambda x: isinstance(x, match_class), self)
- 
+
         # Perform any additional requested filtering
         if filter_cond is not None:
             filtered_list = filter(filter_cond, filtered_list)
@@ -98,7 +98,7 @@ class TelemetryBase():
         # Check for an empty list
         if filtered_list == []:
             return np.array([])
- 
+
         return np.array(filtered_list, dtype=match_class._dtype)
 
     def __iter__(self):
@@ -262,17 +262,17 @@ class TelemetryBase():
 class FDTelemetry(TelemetryBase):
     """
     Implementation of bidirectional telemetry from a file descriptor.
-    
+
     Intended for serial and network streams.
     """
 
     def __init__(self, fd, *args, **kwargs):
         """ Instantiates a telemetry instance on a given fd.
-        
+
         Probably should only be called by derived classes.
-        
+
          - fd: the file descriptor to perform telemetry operations upon
-        
+
         Meaningful parameters passed up to TelemetryBase include: githash,
         service_in_iter, iter_blocks, use_walltime
         """
@@ -327,13 +327,19 @@ class FDTelemetry(TelemetryBase):
         if r:
             # Shouldn't throw an exception-- they just told us
             # it was ready for read.
-            chunk = os.read(self.fd, 1024)
-            if chunk == '':
-                raise RuntimeError("stream closed")
+            try:
+                chunk = os.read(self.fd, 1024)
+                if chunk == '':
+                    raise RuntimeError("stream closed")
 
-            self.recv_buf = self.recv_buf + chunk
+                self.recv_buf = self.recv_buf + chunk
 
-            didStuff = True
+                didStuff = True
+            except OSError as err:
+                # For some reason, we sometimes get a
+                # "Resource temporarily unavailable" error
+                if err.errno != 11:
+                    raise err
 
         if w:
             written = os.write(self.fd, self.send_buf)
@@ -356,7 +362,7 @@ class NetworkTelemetry(FDTelemetry):
     """ TCP telemetry interface. """
     def __init__(self, host="127.0.0.1", port=9000, *args, **kwargs):
         """ Creates a telemetry instance talking over TCP.
-        
+
          - host: hostname to connect to (default localhost)
          - port: port number to communicate on (default 9000)
 
@@ -379,13 +385,13 @@ class SerialTelemetry(FDTelemetry):
     """ Serial telemetry interface """
     def __init__(self, port, speed=115200, *args, **kwargs):
         """ Creates telemetry instance talking over (real or virtual) serial port.
-        
+
          - port: Serial port path
          - speed: Baud rate (doesn't really matter for VCP, defaults 115200)
 
         Meaningful parameters passed up to TelemetryBase include: githash,
         service_in_iter, iter_blocks, use_walltime
-        
+
         Requires the module pyserial to provide OS independance.
         """
 
@@ -401,7 +407,7 @@ class FileTelemetry(TelemetryBase):
     def __init__(self, file_obj, parse_header=False,
              *args, **kwargs):
         """ Instantiates a telemetry instance reading from a file.
-        
+
          - file_obj: the file object to read from
          - parse_header: whether to read a header like the GCS writes from the
            file.
--- a/python/taulabs/telemetry.py
+++ b/python/taulabs/telemetry.py
@@ -327,6 +327,7 @@ class FDTelemetry(TelemetryBase):
         if r:
             # Shouldn't throw an exception-- they just told us
             # it was ready for read.
+            # TODO: Figure out why read sometimes fails when using sockets
             try:
                 chunk = os.read(self.fd, 1024)
                 if chunk == '':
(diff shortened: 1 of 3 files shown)


peabody124:
Oh awesome. I'm super stoked about enums. That will make code so much clearer and it much easier to see things like flight mode

peabody124:
Awesome. Im very stoked about enums

python/taulabs/telemetry.py
```diff
@@ -323,13 +323,19 @@ def _do_io(self, finish_time):
         if r:
             # Shouldn't throw an exception-- they just told us
             # it was ready for read.
-            chunk = os.read(self.fd, 1024)
-            if chunk == '':
-                raise RuntimeError("stream closed")
+            try:
```
mlyle: I'm a little concerned about this-- worries me it's masking a bug with when things are ready for reading.  EAGAIN after they told us it's ready to read?  How odd.  Maybe I am calling this at inappropriate times.

python/taulabs/uavo.py
```diff
@@ -80,6 +85,29 @@ def from_bytes(cls, data, timestamp, instance_id, offset=0):
 
         return cls._make(field_values)
 
+    def __repr__(self):
+        """ String representation of the contents """
+        rep = self.__class__.__name__ + '('
+        for field in self._fields:
+            field_value = getattr(self, field)
+            # Show ENUMs in a fancy way (string and value)
+            if hasattr(self, 'ENUM_' + field):
```
mlyle: I think I'd prefer having an array of enums based on field index than all of this introspection and string manipulation.  But not a dealbreaker at all.

python/taulabs/uavo.py
```diff
@@ -183,18 +212,64 @@ def make_class(xml_file):
...
             info['type_val'] = type_enum_map[info['type']]
+
+            # Get parent
+            if info['parent'] is not None:
+                parent_name = info['parent'].split('.')
+                field_name = RE_SPECIAL_CHARS.sub('', parent_name[1])
+                try:
+                    parent_class = globals()['UAVO_' + parent_name[0]]
+                    parent_options = getattr(parent_class, 'ENUM_' + field_name)
+                    if len(info['options']) == 0:
```
mlyle: I'm really glad you handled this case.  The only additional thing needed is to adjust the hash to handle inheritance / sparse options.

python/taulabs/uavo.py
```diff
@@ -269,10 +344,25 @@ class tmpClass(namedtuple(name, tuple_fields), UAVTupleClass):
     # child classes don't get a dict / keep all their namedtuple goodness
     tuple_class = type(name, (tmpClass,), { "__slots__" : () })
 
+    # Set the default values for the constructor
+    defaults = [name, 0, uavo_id]
+    defaults.extend(f['defaultvalue'] for f in fields)
+    tuple_class.__new__.__defaults__ = tuple(defaults)
```
mlyle: Super slick.  I've wanted to do this.  Thank you.

python/taulabs/telemetry.py
BrainFPV: Yeah.. I have no idea how this can occur. It may be a bug in Python or something.

python/taulabs/uavo.py
mluessi: The ENUMs are also useful for setting values. E.g., one can use `UAVO_HwSparky2.ENUM_FlexiPort.Disabled` to get a specific value when manipulating UAVOs in code.

python/taulabs/uavo.py
mluessi: So the hash in Python is currently broken? This was not caused by this PR, correct? What would need to be done to fix it?

python/taulabs/uavo.py
mluessi: Some serious Python magic :)

python/taulabs/telemetry.py
mluessi: PS: I never saw it when parsing log files, only when connecting to the POSIX simulation using TCP

python/taulabs/uavo.py
mlyle: The hash for whenever there's a sparse enum includes the next index before there's a gap.  So it only affects the hw* objects.

```
        if(info->fields[n]->type == FIELDTYPE_ENUM) {
            QStringList options = info->fields[n]->options;
            int nextIdx = 0;

            for (int m = 0; m < options.length(); m++) {
                int idx = findOptionIndex(info->fields[n], m);

                if (idx < 0) abort();

                // Not contiguous options.  Update with next value.
                if (idx != nextIdx) {
                    hash = updateHash((quint32) idx, hash);
                }

                nextIdx = idx+1;

                hash = updateHash(options[m], hash);
            }
```

is how it works on the C side.  now that you understand enums it's possible to fix.   It's issue #1804

python/taulabs/telemetry.py
peabody124: Oh that's interesting. Was it when the simulator crashed?

python/taulabs/uavo.py
mlyle: Sure, I'm not saying they should go away-- they should definitely be accessible symbolically.

mluessi: Just to make sure, you are suggesting to add a list / tuple with the same length as `_fields` that also contains the enums (and `None` for things that are not enums)? Performance-wise, I'm not sure if this is needed, as the `__repr__` method is only used when printing objects and AFAIK `hasattr` is quite fast.

python/taulabs/uavo.py
mluessi: Thanks, I will look into this and add it.

mluessi:
Fixed the hash calculation. I had introduced some more bugs by "forgetting" that dicts are not ordered and by using a regex to modify the enum options (so they could be used as attributes in code). The enums and the reverse mappings are now just dicts and can be accessed like this:

    print uavo.UAVO_HwSparky2.ENUM_Magnetometer['Internal']
    0
    print uavo.UAVO_HwSparky2.ENUMR_Magnetometer[2]
    ExternalAuxI2C

Fixes #1804

python/taulabs/uavo.py
```diff
@@ -269,10 +346,25 @@ class tmpClass(namedtuple(name, tuple_fields), UAVTupleClass):
...
+
+    # Add enums
+    for field in fields:
+        if field['type'] == 'enum':
+            enum = field['options']
+            mapping = dict((key, value) for key, value in enum.iteritems())
+            reverse_mapping = dict((value, key) for key, value in enum.iteritems())
+            setattr(tuple_class, 'ENUM_' + field['name'], mapping)
+            setattr(tuple_class, 'ENUMR_' + field['name'], reverse_mapping)
```
mluessi: Instead of using a two dicts, we could just use one dict that contains the mapping from integer to string and vice versa. May be easier to use. I.e., `ENUM_FlightMode[0]` would return a string while `ENUM_FlightMode['Manual']` would return an int. Thoughts?

mlyle: I think whaty ou've got is great.

python/taulabs/telemetry.py
mlyle: I know we don't like TODO comments but I hate the idea of blocking this PR / a useful fix on this anomaly.  Maybe put TODO: in front of the above comment and call it good.

python/taulabs/uavo.py
mlyle: It's probably not needed.  I just don't really like introspection when it's not completely necessary.  Anyways, not a dealbreaker at all.

python/taulabs/telemetry.py
mluessi: @peabody124 no, the simulator works normal and successive read calls are successful. I can add a note to the code.

mluessi:
Added a comment about the read failure.

mluessi:
Is there anything else that needs to be done? If not, this should be good to merge.

kubark42:
@mluessi The branch has merge conflicts.

mluessi:
@kubark42 rebased

python/taulabs/telemetry.py
```diff
@@ -327,13 +327,20 @@ def _do_io(self, finish_time):
...
-            if chunk == '':
-                raise RuntimeError("stream closed")
+            # TODO: Figure out why read sometimes fails when using sockets
+            try:
+                chunk = os.read(self.fd, 1024)
+                if chunk == '':
+                    raise RuntimeError("stream closed")
 
-            self.recv_buf = self.recv_buf + chunk
+                self.recv_buf = self.recv_buf + chunk
```
kubark42: Ugh, python. This is exactly why I hate semantic white-spacing. :p

mluessi: ? this is correct (needs be indented)

kubark42: Yah, realized it's correct. But it's annoyingly hard to disambiguate between accidental IDE style changes and intentional scope changes. No matter, just ranting.

kubark42:
LGTM, and everyone seems to like it. Win!

PR #1982, opened 2015-10-09
Convert Sparky1 to PIOS HAL

fujin:
* I originally opened this against the limbo fork d-Ronin (https://github.com/d-ronin/dronin/pull/19) as a precursor to my practice evaluation with the MOTOLAB
  closed hardware target (tornado f3 / motofc) but figured I might as
  well try my luck here :ocean:
* Converts the sparky1 target board-info / pios_board  to pios_hal per sparky2
  and others..

### DIFF
--- a/flight/targets/sparky/fw/pios_board.c
+++ b/flight/targets/sparky/fw/pios_board.c
@@ -38,6 +38,7 @@
 #include "board_hw_defs.c"
 
 #include <pios.h>
+#include <pios_hal.h>
 #include <openpilot.h>
 #include <uavobjectsinit.h>
 #include "hwsparky.h"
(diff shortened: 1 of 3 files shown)


peabody124:
Awesome fujin, thanks a ton for this. I'm gonna have to pop blades back onto SeeingSpark!

flight/targets/sparky/fw/pios_board.c
```diff
@@ -167,12 +168,6 @@ static const struct pios_hmc5883_cfg pios_hmc5883_external_cfg = {
 };
 #endif /* PIOS_INCLUDE_HMC5883 */
 
-/* One slot per selectable receiver group.
- *  eg. PWM, PPM, GCS, SPEKTRUM1, SPEKTRUM2, SBUS
- * NOTE: No slot in this map for NONE.
- */
-uintptr_t pios_rcvr_group_map[MANUALCONTROLSETTINGS_CHANNELGROUPS_NONE];
-
 #define PIOS_COM_TELEM_RF_RX_BUF_LEN 512
```
peabody124: so a lot of these constants can be deleted, actually. See https://github.com/TauLabs/TauLabs/pull/1759/files#diff-fddc7ccc67186b59e95c2e594ed8468b for an example on Sparky2.

flight/targets/sparky/fw/pios_board.c
```diff
@@ -319,22 +218,7 @@ static void PIOS_Board_configure_hsum(const struct pios_usart_cfg *pios_usart_hs
...
-			PIOS_DELAY_WaitmS(200);
-		}
-		PIOS_WDG_Clear();
-		PIOS_DELAY_WaitmS(200);
-		PIOS_WDG_Clear();
-		PIOS_DELAY_WaitmS(200);
-		PIOS_WDG_Clear();
-		PIOS_DELAY_WaitmS(100);
-	}
+		PIOS_HAL_Panic(PIOS_LED_ALARM, code);
```
peabody124: I would have replaced the actual calls, but this is consistent with how it was done elsewhere.

peabody124:
So first pass through except for the constants which can be dropped this looks pretty good. I'll try and test it in the next few days.

flight/targets/sparky/fw/pios_board.c
fujin: Cool will do

flight/targets/sparky/fw/pios_board.c
fujin: I can do that

fujin: I didn't end up doing this. If you feel strongly about it I can update. I would prefer to keep the implementation consistent with the other ones so far, or update them all to be consistent. WDYT?

fujin:
@peabody124 I ran the whole file through astyle on a separate commit. Gave up fighting with editor configuration to try to match the loosely defined syntax variation of Linux style used here.

peabody124:
Oh for things like astyle it is really bad practice to combine into a PR with functional changes. If you look in the editor now you can't tell which changes to review.

Generally whenever we do that I recommend a standalone PR for formatting, and include the exact switches you used for astyle (I've used gindent in the past and if you grep for that in the code you'll see my switches). That way people with branches can run the same switch and avoid a lot of hell with merging.

flight/targets/sparky/fw/pios_board.c
peabody124: Totally ok with me to keep panic xonsistent. That was more a style comment.

fujin:
@peabody124 there are two separate commits on the PR (functional / syntax) and I just used the `make astyle_flight` command in the main Makefile, with astyle installed from `make astyle_install`. If you prefer me to split the syntax out into a separate PR that is probably not going to work because the first part only changes a subset of the code (and is subsequently unformatted) and then the following commit reformats according to the guidelines in the Makefile.

fujin:
@peabody124 we also looked at uncrustify in d-Ronin (see https://github.com/d-ronin/dronin/pull/22) and I submitted a change there for an uncrustified-pios_board.c for Sparky1, if you prefer that one, can include here: https://github.com/fujin/dronin/commit/36ba62847bcb460b5a88d39b0084dedc57691e09

It looks more consistent with the rest of the codebase.

kubark42:
Not that it's particularly relevant to this code review, but I'd love to see a decrustify across our codebase.

tracernz:
@kubark42 We actually discussed that on IRC a couple of days ago while @fujin was working on this. I personally think we should consider doing a one-off reformat of all flight code using a properly setup astyle command, and then add the same command to the Jenkins build so it will reject future PRs which don't conform. @trollcop has been doing this for some time with multiwii/baseflight and it seems to work reasonably well. The problem with a big cleanup though is you lose some of the power of `git blame` (and related github view) when real changes are hidden by reformatting commits.

kubark42:
@peabody124 Do you think this is ready for merge? I don't feel like it's my place to merge in changes specific to your boards.

peabody124:
Ok just tested this on hardware. USB port selection works as receiver port works. Code looks good. Merging.

PR #1984, opened 2015-10-12
Remove make gui

mlyle:
Seems unnecessary/excessive.

Removed in d-ronin/dronin#11


kubark42:
Removes functionality added in https://github.com/TauLabs/TauLabs/pull/348.

tracernz:
IIRC this was broken and we discussed it on IRC and decided we didn't want to maintain it at dRonin. I'm happy for this to be reverted if somebody at Tau wants it and will maintain it. ;)

kubark42:
Totally fine! I just wanted the breadcrumb trail for documentation.

PS. Thanks for pointing out it was broken. That's a helpful tidbit that was missing from the PR discussion.

PR #1985, opened 2015-10-12
uavo: Build into libraries instead of per-target

mlyle:
Builds on top of the changes for #1895.

d-ronin/dronin#10


mlyle:
This saves build time and fixes problems with Windows command lines being too long (and provides room for further improvement there should we ever need to further shrink paths.)

Makefile
```diff
@@ -821,6 +813,56 @@ ifneq ($(word 2,$(MAKECMDGOALS)),)
...
+		$(MAKE) -r --no-print-directory \
+		TCHAIN_PREFIX="$(ARM_SDK_PREFIX)" \
+		REMOVE_CMD="$(RM)" \
+		\
+		MAKE_INC_DIR=$(MAKE_INC_DIR) \
+		ROOT_DIR=$(ROOT_DIR) \
+		TARGET=$(TARGET) \
+		OUTDIR=$(OUTDIR) \
+		\
+		PIOS=$(PIOS) \
```
solidgoldbomb: Are all of these variables needed for the uavobjects targets?  Seems like at least the modules dir could be left out.  Maybe others.

ground/uavobjgenerator/uavobjectparser.cpp
```diff
@@ -169,6 +177,24 @@ void UAVObjectParser::calculateAllIds()
     foreach (ObjectInfo *item, objInfo) {
         calculateID(item);
     }
+
+    /* Sort the object info list by ID, now that they're all defined */
+    std::sort(objInfo.begin(), objInfo.end(), [](ObjectInfo *o1, ObjectInfo *o2) {
+            return o2->id < o1->id;
+            });
+
+    /* This is a 64 bit hash to have a bit more protection against collisions
```
solidgoldbomb: How is this new hash being used by the flight/gcs software?

Makefile
```diff
@@ -9,7 +9,7 @@ include $(CURDIR)/make/system-id.mk
 
 # configure some directories that are relative to wherever ROOT_DIR is located
 TOOLS_DIR := $(ROOT_DIR)/tools
-BUILD_DIR := $(ROOT_DIR)/build
+export BUILD_DIR := $(ROOT_DIR)/build
```
solidgoldbomb: My preference would be to keep the root BUILD_DIR as non-exported and to export a more specific directory.  That keeps the management of the top of the build dir in this file while allowing sub make files to play under the dirs that are defined by the top level makefile.

ground/uavobjgenerator/uavobjectparser.cpp
mlyle: It's not yet.  The only purpose right now is to cause a file which is marked as a dependency to change if the actual contents have changed.  But the intention is to perhaps replace the existing sha1 hash with a hash generated here (perhaps a SHA-1, perhaps something more like this), so that A) we're no longer dependent on formatting/versions when calculating the uav hash, and B) we know the hash earlier in the build process and some of the weirder glue in the makefile to handle uavo hashes can be removed.

solidgoldbomb:
Finished reviewing.  Running a few build tests and looking over the generated code.

Makefile
mlyle: Yup, all the makefiles everywhere can get cut down.  This was just borrowed from other targets.

I really want to move these to exports to remove all of this boilerplate.  That way when we change how things work we don't need to plumb in / unplumb variables

solidgoldbomb:
Overall, looks good to me other than the things I mentioned in comments.

Would be good to know what targets have been tested.  Specifically:
* Has this been flight tested on both hard/soft float boards?
* Has this been tested on android?  Seems to build OK but would be good to test there too.

Also looks like you'll need to rebase (or merge from next) before this can be merged.

flight/PiOS/STM32F30x/library.mk
```diff
@@ -12,6 +12,7 @@ CDEFS				+= -DSTM32F30X
 CDEFS				+= -DHSE_VALUE=$(OSCILLATOR_FREQ)
 CDEFS 				+= -DUSE_STDPERIPH_DRIVER
 ARCHFLAGS			+= -mcpu=cortex-m4 -march=armv7e-m -mfpu=fpv4-sp-d16 -mfloat-abi=hard
+FLOATABI			= hard
```
kubark42: What is this doing? Is that relevant to changing the UAVOs? The git commit message doesn't mention this, but it looks like it's specifying to use the hardware FPU?

Makefile
solidgoldbomb: This is really only boilerplate if it gets blindly copy/pasted without being appropriately limited.  The set of paths passed into the submakefiles was originally intended to form a basic API to describe where the submakefiles should be looking/writing.  Some attempt was made to pass in only those paths that made sense for each build type.

Deleting these explicit parameters and exposing the root paths as global variables (via export) would throw away any scoping for submakefiles.  It sounds like that's your ultimate intent with this, right?  I'm not fond of that idea.

Could you seek out another reviewer with makefile experience to take a look at this particular aspect of the changes (and the related export of BUILD_DIR) and write their comments here?  I'll defer to whatever you and your 2nd reviewer decide so as to not hold up this merge any further.

mlyle: Well, I plan to move the BUILD_DIR one.   You made a good point that it makes sense to make sure that things are contained.

OTOH, the makefile stuff is rotting.  Most of the lines of target makefiles are dead and have no purpose.  They've been unmaintained.  They *have* been carelessly copy and pasted for a long time.  I really would like to refactor and clean the makefiles in future PRs, but I don't want to do it this way.  Instead, I'd like to export a few more paths and let common included files calculate a lot more.

mlyle: e.g. picking the first one of these I find in the makefile

```
ut_$(1)_%: $$(UT_OUT_DIR)
        $(V1) mkdir -p $(UT_OUT_DIR)/$(1)
        $(V1) cd $$(UT_ROOT_DIR) && \
                $$(MAKE) -r --no-print-directory \
                BUILD_TYPE=ut \
                BOARD_SHORT_NAME=$(1) \
                TCHAIN_PREFIX="" \
                REMOVE_CMD="$(RM)" \
                \
                MAKE_INC_DIR=$(MAKE_INC_DIR) \
                ROOT_DIR=$(ROOT_DIR) \
                BOARD_ROOT_DIR=$$(BOARD_ROOT_DIR) \
                BOARD_INFO_DIR=$$(BOARD_ROOT_DIR)/board-info \
                TARGET=$$(TARGET) \
                OUTDIR=$$(OUTDIR) \
                \
                PIOS=$(PIOS) \
                OPUAVOBJ=$(OPUAVOBJ) \
                OPUAVTALK=$(OPUAVTALK) \
                OPMODULEDIR=$(OPMODULEDIR) \
                FLIGHTLIB=$(FLIGHTLIB) \
                SHAREDAPIDIR=$(SHAREDAPIDIR) \
                \
                GTEST_DIR=$(GTEST_DIR) \
                \
                $$*
```
A lot of these are dead.  BUILD_TYPE/BOARD_SHORT_NAME are abused.  PIOS is dead / wrong for unit test targets.  Some of the other libs are dead.

It takes a lot of work to make sure these are really dead and I question the benefits from narrowing the scoping of a lot of these things that I think it's fine for all flight makefiles to know.

For now, I'll cut down this new one when I get to it, but I hope we could reach agreement to just export some of the more frequently used ones of these and cut down the boilerplate.

solidgoldbomb: Agreed, this sort of list is hard to keep up to date for similar reasons as pruning the set of unneeded ```#include``` lines from source files.  There's no explicit failure if you allow something that is no longer needed.  Or even if you allow something that is not needed and subsequently becomes wrong.

I agree that the submakefiles have a lot of unnecessary boilerplate lines that have been cut/pasted over time.  Getting that cleaned up would be a win for sure.

I can get behind nearly any approach that still provides containment/scoping for the makefiles in the targets dir.  If that ends up being something like exporting some top-level dirs to a common makefile that builds the scoped definitions and then unsets the top-level exports, that sounds good to me.  I'm just a bit wary of going back to the bad-old-days where these top-level dirs were globals throughout the build.

The overall intention of having some sort of API/scoping for the submakefiles is the thing I'm hoping to preserve.  I think there are a lot of different ways to implement it and there's lots of room for moving to new/better ways in another PR.

In this PR, the export of the BUILD_DIR and extra parameters in the new lib build target are the only sticking points for me to merge.  Like I said before though, I'll still defer to a 2nd reviewer on both points if that's what they recommend.

mlyle: OK.  I appreciate that.

For this one, I took out the BUILD_DIR export and cut down the lines a bit.

But I had to hack to find the appropriate uavobjects lib dir using path relative to the output path.  Probably need to ultimately have a lib path that gets appropriately exported, with subdirs for arch, but there's some more infrastructure needed to do that.

solidgoldbomb:
Thanks for making that change.  Looking forward to the cleanup of the submakefiles in future PRs.

One more thing that I noticed while running a test build (missed it the first time around).  The invocation of ```ar``` doesn't obey the ```V=``` setting and dumps a huge blob (~60 lines) of text to the console during the build.  Would be good to get that cleaned up by adding a template to wrap ```$(AR)``` in firmware-defs.mk.

Would you prefer to have this merged now and leave the output cleanup for a separate PR or do you want to fix it in this PR?  I'm OK either way but would like to see that fixed up.

mlyle:
I dropped the 'v' from the ar command line.

Still no actual flight test on this, but when I compared the assembly output the only thing I saw change was the order / address of uavo functions in the binary.  Other than changes to addresses, the firmware is byte-for-byte identical on at least F4 targets.

solidgoldbomb:
That's better without the ```v``` but it really should get a proper template that obeys the ```V=0/V=1``` parameters to make so that it dumps a brief summary line like ```AR  uavobjects_armhardfp/libuavobject.a``` by default and dumps out the full command line when the user specifies ```V=1```.

I think the biggest risk in this change (building once) would be if we previously used different optimization flags on different targets and now we have effectively forced the same flags on all uavo code on all targets (modulo the hard/soft float split).  It might be worth checking the code size delta on the F1 targets just to make sure they haven't bloated.

Code looks OK to me.  Your call on whether you're comfortable with current test coverage.

mlyle:
Thanks for the review.  Yah, this might as well see a little more test before merging-- I have no immediately dependent work.

It also appears to just be re-arranged code on F1.

mlyle:
Flight tested on sparky2.  Bench-tested with motors spinnin' on CC3D.

mlyle:
jenkins, test this please

tracernz:
The ar output is still not suppressed when building with normal verbosity as per @solidgoldbomb.
>One more thing that I noticed while running a test build (missed it the first time around). The invocation of ar doesn't obey the V= setting and dumps a huge blob (~60 lines) of text to the console during the build. Would be good to get that cleaned up by adding a template to wrap $(AR) in firmware-defs.mk.

Additionally, this requires all of the UAVO synthetics to be rebuilt (to [remove the dependence on openpilot.h -> pios_config.h](https://github.com/TauLabs/TauLabs/pull/1985/files#diff-f658c23414618e2c99bad0db11190a0fL37)) which will require a clean when jumping between branches on opposite sides of this merge.

tracernz:
Why did we remove all the uavobjects_<target> targets? Why do we want to build all of the UAVOs for every target?

PR #1986, opened 2015-10-12
python: Allow filtering of objects

mlyle:
This greatly simplifies looking at a range of data during a navigation
or RTH segment.

I do things like this at the shell.py command line:
```
gp=uavo_list.as_numpy_array(UAVO_GPSPosition, lambda x: x.time>85 and x.time<115)
ba=uavo_list.as_numpy_array(UAVO_BaroAltitude, lambda x: x.time>85 and x.time<115)
pa=uavo_list.as_numpy_array(UAVO_PositionActual, lambda x: x.time>85 and x.time<115)
```

### DIFF
--- a/python/taulabs/telemetry.py
+++ b/python/taulabs/telemetry.py
@@ -80,7 +80,7 @@ class TelemetryBase():
 
         self.eof = False
 
-    def as_numpy_array(self, match_class):
+    def as_numpy_array(self, match_class, filter_cond=None):
         """ Transforms all received instances of a given object to a numpy array.
         
         match_class: the UAVO_* class you'd like to match.
@@ -91,6 +91,10 @@ class TelemetryBase():
         # Find the subset of this list that is of the requested class
         filtered_list = filter(lambda x: isinstance(x, match_class), self)
  
+        # Perform any additional requested filtering
+        if filter_cond is not None:
+            filtered_list = filter(filter_cond, filtered_list)
+
         # Check for an empty list
         if filtered_list == []:
             return np.array([])


PTDreamer:
Needs rebasing

peabody124:
Could you give a code snippets how to use it? Perhaps in the commit message? This combined with the enum support will make some things much easier!

mlyle:
Updated above.

mlyle:
Did something change with the github/organization settings?  It looks like *every* PR needs a rebase now.

PTDreamer:
> Did something change with the github/organization settings? It looks like every PR needs a rebase now

I added those protections we talked about on IRC. Should only affect force pushes  to next, and the impossibility to merge PRs with failed tests.

tracernz:
> I added those protections we talked about on IRC. Should only affect force pushes to next, and the impossibility to merge PRs with failed tests.

@PTDreamer: Thanks for doing that. I expect whatever the issue is, it will only apply to existing PRs and not new ones.

kubark42:
@mlyle So in a nutshell, this allows me to choose time windows in the data? That's a million times better than my old approach!

```
startTime = 0;
endTime = Inf;

# Read data from the UAVObject
accels = uavos.as_numpy_array(UAVO_Accels)

# Get only the indices that are in the time of interest
accels_idx_of_interest = (accels['time']>startTime) & (accels['time']<endTime)

# Extract the data of interest and place into an array
accels_x    = accels['x'][accels_idx_of_interest,0]
accels_y    = accels['y'][accels_idx_of_interest,0]
accels_z    = accels['z'][accels_idx_of_interest,0]
accels_time = accels['time'][accels_idx_of_interest]
```

mlyle:
@kubark42 yes, or filter on anything else / any other field, like get all the points of high acceleration.

kubark42:
Oh, how cool! Could you give some example of that as well? I'm truly not a pythonista and don't know much about what ```lambda x``` is accomplishing. Is that like a private function in Matlab? Could it be defined *a priori*, instead of on each line? Can it accept variables as arguments?

mlyle:
```foo(lambda x: x.time>85 and x.time<115)```

is equivalent to 

```
def func(x):
    return x.time > 85 and x.time < 115

foo(func)
```

So you could do

```pa=uavo_list.as_numpy_array(UAVO_PositionActual, lambda x: ((x.North ** 2) + (x.East ** 2)) ** 0.5 > 15.0```

to get all the points at least 15 meters away from home.

**Comment by peabody124** [comment]
That is so awesome

**Comment by kubark42** [comment]
Maybe @mluessi can show me when he gets here, but I'm still unclear on what is the right way to accomplish the following two pseudo-codes:

```
bob = lambda x: x.time>85 and x.time<115
uavo_list.as_numpy_array(UAVO_A, bob)
uavo_list.as_numpy_array(UAVO_B, bob)
.
.
.
uavo_list.as_numpy_array(UAVO_Y, bob)
uavo_list.as_numpy_array(UAVO_Z, bob)
```

Or 

```
startTime = 0
endTime = Inf
gp=uavo_list.as_numpy_array(UAVO_GPSPosition, lambda x: x.time>startTime and x.time<endTime)
```

mlyle:
Either will work.  Or even an inline local def of a function if you don't like the lambda syntax (and functions can see variables from their enclosing scopes)

kubark42:
Thanks, this is so awesome.

mluessi:
Yes, @kubark42 Python is awesome ;)

kubark42:
@mluessi but, but.... whitespace! :D

PR #1987, opened 2015-10-12
Actuator/mixer refactor

mlyle:
Seeking additional code clarity and decreased resource utilization.

. Principal changes are attempting to factor more things into functions and simplify logic. The parallel array of mixer types is eliminated. The settings objects are moved to globals instead of passed around.

Needs thorough test and review.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -13,6 +13,7 @@
  * @file       actuator.c
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2015
+ * @author     dRonin, http://dronin.org, Copyright (C) 2015.
  * @brief      Actuator module. Drives the actuators (servos, motors etc).
  *
  * @see        The GNU Public License (GPL) Version 3
@@ -66,7 +67,6 @@
 
 // Private types
 
-
 // Private variables
 static struct pios_queue *queue;
 static struct pios_thread *taskHandle;
@@ -76,28 +76,29 @@ static volatile bool actuator_settings_updated;
 // used to inform the actuator thread that mixer settings are changed
 static volatile bool mixer_settings_updated;
 
+// The actual mixer settings data, pulled at the top of the actuator thread
+MixerSettingsData mixerSettings;
+
+// Ditto, for the actuator settings.
+ActuatorSettingsData actuatorSettings;
+
 // Private functions
 static void actuatorTask(void* parameters);
-static float scaleChannel(float value, float max, float min, float neutral);
-static void setFailsafe(const ActuatorSettingsData * actuatorSettings, const MixerSettingsData * mixerSettings);
+static float scaleChannel(float value, int idx);
+static void setFailsafe();
 static float ThrottleCurve(const float input, const float* curve, uint8_t num_points);
 static float CollectiveCurve(const float input, const float* curve, uint8_t num_points);
-static bool set_channel(uint8_t mixer_channel, float value, const ActuatorSettingsData * actuatorSettings);
-static void actuator_update_rate_if_changed(const ActuatorSettingsData * actuatorSettings, bool force_update);
+static bool set_channel(uint8_t mixer_channel, float value);
+static void actuator_update_rate_if_changed(bool force_update);
 static void MixerSettingsUpdatedCb(UAVObjEvent * ev);
 static void ActuatorSettingsUpdatedCb(UAVObjEvent * ev);
 float ProcessMixer(const int index, const float curve1, const float curve2,
-		   const MixerSettingsData* mixerSettings, ActuatorDesiredData* desired,
-		   const float period);
+		   ActuatorDesiredData *desired);
+static float MixChannel(int ct, ActuatorDesiredData *desired,
+		float curve1, float curve2);
 
-//this structure allows easy access to the UAVObjects for one mixer.
-typedef struct {
-	typeof(((MixerSettingsData *) NULL)->Mixer1Type) type;
-	typeof(((MixerSettingsData *) NULL)->Mixer1Vector[0]) *vector;
-} MixerTable;
-
-MixerTable MixerEntry_t[ACTUATORCOMMAND_CHANNEL_NUMELEM];
-
+static MixerSettingsMixer1TypeOptions GetMixerType(int idx);
+static int8_t *GetMixerVector(int idx);
 
 /**
  * @brief Module initialization
@@ -144,6 +145,45 @@ int32_t ActuatorInitialize()
 }
 MODULE_INITCALL(ActuatorInitialize, ActuatorStart);
 
+float GetCurve2Source(ActuatorDesiredData *desired, MixerSettingsCurve2SourceOptions source) {
+	float tmp;
+
+	switch (source) {
+		case MIXERSETTINGS_CURVE2SOURCE_THROTTLE:
+			return desired->Throttle;
+			break;
+		case MIXERSETTINGS_CURVE2SOURCE_ROLL:
+			return desired->Roll;
+			break;
+		case MIXERSETTINGS_CURVE2SOURCE_PITCH:
+			return desired->Pitch;
+			break;	
+		case MIXERSETTINGS_CURVE2SOURCE_YAW:
+			return desired->Yaw;
+			break;
+		case MIXERSETTINGS_CURVE2SOURCE_COLLECTIVE:
+			ManualControlCommandCollectiveGet(&tmp);
+			return tmp;
+			break;
+		case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY0:
+		case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY1:
+		case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY2:
+		case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY3:
+		case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY4:
+		case MIXERSETTINGS_CURVE2SOURCE_ACCESSORY5:
+			(void) 0;
+			AccessoryDesiredData accessory;
+
+			if(AccessoryDesiredInstGet(mixerSettings.Curve2Source - MIXERSETTINGS_CURVE2SOURCE_ACCESSORY0,&accessory) == 0)
+				return accessory.AccessoryVal;
+			return 0;
+			break;
+	}
+
+	/* Can't get here */
+	return 0;
+}
+
 /**
  * @brief Main Actuator module task
  *
@@ -170,20 +210,18 @@ static void actuatorTask(void* parameters)
 	FlightStatusData flightStatus;
 
 	/* Read initial values of ActuatorSettings */
-	ActuatorSettingsData actuatorSettings;
 	actuator_settings_updated = true;
 	ActuatorSettingsGet(&actuatorSettings);
 
 	/* Read initial values of MixerSettings */
-	MixerSettingsData mixerSettings;
 	mixer_settings_updated = true;
 	MixerSettingsGet(&mixerSettings);
 
 	/* Force an initial configuration of the actuator update rates */
-	actuator_update_rate_if_changed(&actuatorSettings, true);
+	actuator_update_rate_if_changed(true);
 
 	// Go to the neutral (failsafe) values until an ActuatorDesired update is received
-	setFailsafe(&actuatorSettings, &mixerSettings);
+	setFailsafe();
 
 	// Main task loop
 	lastSysTime = PIOS_Thread_Systime();
@@ -197,65 +235,18 @@ static void actuatorTask(void* parameters)
 		/* Process settings updated events even in timeout case so we always act on the latest settings */
 		if (actuator_settings_updated) {
 			actuator_settings_updated = false;
-			ActuatorSettingsGet (&actuatorSettings);
-			actuator_update_rate_if_changed (&actuatorSettings, false);
+			ActuatorSettingsGet(&actuatorSettings);
+			actuator_update_rate_if_changed(false);
 		}
 		if (mixer_settings_updated) {
 			mixer_settings_updated = false;
-			MixerSettingsGet (&mixerSettings);
+			MixerSettingsGet(&mixerSettings);
 
-			for (int i=0; i<ACTUATORCOMMAND_CHANNEL_NUMELEM; i++) {
-				switch(i) {
-				case 0:
-					MixerEntry_t[i].type = mixerSettings.Mixer1Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer1Vector;
-					break;
-				case 1:
-					MixerEntry_t[i].type = mixerSettings.Mixer2Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer2Vector;
-					break;
-				case 2:
-					MixerEntry_t[i].type = mixerSettings.Mixer3Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer3Vector;
-					break;
-				case 3:
-					MixerEntry_t[i].type = mixerSettings.Mixer4Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer4Vector;
-					break;
-				case 4:
-					MixerEntry_t[i].type = mixerSettings.Mixer5Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer5Vector;
-					break;
-				case 5:
-					MixerEntry_t[i].type = mixerSettings.Mixer6Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer6Vector;
-					break;
-				case 6:
-					MixerEntry_t[i].type = mixerSettings.Mixer7Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer7Vector;
-					break;
-				case 7:
-					MixerEntry_t[i].type = mixerSettings.Mixer8Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer8Vector;
-					break;
-				case 8:
-					MixerEntry_t[i].type = mixerSettings.Mixer9Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer9Vector;
-					break;
-				case 9:
-					MixerEntry_t[i].type = mixerSettings.Mixer10Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer10Vector;
-					break;
-				default:
-					// We can never get here unless there are mixer channels not handled in the above. Fail out.
-					PIOS_Assert(0);
-				}
-			}
 		}
 
 		if (rc != true) {
 			/* Update of ActuatorDesired timed out.  Go to failsafe */
-			setFailsafe(&actuatorSettings, &mixerSettings);
+			setFailsafe();
 			continue;
 		}
 
@@ -273,18 +264,17 @@ static void actuatorTask(void* parameters)
 		MixerStatusGet(&mixerStatus);
 #endif
 		int nMixers = 0;
-		MixerTable *mixers = MixerEntry_t;
 
 		for(int ct=0; ct < MAX_MIX_ACTUATORS; ct++)
 		{
-			if(mixers[ct].type != MIXERSETTINGS_MIXER1TYPE_DISABLED)
+			if(GetMixerType(ct) != MIXERSETTINGS_MIXER1TYPE_DISABLED)
 			{
 				nMixers ++;
 			}
 		}
 		if((nMixers < 2) && !ActuatorCommandReadOnly()) //Nothing can fly with less than two mixers.
 		{
-			setFailsafe(&actuatorSettings, &mixerSettings); // So that channels like PWM buzzer keep working
+			setFailsafe(); // So that channels like PWM buzzer keep working
 			continue;
 		}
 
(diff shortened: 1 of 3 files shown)


PTDreamer:
LGTM
Seems to need rebasing though

flight/Modules/Actuator/actuator.c
```diff
@@ -532,10 +426,24 @@ static float scaleChannel(float value, float max, float min, float neutral)
...
+			return 0;
+	}
+
+}
+
 /**
  * Set actuator output to the neutral values (failsafe)
  */
-static void setFailsafe(const ActuatorSettingsData * actuatorSettings, const MixerSettingsData * mixerSettings)
+static void setFailsafe()
```
kubark42: I don't know how I feel about this kind of change. I like it when a function can be wholly tested for correctness, based on inputted and returned values. I don't think this change makes things worse, because the function already depended on external bits. But I wonder if it's possible to move generally more in that direction instead of doubling down on depending on UAVO get/set operations.

flight/Modules/Actuator/actuator.c
```diff
@@ -691,7 +589,7 @@ static bool set_channel(uint8_t mixer_channel, float value, const ActuatorSettin
                             }
 			}
 			PIOS_Servo_Set(mixer_channel,
-							buzzOn?actuatorSettings->ChannelMax[mixer_channel]:actuatorSettings->ChannelMin[mixer_channel]);
+							buzzOn?actuatorSettings.ChannelMax[mixer_channel]:actuatorSettings.ChannelMin[mixer_channel]);
```
kubark42: Not your fault in the original, but could we get some spaces here as long as that line is being changed?

flight/Modules/Actuator/actuator.c
```diff
@@ -714,31 +612,180 @@ static bool set_channel(uint8_t mixer_channel, float value, const ActuatorSettin
...
+			}
+			return -1;
+		/* No default-- make sure we handle all types positively here */
+	} 
+
+	/* Can't get here. */
+	return -1;
+}
+
+static MixerSettingsMixer1TypeOptions GetMixerType(int idx)
```
kubark42: I was all excited when I saw you removed my code above because I thought you had come up with some awesome way to get rid of the ugliness I introduced. So sad to discover it's not gone yet! :D

flight/Modules/Actuator/actuator.c
```diff
@@ -191,76 +214,34 @@ static void actuatorTask(void* parameters)
 	{
 		PIOS_WDG_UpdateFlag(PIOS_WDG_ACTUATOR);
 
-		// Wait until the ActuatorDesired object is updated
-		bool rc = PIOS_Queue_Receive(queue, &ev, FAILSAFE_TIMEOUT_MS);
+		bool rc=false;
```
kubark42: Spacing around ```=```.

flight/Modules/Actuator/actuator.c
```diff
@@ -191,76 +214,34 @@ static void actuatorTask(void* parameters)
...
 
-		/* Process settings updated events even in timeout case so we always act on the latest settings */
-		if (actuator_settings_updated) {
-			actuator_settings_updated = false;
-			ActuatorSettingsGet (&actuatorSettings);
-			actuator_update_rate_if_changed (&actuatorSettings, false);
+		// If it's our first iteration, fall down to getting the
+		// config and enabling failsafe.
+		if (!first) {
```
kubark42: Why do this inside the loop, instead of immediate before starting the loop? That would save the continual evaluation of the ```if()``` clause, right?

flight/Modules/Actuator/actuator.c
```diff
@@ -191,76 +214,34 @@ static void actuatorTask(void* parameters)
...
-					MixerEntry_t[i].type = mixerSettings.Mixer10Type;
-					MixerEntry_t[i].vector = mixerSettings.Mixer10Vector;
-					break;
-				default:
-					// We can never get here unless there are mixer channels not handled in the above. Fail out.
-					PIOS_Assert(0);
-				}
-			}
+
+		/* Process settings updated events even in timeout case so we always act on the latest settings */
```
kubark42: Not that it's important, but why use ```/* */``` style comments instead of ```//```?

flight/Modules/Actuator/actuator.c
```diff
@@ -297,126 +277,36 @@ static void actuatorTask(void* parameters)
...
-					}
-				}
-			} else {
-				status[ct] = -1;
-			}
+			status[ct] = MixChannel(ct, &desired, curve1, curve2);
 
 			// Motors have additional protection for when to be on
-			if(mixers[ct].type == MIXERSETTINGS_MIXER1TYPE_MOTOR) {
+			if(GetMixerType(ct) == MIXERSETTINGS_MIXER1TYPE_MOTOR) {
```
kubark42: This is evaluated twice in the ```while``` loop, here and at https://github.com/TauLabs/TauLabs/pull/1987/files#diff-6756853a8b52a94d64d1f12e4377cb31R260. Do you see a way to reduce that to just once?

flight/Modules/Actuator/actuator.c
```diff
@@ -458,13 +347,14 @@ static void actuatorTask(void* parameters)
...
 float ProcessMixer(const int index, const float curve1, const float curve2,
-		   const MixerSettingsData* mixerSettings, ActuatorDesiredData* desired, const float period)
+		   ActuatorDesiredData *desired)
 {
-	float result = (((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE1] * curve1) +
-	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_THROTTLECURVE2] * curve2) +
-	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_ROLL] * desired->Roll) +
-	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_PITCH] * desired->Pitch) +
-	                ((float)MixerEntry_t[index].vector[MIXERSETTINGS_MIXER1VECTOR_YAW] * desired->Yaw)) * (1.0f / MULTIROTOR_MIXER_UPPER_BOUND);
+	int8_t *vector = GetMixerVector(index);
```
kubark42: Doesn't this lead to many more calls than the original code? It seems to me that this was evaluated once per settings change, but now it's being evaluated every loop execution, right?

flight/Modules/Actuator/actuator.c
```diff
@@ -66,38 +67,35 @@
...
+MixerSettingsData mixerSettings;
+
+// Ditto, for the actuator settings.
+ActuatorSettingsData actuatorSettings;
 
 // Private functions
 static void actuatorTask(void* parameters);
-static float scaleChannel(float value, float max, float min, float neutral);
-static void setFailsafe(const ActuatorSettingsData * actuatorSettings, const MixerSettingsData * mixerSettings);
+static float scaleChannel(float value, int idx);
```
kubark42: I realize it's legacy, but I think there was an attempt to remove all CamelCase in the flight firmware. @solidgoldbomb and @peabody124 might weigh in here.

flight/Modules/Actuator/actuator.c
mlyle: It's unfortunate and a factoring I prefer to avoid, but seemed preferable to the alternative where settings-related code was duplicated and could be erroneously changed only in one place or another.  As it stands, the if evaluation is very very cheap compared to everything involved in draining a queue (locks, likely task switch).  Like 3 to 5 orders of magnitude cheaper.

flight/Modules/Actuator/actuator.c
mlyle: I don't know, it's what my fingers did.  I never know how long a comment is going to be until I'm done typing ;)

flight/Modules/Actuator/actuator.c
mlyle: A short wrapper makes it testable that way.  I could plumb everything into the actual calls but I fear it makes the code harder to read from being so verbose.  Upside is it would move even more of the memory consumption from heap.

flight/Modules/Actuator/actuator.c
mlyle: Not really.. but it's fairly cheap.  This rips out the whole type caching mechanism because looking up the type is fairly cheap (and if it's a performance problem GetMixerType can be made even cheaper).

That big long switch becomes:

```
  58 0000 0928                  cmp     r0, #9
  59 0002 2ED8                  bhi     .L5
  60 0004 DFE800F0              tbb     [pc, r0]
  61                    .L7:
  62 0008 05                    .byte   (.L6-.L7)/2
  63 0009 09                    .byte   (.L8-.L7)/2
  64 000a 0D                    .byte   (.L9-.L7)/2
  65 000b 11                    .byte   (.L10-.L7)/2
  66 000c 15                    .byte   (.L11-.L7)/2
  67 000d 19                    .byte   (.L12-.L7)/2
  68 000e 1D                    .byte   (.L13-.L7)/2
  69 000f 21                    .byte   (.L14-.L7)/2
  70 0010 25                    .byte   (.L15-.L7)/2
  71 0011 29                    .byte   (.L16-.L7)/2
```

TBB implements a jump table that takes 3-5 cycles.

It could be made cheaper at the expense of legibility with pointer arithmetic and DONT_BUILD_IF type asserts (it'd inline then).

flight/Modules/Actuator/actuator.c
mlyle: Yah, I don't like the structure.  It saves memory, but it prevents mixing arbitrary things-- e.g. mixing gimbal roll and pitch for something like the SuperSimpleGimbal.  However I wanted no functional change in this PR.

flight/Modules/Actuator/actuator.c
mlyle: It's preserving the existing naming and keeps the diff down.  Also, it's unfortunate because the actual name of the uavos and the stuff inside and their accessors are all camelcased.

OTOH the above "taskHandle" is clearly badness.

flight/Modules/Actuator/actuator.c
mlyle: If we really care about performance it wouldn't be insane to completely unroll the loops and waste a little bit of flash which would be faster than the old code.

I'm biased to not care:  On Cortex-M4 performance doesn't matter; on M3 this compares to ~50 cycles for each of the below multiplies (6) and ~20 cycles for each of the adds (4). So maybe it's a 10% penalty for runtime for this one function?

flight/Modules/Actuator/actuator.c
mlyle: I've pushed another change with uncrustify applied to the file with proposed settings that match most of the flight mode style.  Perhaps we consider that?  If we don't want to reindent that I can pull that and just reindent those two lines.

mlyle: Note things in this area have pre-existing mixed use of spaces / tabs, etc, are completely broken for style for other reasons.

flight/Modules/Actuator/actuator.c
```diff
@@ -170,98 +206,42 @@ static void actuatorTask(void* parameters)
...
 	// Main task loop
 	lastSysTime = PIOS_Thread_Systime();
-	while (1)
-	{
+	while (1) {
 		PIOS_WDG_UpdateFlag(PIOS_WDG_ACTUATOR);
 
-		// Wait until the ActuatorDesired object is updated
-		bool rc = PIOS_Queue_Receive(queue, &ev, FAILSAFE_TIMEOUT_MS);
+		bool rc = false;
```
kubark42: Not your fault, but I would be a proponent for refactoring this variable name to be something sensible. ```rc``` means as much as ```tmp1``` in this context, and it actually is harder to search for than ```tmp1``` because the string "rc" shows up in lots of unrelated spots, like "ActuatorCommand"

flight/Modules/Actuator/actuator.c
kubark42: All these settings objects are now global, so the ```settings_updated``` variable can trigger a function update such as

```
settingsUpdate()
{
	ActuatorSettingsGet(&actuatorSettings);
	actuator_update_rate_if_changed(false);
	MixerSettingsGet(&mixerSettings);
}
```

and then ```settingsUpdate``` could be called before the loop begins, as well as inside the loop, right?

Or am I missing something?

flight/Modules/Actuator/actuator.c
kubark42: Is there a general rule for when to use ```/* */``` vs ```//```?

flight/Modules/Actuator/actuator.c
kubark42: Agreed. I think historically the general rule has been to fix style problems whenever the nearby functions are changed. But this process is taking so long, and we're not using astyle for new functions/files, so I think the problem is getting worse faster than it's getting better.

flight/Modules/Actuator/actuator.c
kubark42: Fair enough. What about the new functions, ```MixChannel()``` and ```GetMixerVector()```

flight/Modules/Actuator/actuator.c
mlyle: That would cause undefined behavior unless actuatorSettings is global, because it means that we could be running the loop when we get that update callback and internal data structures change.  Unless something is declared "volatile", the optimizer is allowed to make optimizations on the assumption that actuatorSettings doesn't change within code unless it is passed to something that could change it within that thread of execution (that is, a function that changes it or a function in another module that escapes analysis).   Undefined behavior means the code can do literally anything, not necessarily something one would suspect from inspecting the input source (e.g. run "both halves" of an if/else statement, or..).

Declaring it volatile is really bad for performance (generates frequent loads) and even so takes a lot of analysis on conditionals that state changes don't do something bad.

mlyle: Or in other words, the pattern this thing uses with a flag for communication and running the update once at the top of the task's loop is a good one.

mlyle: P.S. There is plenty of code that violates this in the codebase that very possibly does bad things at runtime.

kubark42: > That would cause undefined behavior unless actuatorSettings is globa

Okay, I'm being thick here. ```actuatorSettings``` data is global, as per https://github.com/mlyle/taulabs/blob/actuator-refactor/flight/Modules/Actuator/actuator.c#L80.

flight/Modules/Actuator/actuator.c
```diff
@@ -66,38 +66,35 @@
...
 
-// used to inform the actuator thread that actuator update rate is changed
-static volatile bool actuator_settings_updated;
-// used to inform the actuator thread that mixer settings are changed
-static volatile bool mixer_settings_updated;
+// used to inform the actuator thread that actuator / mixer settings are updated
+static volatile bool settings_updated;
+
+// The actual mixer settings data, pulled at the top of the actuator thread
+MixerSettingsData mixerSettings;
```
kubark42: This needs to be ```static```.

flight/Modules/Actuator/actuator.c
```diff
@@ -66,38 +66,35 @@
...
+
+// Ditto, for the actuator settings.
+ActuatorSettingsData actuatorSettings;
```
kubark42: This needs to be ```static```.

flight/Modules/Actuator/actuator.c
mlyle: good eye.

flight/Modules/Actuator/actuator.c
kubark42: P.S. Don't mean to bikeshed anything, this isn't a deal-breaker. I just want to understand what is going on here so I'm not tempted to fix it on my own time. ;)

flight/Modules/Actuator/actuator.c
mlyle: I can rename all the functions in the file if you want.. I like conforming to existing local conventions but yah, everything is mixed :/

flight/Modules/Actuator/actuator.c
mlyle: I meant to say volatile.  Sorry.

kubark42: Okay, this still isn't making sense to me. Wrapping up those three lines of code into a function shouldn't confuse the compiler. I get the feeling we're having a bit of a comms barf right now, so I'll prepare a gist with the full proposed solution.

flight/Modules/Actuator/actuator.c
kubark42: Honestly, I prefer camel case. But @solidgoldbomb and @peabody124 feel pretty strongly about not having it, so when introducing new functions there's been an effort to move toward Linux kernel style naming.

flight/Modules/Actuator/actuator.c
mlyle: I am half awake.  I understood your request to say that it happen in the callback itself.

I've reread and now am unsure-- if you're saying to just to factor to a call and to put the call before and conditional in the while, I thought about that.  I picked this way, and they seem equivalent.

flight/Modules/Actuator/actuator.c
mlyle: I'd rather rename all or none, because I really hate things being mixed in a file.

flight/Modules/Actuator/actuator.c
mlyle: I think this is what you meant.  I was confused because of the existing SettingsUpdated() call that is a callback.

flight/Modules/Actuator/actuator.c
mlyle: Technically we're supposed to use Linux Kernel style which says "never use // comments" heh.

flight/Modules/Actuator/actuator.c
kubark42: <facepalm>

Totally my fault. I missed that there was another function called ```SettingsUpdatedCb()```, which is where the confusion came from. I think you got what I meant in the end.

The advantage of the call before is we can eliminate the ```first``` variable and its conditional. I like to reduce complexity inside the while loop, not only for efficiency reasons but because it makes it easier to understand what the loop is intended to do. I admit to getting a little carried away by being offended at the notion of running something once, and then checking forever afterwards. It feels icky, but it truly is not a big impact on processor.

flight/Modules/Actuator/actuator.c
kubark42: I'm fine with whichever. Let's give @peabody124 and/or @solidgoldbomb a day to chime in and if we don't hear from them then we can assume it's not important and merge anyway.

mlyle: I'll go ahead and rename functions (ignoring variables because that's just too much diff).  Every single possible convention is currently used and it's yucky.

kubark42: Agreed. [Start a new convention to fix the old ones?](https://xkcd.com/927/)

flight/Modules/Actuator/actuator.c
mlyle: Do you prefer this factor: https://github.com/mlyle/taulabs/commit/752c1f705e62e3dfe7a8f0a0d6d5998b8aa711bb

mlyle: Note it can use a several-millisecond old configuration object because of the re-order of the config fetch with respect to the queue wait.

mlyle:
I also added the DONT_BUILD_IF thing.  When it's deliberately "wrong" you get something like:

/Users/mlyle/taulabs/flight/Modules/Actuator/actuator.c:766:14: error: size of array 'static_assertion_OutputModeAssumptions' is negative
 DONT_BUILD_IF(!OUTPUT_MODE_ASSUMPTIONS, OutputModeAssumptions);

when the assumptions fail.

mlyle:
OK, so I have a different set of fixes checked in.

Note that this is just a prelude to additional work in the mixer.

P.S. I've thought more about the `a * x^b` basis function in the actuator fitting-- I don't like it.  The a commutes with the mixer and the PID coefficients already; I really still don't understand its purpose :/  I could understand like `a + (1 -a) x^b` as a basis for offset or similar.

kubark42:
> I also added the DONT_BUILD_IF thing.

Neat! Very good pattern.

> I've thought more about the a * x^b basis function in the actuator fitting...

Agreed that 1) ```a``` can be grouped in with the rate PID and 2) ```a``` is very close to 1.0 so has a limited impact. On the one hand, the ideal is to have the actuators be as well mapped as possible and so I would advocate for the term to be kept. On the other hand, it's well known that these models are only valid for hovering flight, so we have not clue if this is making things better, worse, or unchanged when in lateral flight.

If I were to do your suggested fitting, I would actually use ```c*(a+(1-a)*x^b)``` and then abstract out ```c``` inside the UAVO. That would give a more optimal set of parameters. Note, however, that in general I am uncomfortable with too many parameters because that leads to overfitting: "With four parameters I can fit an elephant, and with five I can make him wiggle his trunk."[*]

[*] Turns out you really can! https://publications.mpi-cbg.de/Mayer_2010_4314.pdf

flight/PiOS.posix/openpilot.h
```diff
@@ -1,12 +1,11 @@
 /**
  ******************************************************************************
- * @addtogroup OpenPilotSystem OpenPilot System
- * @{
- * @addtogroup OpenPilotCore OpenPilot Core
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
  * @{
+ *
  * @file       openpilot.h
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
```
kubark42: As long as poking around here, maybe update the copyright?

kubark42:
This looks pretty reasonable. What has it been tested on?

mlyle:
I agree; this one needs test.  I will get that done soon.

mlyle:
Flown on sparky2.

Benchtested with motors spinning on CC3D.

mlyle:
confirmed correct failsafe behavior across targets, too.

kubark42:
Jenkins managed to confirm that things build properly before going silent. Giving this a try.

PR #1990, opened 2015-10-14
Naze: Configure gyro sample-rate and filter cutoff

tracernz:
Looks like this was missed out when we originally merged the target. It can run okay up to (and including) 666 Hz sample-rate (with 99% CPU usage reported) but after that it fails to operate correctly. Thoughts on removing the (unusable) higher rate options?

### DIFF
--- a/flight/targets/naze32/fw/pios_board.c
+++ b/flight/targets/naze32/fw/pios_board.c
@@ -528,6 +528,34 @@ void PIOS_Board_Init(void) {
 			break;
 	}
 
+	// the filter has to be set before rate else divisor calculation will fail
+	uint8_t hw_mpu_dlpf;
+	HwNazeMPU6000DLPFGet(&hw_mpu_dlpf);
+	enum pios_mpu60x0_filter mpu_dlpf = \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_256) ? PIOS_MPU60X0_LOWPASS_256_HZ : \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_188) ? PIOS_MPU60X0_LOWPASS_188_HZ : \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_98) ? PIOS_MPU60X0_LOWPASS_98_HZ : \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_42) ? PIOS_MPU60X0_LOWPASS_42_HZ : \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_20) ? PIOS_MPU60X0_LOWPASS_20_HZ : \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_10) ? PIOS_MPU60X0_LOWPASS_10_HZ : \
+			    (hw_mpu_dlpf == HWNAZE_MPU6000DLPF_5) ? PIOS_MPU60X0_LOWPASS_5_HZ : \
+			    pios_mpu6050_cfg.default_filter;
+	PIOS_MPU6050_SetLPF(mpu_dlpf);
+
+	uint8_t hw_mpu_samplerate;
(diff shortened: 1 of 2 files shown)


kubark42:
99%, wow. 

I'd suggest not supporting it in the UAVO if the board itself can't support it. There's not much to be gained in the scenario that we leave the higher rates, but there's a lot to lose.

mlyle:
What's does "runs OK" mean?  Do we know for sure it's not dropping samples?

I think I'd prefer to take 666Hz away too based on this.

tracernz:
@kubark42 I suspect the CC3D has the same issue as it has rates up to 8 kHz available too.
@mlyle Runs ok means the update time was about right for 666 Hz but I'm not happy with the risk of it going bad.

I've removed rates above 500 Hz, and while I was changing the UAVO hash I figured I might as well rename the fields correctly.

kubark42:
Okay to merge, but I'd get a warmer, fuzzier feeling if we had a flight report at the 500Hz rate.

mlyle:
The previous default was 500, so it's what all Naze flight reports have been at.  Do you think we need a re-test from cutting down the options?

mlyle:
Oh, woops, missed the new code.  What rate were we running at without configuration? heh.

tracernz:
Actually it ignored the UAVO before and thus the previous default was 400 Hz.

kubark42:
> Actually it ignored the UAVO before and thus the previous default was 400 Hz.

400Hz or 500Hz?

There's a functional change here which *should* result in a return to default, but I'm hoping for a flight confirmation to get the warm fuzzies since this potentially changes the default.

tracernz:
The only possible rate before this PR was 400 Hz. This is set at https://github.com/tracernz/TauLabs/blob/naze-gyrorate/flight/targets/naze32/fw/pios_board.c#L205 and written to the gyro at https://github.com/tracernz/TauLabs/blob/naze-gyrorate/flight/PiOS/Common/pios_mpu6050.c#L204 during initialisation. It was then never written again because the code to read the UAVO and update the sample-rate were missing (they are added by this PR).

I would also be happier if this is flight-tested but it is unlikely I will get a chance for at least a few days. Happy to wait for merge until then.

kubark42:
Okay, this default change is actually important than I first realized. I think we should hold off until @tracernz confirms a successful flight test. 

Otherwise, LGTM.

kubark42:
Thanks for the clarification. Since this introduces a new default (500Hz) and removes the known-good update rate (400Hz) as an option, let's get a good flight test and then I'll merge.

tracernz:
@kubark42 I've just flight tested this with the gyro set to 500 Hz. It seemed to perform just fine in my driveway and successfully completed an autotune. I then took it out the back to the field and gave it a quick shakedown with the [autotuned gains which were a little suspect](http://forum.taulabs.org/viewtopic.php?f=27&t=268&p=7886#p7886). I think this is safe to merge.

mlyle:
jenkins: test this please

kubark42:
@tracernz could you do one last test where you do autotune without this patch, so it's running at 400Hz? While I agree that if rate works, that attitude ultimately *will* work, I agree it's suspect that something which worked before changes behavior now. It'd be good to get an idea if this patch is affecting that or not.

mlyle:
@kubark42 autotune is unreliable on all targets.  It often spits out bogus 8-10ms taus.

Attitude works on naze32-- that's how he was able to fly the autotune sequence.  Just the results he got was bogus which there's a fair amount of on the forum thread.

Then on the GCS side, there's a for loop with 30 iterations and no convergence check (as @mluessi pointed out).  If you get a <13ms tau, whether bogus or real, sliding the damping/noise sensitivity sliders around gives you effectively random tunes (Kp/Ki bounce around seemingly randomly).

mlyle:
jenkins, test this please

kubark42:
Fair enough. I don't use autotune, so I didn't realize it gives a fair number of inconsistent results.

PR #1992, opened 2015-10-16
flight/uavo: remove reference to flight plan

mlyle:
It's unused; the flight controller did not file a flight plan.


mluessi:
LGTM. Hopefully something similar will come back at some point in the future when it is actually implemented ;)

kubark42:
Damned pilots, never filing flight plans. 

Thanks for getting it out.

PR #1994, opened 2015-10-17
Add uncrustify to make system

mlyle:
Should supplant astyle, but astyle is not removed in this PR.


make/tools.mk
```diff
@@ -439,6 +439,45 @@ astyle_clean:
 	$(V0) @echo " CLEAN        $(ASTYLE_BUILD_DIR)"
 	$(V1) [ ! -d "$(ASTYLE_BUILD_DIR)" ] || $(RM) -r "$(ASTYLE_BUILD_DIR)"
 
+# Set up uncrustify tools
+UNCRUSTIFY_DIR := $(TOOLS_DIR)/uncrustify-0.61
+UNCRUSTIFY_BUILD_DIR := $(DL_DIR)/uncrustify
+libusb_win_install: LIBUSB_WIN_FILE := libusb-win32-bin-1.2.6.0.zip
```
mluessi: Should this be here (doesn't seem to be related to uncrustify)

make/uncrustify.cfg
```diff
@@ -0,0 +1,148 @@
+# first attempt at an uncrustify config file for dRonin
```
mluessi: ;)

make/tools.mk
```diff
@@ -439,6 +439,45 @@ astyle_clean:
...
+
+.PHONY: uncrustify_install
+uncrustify_install: | $(DL_DIR) $(TOOLS_DIR)
+uncrustify_install: UNCRUSTIFY_URL := http://downloads.sourceforge.net/project/uncrustify/uncrustify/uncrustify-0.61/uncrustify-0.61.tar.gz
```
PTDreamer: Wouldn't it be better to follow the project's gitHub master branch instead of having to update this on every new release?

make/tools.mk
```diff
@@ -439,6 +439,45 @@ astyle_clean:
 	$(V0) @echo " CLEAN        $(ASTYLE_BUILD_DIR)"
 	$(V1) [ ! -d "$(ASTYLE_BUILD_DIR)" ] || $(RM) -r "$(ASTYLE_BUILD_DIR)"
 
+# Set up uncrustify tools
+UNCRUSTIFY_DIR := $(TOOLS_DIR)/uncrustify-0.61
+UNCRUSTIFY_BUILD_DIR := $(DL_DIR)/uncrustify
+libusb_win_install: LIBUSB_WIN_FILE := libusb-win32-bin-1.2.6.0.zip
+
+.PHONY: uncrustify_install
```
PTDreamer: new target was not added to top level makefile help

make/tools.mk
mlyle: I don't think so-- I think things should be reproducible and we should only move between versions of tools if we really mean to.  If they change something that alters interpretation of the config file, etc, we could break.

make/tools.mk
mlyle: Added to help, and marked astyle as 'deprecated' in the help.

make/uncrustify.cfg
```diff
@@ -0,0 +1,147 @@
+# run with FLIGHT=some_file.c make uncrustify_flight
+
```
kubark42: @mlyle do you know of a handy online calculator for visualizing the effects of this uncrustify configuration?

mlyle: There's a program called 'universalindentgui' that you can load this into and live indent a source file.

PR #1995, opened 2015-10-20
AQ32 - Add capability to log to OpenLog device.

jihlein:
Since the AQ32 hardware has no SPI flash chip, this flight firmware update adds the capability to send log files to an OpenLog device via a serial port.  The OpenLog device must be set at 115200 baud.

This has been tested on AQ32 hardware, on uart4 and uart6.  Uart3 was not tested due to the fact it was already hooked up to something else.  Log rates of 25, 50, and 100 Hz were tested, and log files from all three rates were successfully parsed using LogConvert.m and Octave-4.0.

The modifications allow for logging to either the OpenLog or SPI Flash if both are available.  I've tested that feature on a Quanton board and it works fine.  Logs written to either the OpenLog or SPI flash can be successfully parsed.  I intend to make another PR with the Quanton files once this one is merged.  No reason why the same mods would not work on Colibri and Sparky2 for that matter.

It also works on F3 targets, I've tested it on my private builds for Naze32Pro and Dodo.


flight/Modules/Logging/logging.c
```diff
@@ -83,9 +83,12 @@ static void writeHeader();
 // Local variables
 static uintptr_t logging_com_id;
 static uint32_t written_bytes;
+static bool destinationSPIflash;
```
tracernz: Underscores are the preferred style. ;)

flight/Modules/Logging/logging.c
```diff
@@ -117,7 +120,15 @@ int32_t LoggingInitialize(void)
 
 	LoggingStatsInitialize();
 	LoggingSettingsInitialize();
-
+	
+	LoggingSettingsLogDestinationOptions logDestination;
```
tracernz: Underscores again.

jihlein:
Old habits....

tracernz:
Yeah, it's a pain that most of the existing stuff is a mixture of camel-case and underscores too.

kubark42:
@tracernz In theory, the only thing that should be camelcase in flight code is the UAVObjects. And I kind of vote to keep those always camelcase, because it highlights their specialness.

mluessi:
It seems like this only creates a single openlog file instead of starting a new file every arm/disarm cycle. Is this correct? If so, i think we should add the corresponding openlog commands to start a new file on arm (if the settings are configured accordingly). We could even use the GPS time (if available) to create a meaningful file name.

jihlein:
The OpenLog v3 light code, which logs the fastest, does not support escape sequence commands, so there is no way to create a filename, other then the roll over on every power cycle.  While dropping back to OpenLog v3 will return the escape sequences, I don't know it will keep up with the data stream at 115200.  

The receive data log buffer is 4 times larger in the light version, the escape sequences were sacrificed to make more memory available.

tracernz:
@kubark42 I agree the distinction for UAVOs is a good idea. We should document this explicitly though because it is otherwise confusing for new developers. I've just had a quick stab at adding this to the coding standards page (and altering the structure of the page a little to make adding more notes easier). Please take a quick look: https://github.com/TauLabs/TauLabs/wiki/Development-Coding-standards :).

jihlein:
Looks and reads well.

shared/uavobjectdefinition/loggingsettings.xml
```diff
@@ -1,6 +1,7 @@
 <xml>
 	<object name="LoggingSettings" singleinstance="true" settings="true">
 		<description>Settings for the logging module</description>
+		<field name="LogDestination" units="" type="enum" options="SPIFlash,OpenLog" elements="1" defaultvalue="OpenLog"/>
```
mluessi: Wouldn't it have been better to leave the default `SPIFlash`? Also, as far as I can tell it crashes when `OpenLog` is set but the port is not setup for this.

jihlein: Any board can log to the OpenLog, but only some boards have the SPI Flash chip, so OpenLog seemed to be the logical choice for a default.

Which board crashes?  I haven't seen that but would like to try and replicate it.

This could be setup such that if a port is assigned to the OpenLog, it automatically determines that it should log to the device, doing away with the LogDestination object.  Looking into that and also removing the destination tie into PIOS_HAL.c.

kubark42: > Any board can log to the OpenLog, but only some boards have the SPI Flash chip, so OpenLog seemed to be the logical choice for a default.

I should have caught this. We have an existing base of users who are used to this working with flash chips. Most TL boards have flash chips, so while there's a certain amount of logic to this stance, it does switch away from a known-good default to one which potentially will cause (and has caused?) issues down the road.

mlyle: There's discussion of this in #2000.

jihlein:
Will address all these issues in upcoming PR........

jihlein:
PR #2006 addresses the issues discussed here.

mlyle:
@jihlein Thanks for putting #2006 together and fixing this up / making this better.

I read the header of this-- do you have a Quanton PIOS_HAL implementation, then?  @dustin wants to do some work in this area (getting MSP working on Quanton).

jihlein:
I don't have a PIOS_HAL implementation for Quanton.  It wouldn't be too hard to put one together for the uart ports, but the rcvr port does not fit the interface well.  Too many receiver port options with ADC pin use, etc.  Need to think about the Quanton receiver port a little more.
Also regarding PIOS_HAL, did you see my comment that PIOS_HAL breaks Frsky telemetry on F3 targets?  Easy to fix with another input argument to the port configure routine, but there are issues going in that direction from what I understand.

Date: Tue, 3 Nov 2015 09:07:47 -0800

mlyle:
Yes, I saw your comment.  I'm still ruminating.  It's problematic to have to specify every inversion possibility at the port level.

mlyle:
Also, I don't really care so much about receiver ports.  OK, I will work with @dustin on getting quanton ported over; I just didn't want to duplicate work.

jihlein:
I can have a Quanton port to PIOS_HAL  ready in a few days if it helps with your workloads.  Just say the word.
-JI

Date: Tue, 3 Nov 2015 09:47:56 -0800

jihlein:
Looked thru some of my old files, I have a Quanton PIOS_HAL port 90% complete FWIW.  Needs a few compile errors fixed and testing.

dustin:
@jihlein That's great news!  Thanks so much for that work.

PR #1999, opened 2015-10-22
Basic MSP support

dustin:
This is enough to make a default MWOSD reasonably happy talking to a naze32.

It allows UART1 to be used, which is important for OSDoge installs where wiring to UART1 is useful, but switching to the USB port of the naze for config is still necessary.  UAVTalk is autodetected and the MSP code will yield and allows UAVTalk to take over.

There's still more that can be done, but this is a pretty decent base and an example board support.


DTFUHF:
This is very cool stuff.

mlyle:
jenkins, add to whitelist

mlyle:
jenkins, test this please

mlyle:
I have collaborated on this some so probably shouldn't be the final reviewer.  But I'll skim this tomorrow.

Also probably will send a PR-to-this-PR to support other PIOS_HAL targets for MSP.

dustin:
Did a quick update.  I'd mixed up some module placement.

dustin:
One more TODO worth mentioning:

* Flight mode mapping

This is particularly rough with MWOSD since flight mode isn't text, but a character, but we can do some basic mapping (which implies arm state).

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -0,0 +1,500 @@
...
+ * Main task routine
+ * @param[in] parameters parameter given by PIOS_Thread_Create()
+ */
+static void uavoMSPBridgeTask(void *parameters)
+{
+	while (1) {
+		uint8_t b = 0;
+		uint16_t count = PIOS_COM_ReceiveBuffer(msp->com, &b, 1, PIOS_QUEUE_TIMEOUT_MAX);
+		if (count) {
+			if (!msp_receive_byte(msp, b)) {
```
mluessi: Is it a good idea to just stop the task? Under what condition does that occur? It seems like the is an automatic UAVTalk detection, does that mean the same port will automatically switch to UAVTalk telemetry when it is connected to it? Does this work?

mluessi:
Otherwise LGTM

DTFUHF:
Dustin thanks for the all the work, this is a very nice feature to have :)

So it looks like if multiple unknown bytes are received, the task kills itself.

Some thoughts:

- How does the the task restart itself? On next reboot?
- Do any other tasks in TauLabs end, at all?
- Would encountering repeated communication errors, or an unsupported MSP version have the same behaviour?
- Is there a better way of sharing a port? Are there any other cases of port sharing in Taulabs? Are there any other targets, or use cases where port sharing is necessary?
- Maybe port sharing could be done at the board level for this target specifically, so the port could be used for other modules/pios as well?

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: Task stops only under the condition that it has detected UAVTalk is incoming.

Stopping the task makes sense to me since it can't properly share the reads (not particularly easy to unread the sequence of bytes it uses for detection) and under any condition where UAVTalk is going to be seen on that port, it's probably not going to be talking MSP again unless I make UAVTalk also detect MSP and switch back, which I don't think is desirable.

This does work, though.  I tested on UART1 on my naze (which is where OSDoge plugs in).  OSD was updating fine, and then I unplugged the OSD and plugged in USB and attached to GCS and was able to do configuration and stuff.  That's the sole use case for switching.

DTFUHF:
MSP seems to be working :) after enabling the module and setting MainPort = MSP.

However, I could not get the naze to connect to GCS again, it needed to be reflashed to restore default settings.

kubark42:
Cool stuff! @dustin, do you think you could also provide some UI support for this? We're developing a bad habit of pushing users into the UAVOBrowser for configuring basic functionality.

kubark42:
> Also probably will send a PR-to-this-PR to support other PIOS_HAL targets for MSP.

Might reduce workload to get #2001 up and running at the same time.

dustin:
The task does not restart itself.  You need to reboot after you do configuration before OSD will work again.

I'm not certain if any other tasks end.  I do know that if you just return from a task, the whole system dies, but calling `PIOS_Thread_Delete` before returning seems to do the right thing.

It's not multiple unknown bytes, but specifically what appears to be a UAVTalk start sequence, i.e. `0x3c 0x2x 0xxx 0x0x`.  That's 16 bits that need to be in the right place over four bytes, which is certainly possible to occur accidentally, but not terribly likely.  One thing I didn't do here that would be easy and help defend against that is to go into `MSP_DISCARD` instead of `MSP_IDLE` on unknown commands.  Not sure why I didn't do that in the first place, but that would mean that once it gets in sync with MSP, it'd have to be seeing something that wasn't MSP before it'd ever attempt to yield.

I don't know if there are any other targets where it's necessary.  It makes the most sense anywhere there's a hardware UART that's used for UAVTalk instead of USB magic and you want to use that UART for an in-flight peripheral.

We could possibly enable port sharing as a PIOS directive.  i.e. `#define PIOS_SHARE_MSP` or similar and use that to decide whether to even check for UAVTalk during `MSP_IDLE`.


So, the two concerns I've seen so far are:

1) Perhaps don't try to detect UAVTalk if the board doesn't share a UART with gcs.
2) Ending tasks makes people nervous.

dustin:
@DTFUHF Oh.  That's actually kind of a big caveat I missed -- you can't connect at 57600 anymore -- have to change GCS to 115200 in order to reconnect.  That's a bit of an annoyance.

@kubark42 The only UI that really needs to be added is a checkbox in the modules list.  I'm not entirely sure how to do that, but I assume it's easy.  Might want to reword the existing thing for mavlink since there would be two options for minimosd now.

DTFUHF:
@dustin OK, thanks, this works. GCS baudrate needs to be changed in Tools -> Options -> Serial Telemetry to 115200.

I noticed also that MSPSpeed has no effect at this time.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -0,0 +1,500 @@
...
+				TaskMonitorRemove(TASKINFO_RUNNING_UAVOMSPBRIDGE);
+				PIOS_Thread_Delete(task);
```
mlyle: On ChibiOS this waits for task to complete, which won't happen.  I really think this should sleep forever instead, since free() is a nop and resources won't be reclaimed.

dustin:
OK, so current gating TODO:

* MSP Speed Setting
* Use MSP_DISCARD for unhandled commands
* Make UAVTalk detection a board-specific flag since it's not needed everywhere.
* Infinite loop instead of exit (though, IMO, exit should be possible, it's still an edge case)

Also, before it's really usable, I should probably implement "boxes," which is kind of orthogonal to the bulk of this work.  Currently, AHI, voltage, current, and throttle position will be displayed correctly on the screen as you fly around in "disarmed" state.

dustin:
OK.  Discussion with @mlyle we decided against making uavtalk detection optional.  I've got a couple updates coming that solve a few of the issues.  MSP_DISCARD will make it a lot more resilient against large commands that have something that looks like UAVTalk causing it to trigger, but that's a bit more work and I'll want to test it.  Incoming isn't tested yet.  Still open:

* MSP_DISCARD for large messages
* Boxes (I think this actually kind of matters)

mlyle:
jenkins, test this please

flight/targets/naze32/fw/pios_board.c
```diff
@@ -352,9 +357,23 @@ void PIOS_Board_Init(void) {
 
 
 	/* UART1 Port */
-#if defined(PIOS_INCLUDE_TELEMETRY) && defined(PIOS_INCLUDE_USART) && defined(PIOS_INCLUDE_COM)
-        PIOS_HAL_ConfigureCom(&pios_usart_main_cfg, PIOS_COM_TELEM_RF_RX_BUF_LEN, PIOS_COM_TELEM_RF_TX_BUF_LEN, &pios_usart_com_driver, &pios_com_telem_serial_id);
+#if (defined(PIOS_INCLUDE_TELEMETRY) || defined(PIOS_INCLUDE_MSP_BRIDGE)) && defined(PIOS_INCLUDE_USART) && defined(PIOS_INCLUDE_COM)
```
mlyle: indentation

mlyle:
Other than the indentation, this looks great to me / ready for merge.

flight/targets/naze32/fw/pios_board.c
dustin: I think I fixed this up.

dustin:
https://www.youtube.com/watch?v=_v3TA73zNkc

dustin:
The only bits that are currently undesirable:

* If your TX is ever not connected, throttle doesn't report properly (MWOSD autoscales and that 0xffff it starts sending isn't normally achievable)
* I've not tested with a proper battery setup.

For the first issue, there are a few ideas, but they get kind of ugly.  Not sending clearly invalid channel data causes throttle to appear stuck.  Synthesizing a value might not be less confusing. Perhaps sending the idle values for all channels would make sense.  That's not guaranteed to report 0%, but it'll be close.  Forcing the value to be 1000 might work, but it also might make it impossible hit 0% again after recovery (which probably matters less).  Besides, blinking RSSI ought to be enough for anybody.

The second issue is the kind of thing I'd like to work out after flying some, which will require some branch merging on current rigs.  @tracernz has done some nice naze ADC work, and I've got a couple of naze+osdoge stacks that could benefit from this.  They both have mag and baro as well.  One has a GPS.   If we can get all that stuff running, we should be able to get proper end-to-end testing of all the important bits.

mlyle:
I think probably the first bug should be fixed before merge.

My favorite solutions in order are - send the channel neutral value from config, send 1000, stop sending throttle.  But I'd merge any of them.

I don't think this should have to wait for ADC support or test of all battery parameters.  "Does something useful" and "no known bad bugs" and "author has an intention to play with ADC support" are plenty for me.

dustin:
Is there a straightforward way to say "RC stuff is invalid?"  Do I just look for out of range values?

mlyle:
Yah, I think the best thing you can do is look for very large values :/  There are specific constants for specific errors but I think it's a mistake to compare against those (what if someone adds a new error?)

dustin:
I've not tried the above patch, but it looks like the thing I understand as what we discussed.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -0,0 +1,603 @@
...
+	// TODO
+}
+
+static void _msp_send_channels(struct msp_bridge *m)
+{
+	ManualControlCommandData manualState;
+	ManualControlCommandGet(&manualState);
+
+	// MSP RC order is Roll/Pitch/Yaw/Throttle/AUX1/AUX2/AUX3/AUX4
+	int channelMap[] = {
```
mlyle: I am embarrassed for the compiler that this is necessary, but static const uint8_t please?

It's putting it on stack:

```
 843 022a 0FCF                  ldmia   r7!, {r0, r1, r2, r3}
 844 022c 05AE                  add     r6, sp, #20
 845 022e 0FC6                  stmia   r6!, {r0, r1, r2, r3}
 846 0230 97E80F00              ldmia   r7, {r0, r1, r2, r3}
 847 0234 86E80F00              stmia   r6, {r0, r1, r2, r3}
```

and using 4 bytes per index of flash:

```
 1032                   .LC0:
 1033 0000 01000000             .word   1
 1034 0004 02000000             .word   2
 1035 0008 03000000             .word   3
 1036 000c 00000000             .word   0
 1037 0010 04000000             .word   4
 1038 0014 06000000             .word   6
 1039 0018 07000000             .word   7
 1040 001c 08000000             .word   8
```

mlyle: Also channel_map instead of channelMap

dustin:
Updated the infinite loop to undo the global storage of the task and rebased tl/next

dustin:
I just tested the stick value thing.  Hits 0% on the OSD before the RSSI even drops fully.  Percentage is still correct after restarting radio.

mlyle:
Congratulations!  Thanks for all your work on this.

PR #2000, opened 2015-10-22
fix the build

mlyle:
Jenkins is broken, so this stuff snuck in

### DIFF
--- a/flight/targets/sparky/fw/pios_board.c
+++ b/flight/targets/sparky/fw/pios_board.c
@@ -380,7 +380,7 @@ void PIOS_Board_Init(void)
 	uint8_t hw_mainport;
 	HwSparkyMainPortGet(&hw_mainport);
 	PIOS_HAL_ConfigurePort(hw_mainport, &pios_main_usart_cfg,
-	                       &pios_usart_com_driver, NULL, NULL, NULL,
+	                       &pios_usart_com_driver, NULL, NULL, NULL, NULL,
 	                       PIOS_LED_ALARM,
 	                       &pios_main_dsm_hsum_cfg, &pios_main_dsm_aux_cfg,
 	                       hw_DSMxMode, NULL, NULL, false);
@@ -391,7 +391,7 @@ void PIOS_Board_Init(void)
 	PIOS_HAL_ConfigurePort(hw_flexiport, &pios_flexi_usart_cfg,
 	                       &pios_usart_com_driver,
 	                       &pios_i2c_flexi_id,
-	                       &pios_i2c_flexi_cfg, NULL,
+	                       &pios_i2c_flexi_cfg, NULL, NULL,
 	                       PIOS_LED_ALARM,
 	                       &pios_flexi_dsm_hsum_cfg, &pios_flexi_dsm_aux_cfg,
 	                       hw_DSMxMode, NULL, NULL, false);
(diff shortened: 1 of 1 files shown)


mluessi:
Somehow this is a disadvantage of having one large `PIOS_HAL_ConfigurePort` function. Whenever something is added, all targets need to be updated, even those that don't use the added feature. Maybe instead of parameters, the information could be passed using a struct (and each target would memset it to all zero before passing it). Like this, parameters could be added without changing the function signature every time. Naturally, this shouldn't be done in this PR (which LGTM).

peabody124:
@mluessi I agree about a configuration structure, similar to what we use for other driver init methods. It also makes it clearer what each value means when you have something like ``` .tx_buf_enable = 0```

mluessi:
@peabody124 yes, `PIOS_HAL_ConfigurePort` currently makes it very easy to use NULL at the wrong location.

mluessi:
PS: Pull request 2000 yay!

kubark42:
jenkins, test this please.

kubark42:
Okie, Jenkins seems to be down still. @lilvinz, can you give it a kick?

In the meantime, merging since this seems to be fixing a breakage in next and only affects one board.

mlyle:
> @mluessi I agree about a configuration structure, similar to what we use for other driver init methods. It also makes it clearer what each value means when you have something like .tx_buf_enable = 0

The downside here is that these configuration structures would need to be in ram on the initialization stack, because many parts are dynamically calculated.

Really PIOS_HAL should not be changing so much.  The most recent was my bad for not finding during review-- the logging settings should be accessed by the actual logging task and the way that works doesn't seem to make sense for the different targets (AQ32 looks at configuration for where log should go for flash vs. openlog, when AQ32 doesn't have flash... but the other targets don't?  The HW target looks at a non-HW specific variable?)  The main thing I want to do is remove some of the sbus, DSM, etc specific configuration that can be calculated at runtime so it's not such an argument-list-of-death.

The previous one was adding a fairly fundamental type (PWM).  It highlights a vulnerability with how we do CI/PR building (something that has been open a week should be re-tested before merge... people who are doing test on PRs should be doing it merged or rebased to next).

jihlein:
(AQ32 looks at configuration for where log should go for flash vs. openlog, when AQ32 doesn't have flash... but the other targets don't? The HW target looks at a non-HW specific variable?)

For the AQ32 target I guess looking at the logging configuration is not 100% necessary, it's just an extra check for consistency.  But on a target that can support both the OpenLog and SPIFlash, the check is necessary so we get the proper value stored in pios_com_logging_id.  That's the reason for adding logging destination to the list of PIOS_HAL_ConfigurePort arguments.  For Quanton to use the OpenLog or SPI Flash, this is necessary (should have that PR ready either tonight or tomorrow).

In this case, is there a better way to do this?  Would it be better to create  pios_com_logging_id_spi_flash and pios_com_logging_id_openlog, and decide which one to use in the logging module based on the destination setting?  Or am I misunderstanding the issue here?

mlyle:
@jihlein I think complicating PIOS_HAL is probably not worth it.  There's separate pressures pushing for the PIOS_HAL argument list to get cut down, and the whole point of PIOS_HAL is to prevent configuration boilerplate from being duplicated between targets.

The logging module could look at the configuration, but I'm not sure that we even need another configuration parameter.  One option would be to use openlog if a port is configured for it; else to use SPI flash if onboard; else to raise a configuration error (or silently not log).

Separate IDs would be one way to accomplish this.  Alternatively, PIOS_COM could be taught to tell us "what kind" of stream this is (USB, real serial, flash?) via an API and logging could decide what to do depending on where it's streaming, and the target could decline to initialize the logging flash stream if there's already one pointing to openlog.

jihlein:
I could probably come up with the necessary changes to use pios_com_logging_spi_flash_id and pios_com_logging_openlog_id and select between the two in the logging module.  This would remove the destination parameter from the PIOS_HAL, and also remove the need to know the logging destination in PIOS_BOARD.c.   Anything beyond that is probably beyond my knowledge/comfort level at this point.  Sound like a plan?

mlyle:
Sounds great.  I really appreciate it.

jihlein:
PR #2006 removes the destination parameter from PIOS_HAL.

PR #2005, opened 2015-10-24
actuator: cut down stack 200 bytes on F1

mlyle:
The actuator refactor freed up a bunch of stack use but used some more
heap.  -fstack-usage confirms 200 less stack used with that refactor.

Necessary for autotune to work again.  Results in additional heap free.  Needs additional testing before merge.

### DIFF
--- a/flight/targets/coptercontrol/fw/pios_config.h
+++ b/flight/targets/coptercontrol/fw/pios_config.h
@@ -103,7 +103,7 @@
 #define CPULOAD_LIMIT_CRITICAL		95
 
 /* Task stack sizes */
-#define PIOS_ACTUATOR_STACK_SIZE        800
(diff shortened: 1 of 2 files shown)


tracernz:
Can't wait to get this memory back. Should be able to do voltage & current ADCs + MSP telemetry with an extra 200 bytes.

peabody124:
@tracernz is this flight tested? would someone mind doing a build with diagnostics enabled and post the stack info remaining?

tracernz:
@peabody124 No I haven't. Should have done it yesterday while I was testing autotune. I should get a chance later next week.

tracernz:
Tried without this PR on Naze, couldn't get stackfree to go below 316 bytes on the bench. With this PR 116 bytes free as expected. It's raining outside so no flight test.

PR #2006, opened 2015-10-24
Flight - OpenLog integration cleanup.

jihlein:
Fixes/streamlines some of the issues with the original pull request.

Dropped the need for the destination UAVObject.  On a board without SPI Flash such as the AQ32, the logging module will only become active if logging is enabled and OpenLog is assigned to one of the available ports.  Otherwise the module does not load.

On a board like Quanton, logging destination will be SPI Flash, unless OpenLog is assigned to one of the available ports, in which case the log data will be sent to the OpenLog.

So the complete change could be tested, I added the Quanton implementation to the PR.

Test Results (results assume logging has already been enabled):

AQ32 - with OpenLog not assigned to a ports, the logging module is inactive.  Once OpenLog is assigned to a port and the board is reset, the logging module is loaded and active.  As per the rest of the setup, logging takes place when the board is armed.  Tests run on uart4 and uart6.

Quanton - with OpenLog not assigned to a port, the logging module is active, and log data is sent to SPI Flash when armed.  The logged file can be downloaded via the GCS.  With OpenLog assigned to a port, after a reset the logging module is still active, and log data is sent to the OpenLog upon system arming.  Tests run on uart2 and uart5.

This implementation is quite a bit easier to setup, and should take care of previous issues/concerns.

Colibri and Sparky2 implementation should be straight forward, but I did not include those boards in this PR because I can not test them.

### DIFF
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -97,9 +97,15 @@ extern uintptr_t streamfs_id;
  */
 int32_t LoggingInitialize(void)
 {
-	// TODO: make selectable
-	logging_com_id = PIOS_COM_LOGGING;
-
+	if (PIOS_COM_OPENLOG) {
+		logging_com_id = PIOS_COM_OPENLOG;
+		destination_spi_flash = false;
+	}
+	else if (PIOS_COM_SPIFLASH) {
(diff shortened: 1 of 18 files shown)


mlyle:
Looks good to me, and no objections.

PR #2007, opened 2015-10-24
Battery: Disallow negative voltage and current

tracernz:
The PIOS_ADC_PinGet function returns -1.0f upon failure (e.g. the pin asked is out of range for the ADC pins configured on the board, since GCS currently allows configuring anything from ADC0-ADC8 regardless of board capabilities). Currently this results in -1.0f being scaled by the battery scaling factor and showing as a negative voltage or current which is obviously undesirable. 

There should be no condition whereby the ADC can measure a negative voltage so a real reading should not be negative. I moved the scaling/offset stuff until after the negative check to cover the case where the user might wish to offset the values into the negative range for whatever reason.

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -130,8 +130,15 @@ static void batteryTask(void * parameters)
 
 		// handle voltage
 		if (voltageADCPin >= 0) {
-			flightBatteryData.Voltage = ((float) PIOS_ADC_GetChannelVolt(voltageADCPin)) / batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_VOLTAGE] * 1000.0f +
-							batterySettings.SensorCalibrationOffset[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONOFFSET_VOLTAGE]; //in Volts
+			flightBatteryData.Voltage = (float)PIOS_ADC_GetChannelVolt(voltageADCPin);
+
+			// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns -1 on error)
+			if(flightBatteryData.Voltage < 0.0f)
+				flightBatteryData.Voltage = 0.0f;
+
+			// scale to actual voltage
+			flightBatteryData.Voltage *= 1000.0f / batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_VOLTAGE];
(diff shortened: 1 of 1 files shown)


flight/Modules/Battery/battery.c
```diff
@@ -146,8 +153,16 @@ static void batteryTask(void * parameters)
 
 		// handle current
 		if (currentADCPin >= 0) {
-			flightBatteryData.Current = ((float) PIOS_ADC_GetChannelVolt(currentADCPin)) / batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_CURRENT] * 1000.0f +
-							batterySettings.SensorCalibrationOffset[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONOFFSET_CURRENT]; //in Amps
+			flightBatteryData.Current = (float)PIOS_ADC_GetChannelVolt(currentADCPin);
+
```
mlyle: Fix this indent-of-death line and i'll merge :)

tracernz: Fixed. ;)

kubark42:
I'm just now seeing this, and hard to review on my phone, but at first glance I don't like what this does. It turns an error into a plausible reading. That seems bad to me. And it makes it hard to track downstream, since a real 0 reading is now indistinguishable from an error reading.

tracernz:
Do you think it would be a good idea to set a configuration error alarm as well? A negative reading does bad things like causing infinite flight time remaining, negative consumed energy, and is obviously bad for telemetry layers which cast the values to unsigned integral types so I definitely want to avoid negative values propagating past the battery module, hence this PR. I am happy to add a configuration alarm in a subsequent PR. :)

mlyle:
It's also good for battery usage to not run backwards / "fill up" on invalid values, which the old code would do.

mlyle:
Note that offsets can produce negative values still, and it's altogether possible to have real negative currents for short times, though.. so the casts to unsigned int are bad.

kubark42:
Agreed with all those points, and in general the notion that we shouldn't allow error values to influence our system. My only beef is that before this PR we had some insight into a failed configuration, but now that is gone. If we have to choose between having the battery management system do wrong things vs. having a distinct error value I would definitely choose to give up the error value. But we can definitely have both, right?

The error value we choose probably isn't important, so long as we can distinguish between good and bad. I would think an error flag is a good way-- maybe sanity check could handle this? Then leaving the error value at 0 isn't critical.

tracernz:
I agree, and thanks for your input. I'll write a subsequent PR to flag an error when the ADC driver returns a negative value.

PR #2008, opened 2015-10-24
MSP updates

dustin:
Bug fix:  roll and pitch axes were both reversed

Board support:   +sparky1

Feature support:  +baro altitude

Also, hardware detection enhancements for more accuracy with less memory overhead.


dustin:
Tested this on sparky.  Orientation seems correct, altitude seems to match uavos.  I don't think there's any more to do here.

mlyle:
Cool! Just waiting for Jenkins, then.

PR #2011, opened 2015-10-24
mpu*/gyro settings: adjust for sanity on CC/naze/sparky

mlyle:
All 3 targets had a default filter cutoff frequency over the
nyquist frequency of the sampling rate.

Also increased the default gyro scale to be +/- 1000 deg/sec on F1
targets.

### DIFF
--- a/shared/uavobjectdefinition/hwcoptercontrol.xml
+++ b/shared/uavobjectdefinition/hwcoptercontrol.xml
@@ -11,10 +11,10 @@
 
 		<field name="DSMxMode" units="mode" type="enum" elements="1" parent="HwShared.DSMxMode" defaultvalue="Autodetect"/>
 
-		<field name="GyroRange" units="deg/s" type="enum" elements="1" options="250,500,1000,2000" defaultvalue="500"/>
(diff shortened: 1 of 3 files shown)


mlyle:
jenkins, test this please

tracernz:
This seems sane. I personally have been setting the filter at 188 Hz anyhow.

kubark42:
What's the reason to change the gyro rate?

mlyle:
@kubark42 The range?   Basically when I've measured, +/-500 deg/sec doesn't seem to reduce absolute noise spectrum much (barely detectable at rest, let alone in flight) vs. the additional gain of +/-1000 deg/sec.  It's just a change to the default, so it doesn't break existing setups.  And it's possible for a transient or motor failure to hit 1000 deg/sec.

It probably should be turned up on all targets.  I've consistently flown +/- 1000 or +/-2000 on all vehicles I've built with no performance degradation.

kubark42:
@mlyle, yes, sorry, I meant range. (And corrected it above). I tend to agree, and I think I read somewhere that APM uses 2000deg/s by default. I've never been able to see the difference in flight between 500deg/s and 2000deg/s. But I'm not a good pilot so I'm not a good judge.

It's probably worth rolling out these kind of upgrades slowly, so I agree with starting on the F1 targets first.

PR #2014, opened 2015-10-24
Sparky: Remove GenericI2CSensor (to free up some flash)

tracernz:
Due to an unfortunate mishap the Sparky build has ended up breaking on next (two PRs were merged which added a small amount of code, and built fine separately in Jenkins, but together they were too much for the tiny amount of space left).


mlyle:
<3

tracernz:
FWIW: The flash overran by 8 bytes. ;)

Issue #2013, opened 2015-10-24
Autotune on Naze32 produces incorrect values


dustin:
There've been a couple other reports of this, but I tried a couple of autotunes with various tweaks on naze32 last night and got unrealistically low taus.

mlyle:
need to determine if this affects CC3D also

tracernz:
Note: I have reports of this prior to #1990. ;) @dustin tested with the gyro rate at both 333 Hz and 500 Hz with the same result.

peabody124:
Could you test it on master (previous release) and see if it worked then. It would be nice to distinguish if it never worked v a regression. Also if anyone else can reproduce it to make sure it isn't a frame/board specific event.

VArcht:
Tested this on CC3D with taulabs_master_20150131, same result 0.00677 tau.

tracernz:
EMAX Nighthawk (My test rig)
Cobra 2204/1960 kv
Rotorgeeks 12A
DAL 5040
Dinogy 3S 1500 65C

Naze32
```
Measured properties
Gain	Bias	Tau(s)	Noise
Roll	9.55547	0.0169047	0.007	293.646
Pitch	9.67937	-0.00273166	0.007	205.503


Tuning aggressiveness
Damping	1.1
Noise sensitivity	1 %
Natural frequency	17.6


Computed Values
RateKp	RateKi	RateKd
Roll	0.00491452	0.0439584	1.93644e-05
Pitch	0.00434183	0.038836	1.71078e-05
Outer Kp	16.3843	-	-
Derivative cutoff	51
```

Sparky
```
Measured properties
Gain	Bias	Tau(s)	Noise
Roll	9.80113	0.0124977	0.0314	116.044
Pitch	9.39549	-0.016818	0.0314	92.09


Tuning aggressiveness
Damping	1.1
Noise sensitivity	1 %
Natural frequency	6.1


Computed Values
RateKp	RateKi	RateKd
Roll	0.00179425	0.00553777	4.14406e-05
Pitch	0.00269185	0.00830813	6.21718e-05
Outer Kp	5.67494	-	-
Derivative cutoff	20
```

The Sparky tune seems to fly well, the Naze tune obviously doesn't.

peabody124:
@tracernz Ok! That is hugely useful and confirms that naze/cc3d is definitely diverging and hitting the lower limit on tau

VArcht:
[...]

PR #2016, opened 2015-10-24
GCS:Autotune: Warning for very low Tau (i.e. diverged), flight controller name posted to forum

tracernz:
Two parts to this one:
* A warning is generated if Tau is less than 0.008 s. The lower limit for Tau is 0.007 s according to @peabody124, and this occurs when the identification diverges, resulting in poor tunes as per #2013.
* The flight controller short name is included in forum posts to aid in identifying future issues. The lack of this information was identified whilst discussing #2013 on IRC.


peabody124:
looks sensible to me. i assume that this is tested?

tracernz:
Yes, I have tested both.

tracernz:
At @VArcht's suggestion I have added a little bit more code to include the firmware version info in autotune forum posts. This should help even more with tracking down issues in future.

(Tune is from a board sitting on the bench :P)
screenshot from 2015-10-25 14 12 06

PR #2018, opened 2015-10-25
mavlink: reduce memory usage by a couple hundred bytes

mlyle:
Whether enabled or not.  Remove a serial-transmit-buffer when enabled
and save an unnecessary copy.  Only allocate the very large message
structure when the task is running.

Needs test/verification still.

### DIFF
--- a/flight/Modules/UAVOMavlinkBridge/UAVOMavlinkBridge.c
+++ b/flight/Modules/UAVOMavlinkBridge/UAVOMavlinkBridge.c
@@ -86,9 +86,7 @@ static bool module_enabled = false;
 
 static uint8_t * stream_ticks;
 
-static mavlink_message_t mavMsg;
-
-static uint8_t * serial_buf;
+static mavlink_message_t *mav_msg;
 
 static void updateSettings();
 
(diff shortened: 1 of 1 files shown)


kubark42:
Awesome! Looking forward to merging this.

flight/Modules/UAVOMavlinkBridge/UAVOMavlinkBridge.c
```diff
@@ -125,7 +123,8 @@ static int32_t uavoMavlinkBridgeInitialize(void) {
 		module_enabled = true;
 		updateSettings();
 
-		serial_buf = PIOS_malloc(MAVLINK_MAX_PACKET_LEN);
+		mav_msg = PIOS_malloc(sizeof(*mav_msg));
```
tracernz: Might be a good idea to ensure that the malloc succeeded (and same for the existing one right below) and disable the module if not?

kubark42: @tracernz good catch. I think in general we throw a PIOS_Assert. It's a really big deal if the malloc fails.

tracernz: For core code like drivers we do an assert, but for optional modules it looks like we often just don't start the module, e.g. UAVOFrSkySensorHubBridge, UAVOHottBridge. I agree with throwing an assert(0) in core code like drivers, but for optional modules it seems reasonable to flag an error and disable the module, allowing the user to correct it without losing all the module and board settings.

kubark42: I see. On the one hand, since allocating memory is fundamental to success, and there's no guarantee that we always check in core modules, being conservative and always locking up when a malloc fails on boot seems a good strategy. On the other hand, I agree that we shouldn't force the user to lose all settings just because of a malloc failure.

tracernz:
Code looks good. Once a check on the malloc results is added I'm happy for it to merge.

peabody124:
bump @mlyle want to do this and we can merge it or do you not want to do that?

PR #2019, opened 2015-10-25
frskysensorhub: save a few bytes of BSS.

mlyle:


flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
```diff
@@ -178,15 +178,15 @@ static const uint8_t frsky_rates[] = {
...
+	uint32_t frsky_port;
 
-static bool module_enabled;
+	uint8_t frame_ticks[MAXSTREAMS];
 
-static uint8_t *frame_ticks;
-
-static uint8_t *serial_buf;
+	uint8_t serial_buf[FRSKY_MAX_PACKET_LEN];
+} *frsky_state;
```
kubark42: Seems very weird to call this struct ```frsky_state```.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
```diff
@@ -256,6 +254,9 @@ static void uavoFrSKYSensorHubBridgeTask(void *parameters)
 	GPSPositionData gpsPosData;
 	BaroAltitudeData baroAltitude;
 	FlightStatusData flightStatus;
+
+	uint8_t *serial_buf = frsky_state->serial_buf;
```
kubark42: Is this just to create a shorthand variable? I don't immediately see what's wrong with ```frsky_state->serial_buf```, it's pretty compact like that. And ```PIOS_COM_SendBuffer(frsky_state->frsky_port, serial_buf, msg_length);``` already includes the use of ```frsky_state``` once so having it in the function a second time won't be awkward.

kubark42:
Nice. I made a couple comments, let me know what you think and I'll merge.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
mlyle: I'm fine with any convention for these.  It's a pattern I'd like to start using for modules instead of globals-- a state structure with a pointer, so these things only consume 4 bytes of BSS instead of a bunch more memory.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
mlyle: Yah, it was to avoid having to change a bunch of the below lines to frsky_state->serial_buf ; this does likely waste 4 bytes of stack space but only when the module is running (probably the fact that frsky_state doesn't change escapes analysis).

There's a lot of frsky pack calls that would need to be updated but I'm willing to change them if you'd like.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
kubark42: I think we've done this from time to time, but it's never perfectly stuck.

Maybe call them-- and I'm only half tongue in cheek-- ```local_global```?

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
kubark42: Ah, I see. If you don't think it's too much of a distraction, I think the code will be cleaner, esp. since all those lines are already being changed in this PR so it won't distract from review.

P.S. Strange to me that the compiler won't optimize out this kind of thing. Place-holder variables seem like some of the easiest low-hanging fruit.

mlyle: There's a bunch of additional lines to change, but I'll change them.  (It's not just the send calls, but all the buffer packing calls).

The optimizer can't do it because it's difficult to know if the frsky_state pointer has changed.  It is scoped to the module, and there is only one place that calls it... but the compiler can't know if when we call into another module if it re-enters that code there and changes frsky_state under us.

flight/Modules/UAVOFrSKYSensorHubBridge/UAVOFrSKYSensorHubBridge.c
mlyle: It's helpful if they have unique names because they're easier to get to with a debugger.  So the intention here is the state information for the frsky module (of course, we have 2.5 frsky modules heh).

I could call it shub_global.  If this sounds good I'll try and get this done before I'm away for a few days.

mlyle:
Updated per comments.  There's a little more diff but not as bad as I thought.

kubark42:
PR looks fine. The one thing I don't like about this is that we're making our code less maintainable in order to save a few bytes of memory. I don't think we should be making these tradeoffs, since RAM will always be getting cheaper whereas the costs of developer maintenance will always be getting higher.

However, until we make an official decision to discontinue support of memory-constrained targets, changes like this are valuable and have their place.

mlyle:
I actually like encapsulating all this stuff in structures for some reasons beyond allowing it to save bss.  1. It makes it obvious you're using a global.  2. It provides namespace protection/avoids shadowing.  3. It makes it really easy/is the first step to transition this stuff in the future to allow multiple instances instead of just one in the module.

PR #2020, opened 2015-10-26
Battery: ADC alarms, disallow negative voltage or current

tracernz:
This adds ADC alarms to the battery module as requested by @kubark42 in #2007. Alarms added are:
* Critical: invalid ADC pin selected for either voltage or current.
* Warning: ADC calibration offset causes negative voltage or current.

These alarms are displayed using the previously unused "ADC In" element in the SystemHealth gadget SVG. The SVG didn't however contain the coloured boxes and error symbol(red cross) so I added these. The linear SVG didn't contain the ADC element so I added it. While editing the linear SVG I also took the liberty of correcting the font on the neighbouring "Altitude Hold" alarm and adding the missing error symbol to it (originally added in 49541da).

In the flight code __negative voltages or currents are clipped to 0__ (after the calibration offset is applied) to ensure we don't encounter undefined behaviour in downstream modules (e.g. telemetry) when they cast the floats to unsigned integers, as mentioned in #2007. The lowest remaining stack space I saw for the Battery task was 60 bytes while trying various combinations of settings, the same as current next without this PR.

The alarms look like this in GCS:
screenshot from 2015-10-26 16 31 01
Note: the above typo, "Error" instead of "Critical", has been fixed.
screenshot from 2015-10-26 16 39 27
screenshot from 2015-10-26 16 52 58

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -128,17 +128,32 @@ static void batteryTask(void * parameters)
 				currentADCPin = -1;
 		}
 
+		bool adc_pin_invalid = false;
+		bool adc_offset_invalid = false;
+
 		// handle voltage
 		if (voltageADCPin >= 0) {
-			flightBatteryData.Voltage = (float)PIOS_ADC_GetChannelVolt(voltageADCPin);
+			float adc_voltage = (float)PIOS_ADC_GetChannelVolt(voltageADCPin);
+			float scaled_voltage = 0.0f;
 
-			// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns -1 on error)
-			if(flightBatteryData.Voltage < 0.0f)
-				flightBatteryData.Voltage = 0.0f;
+			// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns negative on error)
+			if(adc_voltage < 0.0f)
+				adc_pin_invalid = true;
+			else {
+				// scale to actual voltage
+				scaled_voltage = (adc_voltage * 1000.0f
+						/ batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_VOLTAGE])
+						+ batterySettings.SensorCalibrationOffset[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONOFFSET_VOLTAGE]; //in Volts
 
-			// scale to actual voltage
-			flightBatteryData.Voltage *= 1000.0f / batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_VOLTAGE];
-			flightBatteryData.Voltage += batterySettings.SensorCalibrationOffset[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONOFFSET_VOLTAGE]; //in Volts
+				// disallow negative values as these are cast to unsigned integral types
+				// in some telemetry layers
+				if(scaled_voltage < 0.0f) {
+					scaled_voltage = 0.0f;
+					adc_offset_invalid = true;
+				}
+			}
+
+			flightBatteryData.Voltage = scaled_voltage;
 
 			// generate alarms and warnings
 			if (flightBatteryData.Voltage < batterySettings.VoltageThresholds[FLIGHTBATTERYSETTINGS_VOLTAGETHRESHOLDS_ALARM])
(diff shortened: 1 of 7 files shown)


tracernz:
And just looking at the screenshots above I realised I'd made a typo in the alarm HTML, doh. Just force pushed a corrected commit.

flight/Modules/Battery/battery.c
```diff
@@ -128,18 +128,28 @@ static void batteryTask(void * parameters)
...
+		bool adc_offset_invalid = false;
+
 		// handle voltage
 		if (voltageADCPin >= 0) {
 			flightBatteryData.Voltage = (float)PIOS_ADC_GetChannelVolt(voltageADCPin);
 
 			// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns -1 on error)
 			if(flightBatteryData.Voltage < 0.0f)
-				flightBatteryData.Voltage = 0.0f;
+				adc_pin_invalid = true;
```
kubark42: I think it might be cleaner to do

```
// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns -1 on error)
if(flightBatteryData.Voltage < 0.0f) {
	flightBatteryData.Voltage = 0.0f;
	adc_pin_invalid = true;
} else {
	// scale to actual voltage
	flightBatteryData.Voltage *= 1000.0f / batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_VOLTAGE];
	flightBatteryData.Voltage += batterySettings.SensorCalibrationOffset[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONOFFSET_VOLTAGE]; //in Volts
}
```

shared/uavobjectdefinition/systemalarms.xml
```diff
@@ -21,10 +21,11 @@
 				<elementname>FlightTime</elementname>
 				<elementname>I2C</elementname>
 				<elementname>GPS</elementname>
-                		<elementname>AltitudeHold</elementname>
```
kubark42: Bless your kind soul for fixing this whitespace!

flight/Modules/Battery/battery.c
```diff
@@ -157,12 +167,19 @@ static void batteryTask(void * parameters)
 
 			// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns -1 on error)
 			if(flightBatteryData.Current < 0.0f)
-				flightBatteryData.Current = 0.0f;
+				adc_pin_invalid = true;
```
kubark42: Likewise, would be cleaner to check the ```<``` just once, and not to compute the actual current if it's not necessary.

ground/gcs/src/plugins/systemhealth/html/ADC-Critical.html
```diff
@@ -0,0 +1,13 @@
+<html>
+  <head>
+    <title></title>
+    <meta content="">
+    <style></style>
+  </head>
+  <body style="color: black">
+    <h1>ADC: Critical</h1>
+    <p>
+      Invalid ADC channel selected in battery module.
```
kubark42: Is this absolutely the only way the error can get triggered?

kubark42:
Thanks for doing this. It's a valuable upgrade.

flight/Modules/Battery/battery.c
tracernz: The batterySettings.SensorCalibrationFactor can also cause the voltage or current to be negative, so I wanted to check it after the scaling to avoid the situation where we pass on a negative value to downstream layers. The reason I check it before the scaling is to differentiate between an ADC error and the calibration offset causing it to go negative, you are right that the scaling doesn't need to go ahead when there is an ADC error so I will look at cleaning it up. Thanks. :)

shared/uavobjectdefinition/systemalarms.xml
tracernz: ;-)

ground/gcs/src/plugins/systemhealth/html/ADC-Critical.html
tracernz: I shall double-check that. :+1:

flight/Modules/Battery/battery.c
kubark42: Honestly, I think the original code is poor in that it is constantly changing a variable's value throughout the function. IMHO, this code would all be cleaner if temporary variables are used. That would make the checks and their intention much more obvious, and thus maintainable.

tracernz: Hopefully the compiler will be smart enough to not use any more stack if I add a temporary variable, since the UAVO data struct is also in the local stack frame. I want to fit this module on the F1 targets very soon. ;-)

tracernz:
@kubark42 Thanks for your feedback. I've attempted to address your comments and would appreciate if you can take another look. Hopefully the code is a little clearer now.

flight/Modules/Battery/battery.c
```diff
@@ -128,17 +128,32 @@ static void batteryTask(void * parameters)
...
-			flightBatteryData.Voltage += batterySettings.SensorCalibrationOffset[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONOFFSET_VOLTAGE]; //in Volts
+			float adc_voltage = (float)PIOS_ADC_GetChannelVolt(voltageADCPin);
+			float scaled_voltage = 0.0f;
+
+			// A negative result indicates an error (PIOS_ADC_GetChannelVolt returns negative on error)
+			if(adc_voltage < 0.0f)
+				adc_pin_invalid = true;
+			else {
+				// scale to actual voltage
+				scaled_voltage = adc_voltage * 1000.0f
```
kubark42: What would you say about putting a parentheses around this, so it's 
```
... = (adc_voltage * 1000.0f 
   / batterySettings.SensorCalibrationFactor[FLIGHTBATTERYSETTINGS_SENSORCALIBRATIONFACTOR_VOLTAGE])
   +...
```

That might make the code less brittle in case someone moves things around and accidentally messes up the order of operations.

kubark42:
Yup, aside from that one question about the ```()```, looks good to me. And I like how you've used the warning and critical errors to express the difference between a bad pin and a bad calibration value.

tracernz:
Thanks for your review @kubark42. Parentheses added as suggested. ;-)

kubark42:
Awesome!

PR #2022, opened 2015-10-27
GCS: Clean up various errors during startup

tracernz:
As per title this cleans up most of the warnings printed to stderr/stdout during GCS startup. A brief list of changes:
* Added a check for file existence rather than blindly attempting to load a file which doesn't exist in SvgImageProvider
* Add missing scenesize to the pitch scale marker QML object
* Remove a debug print from the PFD widget
* Remove the unused configuration stuff from TelemetryScheduler (it's config was deleted at runtime due to being incomplete, hence the source of the associated error message).
* Convert the colour profile for the autotune icons to prevent libpng warning about invalid profile.
* Correct a typo in the action manager group name for RfmBindWizard
* Set the fallback 3D model in ModelView plugin at init

There remains an SSL error generated by the XmlListModel url accesses in the GithubNewsPanel and ForumNewsPanel QML. SSLv2 was removed from OpenSSL some years ago because it is insecure. [It looks like only TLS1.0 or SSL3 should be used by default](http://doc.qt.io/qt-5/qsslconfiguration.html#defaultConfiguration) but this mustn't apply to QtQuick?

The "Board not found" warning is generated in https://github.com/Taulabs/TauLabs/tree/next/ground/gcs/src/plugins/config/configattitudewidget.cpp#L106 and TBH it doesn't appear to be particularly helpful to detect an invalid board type here? Thoughts on removing the message?

I would appreciate your review @PTDreamer. :-)

Before:
```
mike@mike-desktop:~/Dev/TauLabs/TauLabs$ ./build/ground/gcs/bin/taulabsgcs
m_crashReporterChar:  /home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/bin/crashreporterapp
"gtk+"
Starting the Udev client
*** addAction(): Unknown group:  "Bind Wizard"
UavTalkRelay SERVER listening on  QHostAddress("0.0.0.0") 2000
qt.network.ssl: QSslSocket: cannot resolve SSLv2_client_method
qt.network.ssl: QSslSocket: cannot resolve SSLv2_server_method
Loaded stylesheet: "/home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/share/taulabs/stylesheets" "wide"
"UAVGadgetInstanceManager::createOptionsPages - failed to create options page for configuration TelemetrySchedulerGadget:default, configuration will be removed."
void PfdQmlGadgetWidget::setQmlFile(QString) "/home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/share/taulabs/pfd/default/Pfd.qml"
Cannot open file 'pfd.svg', because: No such file or directory
file:///home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/share/taulabs/pfd/default/SvgElementImage.qml:18: TypeError: Cannot read property 'height' of undefined
file:///home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/share/taulabs/pfd/default/SvgElementImage.qml:17: TypeError: Cannot read property 'width' of undefined
GLC_Context::GLC_Context
GLC_Context::chooseContext
GLC_ContextSharedData::GLC_ContextSharedData()
ModelView: aircraft file not found.
Board not found
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
```

After:
```
mike@mike-desktop:~/Dev/TauLabs/TauLabs$ ./build/ground/gcs/bin/taulabsgcs
m_crashReporterChar:  /home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/bin/crashreporterapp
"gtk+"
Starting the Udev client
UavTalkRelay SERVER listening on  QHostAddress("0.0.0.0") 2000
qt.network.ssl: QSslSocket: cannot resolve SSLv2_client_method
qt.network.ssl: QSslSocket: cannot resolve SSLv2_server_method
Loaded stylesheet: "/home/mike/Dev/TauLabs/TauLabs/build/ground/gcs/share/taulabs/stylesheets" "wide"
GLC_Context::GLC_Context
GLC_Context::chooseContext
GLC_ContextSharedData::GLC_ContextSharedData()
Board not found
```


PTDreamer:
> There remains an SSL error generated by the XmlListModel url accesses in the GithubNewsPanel and ForumNewsPanel QML. SSLv2 was removed from OpenSSL some years ago because it is insecure. It looks like only TLS1.0 or SSL3 should be used by default but this mustn't apply to QtQuick?

According to https://bugreports.qt.io/browse/QTBUG-43173 you should be able to use qputenv("QT_LOGGING_RULES", "qt.network.ssl.warning=false");

> The "Board not found" warning is generated in https://github.com/Taulabs/TauLabs/tree/next/ground/gcs/src/plugins/config/configattitudewidget.cpp#L106 and TBH it doesn't appear to be particularly helpful to detect an invalid board type here? Thoughts on removing the message?

Yeh just remove it please

PTDreamer:
PR looks great, thank you for your work.
Will wait for the "board not found" deletion and merge it.

tracernz:
Thanks very much for your review @PTDreamer ! I've removed the "board not found" message. I will leave the SSL warnings in place for now as there are quite a number of them generated on OSX and I suspect that's why the Github panel on the welcome page doesn't work. Must be something wrong with OpenSSL on OSX...?

PTDreamer:
> Thanks very much for your review @PTDreamer !

No, thank you!



> I will leave the SSL warnings in place for now as there are quite a number of them generated on OSX and I suspect that's why the Github panel on the welcome page doesn't work. Must be something wrong with OpenSSL on OSX...?

The GitHub panel also shows nothing about 50% of the times on mint64
The ssl context can be globally set to use only the "secure" protocols, maybe that is something we can try later.

Merging...

PR #2023, opened 2015-10-28
GCS:Autotune: Don't crash when forum share is attempted while disconnected

tracernz:
Fixes a potential problem introduced by me in #2016 where a null pointer dereference occurs when the forum sharing button is clicked while telemetry is disconnected. This should be improbable but it's worth making sure GCS doesn't crash anyhow and I should have recognised that it is possible when writing the code.


kubark42:
LGTM.

PR #2024, opened 2015-10-28
Changed top level makefile to allow compile with MSVC

PTDreamer:
Gnu make can't be used with MSVC so this change makes the top level makefile use "jom" for the gcs targets.
This also adds a .bash_profileMSVC file which needs to be renamed and used instead of the "normal" .bash_profile.

This needs to be used with Qt with msvc2013 options, and  VS2013 community (free https://www.visualstudio.com/en-us/news/vs2013-community-vs.aspx).

P.S.-This was only tested on a 64bit machine
P.S.2.-VS2015 is now available, changes need to be made to the .bash_profile paths to make that work.
P.S.3-MSVC builds are currently broken, I'm working on a set of fixes.

EDIT: Builds need to use QT_SPEC=win32-msvc2013 (e.g. make gcs QT_SPEC=win32-msvc2013 V=1)
-jX switch is not passed to jom. jom is supposed to use the appropriate number of processes, if one want's to force X just edit the GMAKE exported var on the bash_profile file.

### DIFF
--- a/make/winx86/.bash_profileMSVC
+++ b/make/winx86/.bash_profileMSVC
@@ -1,0 +1,77 @@
+# bashrc for TauLabs development on Windows using MSVC
+
+QT_VERSION_FULL=5.5.0
+QT_VERSION=`echo "$QT_VERSION_FULL" | cut -d '.' -f1-2`
+QT_BASEDIR="/C/Qt"
+
+if [ -d "$QT_BASEDIR/Qt$QT_VERSION_FULL" ] ; then
+  QT_BASEDIR="$QT_BASEDIR/Qt$QT_VERSION_FULL"
+fi
+
+CROSS_COMPILE=false
+if [ $(uname -m)="x86_64" ]; then
+  VS_PLATFORM="amd64"
+  WK_PLATFORM="x64"
+  VC_PLATFORM="/amd64"
+  if [ $CROSS_COMPILE = true ]; then
+    VS_PLATFORM="amd64_x86"
+    WK_PLATFORM="x86"
+    VC_PLATFORM=""
+  fi
+else
+  VS_PLATFORM="x86"
+  WK_PLATFORM="x86"
+  VC_PLATFORM=""
+  if [ $CROSS_COMPILE = true ]; then
+    VS_PLATFORM="x86_amd64"
+    WK_PLATFORM="x64"
+    VC_PLATFORM="/amd64"
+  fi
+fi
+
+QT_MSVCVER="msvc2013_64"
+QT_MINGWVER="mingw492_32"
+
+VS_PATH="/C/Program Files (x86)/Microsoft Visual Studio 12.0"
+VS_INCLUDE_PATH="C:/Program Files (x86)/Microsoft Visual Studio 12.0"
+WK_PATH="/C/Program Files (x86)/Windows Kits/8.1"
+WK_INCLUDE_PATH="C:/Program Files (x86)/Windows Kits/8.1"
+
+test_and_add_to_path ()
+{
+  if which $2 >/dev/null 2>&1 ; then
+    return
+  fi
+
+  if [ -f "$1/$2" ] ; then
+    PATH=$1\:$PATH
+    return
+  fi
+}
+test_and_add_to_path "$QT_BASEDIR/Tools/$QT_MINGWVER/bin" "mingw32-make.exe"
+PATH="$QT_BASEDIR/$QT_VERSION/$QT_MSVCVER/bin"\:$PATH
+PATH="$VS_PATH/VC/BIN/$VS_PLATFORM"\:$PATH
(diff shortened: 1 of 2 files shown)


make/winx86/.bash_profileMSVC
```diff
@@ -0,0 +1,77 @@
...
+if ! which make >/dev/null 2>&1 ; then
+  alias make=mingw32-make
+fi
+
+export INCLUDE="$VS_INCLUDE_PATH/VC/INCLUDE;$VS_INCLUDE_PATH/VC/ATLMFC/INCLUDE;$WK_INCLUDE_PATH/include/shared;$WK_INCLUDE_PATH/include/um;$WK_INCLUDE_PATH/include/winrt;"
+export LIB="$VS_INCLUDE_PATH/VC/LIB$VC_PLATFORM;$VS_INCLUDE_PATH/VC/ATLMFC/LIB$VC_PLATFORM;$WK_INCLUDE_PATH/lib/winv6.3/um/$WK_PLATFORM;"
+export LIBPATH="$VS_INCLUDE_PATH/VC/LIB$VC_PLATFORM;$VS_INCLUDE_PATH/VC/ATLMFC/LIB$VC_PLATFORM;$WK_INCLUDE_PATH/References/CommonConfiguration/Neutral;/C/Program Files (x86)/Microsoft SDKs/Windows/v8.1/ExtensionSDKs/Microsoft.VCLibs/12.0/References/CommonConfiguration/neutral;"
+export MACHINE_TYPE=$(uname -m)
+export VCINSTALLDIR=$VS_INCLUDE_PATH
+export GMAKE="MAKEFLAGS= jom -j3"
```
tracernz: So just to clarify, we will always build with 3 threads unless this line is changed?

tracernz:
I wonder if it would be better to move the decision for the value of GMAKE into the Makefile and introduce an env. var. like USE_MSVC. Then we could build with e.g. `make gcs MSVC=yes` and then in the Makefile it would do something like
```Makefile
JOB_NUM = some magic to find j parameter?

ifeq($(WINDOWS),1)
    ifeq($(USE_MSVC),yes)
        GMAKE = MAKEFLAGS= jom -j$(JOB_NUM)
        QT_SPEC = win32-msvc2013
    else
        GMAKE = $(MAKE) --no-print-directory -w 
    endif # WINDOWS
endif # USE_MSVC
```

[There are ways to get the j parameter](http://blog.jgc.org/2015/03/gnu-make-insanity-finding-value-of-j.html), but it's debatable if they're a good idea or not.. :/

tracernz:
And also, if we decide to go with my suggestion above, we would need to look at merging the bash_profile into the existing one.

make/winx86/.bash_profileMSVC
PTDreamer: The ground targets yes

PTDreamer:
> I wonder if it would be better to move the decision for the value of GMAKE into the Makefile and introduce an env. var. like USE_MSVC. Then we could build with e.g. make gcs MSVC=yes and then in the Makefile it would do something like

It is impossible pass the "j" value like that because it is make itself which is responsible for controlling sub makes in terms of parallelism. Imagine we issue make on the toplevel makefile limiting the number of parallel builds to N. Sub make can't use N, it will only be able to use N-x being x other targets or sub makes.

Moving GMAKE var to the makefile is a possibility but IMO has no obvious advantage since you will always require two different bash profiles. The two required PATHs are simply impossible to reconcile.

mlyle:
> It is impossible pass the "j" value like that because it is make itself which is responsible for controlling sub makes in terms of parallelism. Imagine we issue make on the toplevel makefile limiting the number of parallel builds to N. Sub make can't use N, it will only be able to use N-x being x other targets or sub makes.

Both gmake and jom have a jobserver where they coordinate subprocesses between them.  The value passed on the command line with -j and the number of "slots" from the toplevel make are both maximums  (so you can use -j1 to limit subprocesses to not use parallelism, and the jobs they *do* use will be drawn from the top-level concurrency pool).

Unfortunately gmake and jom-style concurrency pools don't cooperate, though.

PTDreamer:
> Both gmake and jom have a jobserver where they coordinate subprocesses between them. The value passed on the command line with -j and the number of "slots" from the toplevel make are both maximums (so you can use -j1 to limit subprocesses to not use parallelism, and the jobs they do use will be drawn from the top-level concurrency pool).
> 
> Unfortunately gmake and jom-style concurrency pools don't cooperate, though.

Sure that is why you can't force --jobs on a submake, it needs to be the parent make to handle that.
If you do force it you are creating a new jobserver instance using a new pipe.

PTDreamer:
Just so I know we are on the same page.
The problem here is that the content of MAKEFLAGS is not properly handled by jom which ends up quitting because of some of the option passed, hence why it is deleted on the code above.
The discussion here is the extraction of the max number of jobs as passed from the command line and use it on a sub target. This IMO breaks the parallelism mechanics of make and shouldn't be used.

Issue #2003, opened 2015-10-23
OpenOCD configs are stale


kubark42:
OpenOCD 0.9.0 no longer works with the in-project config files. This is true for both the bootloader and the debug file. Fortunately, everything we need is already in latest openocd. ```stm32f4discovery.cfg``` for debugging and ```stm32f4x_stlink.cfg``` for the flashing tool. I'm happy to say that in both cases I am able to flash and debug with 0.9.0.

Apparently our current approach is Not Good(TM). From one of the OpenOCD maintainers:

> PaulFertser: I hope the OpenOCD project manages to balance between backwards compatibility and going forward.
> [4:25pm] PaulFertser: Some things just need to be broken. And popular adapters and targets just need to be upstreamed, then it would be maintained automatically.
> [4:25pm] PaulFertser: So using out-of-tree configs is sort of asking for trouble.

His point is that properly installed openocd already knows where to find its config files and we shouldn't try to set up the flashing operations ourselves. For the flashing tool, we need to "source [find target/stm32f4x_stlink.cfg]". For the debugging tool, I'm less sure how this should happen.

In both cases, I'm not familiar enough with OpenOCD to guess at what level these changes should occur. In my experimentations, I can replace all of stm32f4xx.stlink.cfg with 
```
#
# STM32f4x stlink pseudo target
#

source [find target/stm32f4x_stlink.cfg]
```

and flashing works. I don't know how this change should happen for the debugger side, though.

@solidgoldbomb, @lilvinz this is something you know more about than I think anyone else here. Got a second to comment?

tracernz:
Personally I just run `opened -f interface/stlink-v2.cfg -f target/stm32f4x.cfg`, no in-tree config files required, and you can just change f4x for f3x or f1x as required. `target/stm32f4x_stlink.cfg` is deprecated so probably best we don't use that one.

paulfertser:
There's absolutely no difference config-wise between flashing and debugging. _stlink configs are deprecated but if you really need something that would work with 0.7.0 as well as with 0.9.0, that's probably a reasonable option.

Another option is to direct the users towards OpenOCD README and let them run it however they want, e.g. some people will use other JTAG/SWD adapters etc.

If your firmware might go to sleep mode or remap SWD pins, you can physically connect the SRST line and use "reset_config srst_only connect_assert_srst" to be able to attach to the target no matter what.

PS: I'm not an OpenOCD maintainer.

PR #2027, opened 2015-10-30
[WIP] OpenOCD: Upgrade OpenOCD to 0.9.0, refresh configs

tracernz:
Addresses #2003.
I wanted to get this up for comment, especially regarding:
- Removing the OpenPilot JTAG programmer configs
- The FreeRTOS CM4F FPU patch, since we're using ChibiOS for most(all?) F4 targets


flight/targets/aq32/board-info/board-info.mk
```diff
@@ -9,8 +9,8 @@ BOARD               := STM32F4xx_AQ32
 MODEL               := HD
 MODEL_SUFFIX        := 
 
-OPENOCD_JTAG_CONFIG := stlink-v2-norst.cfg
-OPENOCD_CONFIG      := stm32f4xx.stlink.cfg
+OPENOCD_JTAG_CONFIG ?= stlink-v2.cfg
```
mluessi, line 12: Good idea to make it possible to set this externally. I noticed that some of the newer nucleo boards need `stlink-v2-1.cfg`.

mluessi:
Awesome :). Tested on Linux and an F4 target, works perfectly.

One thing: I wonder if it would be better to have the `_debug` target not download the firmware. I often find myself starting debug sessions repeatedly using the same firmware version, so downloading it each time will make it take much longer. Maybe there is some make magic that we can use so the `_debug` target knows the firmware has already been downloaded?

make/tools.mk
```diff
@@ -260,11 +260,11 @@ OPENOCD_BUILD_DIR := $(DL_DIR)/openocd-build
...
-openocd_install: OPENOCD_REV     := cf1418e9a85013bbf8dbcc2d2e9985695993d9f4
+openocd_install: OPENOCD_TAG     := v0.9.0
 openocd_install: OPENOCD_OPTIONS := --enable-maintainer-mode --prefix="$(OPENOCD_DIR)" --enable-buspirate --enable-stlink
 
 ifeq ($(OPENOCD_FTDI), yes)
-openocd_install: OPENOCD_OPTIONS := $(OPENOCD_OPTIONS) --enable-ft2232_libftdi
+openocd_install: OPENOCD_OPTIONS := $(OPENOCD_OPTIONS) --enable-ftdi
 endif
 
 ifeq ($(UNAME), Darwin)
```
kubark42, line 261: What do you think about grabbing this from `brew` instead of compiling locally?

P.S. Running `make openocd_install` fails, but getting it from `brew` works fine.

tracernz, line 261: I haven't tested on my Mac yet so it may need a little work. Getting it from brew sounds reasonable, ~~although we'd need to add that as a toolchain pre-req~~. I like the idea of using brew in general though. :-D

tracernz:
@mluessi I'll see what I can come up with re:the debug loading firmware. :-)

tracernz:
@kubark42 I fixed the OSX build so it should be okay now. I also tried out using Homebrew instead. I installed using the script with `OOCD_INSTALL_DIR=/Users/mike/dev/TauLabs/tools/openocd brew install tau-open-ocd.rb` and it worked so that's promising. The advantage would be that it manages dependencies automatically which is nice.
tau-open-ocd.rb (a hacked version of the stock open-ocd.rb)
```ruby
class TauOpenOcd < Formula
  desc "On-chip debugging, in-system programming and boundary-scan testing"
  homepage "http://sourceforge.net/projects/openocd/"
  url "https://downloads.sourceforge.net/project/openocd/openocd/0.9.0/openocd-0.9.0.tar.bz2"
  sha256 "837042ac9a156b9363cbffa1fcdaf463bfb83a49331addf52e63119642b5f443"

  option "without-hidapi", "Disable building support for devices using HIDAPI (CMSIS-DAP)"
  option "without-libftdi", "Disable building support for libftdi-based drivers (USB-Blaster, ASIX Presto, OpenJTAG)"
  option "without-libusb",  "Disable building support for all other USB adapters"

  depends_on "pkg-config" => :build
  depends_on "libusb" => :recommended
  # some drivers are still not converted to libusb-1.0
  depends_on "libusb-compat" if build.with? "libusb"
  depends_on "libftdi" => :recommended
  depends_on "hidapi" => :recommended

  def install
    # all the libusb and hidapi-based drivers are auto-enabled when
    # the corresponding libraries are present in the system
    install_dir = ENV["OOCD_INSTALL_DIR"]
    args = %W[
      --disable-dependency-tracking
      --prefix=#{install_dir}
      --enable-dummy
      --enable-buspirate
      --enable-jtag_vpi
      --enable-remote-bitbang
    ]

    if build.with? "libftdi"
      args << "--enable-usb_blaster_libftdi"
      args << "--enable-presto_libftdi"
      args << "--enable-openjtag_ftdi"
      args << "--enable-legacy-ft2232_libftdi"
    end

    ENV["CCACHE"] = "none"

    system "./bootstrap", "nosubmodule" if build.head?
    system "./configure", *args
    system "make", "install"
  end
end
```

mluessi:
+1 for merge

kubark42:
@tracernz, what's the status of the ```flight/Project/OpenOCD/0005-clang-fix-build.patch``` vs just using Homebrew? If we need to compile our own openocd to get good ChibiOS support, shouldn't that patch be included in the brew recipe?

tracernz:
@kubark42 Homebrew works okay without that patch but I haven't had time to dig into it and see why yet. We don't currently have a patch to bring floating point support to ChibiOS debugging but we could include it in the brew script. My PC etc. are on a moving truck right now but I can do some more work next week.

tracernz:
I tried to do the Windows side of this but ran into some problems, looks like the newest msys git is shell is broken on x64 machines: https://hub.spigotmc.org/jira/browse/BUILDTOOLS-179 http://sourceforge.net/p/msys2/tickets/208/

tracernz:
The Windows toolchain setup we recommend doesn't include many of the pre-reqs required to compile OpenOCD, and it's non-trivial to either fetch them in our Makefile, or ask the user to install them. I'd suggest just hosting the binary from http://www.freddiechopin.info/download/category/4-openocd (in zip format rather than 7-zip, since that would be another pre-req) and downloading/extracting it with our Makefile. Ideas?

kubark42:
@tracernz Hosting it sounds like a good idea. 

Tested and works\* on OSX. 

[*] I can't get the bootloaders to flash onto a FlyingF3, DiscoveryF4, or our in-house F4 target. Have you been able to do a `bl_XXX_program` or `ef_XXX_program`?

tracernz:
I shall test ef and bl, and finish off the PR. Perhaps the bl and ef targets need some additional work. Thanks for testing!

kubark42:
@tracernz Any chance to test this yet?

PR #2028, opened 2015-10-30
Make: UAVO library fixups

tracernz:
This addresses some of the issues created by #1985. Specifically:
* ar output is suppressed
* UAVO libs are cleaned by all_fw_clean and uavobjects_clean

I think it makes sense to clean them with all_fw_clean since the fw targets cause them to be generated but also by uavobjects_clean since they are after all UAVOs. Thoughts?


kubark42:
> I think it makes sense to clean them with all_fw_clean since the fw targets cause them to be generated but also by uavobjects_clean since they are after all UAVOs. Thoughts?

Agreed. I see very little downside to this.

mlyle:
Thanks for the cleanup!

PR #2029, opened 2015-10-30
GCS: Fix more warnings, print warnings and above to stderr

tracernz:
This is a small collection of things not big enough on their own for seperate PRs so apologies for that.

I ran GCS for a while with a debugger connected and tried various things with QT_FATAL_WARNINGS enabled to trap anything generating warnings at runtime. This PR addresses everything I found as well as a typo in the setup wizard (/s/TauLaps/Tau Labs/).

Additionally it prints messages of warning severity or above to stderr. Once the debug plugin was loaded warning messages were redirected to the gadget and not shown in stderr. If debugging with QT_FATAL_WARNINGS enabled this makes it impossible to see the message. Additionally this could provide useful output to the terminal for bug reports.

Finally it fixes up a couple of things I missed while editing the linear system health widget SVG to add ADC in #2020, e.g. the disconnected overlay not being resized:
screenshot from 2015-10-30 22 00 29


mluessi:
LGTM

ground/gcs/src/plugins/telemetryscheduler/telemetryschedulergadgetwidget.cpp
```diff
@@ -1089,7 +1089,6 @@ void QFrozenTableViewWithCopyPaste::init()
                                    "color: #FF0000;"
                                    "background-color: #8EDE21;"
                                    "selection-background-color: #999}"); //for demo purposes
-    frozenTableView->setSelectionModel(selectionModel());
```
kubark42: Was the telemetry scheduler tested after this change? Do you recall what was the warning Qt was giving you?

tracernz: "setSelectionModel() failed: Trying to set a selection model, which works on a different model than the view". This happens because it is trying to take the selection model from the parent object and set the child object (frozenTableView) but the parent object doesn't share the model ([set on the child object right above](https://github.com/tracernz/TauLabs/blob/6c0975773631ef8aa7299c2b59b377fd8ad171b3/ground/gcs/src/plugins/telemetryscheduler/telemetryschedulergadgetwidget.cpp#L1080)). I'm fairly certain this doesn't make sense and it was failing anyhow?

tracernz: PS. It appears to work fine. ;-)
untitled

PR #2030, opened 2015-10-30
Throttled UAVO Events

mluessi:
As discussed in #1928, this add the option to add a minimum interval to callbacks and queue connections. This should be useful for things like logging or telemetry where we are only interested in updates at a slow rate. Handling this in the UAVO manager should be more efficient than discarding the unwanted events in the callback.

Credit goes to @mlyle for [figuring out a way](https://github.com/TauLabs/TauLabs/pull/1928#discussion_r41660621) to implement this without needing more memory for callbacks without throttling.

### DIFF
--- a/flight/Libraries/math/misc_math.c
+++ b/flight/Libraries/math/misc_math.c
@@ -292,6 +292,23 @@ float linear_interpolate(float const input, float const * curve, uint8_t num_poi
 }
 
 
+/**
+ * Return a psedorandom integer from 0 to interval
+ * Based on the Park-Miller-Carta Pseudo-Random Number Generator
+ * http://www.firstpr.com.au/dsp/rand31/
+ */
+uint16_t randomize_int(uint16_t interval)
+{
+	static uint32_t seed = 1;
+	uint32_t hi, lo;
+	lo = 16807 * (seed & 0xFFFF);
+	hi = 16807 * (seed >> 16);
+	lo += (hi & 0x7FFF) << 16;
+	lo += hi >> 15;
+	if (lo > 0x7FFFFFFF) lo -= 0x7FFFFFFF;
+	seed = lo;
+	return (uint16_t)( ((float)interval * (float)lo) / (float)0x7FFFFFFF );
+}
 
 /**
  * @}
--- a/flight/UAVObjects/inc/uavobjectmanager.h
+++ b/flight/UAVObjects/inc/uavobjectmanager.h
@@ -92,7 +92,7 @@ typedef enum {
 	EV_UPDATED_MANUAL = 0x04, /** Object update event manually generated */
 	EV_UPDATED_PERIODIC = 0x08, /** Object update from periodic event */
 	EV_UPDATE_REQ = 0x10, /** Request to update object data */
-	EV_UPDATED_THROTTLED_DIRTY = 0x80 /** Indicates a throttled object has been updated but not sent **/
+	EV_UPDATED_THROTTLED_DIRTY = 0x20 /** Indicates a throttled object has been updated but not sent **/
 } UAVObjEventType;
 
 /**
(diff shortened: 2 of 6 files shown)


flight/Libraries/math/misc_math.c
```diff
@@ -292,6 +292,23 @@ float linear_interpolate(float const input, float const * curve, uint8_t num_poi
 }
 
 
+/**
+ * Return a psedorandom integer from 0 to interval
+ * Based on the Park-Miller-Carta Pseudo-Random Number Generator
+ * http://www.firstpr.com.au/dsp/rand31/
+ */
+uint16_t randomize_int(uint16_t interval)
```
kubark42: We should unit test this.

:innocent:

flight/UAVObjects/inc/uavobjectmanager.h
```diff
@@ -92,7 +92,7 @@ typedef enum {
 	EV_UPDATED_MANUAL = 0x04, /** Object update event manually generated */
 	EV_UPDATED_PERIODIC = 0x08, /** Object update from periodic event */
 	EV_UPDATE_REQ = 0x10, /** Request to update object data */
-	EV_UPDATED_THROTTLED_DIRTY = 0x80 /** Indicates a throttled object has been updated but not sent **/
+	EV_UPDATED_THROTTLED_DIRTY = 0x20 /** Indicates a throttled object has been updated but not sent **/
```
kubark42: I think the reason we chose this to be ```0x80``` was to give space for other commands to grow in the space, and leave the "state" flag as the MSB. I have no objections to it moving, though.

flight/UAVObjects/inc/uavobjectmanager.h
```diff
@@ -199,7 +199,9 @@ void UAVObjSetTelemetryGcsUpdateMode(UAVObjMetadata* dataOut, UAVObjUpdateMode v
 int8_t UAVObjReadOnly(UAVObjHandle obj);
 int32_t UAVObjConnectQueue(UAVObjHandle obj_handle, struct pios_queue *queue, uint8_t eventMask);
 int32_t UAVObjDisconnectQueue(UAVObjHandle obj_handle, struct pios_queue *queue);
+int32_t UAVObjConnectQueueThrottled(UAVObjHandle obj_handle, struct pios_queue *queue, uint8_t eventMask, uint16_t interval);
 int32_t UAVObjConnectCallback(UAVObjHandle obj_handle, UAVObjEventCallback cb, uint8_t eventMask);
+int32_t UAVObjConnectCallbackThrottled(UAVObjHandle obj_handle, UAVObjEventCallback cb, uint8_t eventMask, uint16_t interval);
 int32_t UAVObjDisconnectCallback(UAVObjHandle obj_handle, UAVObjEventCallback cb);
```
kubark42: Confirmed with @mluessi that we don't need additional disconnects to match the new connnects.

flight/UAVObjects/uavobjectmanager.c
```diff
@@ -1651,6 +1676,19 @@ static int32_t sendEvent(struct UAVOBase * obj, uint16_t instId,
 	LL_FOREACH(obj->next_event, event) {
 		if (event->eventMask == 0
 			|| (event->eventMask & triggered_event) != 0) {
+			if (event->hasThrottle) {
```
kubark42: Might be nice to have a comment here describing what we expect from the if loop.

flight/UAVObjects/uavobjectmanager.c
```diff
@@ -54,10 +55,18 @@ typedef void* InstanceHandle;
...
 	struct pios_queue         *queue;
 	UAVObjEventCallback       cb;
-	uint8_t                   eventMask;
+	uint8_t                   hasThrottle : 1;
+	uint8_t                   eventMask : 7;
 	struct ObjectEventEntry * next;
 };
 
+struct ObjectEventEntryThrottled {
+	struct ObjectEventEntry   entry;
```
tracernz: Should add a comment that this member MUST be the first or everything will break.

mlyle:
I like this. <3  Happy to merge once other feedback is addressed.

flight/UAVObjects/inc/uavobjectmanager.h
mluessi: Changed it to `0x40`, so we have one bit still free :)

mluessi:
Addressed comments

mlyle:
Thanks very much for doing this-- I want to refactor telemetry and it'll be a big help.

PR #2031, opened 2015-10-30
Logging Module Improvements V2 (WIP)

mluessi:
Based on the discussion of the previous PR #1928, this now uses a single queue and throttled callbacks introduced in #2030. 

However, as noted in #1928, the queue doesn't actually store the UAVO values and can't be used as a buffer. To make this work properly, I propose to add a RAM ring buffer and the callback directly calls `UAVTalkSendObjectTimestamped` with an UAVTalk connection that simply writes the data into the ring buffer (if there is sufficient space). The logging task would then retrieve the data from the ring buffer and write it to the flash file system (or serial port). This would have the advantage that the callback would never block and the timestamp would be created at the time the callback occurs.

### DIFF
--- a/flight/Modules/Logging/logging.c
+++ b/flight/Modules/Logging/logging.c
@@ -30,36 +30,27 @@
 #include "openpilot.h"
 #include "modulesettings.h"
 #include "pios_thread.h"
+#include "pios_queue.h"
+#include "pios_mutex.h"
+#include "uavobjectmanager.h"
+#include "misc_math.h"
 #include "timeutils.h"
 #include "uavobjectmanager.h"
 
 #include "pios_streamfs.h"
 #include <pios_board_info.h>
 
-#include "airspeedactual.h"
-#include "attitudeactual.h"
-#include "accels.h"
-#include "actuatorcommand.h"
 #include "flightstatus.h"
-#include "gyros.h"
-#include "baroaltitude.h"
-#include "flightstatus.h"
-#include "gpsposition.h"
-#include "gpstime.h"
-#include "gpssatellites.h"
-#include "magnetometer.h"
-#include "manualcontrolcommand.h"
-#include "positionactual.h"
 #include "loggingsettings.h"
 #include "loggingstats.h"
-#include "velocityactual.h"
-#include "waypoint.h"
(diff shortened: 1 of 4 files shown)


mlyle:
jenkins, test this please

mluessi:
I came up with the following way to implement this: Use two buffers, about 1kB or so each. The callback writes the UAVO data to one buffer using a mutex to protect the buffer from multiple tasks writing to it at the same time. It keeps track of the buffer fill status and when the buffer is more than "1kb - max(UAVO size)" full, it switches to the other buffer and informs the logging task (using a semaphore) that the previous buffer can be written to flash / serial port.

This method has the advantage that no copies are needed, i.e., the logging task doesn't have to copy the data into another buffer when writing. Writing to flash also never blocks the callback from writing to the buffer. Naturally, there are some more things to take care of. E.g. when the buffer switch occurs, it is possible that the task hasn't finished writing the previous buffer. In this case, the callback has to discard objects until the buffer can be switched (this will only happen if the logging rate is too high for the storage method being used). Writing larger chunks of data may also allow us to add some optimizations to the flashfs to write whole pages, which can be faster.

Thoughts?

mlyle:
I think this should merge as-is.  there's clever paged circular buffer things we could do, etc, if we want to take that approach... and i'm freein' up enough ram in other things I don't feel so bad about it.  But i'm concerned about this rotting or accumulating conflict with other diff.

lolorc:
+1, looking forward to test this

mluessi:
It definitely needs testing. I did some bench tests, but this version has received far less testing than the previous more complex version. Also, one thing I was thinking of doing is updating / registering the callbacks when logging starts instead of at boot, this would allow changing the logging period without a reboot.

lolorc:
I fly tested with quanton, I've got some records on my external logger I need to analyze. especially one crash with the quad going mental when enabling AH.

peabody124:
I'll likely test this over the weekend and if it works merge if no objections.

mlyle:
Note that there was significant further work on this and bugs fixed before it merged @ dR.

peabody124:
Tested and it works nicely.

PR #2033, opened 2015-10-31
MSP Updates: current, heading, and stick normalization

dustin:
The MSP stuff was previously mostly bench tested.  There were a few things that I noticed when I actually tried flying it for the first time as well as having tried an s.bus rx for the first time in quite a while.

* current values are off by an order of magnitude with MWOSD configured as baseflight.  For multiwii, it'd probably be right, but more resolution seems better
* Heading was not implemented.  Turns out it was rather trivial.
* s.bus stick values sent directly to MWOSD makes all the values unusable in the OSD.  Need to change how sticks are sent.

### DIFF
--- a/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
+++ b/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
@@ -35,6 +35,7 @@
 #include "flightbatterystate.h"
 #include "gpsposition.h"
 #include "manualcontrolcommand.h"
+#include "manualcontrolsettings.h"
 #include "attitudeactual.h"
 #include "airspeedactual.h"
 #include "actuatorsettings.h"
@@ -125,6 +126,7 @@ typedef enum {
 } msp_state;
 
 struct msp_bridge {
+	ManualControlSettingsData cmdSettings;
 	uintptr_t com;
 
 	msp_state _state;
@@ -146,6 +148,7 @@ struct msp_bridge {
 #define TASK_PRIORITY               PIOS_THREAD_PRIO_LOW
 
 static bool module_enabled;
+static volatile bool cmd_settings_updated = true;
 extern uintptr_t pios_com_msp_id;
 static struct msp_bridge *msp;
 static int32_t uavoMSPBridgeInitialize(void);
@@ -335,42 +338,54 @@ static void _msp_send_altitude(struct msp_bridge *m)
 	msp_send(m, MSP_ALTITUDE, data.buf, sizeof(data));
 }
 
+static uint16_t _msp_scale_rc(struct msp_bridge *m,
+			      ManualControlCommandData manualState,
+			      int channel_num)
+{
+	int val = manualState.Channel[channel_num];
+	int min = m->cmdSettings.ChannelMin[channel_num];
+	int max = m->cmdSettings.ChannelMax[channel_num];
+	int idle = m->cmdSettings.ChannelNeutral[channel_num];
+
+	// Min and max may be reversed.
+	if (min > max) {
+		int tmp = min;
+		min = max;
+		max = tmp;
+	}
+	// If the value is outside of the given range, return the
+	// neutral value.
+	if (val < min || val > max) {
+		return idle;
+	}
+
+	return (((val - min) * 1000) / (max - min)) + 1000;
+}
+
+// MSP RC order is Roll/Pitch/Yaw/Throttle/AUX1/AUX2/AUX3/AUX4
 static void _msp_send_channels(struct msp_bridge *m)
 {
 	ManualControlCommandData manualState;
 	ManualControlCommandGet(&manualState);
-
-	// MSP RC order is Roll/Pitch/Yaw/Throttle/AUX1/AUX2/AUX3/AUX4
-	static const uint8_t channel_map[] = {
-		MANUALCONTROLCOMMAND_CHANNEL_ROLL,
-		MANUALCONTROLCOMMAND_CHANNEL_PITCH,
-		MANUALCONTROLCOMMAND_CHANNEL_YAW,
-		MANUALCONTROLCOMMAND_CHANNEL_THROTTLE,
-		MANUALCONTROLCOMMAND_CHANNEL_FLIGHTMODE,
-		MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY0,
-		MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY1,
-		MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY2,
-	};
+	if (cmd_settings_updated) {
+		cmd_settings_updated = false;
+		ManualControlSettingsGet(&m->cmdSettings);
+	}
 
 	union {
 		uint8_t buf[0];
 		uint16_t channels[8];
-	} data;
-
-	int throttle = manualState.Channel[MANUALCONTROLCOMMAND_CHANNEL_THROTTLE];
-	if (throttle > 500 && throttle < 3000) {
-		// Normal stuff.
-		for (int i = 0; i < 8; i++) {
-			data.channels[i] = manualState.Channel[channel_map[i]];
-		}
-	} else {
-		// Out of bound values indicate rc is not connected.
-		ActuatorSettingsData actuatorSettings;
-		ActuatorSettingsGet(&actuatorSettings);
-		for (int i = 0; i < 8; i++) {
-			data.channels[i] = actuatorSettings.ChannelNeutral[channel_map[i]];
-		}
-	}
+	} data = {
+		.channels = {
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_ROLL),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_PITCH),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_YAW),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_THROTTLE),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_FLIGHTMODE),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY0),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY1),
+			_msp_scale_rc(m, manualState, MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY2),
+		}};
 
 	msp_send(m, MSP_RC, data.buf, sizeof(data));
 }
@@ -542,6 +557,10 @@ static void setMSPSpeed(struct msp_bridge *m)
 	}
 }
 
+static void _msp_cmd_settings_update_cb(UAVObjEvent * ev)
+{
+	cmd_settings_updated = true;
+}
 
 /**
  * Module initialization routine
@@ -562,6 +581,8 @@ static int32_t uavoMSPBridgeInitialize(void)
 			msp->com = pios_com_msp_id;
 
 			setMSPSpeed(msp);
+			ManualControlSettingsConnectCallback(_msp_cmd_settings_update_cb);
+
 			module_enabled = true;
 			return 0;
 		}
--- a/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
+++ b/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
@@ -216,9 +216,11 @@ static void _msp_send_attitude(struct msp_bridge *m)
 
 	AttitudeActualGet(&attActual);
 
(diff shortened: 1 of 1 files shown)


tracernz:
Also noticed this when I had it connected to Baseflight Configurator. LGTM.

dustin:
I tested this on the bench, and then flew with it last night.

dustin:
I added the work I did on heading and the stick updates.  I've not had a chance to test these yet on a real FC (except the current one).

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -333,42 +333,50 @@ static void _msp_send_altitude(struct msp_bridge *m)
...
+{
+	uint16_t val = manualState.Channel[channel_num];
+	uint16_t min = actuatorSettings.ChannelMin[channel_num];
+	uint16_t max = actuatorSettings.ChannelMax[channel_num];
+	uint16_t idle = actuatorSettings.ChannelNeutral[channel_num];
+
+	// Min and max may be reversed.  If the value is outside of
+	// the given range, return the neutral value.
+	uint16_t lower = min < max ? min : max;
+	uint16_t higher = max > min ? max : min;
```
mluessi: I would use if/else, as the two conditions are the same (`min < max` implies `max > min`).

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -333,42 +333,50 @@ static void _msp_send_altitude(struct msp_bridge *m)
...
+	if (val > lower || val < higher) {
+		return idle;
```
mluessi: Isn't this always the case? I.e. the value should always be between `lower` and `higher`.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -333,42 +333,50 @@ static void _msp_send_altitude(struct msp_bridge *m)
...
+	}
+
+	double d = 1000.0 / ((double)max - (double)min);
```
mluessi: Avoid using doubles (they are very expensive on STM32). This should be easily doable using fixed point math (also makes F1 targets happy).

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -333,42 +333,50 @@ static void _msp_send_altitude(struct msp_bridge *m)
...
-		MANUALCONTROLCOMMAND_CHANNEL_PITCH,
-		MANUALCONTROLCOMMAND_CHANNEL_YAW,
-		MANUALCONTROLCOMMAND_CHANNEL_THROTTLE,
-		MANUALCONTROLCOMMAND_CHANNEL_FLIGHTMODE,
-		MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY0,
-		MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY1,
-		MANUALCONTROLCOMMAND_CHANNEL_ACCESSORY2,
-	};
+	ActuatorSettingsData actuatorSettings;
+	ActuatorSettingsGet(&actuatorSettings);
```
mluessi: As the settings don't change frequently, you could use a global variable for this and use a callback to update it. This reduces CPU usage but uses more memory (instead using stack, the variable is allocated globally and some memory is used for the callback).

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: No, I ran into an issue where during failsafe the values are returned out of range.  In the specific case of throttle, this causes the OSD to learn that the range is wider than it'd previously seen and instead of going 0-100%, it goes 0-1%.

I don't know if there's a better way to detect this, but @mlyle suggested just looking at the value to ensure it's in range.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: Thanks, did that.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: I redid this with fixed-point math.  Seems to do what I expect.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: OK, I moved this.  Not entirely sure how/when to update it, but it makes sense that the settings won't change much.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
mluessi: The code doesn't make sense to be because it assumes `(val > lower || val < higher)` is an exception whereas I would assume it should always be true, i.e., isn't `lower` supposed to be the minimum and `higher` the maximum?

dustin: @mluessi Thanks, that was a mistake.  That catch saved me a round trip on my testing.  :)

dustin:
OK.  I spent a bit of time on the bench playing with this.  There was some weirdness about channel reversing I didn't expect (i.e. the min and max values are reversed, but the channel value itself isn't reversed), but this does all the needful.

dustin:
Also added a current consumption change @tracernz pointed out.

dustin:
I pushed up the pre-scaled change, although it's not been tested yet.

Pros:

* A lot less code.
* Less duplication of scaling rules.

Cons:

* Doesn't seem to be a way to send flight mode as an aux channel anymore (without re-scaling).
* Using floating point math now.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -218,7 +219,7 @@ static void _msp_send_attitude(struct msp_bridge *m)
 
 	data.att.x = attActual.Roll * 10;
 	data.att.y = attActual.Pitch * -10;
-	data.att.h = 0 * 10;
+	data.att.h = attActual.Yaw;
```
kubark42: Something isn't symmetric here. There is a ```*10``` multiplier for Roll and Pitch, but not for yaw. Is this normal?

No matter if it is or isn't, can you please add comments that explain clearly why there are rescaling multipliers?

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -283,12 +284,13 @@ static void _msp_send_analog(struct msp_bridge *m)
...
 	}
 
-	// TODO:  Verify these with a board that has working power stuff
 	if (batSettings.VoltagePin != FLIGHTBATTERYSETTINGS_VOLTAGEPIN_NONE)
 		data.status.vbat = (uint8_t)lroundf(batState.Voltage * 10);
 
-	if (batSettings.CurrentPin != FLIGHTBATTERYSETTINGS_CURRENTPIN_NONE)
-		data.status.current = lroundf(batState.Current * 10);
+	if (batSettings.CurrentPin != FLIGHTBATTERYSETTINGS_CURRENTPIN_NONE) {
+		data.status.current = lroundf(batState.Current * 100);
```
kubark42: FWIW, when using variables that are scaled such that they are no longer using "normal" units, my personal preference is to see that in the variable name itself. So this could be data.status.current_100. There's a double advantage to this: 1) it's clear to any downstream users that this variable isn't containing a value that can be used unscaled and 2) if there are any pre-existing downstream users of the variable the compiler will fail until they're all refactored, giving the opportunity to make sure each and every scaling is correct.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -333,42 +335,34 @@ static void _msp_send_altitude(struct msp_bridge *m)
...
+		.channels = {
+			_msp_scale_rc(manualState.Roll),
+			_msp_scale_rc(manualState.Pitch),
+			_msp_scale_rc(manualState.Yaw),
+			_msp_scale_rc(manualState.Throttle),
+			_msp_scale_rc(acc0.AccessoryVal),
+			_msp_scale_rc(acc1.AccessoryVal),
+			_msp_scale_rc(acc2.AccessoryVal),
+			1000, // no aux4
+		}};
```
kubark42: Style error with ``}}```?

kubark42:
@dustin I don't quite understand the motivation of this PR. Do you think you could flesh out the top post with a little more meat?

P.S. Nice work on having detailed commit messages. :+1:

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: It's correct, but yeah, that's worthy of a comment.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -333,42 +335,34 @@ static void _msp_send_altitude(struct msp_bridge *m)
 	msp_send(m, MSP_ALTITUDE, data.buf, sizeof(data));
 }
 
+static uint16_t _msp_scale_rc(float percent) {
```
kubark42: Could you please document this math equation? It will be hard to maintain without understanding its origins.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: I'm not sure about this one.  It's just a protocol field local to this function and it's changed meaning between a couple versions (and is configurable-ish on the other end).  The protocol docs might be helpful, but it's not well documented there (in particular, the units are documented for voltage, but not current).  It used to be 1/10 of an amp, but at some point some systems changed it to 1/100th of an amp.  Default baseflight is 1/100th, so really *100 is just right for what worked in my out of the box config.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
dustin: Added a comment to the top indicating what the goal of this is.

kubark42:
Thanks for the fixes, @dustin. This looks good to me and I'm happy to merge it if you think it's ready. My one question is in regards to ```_msp_scale_rc()``` (which, FWIW, is prefixed with an ```_```, which we don't tend to do in Tau Labs). It seems like the function has evolved a bit since you started, where at first you were looking to scale between the actual max/min PWM, whereas now you are just mapping between 1000 and 2000. Either is fine, I just wanted clarification on this before committing.

dustin:
The output is always expected to be 1000-2000 as that's what MSP wants.  The way I was originally doing it was by reading the config and doing the scaling myself.  It was recommended I use the pre-scaled values and try that.

That latest revision occurred after all my testing, though.  I sent it up because people seemed happier with the concept, but I wouldn't advise sending it through until I've actually seen it work correctly.  It was a little tricky getting all the bits working right on a real board.  MWOSD interprets stick values for some operations and getting the right ones in is a step towards some pretty useful functionality (and immediately provides some anyway).

tl;dr: Please hold off on my final testing of this new change.

dustin:
Just tested this.  Pitch is reversed here (which makes sense, as it's marked as a reversed channel).  That's kind of unfortunate, as I was hoping the reversal would be handled before it hits this scale.  This implies the actual reversal management is happening in other places.

mlyle:
> This implies the actual reversal management is happening in other places.

Nope, we just use a different coordinate system, and as a result our idea of pitch is different from almost everyone else's.  You can confirm this by unreversing the channel and finding that it's now "right" for the OSD.

dustin:
OK.  That sounds good, then.

The next patch will be tested and ready to go:

* Pitch is reversed
* Throttle is scaled from 1100 -> 1900 across ≤0 -> 1

dustin:
OK, this is tested on my sparky.  Everything works as expected.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
tracernz: The field names don't imply any particular unit and they're not (and I don't imagine ever will be) exported outside the scope of the module. I think within the scope of this module it's reasonable to assume that local variables will be in MultiWii units. To me current_100 is quite meaningless, does it mean this variable contains a value equivalent to 100 A in whatever scaling is being used, or current multiplied by 100, or current divided by 100, or...?

mlyle: yah, I'm kinda meh on this too.  If it was leaving the module some more caution would be warranted.

kubark42: > ... does it mean this variable contains a value equivalent to 100 A in whatever scaling is being used, or current multiplied by 100, or current divided by 100, or...?

That's a fair question and I don't have a way to answer it. What's clear to me, though, is that it's *not* just current and should be investigated more before making assumptions. 

I especially see this kind of scaling error with degrees and radians. Moving to this kind of naming has saved my team tons of time.

Issue #1772, opened 2015-08-05
Proposal: remove support for CC (not CC3D).


mlyle:
There are very few of the boards out there and probably none are flying.  Supporting CC uses a few bytes of CC3D memory and adds complexity in several places.

@solidgoldbomb @CheBuzz @guilhermito @elafargue @peabody124 any objections?

DTFUHF:
This will probably ease confusion too among new TL adopters.

Is CC original even open source HW?

mlyle:
It appears to have documents in the tree.

tracernz:
This would make it easier to clean up the F1 ADC stuff a bit too.

mlyle:
Also on the list: remove support for I2C PCF8591 ADC (only supported on CC flexiport, right now, and doesn't build by default).

peabody124:
I don't love the idea if it is just to remove a few switch statements. I believe the big memory hit is the ADC FIFO which is dynamically allocated on CC (meaning it probably doesn't support autotune).

That being said, supporting pretty old hardware with unavailable sensors shouldn't block serious enhancements.

mlyle:
So, basically, there's three reasons.  First, the way the sensors is bifurcated makes it a little more complicated to handle non-acro Naze variants in the future.  Second, tracernz is going to do some F1 ADC cleanup for battery monitoring on OpenNaze that it'd be nice to not account for primary flight sensors being on that on CC.

But the really big one is to keep the size of the test matrix sane.  If no one flies it, either it ends up untested or developers spend effort keeping something working that no one uses.  This is different from CC3D where there's a vibrant community around these targets and a large number of initial units.  The number of CC units being flown out there is possibly 0 and is certainly small and rapidly decaying.

SystemsGuy:
It does beg the question - is anyone still flying a CC?  I've had mine in a box for at least two years...

guilhermito:
Yes I'm still flying it, so I would prefer to continue to support it as a target

mlyle:
OK, that's good enough for me-- at least one flying out there.  I'll try again in a couple releases ;)

PR #2035, opened 2015-11-01
Naze: Add ADC and Battery monitoring support

tracernz:
This adds a lightweight ADC driver, intended solely for battery monitoring, on the F1 (without any filtering/buffering, since there isn't enough RAM). The ADC driver and battery module are added to the Naze. 

The ADC driver Init function uses a slightly different interface to other PIOS_INTERNAL_ADC drivers, with a number_of_used_pins parameter added. This avoids the need to have the entire ADC config structure in RAM, saving around 200 bytes that are badly needed on the F1.

On the Naze, the Battery module and ADC driver can run in parallel with the UAVOMSPBridge module, and there is still enough RAM for autotune with all modules disabled (just under 200 bytes heap free). Shown here is a Naze32 running Tau Labs connected to baseflight configurator (with some modifications to the UAVOMSPBridge module on a local branch) and providing battery data.
screenshot from 2015-11-01 12 07 21

### DIFF
--- a/flight/PiOS/STM32F10x/inc/pios_internal_adc_light_priv.h
+++ b/flight/PiOS/STM32F10x/inc/pios_internal_adc_light_priv.h
@@ -1,0 +1,49 @@
+/**
+ ******************************************************************************
+ * @addtogroup PIOS PIOS Core hardware abstraction layer
+ * @{
+ * @addtogroup PIOS_INTERNAL_ADC Internal ADC Functions
+ * @{
+ *
+ * @file       pios_internal_adc_light_priv.h
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013
+ * @brief      ADC private definitions.
+ * @see        The GNU Public License (GPL) Version 3
+ *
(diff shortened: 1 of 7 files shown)


mlyle:
I hate to gate this, but I think we should think about what to do with CC/CC3D when it comes to ADC driver while we're at it.

Basically, (recapping from IRC discussions), the current F1 ADC driver is very resource intensive.  Also, a lot of flash space is used on CC3D to maintain support for original CopterControl.  It certainly wouldn't be a good investment to build in two different ADC drivers.

I'd like to get CC3D doing more stuff; one minimal bit of functionality is monitoring battery usage.  

We discussed this some in #1772, which didn't meet very warm reception.  On the other hand, we're facing either CC3D atrophying because we are out of resources to keep both CC & CC3D in the same target...

tracernz:
Yes, if we want to continue moving CC3D forward alongside Naze we either need to drop CopterControl altogether or move it to a seperate target.

mlyle:
Screw it.

PR #2039, opened 2015-11-03
[WIP] GCS:Config: Map motor channels by labels rather than numbers

tracernz:
Mapping numbers to numbers can be very confusing mental exercise for users. This changes the labels shown in the multirotor vehicle widget to labels such as NW for the North-West motor. In this commit only Quad-X, Quad-P, and Tri are changed.

Quad-X:
screenshot from 2015-11-03 18 37 36
Mixers like this will be difficult to label appropriately in only a couple of letters... Ideas?
screenshot from 2015-11-03 18 38 03


mlyle:
This looks great.

I believe you said you were going to change the labels for some of the unusual ones (eg OCTO-V) to match the convention used in the channel naming, so holding off merging for that.

tracernz:
Yep, will do. Just removed the OP logos from the octocopter images, not sure why they were missed before.

tracernz:
Okay, I've dug deeper and decided to refactor all of this code to avoid duplicating the same stuff over and over, and to avoid using GUI strings to map data. I've started but won't have much time until next week so this can wait until then.

DTFUHF:
This is a great improvement in usability, good work. In my mind this
separates information the developer needs to know (motor ordering) from
information the user needs to know (motor orientation).

I wouldn't sweat the more complicated setups. They are still quite
functional.

kubark42:
While this is being done, are there objections to moving to a standardized notation, by which the NW motor is always #1, and motors increment clockwise from there? The hexa-x and octocopter-v violate this, as they use the NE motor as #1.

tracernz:
None from me. I'll tidy all this up early next week after my exams and update this PR. :-)

ground/gcs/src/plugins/config/cfg_vehicletypes/configmultirotorwidget.cpp
```diff
@@ -90,6 +91,12 @@ void ConfigMultiRotorWidget::setupUI(SystemSettings::AirframeTypeOptions frameTy
         //Enable all necessary motor channel boxes...
         enableComboBoxes(uiowner, CHANNELBOXNAME, 3, true);
 
+        {
+            // Set the labels for motor channel boxes
+            QVector<QString> labels({"NW", "NE", "S"});
```
PTDreamer: Why not use QStringList?

kubark42:
@tracernz I hope to have a big PR landing for implementing #2061. Would you mind holding off merging this until that one is submitted, so we can see how they interrelate?

tracernz:
@kubark42 Yeah, asking you about that was on my todo list this morning. I've been working with hardware the last few days so I haven't done anymore on this. Feel free to go ahead and I'll pick this up again later. I think your plans will tie in nicely with what I want to do anyway. :-D

kubark42:
Thanks. It's being a lot more effort than I originally anticipated. All kinds of things were done before the ```UAVObjectToWidgetRelation()``` functions had been written, so it's all tied into the UI instead of the UAVObjects. Ugh.

tracernz:
Yeah, I was hoping to get away from that too. :/

ground/gcs/src/plugins/config/cfg_vehicletypes/configmultirotorwidget.cpp
tracernz, line 96: Will do.

tracernz:
I'll close this for now pending Kenn's work for #2061.

PR #2040, opened 2015-11-03
GCS: Add TxPID checkbox to modules page

### LINKED ISSUE #1872: TxPID doesn't show up on modules page
Needs a hack like autotune did where it shows up greyed / directs users to the txpid page.

Camerastabilization also.

Thx to @pug398 for calling my attention this.

tracernz:
For indication only, checkbox is always disabled.
Fixes #1872


mlyle:
Merging, but this needs to be more obvious as to what to do.  Users including @dustin have interpreted the gray as "not available on this board."  Probably need to add a bunch of text to the modules page to provide hints as to where to do things, what modules are discouraged/deprecated, etc.

PR #2041, opened 2015-11-03
GCS: Fix segfault when board doesn't have connection diagram

tracernz:
I'm fairly sure this fixes [the crash reported in the forums a couple of days ago](http://forum.taulabs.org/viewtopic.php?p=8020#p8020) but I haven't gotten around to setting up the tools to look at crash dumps yet. I certainly the segfault fixed in this PR while testing #2036.


tracernz:
I should note that the segfault occurs because [m_background is only created when the connection diagram SVG exists](https://github.com/tracernz/TauLabs/blob/no-more-segfault/ground/gcs/src/plugins/setupwizard/connectiondiagram.cpp#L77) but it was always dereferenced when the connection diagram window opens by https://github.com/TauLabs/TauLabs/blob/next/ground/gcs/src/plugins/setupwizard/connectiondiagram.cpp#L60.

PR #2042, opened 2015-11-03
vtol_follower_fsm: rth: don't rise so much immediately

mlyle:
Previously this would climb all the way to the minimum altitude
immediately.  IMO this doesn't match the intent of the initial RTH
pause (rise is in the next state).  It has also caused issues
when a burst of full throttle from the combined effect of altitude
error and the impulse from application of the mode destabilized
flaky quads.

On the other hand, if we're too low, it's advantageous to rise
at least a small amount immediately.  It's also good to not set our
hold position too low from any immediate altimeter error.
So climb 1.5m right away.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -606,17 +606,29 @@ static void go_enable_fly_path()
 }
 
 /**
- * Enable holding position at current location for 10 s. Uses a minimum altitude for
- * the vertical altitude. Configures for hold.
+ * Enable holding position at current location for 10 s. 
  */
 static void go_enable_pause_10s_here()
 {
 	PositionActualData positionActual;
 	PositionActualGet(&positionActual);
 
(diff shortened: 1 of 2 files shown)


mlyle:
Added related tuning change:  lower altitude hold gains a little based on user feedback and what I've seen work best on airframes.  Also ensure we default to have a little more than pure atan-based-attitude compensation, as we're in higher drag regimes (I typically see 140-150% fly best, so 120% is chosen).

As always with my tuning change PRs, I am moving not the entire way to what I see as optimal values.

tracernz:
The code change looks fine and sanity checks okay (NED is non-trivial to deal with in your brain). The gains are outside my expertise. :-P

peabody124:
Yeah it looks like this fixes things, and that I had a bug all this time. I'm trying to figure out if there is a reason I had it as before. Has this been flight tested?

peabody124:
I've tested this in the simulation environment and it does as expected. Code reviews properly. Merging.

PR #2043, opened 2015-11-04
uavotaranis: make stuff actually work, send preludes.

mlyle:


flight/Libraries/frsky_packing.c
```diff
@@ -590,14 +590,22 @@ uint8_t frsky_insert_byte(uint8_t *obuff, uint16_t *chk, uint8_t byte)
...
+	 * and therefore the worst-case is 17 bytes total (the first byte 0x10 won't be
 	 * escaped) */
-	uint8_t tx_data[15];
+	uint8_t tx_data[17];
 	uint16_t chk = 0;
 	uint8_t cnt = 0;
+
+	if (send_prelude) {
+		tx_data[0] = 0x7e;
+		tx_data[1] = 0x98;
```
peabody124: In the old code before I lost this in refactor (sorry about that!) I had a reference for this byte to here:
https://github.com/openLRSng/openLRSng/blob/18a9da163e05af25c3c8fb2bdcd749c097770b9c/frskytx.h#L115
Not sure if worth bringing back.

peabody124:
I pushed my old code to make it available before refactor: https://github.com/peabody124/TauLabs/blob/taulinkmodule_old/flight/Modules/UAVOTaranis/UAVOTaranis.c#L702 and looks like this is restoring the same behavior.

As a minor comment I don't love calling it a prelude because that implies it is something that should always be there (to me) and the normal frisky receivers do not expect it. I'm not exactly sure what it is or what it does. Did you find more information that caused you to pick that name?

flight/Libraries/frsky_packing.c
```diff
@@ -590,14 +590,22 @@ uint8_t frsky_insert_byte(uint8_t *obuff, uint16_t *chk, uint8_t byte)
...
-int32_t frsky_send_frame(uintptr_t com, enum frsky_value_id id, uint32_t value)
+int32_t frsky_send_frame(uintptr_t com, enum frsky_value_id id, uint32_t value,
+		bool send_prelude)
 {
 	/* each call of frsky_insert_byte can add 2 bytes to the buffer at maximum
-	 * and therefore the worst-case is 15 bytes total (the first byte 0x10 won't be
+	 * and therefore the worst-case is 17 bytes total (the first byte 0x10 won't be
 	 * escaped) */
-	uint8_t tx_data[15];
+	uint8_t tx_data[17];
```
peabody124: good catch - i had a theoretical bug here before

mlyle:
The other side consists of the RX sending interrogations of the 0x7e - id format; and then the sensor responding with the packed message.  Both halves of this are sent over the air to the TX.

What the opentx side expects to see is both halves of the conversation.

peabody124:
Finally tested. Works on Taranis again. Thanks for the fix @mlyle

PR #2047, opened 2015-11-09
picoc: use single precision floating point

mlyle:
Reduces the cost of picoc from ~50k to ~30k of flash.

### DIFF
--- a/flight/Modules/PicoC/picoc_library.c
+++ b/flight/Modules/PicoC/picoc_library.c
@@ -193,93 +193,93 @@ struct LibraryFunction PlatformLibrary_string[] =
 #ifndef NO_FP
 void LibSin(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = sin(Param[0]->Val->FP);
+	ReturnValue->Val->FP = sinf(Param[0]->Val->FP);
 }
 
 void LibCos(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = cos(Param[0]->Val->FP);
+	ReturnValue->Val->FP = cosf(Param[0]->Val->FP);
 }
 
 void LibTan(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = tan(Param[0]->Val->FP);
+	ReturnValue->Val->FP = tanf(Param[0]->Val->FP);
 }
 
 void LibAsin(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = asin(Param[0]->Val->FP);
+	ReturnValue->Val->FP = asinf(Param[0]->Val->FP);
 }
 
 void LibAcos(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = acos(Param[0]->Val->FP);
+	ReturnValue->Val->FP = acosf(Param[0]->Val->FP);
 }
 
 void LibAtan(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = atan(Param[0]->Val->FP);
+	ReturnValue->Val->FP = atanf(Param[0]->Val->FP);
 }
 
 void LibSinh(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = sinh(Param[0]->Val->FP);
+	ReturnValue->Val->FP = sinhf(Param[0]->Val->FP);
 }
 
 void LibCosh(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = cosh(Param[0]->Val->FP);
+	ReturnValue->Val->FP = coshf(Param[0]->Val->FP);
 }
 
 void LibTanh(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = tanh(Param[0]->Val->FP);
+	ReturnValue->Val->FP = tanhf(Param[0]->Val->FP);
 }
 
 void LibExp(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = exp(Param[0]->Val->FP);
+	ReturnValue->Val->FP = expf(Param[0]->Val->FP);
 }
 
 void LibFabs(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = fabs(Param[0]->Val->FP);
+	ReturnValue->Val->FP = fabsf(Param[0]->Val->FP);
 }
 
 void LibLog(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = log(Param[0]->Val->FP);
+	ReturnValue->Val->FP = logf(Param[0]->Val->FP);
 }
 
 void LibLog10(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = log10(Param[0]->Val->FP);
+	ReturnValue->Val->FP = log10f(Param[0]->Val->FP);
 }
 
 void LibPow(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = pow(Param[0]->Val->FP, Param[1]->Val->FP);
+	ReturnValue->Val->FP = powf(Param[0]->Val->FP, Param[1]->Val->FP);
 }
 
 void LibSqrt(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = sqrt(Param[0]->Val->FP);
+	ReturnValue->Val->FP = sqrtf(Param[0]->Val->FP);
 }
 
 void LibRound(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
 	/* this awkward definition of "round()" due to it being inconsistently declared in math.h */
-	ReturnValue->Val->FP = ceil(Param[0]->Val->FP - 0.5);
+	ReturnValue->Val->FP = ceilf(Param[0]->Val->FP - 0.5);
 }
 
 void LibCeil(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = ceil(Param[0]->Val->FP);
+	ReturnValue->Val->FP = ceilf(Param[0]->Val->FP);
 }
 
 void LibFloor(struct ParseState *Parser, struct Value *ReturnValue, struct Value **Param, int NumArgs)
 {
-	ReturnValue->Val->FP = floor(Param[0]->Val->FP);
+	ReturnValue->Val->FP = floorf(Param[0]->Val->FP);
 }
 
 /* list of all library functions and their prototypes */


mlyle:
Note that there's still dmul/dadd/etc all in here.  Just the dependency on the double precision library functions is removed (sin, cos, sqrt, etc).

kubark42:
I was initially a little torn about this one, because I'm always wary of subtle changes which can occur when changing precision. Then I found

```
/* list of all library functions and their prototypes */
struct LibraryFunction PlatformLibrary_math[] =
{
	{ LibSin,		"float sin(float);" },
	{ LibCos,		"float cos(float);" },
	{ LibTan,		"float tan(float);" },
	{ LibAsin,		"float asin(float);" },
	{ LibAcos,		"float acos(float);" },
	{ LibAtan,		"float atan(float);" },
	{ LibSinh,		"float sinh(float);" },
	{ LibCosh,		"float cosh(float);" },
	{ LibTanh,		"float tanh(float);" },
	{ LibExp,		"float exp(float);" },
	{ LibFabs,		"float fabs(float);" },
	{ LibLog,		"float log(float);" },
	{ LibLog10,		"float log10(float);" },
	{ LibPow,		"float pow(float,float);" },
	{ LibSqrt,		"float sqrt(float);" },
	{ LibRound,		"float round(float);" },
	{ LibCeil,		"float ceil(float);" },
	{ LibFloor,		"float floor(float);" },
	{ NULL, NULL }
};
```

which makes it pretty clear these are supposed to be floats!

peabody124:
@ernieift is this something you would like to take a look at?

ernieift:
PicoC uses double type for all floating point calculations internally. TokenFloatType and TokenDoubleType are mapped to FPType which is from type double (see type.c).
PicoC has no optimizer for huge formulas or brackets. I guess, it's better to use the double type to avoid rounding errors.
To safe flash memory but have the interpreter still available, you can disable all floating points by define NO_FP in picoc_port.h.
I'm still writing on the F7 port. With its FPU and flash the problem will disappear by itself.

mlyle:
Isn't it better to have single precision trig floating point than none at all?

ernieift:
@kubark42 the math define texts are floats like in the original PicoC source. I think Zik Saleeba used this for better readability and it saves one character per define ;).

ernieift:
@mlyle do we have a memory problem on F4? If so, we could introduce a define switch for smaller targets. Generally i'd better use double precision for GPS and PIDs within the interpreter.

mlyle:
No problems on F4 (well, except for Colibri where we probably shouldn't dare to change the bootloader).  This is to reduce the CPU cost of floating point to something sane (some of these things have very long runtimes in cycles) and to cut things down so that it can fit into F3 (instead of nav).

It seems kinda silly to bring in 18 double precision floating point functions for the interpreter when we literally don't use them anywhere else in the flight code.

ernieift:
When i started the work on PicoC, i quickly realized it won't fit into the F3 with all the other flight modules. Without GPS etc. there are hardly advantages to have PicoC on board.
In order to remove the double precision from PicoC, you have to touch almost every file of the original source. Then you loose the upgrade possiblity and have review all the code.
The interpreter has a very low priority. It won't give the performance gain you may expect here.

mlyle:
I'm not concerned with the __aeabi_muldd, etc, routines, or all of that being done in software, or retaining double precision for operators and intermediate products.

My concern is with those trig, log, and power functions-- they can be expected to take thousands of cycles (which in a low priority task is forever), and they use 20-25k of flash which puts this out of reach of fitting on F3.

dustin:
How is PicoC used in the wild?  Having PicoC working on F3 would make showcasing TauLabs' amazing abilities within reach of a lot of people who want to spend less than ~$90 on a flight controller just to do something fun with lights or build a "laser tag" type functionality.

Is having a larger footprint for what appears to be significantly slower functions worth keeping a hard limit on the targets?

tracernz:
What is the reason we would want double-precision math in PicoC when we don't use it anywhere else in the firmware, and we only have a single-precision FPU?

ernieift:
You have double-precision in PicoC by default. For normal math (+,-,*,/) nothing changes here. The interpreter math.h part a copy of the original clibrary.c from the picoc-project.
Sure, you get more flash by removing the software stm-lib but risk an incompatibility in precision. The user is free to use the trigonometric functions and has influence on PicoCs performance. If there are no calls for this in the script, it is as fast as possible. My suggestion: insert a conditional define (for use sin() / sinf() e.g.) in targets pios_config.h like this:
```
#define PIOS_INCLUDE_PICOC
#define PICOC_FORCE_FLOAT
```
and in picoc_library.c
```
#ifndef PICOC_FORCE_FLOAT
ReturnValue->Val->FP = sin(Param[0]->Val->FP);
#else
ReturnValue->Val->FP = sinf(Param[0]->Val->FP);
#endif
```

BTW xxx->Val->FP is always double type. So you have the type conversion twice. I am concerned what impact it has on the calculations of the interpreter. A few tests were good. If PicoC fits in a F3 target this way, you have my blessing. On a Quanton, i'd rather use higher precision.

ernieift:
To substantiate my concerns something I once picked out something:
http://stackoverflow.com/questions/916081/convert-float-to-double-without-losing-precision

mluessi:
I'm +1 for this change. Given that we don't use double for critical flight code, it makes little sense to use it in picoc, which is usually used for simple things like controlling LEDs.

ernieift:
It makes sense. All other PicoC implementations (independed from CPU power) use double precision math. It is testet and proofed to work. If you don't use these few math functions, nothing changes. But if you use them, your script may not run as expected. The only plus is the smaller code of PicoC in ROM. It needs the same RAM as before. As i said, the native floating type within PicoC is double.

kubark42:
>What is the reason we would want double-precision math in PicoC when we don't use it anywhere else in the firmware, and we only have a single-precision FPU?

PicoC is a generic platform. We have little knowledge of what the user will ask of it, so it makes a lot of sense to default to giving the user the best chance of success. I've already run into problems with single-precision FPU, and am very disappointed that STM's F7 chips do not implement the double-precision hardware FPU.

I like @ernieift's solution to use #defines. This makes a lot of sense to me.

dustin:
Is PicoC the right platform?  It sounds like it's just too big -- only fitting on the largest targets and requiring slow software floating point math even where then (when FPUs are available).

How widely is it actually used in its current form?  The README makes it sound like it was written for this purpose, where it's not very suitable in its current form:

> PicoC is a very small C interpreter for scripting. It was originally written as a script language for a UAV's on-board flight system.

If it's meant to be a script language for a UAV flight system, but only few targets can even fit it and if you use the features that make them not fit they're prohibitively expensive, it should be fixed or replaced.

The problem with #defines (other than being kind of ugly) is that if you've got a real problem with single vs. double precision floating point operations on these causing compatibility or astonishment issues, you've now amplified that by having them both happen int he same codebase.  Suddenly you've got scripts that are written for F4 and are sluggish due to precision running on F3, potentially faster, but with slightly different answers.  This doesn't seem better.

IMO, better is building a scripting language for UAVs that is beneficial to as many of these UAVs as possible to build interesting and useful applications that are efficient and can target as wide an audience as feasible.

mlyle:
The most damning thing I hear about picoc using single precision is that you need to be careful handling world-scale coordinates, which is something we need to do in flight code too.

kubark42:
I don't see the value of making this work on the F3 if it means hobbling any performance on more modern targets.

Precision is important and I'm not sure how we explain that to lay users who might want their LED to blink when they get to a certain geo-location. I'd rather we get to a right answer slowly than the wrong answer quickly.

I'm not sure that replacing PicoC with another scripting approach will resolve anything. The issue is that single-precision FPUs can't do double-precision math without lots of library calls. Plus, @ernieift has been excellent at maintaining the code, and that's got incredible value.

mlyle:
@kubark42 Most of what picoc is used for in practice, from what I can tell, is LED and alarm control.  It can fulfill a lot of roles in this area.  There are exciting F3 targets coming.  F3 is nice for our users in general because of the better UARTs and the cheap boards-- it's a real "sweet spot" for the predominant use case for Tau.

I'm super excited about navigation and increased autonomy and fancier things.. but I also like the basic stuff and don't want to lose track of our core audience there.

Also, I really question the whole point behind F4 and F7.  We are not able to use the increased resources effectively, other than increased code size.  It's nice for developers but there's not really any significant capability that can be derived from getting somewhat more capable than F3.  There's exciting things you can do when you get a lot larger (hence why I need to finish flypi)... but there's not really anything F4 can do that F3 can't currently, other than PicoC.

mlyle:
P.S. if coordinate handling is the real concern, we can write a couple library functions to do it robustly.  It's not trivial to do correctly even with stock double-precision operations in picoc.

ernieift:
For those who don't know what to do with PicoC:
You can write to PathDesired in position hold mode. This gives you the possibility of autonomous, sensor controlled and gps stabilized flight.
Or just take a LiDar and avoid obstacles on the ground...

ernieift:
@mlyle if you really want to make PicoC running on F3, you should remove all floatings from PicoC by "#define NO_FP" and write some translator functions for non-integer UAVOs. But i doubt there is enough space. I already tried this in the beginning.

kubark42:
> There are exciting F3 targets coming. F3 is nice for our users in general because of the better UARTs and the cheap boards-- it's a real "sweet spot" for the predominant use case for Tau.
> .
> .
> .
> Also, I really question the whole point behind F4 and F7. We are not able to use the increased resources effectively, other than increased code size. 

I see where you're coming from, but I don't agree. We're doing okay for manually-flown multi-rotor flight, but there's so much left on the table because of our resource constraints. I'm already using up tons of memory doing data analytics, and I think the community will find our results interesting once they're ready. 

More to the point, we're seeing microcontrollers like the ARM SAM7 coming online, which are the same price as the STM32F4/F7, but have a ton more RAM and processor speed. Increasing developer workload and decreasing utility in order to target chips which are $3 in quantity instead of $5 just isn't where we want to be, IMHO.

mlyle:
> I see where you're coming from, but I don't agree. We're doing okay for manually-flown multi-rotor flight, but there's so much left on the table because of our resource constraints. I'm already using up tons of memory doing data analytics, and I think the community will find our results interesting once they're ready.

Like I said before, mostly useful to developers.  I think we want to be able to support large targets for our own benefit / ease in prototyping / etc.  I don't see functional benefits we can derive from larger processors-- it basically lets us be even more sloppy than we have been so far in resource use.

Right now center of mass of flight controllers is around STM32F3.  We can effectively "stop and regroup" and target larger things and hope to be relevant when things evolve.  But it really, really sucks to need to spend $100 or buy an illegal clone to use Tau, when there's other flight controllers out there for $30 that for many use cases are superior.

kubark42:
(FWIW, the data analytics are an essential part of the autonomous flight that we'll be providing in the LVL 1. I can't say more than that until the PR goes out.)

There's nothing inherently wrong with using resources to save developer time. We do this all the time by using floats where we could use fixed-point. Not saying we haven't been sloppy, but cases like PicoC there's a reasonable rationale for the increased usage.

While I don't necessarily want to discuss whether staying relevant is valuable or not-- people have a wide variety of opinions on this topic-- OpenPilot got ahead of the pack because it targeted future hardware. When everyone was using 8-bit processors, it made autopilots work on 32-bit. The fact that 3DR/PixHawk ultimately trounced it was purely because of the toxic culture and years we wasted with horrible project governance.

My personal opinion is that I don't want to get in a race for "most fun" autopilot. I want to be in a race for "most transformative" autopilot. Fun requires cheap, transformative requires capable. Ideally, we can do both, but when we have to trade the potential to be transformative in order to target cheap hardware I like to go slowly and see a good consensus emerge.

dustin:
> work on the F3 if it means hobbling any performance on more modern targets

Is this the case?  From what I've seen, this will only work better on all the targets, ± a few floating point inaccuracies for a small set of use cases that can have specific workarounds.

Right now, there are a ton of use cases that are just unavailable.  e.g. I know a guy who was trying to build a bunch of hardware to do a quad laser tag application.  What he could do with an external bit of hardware interacting with a flight controller in very limited ways would be nearly uninteresting, but with PicoC, you could do things like simulate damage trivially.  Except the idea of spending $90 on a sparky2 to do this is completely unappealing.  Getting a $30 sparky makes the whole project viable.

There are a lot of use cases.  I haven't seen any actual benefit to requiring double precision math expressed here.  The closest I've seen is the [general floating point](http://0.30000000000000004.com/) issues that you're always going to have, but the specific cases where it matters can be handled specifically (as @mlyle pointed out -- even with double precision, it's not trivial to do the math correctly, so we might as well provide something to do it).

I get that people are excited about autopilot, but that's just one small application.  There are lots of things that *could* be done with an onboard scripting language.  The most exciting ones for me have little to do with operating the aircraft itself.

mlyle:
> My personal opinion is that I don't want to get in a race for "most fun" autopilot. I want to be in a race for "most transformative" autopilot. Fun requires cheap, transformative requires capable. Ideally, we can do both, but when we have to trade the potential to be transformative in order to target cheap hardware I like to go slowly and see a good consensus emerge.

I want to be in a race at all.  I think there's neat things I can get done with a vibrant user community and they're difficult to accomplish otherwise.  Right now there's a big stagnating PR queue, next is broken, people aren't around much, and it's difficult to get work done.

Right now position and environment sensors really suck, and mocap rooms are small. There's two ways to push forward the core controls in a meaningful way, in my opinion:  either figure out how to get meaningful feedback from acro pilots (which is an area I've been working with the user community from IRC and in hopefully supporting some very nice upcoming F3 targets), or go very large so you can do meaningful vision stuff (which is why I'm building flypi).

ernieift:
Allow me the question: What have acro pilots to do with trigonometry?
I also used the Quanton with PicoC for ground vehicles. It quite easy to let the flight controller do all the critical and complicated stuff. I think PicoC as an application layer above a versatile Hardware with an operating system. I like the freedom of the script. Even my kids learned a little programming.
Just put a flight controller on a car and let it find its way to north. This is nothing you want do within a code optimized module. It has simply to run.

kubark42:
I think there are two conversations here. One is about the PR, and the other is a little broader. I appreciate where @dustin and @mlyle are coming from regarding the F3/F4/F7. We don't agree, but that's ok, having a great conversation that reflects several viewpoints is why I like programming here in the first place! :D

Regarding the PR, @mlyle are you okay with @ernieift's reasoning? Is ```#define NO_FP``` enough to accomplish your goals for FlyingF3?

kubark42:
> There are a lot of use cases. I haven't seen any actual benefit to requiring double precision math expressed here.

Not claiming that these are PicoC use cases, but here are a few immediate cases where doubles are necessary

* My memory-optimized linear and circular standard deviation algorithms both fail when using single-point precision. The numbers get too close to 0, and the result returned by the algos is ```0```, when it should be non-zero.
* The example of geocoordinates is non-trivial. I have upgraded our internal fork of Tau Labs to use only the NAV-SOL GPS object. This requires converting from ECEF to NED, which yields bad results in single-precision calculations.
* A fun one from back during my dissertation was calculating the Hamiltonian for an optimal control. We needed 13 decimal places for the optimal system to converge to stability. (Okay, really not doing that one in PicoC!)

mlyle:
> Not claiming that the following are likely PicoC use cases, but here are a few immediate cases where doubles are necessary:

We'd be better served for all of these with double-single precision on our hardware.

mlyle:
> Allow me the question: What have acro pilots to do with trigonometry?

Acro pilots want to fly cheap F3 targets that are in nice form-factors easy to integrate into their craft.  They also want picoc to do various things like drive LEDs or do dogfighting.  If we give it to them this way, we don't need to saddle the codebase with lots of special purpose C code for these tasks, or ignore them and let them drift elsewhere.

dustin:
I guess I don't understand what's actually being done in PicoC that would no longer be possible.  I can easily see what can't be done today that would be possible.

The `NO_FP` thing seems to be the worst case -- now you get some answers differently sometimes.

I also don't quite understand how this makes things worse for F4+ targets.  I get that PicoC isn't properly using the FPU, but... shouldn't it?  Why don't we make better use of the hardware *and* have it work on more hardware?

ernieift:
> I can easily see what can't be done today that would be possible.

You mean F3 support? It is not content of this PR.

mlyle:
> You mean F3 support? It is not content of this PR.

No, it's not.  I am doing this as incremental work for more functionality on F3.  I can do a great big monolithic PR if it's preferred.

This would also make picoc work on Colibri, which is F4 but has a bad flash layout (copied from quanton) that makes it not fit.  Quanton's bootloader has since been updated, but we have not wanted to do that for Colibri because of the large number of users on it that might not fare so well with a bootloader update...  There have been users asking about picoc support for Colibri on the forum , when it was removed by @peabody124 in 175781ee483.http://forum.taulabs.org/viewtopic.php?f=18&t=883&p=7083&hilit=picoc#p7083 .

ernieift:
Updating the boot loader via BU-firmware is not a big deal. As you pointed out. There is much unused flash on the Colibri which is almost a schematic/software blue print from Quanton. I think it overdue for this target. Or we recommend installing the Quanton firmware on Colibri for PicoC users?
There is still a pow() in the core of PicoC (lex.c) which reads float values from the script. If you want to remove all double precision calls from it, you have to change this too. When i wrote a float value into a script i want to get it back as i wrote it.
So i'd prefer removing floats by NO_FP. It safes much more flash by a single define.
Please check it out before you touch a running system. I think there are reasons for the PicoC guys to handle floats this way. Then we can write some special library functions to handle floating UAVOs. It is better for scriptwriters to program only integers than struggling with incorrect values.

mluessi:
How about this: We document that all operations are single precision and if someone really needs double, we add a compile switch to enable that.

mluessi:
PS: At the end, it would be nice to do everything using double precision (and we probably will, in a few years), but we manage to do things like an EKF using single precision and I still don't understand why PicoC needs double. Yes, there are things that may not work well, like GPS coordinate operations, but we can add library functions for that if needed.

peabody124:
> Also, I really question the whole point behind F4 and F7. We are not able to use the increased resources effectively, other than increased code size.

Actually this discussion is exactly why I prefer those :). At least when working on F1 I abhor constantly trying to fit things in that last tiny remaining space, or starting to get crashes because we have to get aggressive on stack sizes and then a small change has subtle side effects. Anyway, that's a digression and ultimately we should do the best we can to support what targets are out there, but still acknowledge there are real limitations of those targets and some things may ultimately not fit.

I think @ernieift makes a good point that there might be a reason picoC was coded accordingly. At the very least, we should make sure there is good functional testing on this before merging (side note: can we have unit tests on picoC?). What is the code size if compiled with ```NO_FP```? That might be a better way to support F3 targets, anyway, by providing scripts but less power.

mlyle:
> > Also, I really question the whole point behind F4 and F7. We are not able to use the increased resources effectively, other than increased code size.

> Actually this discussion is exactly why I prefer those :). At least when working on F1 I abhor constantly trying to fit things in that last tiny remaining space, or starting to get crashes because we have to get aggressive on stack sizes and then a small change has subtle side effects.

Yes, it would be nice to have infinite ram and cycles and be able to be sloppy in various ways.  But right now F3 is still a significant sweet spot.  And as long as we choose to make that tradeoff-- to try and support F3 well-- the added resources on F7 don't do us a lot of good.

> What is the code size if compiled with NO_FP? That might be a better way to support F3 targets, anyway, by providing scripts but less power.

NO_FP is useless-- it prevents you from using UAVOs with floating point values, so you can't access nearly any of the system state.

ernieift:
First, you should try to bring PicoC in running state on F3 with `NO_FP` to see wether the work is worth it.
If you have performance problems and want to use the FPU, why not implement the FPU compatible functions like `sinf()` as `float sinf(float)`, the way you wanted, in addition. There is no ROM problem on PicoC enabled F4-targets.
You may disable the double precision part of math.h by a `#define PICOC_NO_DOUBLE` in Colibri for instance. But don't touch the `pow()` in lex.c or other core functions within PicoC interpreter itself.
What you should not do, is to replace all the `(double)` with `(float)` within PicoC. This makes the interpreter incompatible the the main fork and can cause badly issues.

To the F3 discussion:
You think F3 is the navel of flight controls for now. The F1 was high end in 2010. The guys who want not spend the money for a F4 today are the same who buy the cheap F4 in 2 years instead of a better chip. Then, we reenable all the functions and discuss about script compatibility.

mlyle:
> First, you should try to bring PicoC in running state on F3 with NO_FP to see wether the work is worth it.

This is worth it for Colibri alone.  Also, I'm working on an overlay system where modules can be tacked onto the main firmware image so that the user can select different feature sets. 

> If you have performance problems and want to use the FPU, why not implement the FPU compatible functions like sinf() as float sinf(float), the way you wanted, in addition. 

We could do that... I just really don't see the point.  All the rest of the flight code gets by without using these double precision functions.  We have yet to see an actual end-user use case that uses these functions and requires the precision.

> But don't touch the pow() in lex.c or other core functions within PicoC interpreter itself.

Yes, I have not edited the lexer (even though it's super weird-- base 2/base 8/base 16 floating point literals?).  We already have our own library file so upstream compatibility is not a consideration.

> What you should not do, is to replace all the (double) with (float) within PicoC. This makes the interpreter incompatible to the main fork and can cause badly issues.

Maybe.  The last commit to picoc was in Mar 2014.   There's been stuff sitting in its PR queue indefinitely without action.  We are substantially the only user of picoc.  Upstream compatibility here would be nice to maintain, but really has very little anticipated benefit.

(Really, I'd prefer something like eLua, etc, but in the resources we have, basically only Forth or PicoC fit).

> You think F3 is the navel of flight controls for now. The F1 was high end in 2010. The guys who want not spend the money for a F4 today are the same who buy the cheap F4 in 2 years instead of a better chip. 

Maybe.  The F3 is actually better for most things, because it has nicer UARTs.  And right now we can't really measure the F4 being "better" in any way, other than the increased program memory space.

ernieift:
>  We are substantially the only user of picoc.

Not quite right. There are Magic Lantern and Loxone...

mlyle:
> Not quite right. There are Magic Lantern and Loxone...

Looks like Magic Lantern removed support in 2013:  http://www.learningsurfphotography.com/tag/picoc/ ;)

tracernz:
As far as I can tell PicoC development has stopped. It says in their readme that it is feature complete and no commits have been made since mid-2014, so upstream is not going to be an issue, we're on our own with it. 

I have still yet to see any justification for why PicoC actually needs double-precision math. @kubark42 mentioned some nice-to-have reasons but the same applies to firmware code and we have no plans to use double-precision there.

ernieift:
I am tired of the discussion. I have made proposals to combine both interests but there seems to be a finished remodeling the interpreter and it is presented piecemeal. (My guess)
With this single commit PicoC loses accuracy and not a single LED will be on F3. The interpreter runs on the Colibri therefore not yet.

lolorc:
I totally agree with everything @ernieift said.
While I really value the group effort to implement autotune on F1, I don't see the point in wasting too much time with this and sparky1,
F3 is basically used as a high-end racing platform nowadays.
Also picoC is not really a simple feature to use by all users, when they want leds they use cleanflight :)

The move forward is obviously to carry on with the F4 targets and support more powerful ones.
There's only one F3 target and there are 4 F4 ones btw.

dustin:
> Also picoC is not really a simple feature to use by all users, when they want leds they use cleanflight :)

The general attitude of Tau Labs seems to be degrading into "we only want to support very narrow use cases for users who want to spend a lot of money."  Given the effort I've been putting into telling people that Tau Labs has a generally better platform for all types of flying, this is really depressing.

Tau Labs is, today, better than the multiwii family in basically everything but adoption.  Conversations like this do well to maintain obscurity.

mluessi:
While I don't see a big problem to move PicoC stuff to single precision (or have a compile switch to enable double); I do think the intersection of people who will use PicoC and the people who F3 fly racing quads is very small.

dustin:
> While I don't see a big problem to move PicoC stuff to single precision (or have a compile switch to enable double);

The problem with making it a compile-time feature is that if you do care about floating point precision enough to think double makes a big difference in scripts, but not enough to care that scripts produce the same results on two different builds.

> I do think the intersection of people who will use PicoC and the people who fly F3 racing quads is very small.

This is just perceived constraining of use cases.  What is it about F3 that makes it less appealing for dynamic feature insertion?  These are the guys who are today freaking out over how awesome light effects are.  That's the gateway drug -- you've got hard-coded sets of light features you can use in cleanflight -- what if you could do anything with lights you imagine?  I want to use a knob to change light colors, or location/orientation lights I enable conditionally in a pattern that makes sense to me.

But then you get into more interesting stuff.  What if I want to build a training script for helping learn some acrobatics?  Grade me in how off-center my rolls are, or keep me from going below a certain throttle level when I'm not trying to land (which I did on TX, but I could do more interesting stuff on FC).  There's a lot I could do here.

The whole quadtag thing was targetting F1s, but with a different microcontroller running the extra hardware and having very limited interaction with the actual flight controller -- while you could activate it and stuff, you can't even know what the effect is.  With a bit of picoc (and a similarly priced F3 FC), you could actually have hits register as alarms and show up on the OSD, and then do things like simulate damage by detuning or lowering rates or throttle range.  Even shake the quad a bit when it's hit.

Similarly, integration with lap counters would be *huge*.  The devices exist, but you can't really see what they're doing.  What if you could get lap time, etc... on your OSD with trivial input from the device?

I don't know what all might be interesting, but we can't find out because there's so much thought on what you could do if you only have bleeding edge hardware that we don't make it easy for people in the wild to experiment and show us what amazing things could be done.

tracernz:
>>While I don't see a big problem to move PicoC stuff to single precision (or have a compile switch to enable double);

>The problem with making it a compile-time feature is that it only makes sense if you do care about floating point precision enough to think double makes a big difference in scripts, but not enough to care that scripts produce the same results on two different builds.

Agreed, a compile-time switch breaks the usefulness of an interpreted scripting language (copy-paste code snippets that work on all targets). 

It seems people seem to have an idea that Tau is only good for a very narrow range of use-cases which is unfortunate. Whatever we do it should cater to as many use-cases as possible, especially things which are in-demand e.g. LEDs, and new stuff that nobody has done yet (see @dustin's ideas above for example). 

FWIW the F3 offers much more capable peripherals than the F4/F7 so any suggestion of dropping our focus from F3 does not sit well with me.

kubark42:
I think of Tau Labs as being applicable to a wide variety of vehicle and non-vehicle uses. Platform flexibility is definitely our strength when compared to other autopilots. I've even made a stroboscope with a Quanton! I'd hate to see TL become single use, which is where the danger comes by focussing only on F1/F3 quadcopters. I don't think we're doing the right thing when we have recurrent PRs like https://github.com/TauLabs/TauLabs/pull/2058, whose sole purpose is to slightly bump memory allocation because of in-air FC reboots. Much better IMHO to have 500 byte stack overheads, and set the yellow warning to come on at 200 bytes instead of ~10 bytes.

In addition to leading to decreased safety and increased developer workload, the F1/F3 chips just don't have the memory to do interesting things. I love @mlyle's work at reducing stack and CPU utilization, it's invaluable. I'm personally more attracted to developing all the robotic behaviors I would like to see enabled on the board, and this takes more memory. Things like https://github.com/TauLabs/TauLabs/issues/2061 would be no-brainers if we were only targeting chips with sufficient resources, but frequently they wind up being controversial because the increased abstraction costs RAM which F1/F3 targets don't have.

I'm not arguing against all the wonderful work to reduce resource requirements, but I do feel that it's not sensible to constrain ourselves to 2010 microcontroller limitations in (soon-to-be) 2016.

mlyle:
> I'd hate to see TL become single use, which is where the danger comes by focussing only on F1/F3 quadcopters.

Except right now F3 can do literally everything that F4 can.

> I don't think we're doing the right thing when we have recurrent PRs like #2058, whose sole purpose is to slightly bump memory allocation because of in-air FC reboots. Much better IMHO to have 500 byte stack overheads, and set the yellow warning to come on at 200 bytes instead of ~10 bytes.

We have crummy build infrastructure and tools.  It's trivial to figure out exactly the right amount of stack as a build step and compile fail if you don't have 8 bytes more than this.  It's on my todo-list when I'm not fighting about PRs.  (see -fstack-usage )  It's almost trivial to just pick the right amount at compile time (but you need to build each file twice or have a file of constants).

We're going to have embarrassing stack overflows and logic errors until the build infrastructure catches up.  Unfortunately it is just too hard to get anything done here.

> but I would argue this is not the relevant difference that is holding back TL from even doing robust position hold without significant knowledge of the GCS and UAVOBrowser.

I did fairly decent position hold on two atmegas with crummier sensors and GPSes a decade ago.  I am not sympathetic to the view "we can't do this stuff because we only have 40kbytes of RAM".

mlyle:
One more thing:  this is framed in terms of "developer productivity" vs. "resource utilization".  This is very occasionally true.. but not usually.

We just love to waste resources at the cost of writing even-more-complicated code.  This project copies stuff and allocates additional copies of buffers when there's no reason.  There's a serious love for unnecessary conditionals.  We use single-element queues that we never look at what comes out of them instead of semaphores.  It may increase initial developer productivity, because hey, I can just copy and paste some code that does something similar (like I just removed the pipxtreme system module which is a strict subset of the ordinary one)... but it is a perpetual tax on both resource utilization and developer productivity.  Worse, it increases the feature point complexity of the code which is a further tax and makes it more difficult to assure ourselves of correctness.

Yay, because I tried to bring useful scripting to F3 and Colibri we now have a competing PR which introduces more build variants and subtle variations between targets-- which is the something we don't exactly need more of.

dustin:
> Platform flexibility is definitely our strength when compared to other autopilots. I've even made a stroboscope with a Quanton! I'd hate to see TL become single use, which is where the danger comes by focussing only on F1/F3 quadcopters.

These statements seem to be somewhat in conflict.  Referring to platform flexibility and "autopilots" while arguing against single use is painful.

I like the idea of having a TL autopilot, but it's not the use case I have been putting the most time into.  I was given a sparky specifically for a fixed wing project I want autopilot on (but selfishly placed it in something I can use closer to home in the short term).  I can think of all kinds of scripting things I'd like to do on a plane like this.

But I do think TL is really great on the lower end as well.  The work that makes it great on the low end is *required* for being great on the higher end.  The more things that can be done more cheaply for more people with more aircraft the better off it is.

i.e., there seem to be two camps here that both think the other is trying to specialize too hard.  I think TL can do amazing things in a variety of applications and the engineering effort to increase the applicability of the software is in progress... held up in the pull request queue.

kubark42:
> Except right now F3 can do literally everything that F4 can.

In the sense that the codebase has been written with an eye to that processor, and the F1 as well, I agree. But my claim is that it has been holding us back, and will continue to do so until we move past it. 

Let me be clear, this is just my opinion. I think debate on this is good and healthy, and I think TL should evolve with time. I'm fine with the consensus, even if I don't agree with it.

> I did fairly decent position hold on two atmegas with crummier sensors and GPSes a decade ago.

The Mikrokopter did good position hold, but it wasn't *useful* because it was too hard for the user. We have the same situation, where our position hold is very good, but lay-users don't one uses it because as an overall experience it's nowhere near good enough.

The controls math isn't hard and can fit into tiny slices of memory, but making it user-proof-- so that the user is focussed on the goal, and not the flying-- takes a lot more effort/resources. Things like real geofences that work for the user's expectations instead of circular geofences which are never used because they're too hard to work with. Hiding complexity takes resources.

> I don't know what all might be interesting, but we can't find out because there's so much thought on what you could do if you only have bleeding edge hardware that we don't make it easy for people in the wild to experiment and show us what amazing things could be done.

This is a good point, but I don't think that the F4 is bleeding edge. I had an original Revo in 2012, so that chip is actually quite mature. It was introduced at approximately the same time as the F3.

> We just love to waste resources at the cost of writing even-more-complicated code. 

Excellent point! I'm impressed by your deep architectural thinking on this, and it has only improved the code.

>... held up in the pull request queue.

Yup, that can get extremely frustrating. The reason you don't have good autonomous flight on your fixed-wing is because of this. I wish I had more time to devote to reviewing and clearing out the queue, but work hasn't afforded me that opportunity of late!

Anyway, I think we're all on the same page here. We're debating a fine point about whether the F3 is an appropriate target, which is great, but I sense it's also creating frustration because it's getting wrapped up in the review. As far as this PR goes, I'm inclined to align with the wishes of the module maintainer. Whatever @ernieift is okay with is fine by me.

mlyle:
> We have the same situation, where our position hold is very good, but lay-users don't use it because as an overall experience it's nowhere near good enough. Too much complexity that the user has to tackle.

Our users don't use it because we've made it deliberately hard.  Getting position hold working generally involves little more than an autotune... and clicking on lots of mysterious things in GCS to turn functionality on and off.  Outside of the rate/attitude loop tuning, there's generally not any tuning needed for anything from a 250 racer to a 900 size behemoth-- I've flown and done holds with all of them with default tuning.

Also, our current users mostly don't care.  Almost no one is flying Tau for sedate cameraship flying.  We make it really hard.  Camerastab and accesorydesired are "weird" in various ways from the end user perspective and poorly documented.

It sounds like your preference would be to drop F1 and F3 support because they're a pain and slowing progress down, worry about acro a lot less, etc, to focus on nav and autonomy.  It would certainly make my life a lot easier.  I care about making the existing userbase happy and growing it, though.  And while there's costs of various kinds incurred on the treadmill of making them happy, there's significant benefits from their contributions as well.

Here's the thing-- I'm spending more time arguing and doing project political stuff than the cleanup that I'd like to do... some of which may be "wasted effort" in the direction of supporting F3 but for the most part benefits all targets and reduces maintenance cost everywhere.  This is a far bigger cost to me in productivity than thinking about how to fit things efficiently on F1 and F3.

lolorc:
> I think of Tau Labs as being applicable to a wide variety of vehicle and non-vehicle uses. Platform flexibility is definitely our strength when compared to other autopilots. I've even made a stroboscope with a Quanton! I'd hate to see TL become single use, which is where the danger comes by focussing only on F1/F3 quadcopters. I don't think we're doing the right thing when we have recurrent PRs like #2058, whose sole purpose is to slightly bump memory allocation because of in-air FC reboots. Much better IMHO to have 500 byte stack overheads, and set the yellow warning to come on at 200 bytes instead of ~10 bytes.
> 
> In addition to leading to decreased safety and increased developer workload, the F1/F3 chips just don't have the memory to do interesting things.

can't agree more, :+1:

ernieift:
> Is PicoC the right platform? It sounds like it's just too big -- only fitting on the largest targets and requiring slow software floating point math even where then (when FPUs are available).

What about an additional interpreter next to PicoC like PET-Basic/uBASIC. It can have far less resources and of course limited functionality. Less experienced script writers would have a better understanding here. You could also use a modified minimOSD as serial driven character monitor and print directly to the OSD. All variables then where single float or strings. Variable names are single characters from "a" to "z".
To access UAVOs, you have to write basic functions/commands.

kubark42:
As per https://github.com/TauLabs/TauLabs/pull/2060#issuecomment-158106406, closing in favor of #2060.

dustin:
@kubark42 

> This is a good point, but I don't think that the F4 is bleeding edge. I had an original Revo in 2012, so that chip is actually quite mature. It was introduced at approximately the same time as the F3

F4 is bleeding edge in terms of flight controller availability.  I don't mean literally "this is the newest, hottest chip," but practically, F4 is the new guy, and it arrives at a premium on boards.

I can get F3 flight controllers for under $30 because they've been soaking in the market long enough to have lost their new car smell.  I can get a quanton from China for $75 or so, but it's kind of big and won't fit on my quad.  I'm sure I can get a bootleg sparky2 from somewhere for maybe $50 or so.  Officially, they're $90.  Brain is $100.

I'm sure this will get better over time, but given that Tau Labs isn't just overrun with users now, embracing them where it's reasonable seems like a good thing (e.g. I'm really hoping #2013 can be resolved so F1 users can get a good flying experience, otherwise, maybe we shouldn't be supporting F1).  Making decisions to simplify code for the users while enabling more users just seems like an obvious choice.

peabody124:
@dustin I also hope #2013 can get fixed, I'm just worried in this case there straight up are not the computational cycles to achieve it. I hope #2062 goes some way to improving that

PR #2052, opened 2015-11-11
gcs: output: add 16666Hz option

mlyle:
Stop annoying screeching sounds.  16666Hz chosen because it allows
nearly 100% duty cycle, reasonable resolution (720 counts),  and is
getting up to where human auditory acuity is very low.

### DIFF
--- a/ground/gcs/src/plugins/config/output.ui
+++ b/ground/gcs/src/plugins/config/output.ui
@@ -352,6 +352,11 @@ Leave at 50Hz for fixed wing.</string>
                   <string>12000</string>
                  </property>
                 </item>
+                <item>
+                 <property name="text">
+                  <string>16666</string>
+                 </property>
+                </item>
                </widget>
               </item>
               <item row="1" column="4">
@@ -439,6 +444,11 @@ Leave at 50Hz for fixed wing.</string>
                 <item>
                  <property name="text">
                   <string>12000</string>
+                 </property>
+                </item>
+                <item>
+                 <property name="text">
+                  <string>16666</string>
                  </property>
                 </item>
                </widget>
@@ -530,7 +540,7 @@ Leave at 50Hz for fixed wing.</string>
                 </item>
                 <item>
                  <property name="text">
-                  <string>0</string>
+                  <string>16666</string>
                  </property>
                 </item>
                </widget>
@@ -620,6 +630,11 @@ Leave at 50Hz for fixed wing.</string>
                 <item>
                  <property name="text">
                   <string>12000</string>
+                 </property>
+                </item>
+                <item>
+                 <property name="text">
+                  <string>16666</string>
                  </property>
                 </item>
                </widget>
@@ -757,6 +772,11 @@ Leave at 50Hz for fixed wing.</string>
                 <item>
                  <property name="text">
                   <string>12000</string>
+                 </property>
+                </item>
+                <item>
+                 <property name="text">
+                  <string>16666</string>
                  </property>
                 </item>
                </widget>
@@ -902,6 +922,11 @@ Leave at 50Hz for fixed wing.</string>
                 <item>
                  <property name="text">
                   <string>12000</string>
+                 </property>
+                </item>
+                <item>
+                 <property name="text">
+                  <string>16666</string>
                  </property>
                 </item>
                </widget>


kubark42:
Okay, I love your description, but I still have no clue what this is about. :D

mlyle:
The higher frequencies can be used to generate PWM to drive transistors that in turn drive brushed motors.  0 to 100% duty cycle then corresponds corresponds to motor power.

The highest frequency we supported was 12KHz, which produces quite a bit of audible high frequency vibration on some setups-- both around 12KHz and various sum and difference frequencies.  On my setup, shifting that up to 16.6KHz makes the "whine" completely inaudible, because the motors are less effective at making sound of higher frequencies and the ear is less sensitive to higher frequency sounds.

This is my setup.  Thrust to weight exceeds 3:1, and it's incredibly agile-- there's both very little propeller mass to wait to accelerate and there's no delay from ESCs at all.  Flight controller is a Micro SciSky, a Naze32 clone available for ~$30 including onboard drive transistors and a boost converter to produce a 5V rail from 3.7V nominal lipo.   
xaeys6f

Incidentally, the actuator curve fitting code you submitted is really helpful for both controllability at low power and preventing high power oscillations, with a constant of b=0.65.

jihlein:
Off topic but, do you have STL file for frame and motor part numbers?  Looks like some indoor winter fun.

mlyle:
It's a ton of fun and pretty helpful in testing things out, too. http://jar.lyle.org/~mlyle/littlequad.stl is the version as printed.  I'm likely to revise it soon because the props are too close together-- less than a millimeter of space.  I've printed out of Nylon 910 but ABS may actually be better.

The motors are http://micro-motor-warehouse.com/collections/all-motors/products/cl-0820-15 which is really too expensive but I have not found a less expensive source.  They have connectors on the end-- http://micro-motor-warehouse.com/collections/connectors/products/micro-jst-1-25-sockets is a proper mate that fits into the holes on the board (but notice the alternating orientations of the connectors). 

The other thing to note is that you basically need to touch the antenna with the TX to bind-- there's no front-end module on the scisky so it has basically no output power to acknowledge the bind.

peabody124:
LGTM

xfce:
does the cc3d support 16.6KHz pwm output? I want to custom a cc3d drive brushed motor quad

mlyle:
Yes.

xfce:
atomblushed

custom CC3D brushed controller flished , the same size as Micro SciSky.

mlyle:
What's the CHGING light?  Will it charge over USB?  Very neat board.

xfce:
yes i add a battery charger IC(CHGING LED for charging status,  DONE LED for charge complete), and a 5V dcdc IC on the bottom.
i want to add a opensource dsmx receiver, but can't find any data. maybe next version

dustin:
@xfce I've been doing some work with cc2500 modules that might be fun.  I've been playing with [this module](http://www.aliexpress.com/item/CC2500-SPI-2-4GHz-Wireless-Transceiver-Module/32610339676.html) on my desk with some [avr code](https://github.com/dustin/buggles).  Getting that radio work over would be neat.  There are a lot of variants of that with chip antennas as well.

peabody124:
@xfce quite neat looking

xfce:
@dustin I am a hardware engineer, only know a little code, maybe you can help me integrate the RF receiver in the next PCB version, the bottom still have some room for this.
@peabody124 Thanks, now i am waiting for the PCB and 3D print case. If all  hardware are tested OK, and if you like this I will post the gerber and schematic.

dustin:
@xfce that'd be great, yeah.  hshred mentioned having support in the code
for talking to a cc2500 using frsky's D8 protocol might be neat, but not
very interesting without hardware that supports it.  This board looks quite
interesting for that sort of thing.  Basically, the thing you were thinking
about with dsmx or whatever, but with a differently common protocol.

There are a lot that look like this:
http://www.aliexpress.com/item/10pcs-Lot-CC2500-Wireless-Module-Small-Size-2-4GHz-Wireless-Transceiver-Module-Free-Shipping-Dropshipping/1546243568.html
and
are used for building tiny brushed things.

xfce:
@dustin  I mean dsmx, only because i have devo-7e with deviation.
I think it's better to integrate a separate receiver module, because this unit is a full function CC3D, so if you use these SPI RF module, you have to custom the CC3D firmware.

dustin:
@xfce understood.  But we can roll that kind of thing in similarly to how
the rfm22b is.  If we have code for a cc2500 on SPI and then we have a
board for it, it's a similar concept, but with smaller antennas.

I'm not sure off the top of my head what hardware dsmx uses (nrf24L01?) but
a similar concept could happen there.  Kind of thing is great for brushies.

xfce:
@dustin it's CYRF6936 for dsmx. OK i understand you, I will try to integrate cc2500 or  nrf24L01 the next version, maybe nrf24L01, because crazyflie also use this. what's your advise.

dustin:
I don't have much.  I just think it's a neat idea.  :)  My radios are all
either rfm22b or cc2500.  You should do what works for you, though.

xfce:
OK.
in a word, if you like this, i will add a 2.4G IC on this board next version, then opensource  this, and next work will be you developer

mlyle:
@xfce This would make the frsky radios work, which would make a lot of people very happy.  Would love to work against this board.  Do you plan volume production of it?

xfce:
@mlyle I am just a fan of openpilot/taulabs/dronin/librepilot, doing this just for fun
but if you want some of this board, I think maybe i can help to volume production, as some days ago I DIY OP's revonano, buy materials, contact SMT factory, test the boards, send to my friends all by myself.
BTW the final KylinMicro(I call it) 3D PCB as below, give some advice for the next version,thanks.
topveiw
kylinmicro_t
bottomview(updated, add the cc2500 RF module, use CC3D's Flash SPI2 interface, and PC15 for CS)
micro
and the schematic as bellow,  advice please,thanks

@dustin 
@mlyle

Issue #2051, opened 2015-11-09
#2024 Broke parallel builds on nix


PTDreamer:
#2024

PR #2053, opened 2015-11-11
makefile: restore parallelism to unix environments

mlyle:
Fixes bug #2051, introduced in #2024

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -274,7 +274,7 @@ endif
 .PHONY: gcs
 gcs:  uavobjects
 	$(V1) mkdir -p $(BUILD_DIR)/ground/$@
-	$(V1) ( cd $(BUILD_DIR)/ground/$@ && \
+	+$(V1) ( cd $(BUILD_DIR)/ground/$@ && \
 	  PYTHON=$(PYTHON) $(QMAKE) $(ROOT_DIR)/ground/gcs/gcs.pro -spec $(QT_SPEC) -r CONFIG+="$(GCS_BUILD_CONF) $(GCS_SILENT)" $(GCS_QMAKE_OPTS) && \
 	  $(GMAKE) ; \
 	)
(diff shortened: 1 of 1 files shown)


kubark42:
For those of us unfamiliar with the fiber points of Makefiles, could we get a quick link or explanation for what the + changes?

mlyle:
Presence of $(MAKE) verbatim (not as a subexpansion) in a rule produces various changes of what environment variables are exported, including the stuff necessary for submakes to communicate with the job server.  This was changed to $(GMAKE) by the previous PR-- adding + to the beginning of the rule is an alternate way to trigger the same behavior.

The makefiles need a lot of cleanup and audit; there may be better alternatives ultimately, but this is the minimal fix.

tracernz:
This kind of hidden side-effect is really annoying. :( Relevant docs: https://www.gnu.org/software/make/manual/make.html#MAKE-Variable

mluessi:
+1 for merge

PTDreamer:
I believe I tried this approach and it broke MSVC builds. I have another fix to this same issue on a branch I keep having to rebase https://github.com/TauLabs/TauLabs/commit/09bb3e4e3a94e3bfc3972228478311f850a12d0b
Since this is a small change I wont mind having to fix the rebase again, and deal with possible issues.
The branch I referred is for a GCS autoupdater and since I'm currently dealing with OSX I wont have time to test this with MSVC this next few days.

peabody124:
@PTDreamer since you reported the original bug, can you confirm this fixes it?

peabody124:
screen shot 2015-11-15 at 9 54 20 am

Output on parallel builds looks odd for a time

mlyle:
That's the uavobject libraries.  They don't currently set a target name.

PTDreamer:
> @PTDreamer since you reported the original bug, can you confirm this fixes it?

Not only I reported but I also created it.
@mlyle can you have a look at the commit I referred above and see if you agree with that approach? The thing is that I believe I tried the "+" since it is the most usual way and it broke MSVC compile as far as I can remember.

mlyle:
I'm fine with that approach.

It really sucks to not have fast parallel builds in the meantime though.

PR #2055, opened 2015-11-11
package: restore firmware to linux packages.

### LINKED ISSUE #2054: Jenkins: Firmware missing from recent builds
@lilvinz Would be great if you could take a look at this. Reported by @cGiesen.

mlyle:
Fixes #2054 .  There was an interaction between the packaging scripts
in make standalone triggering an all_clean and the Jenkins packaging scripts
assuming that previous flight targets would still exist.  It's compounded
by the fact that there's basically no error checking on packaging.

### DIFF
--- a/package/Makefile
+++ b/package/Makefile
@@ -77,8 +77,7 @@ help:
 	@echo "       This usually is safe."
 	@echo
 
-# Clean and build uavobjects since all parts depend on them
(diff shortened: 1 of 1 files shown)


mlyle:
Verified linux packages now have firmware blobs appropriately.

tracernz:
LGTM

Issue #2057, opened 2015-11-14
In-flight reset


dustin:
I've had a sparky 1 reset in flight due to the independent watchdog firing.  It seems to be easy to reproduce for me.

Example of the reboot happening in flight here:  http://gfycat.com/SelfassuredMenacingGourami

What you see there is flying, battery alarm goes off (not sure why) then independent watchdog caused a reboot.  This happened three times in ~half hour there.  Possibly happened before in a catastrophic crash that led me to writing code to communicate alarms to the OSD.

This particular flight was running something like this code:  https://github.com/dustin/TauLabs/tree/mspalrm  (tl/next + the work I've done for MSP alarm communication).  The MWOSD side of that is still sitting in a [PR queue](https://github.com/ShikOfTheRa/scarab-osd/pulls).  Basically merge my three changes.

I've got battery monitoring and MSP enabled.

Attached my config:

[f22pretest.txt](https://github.com/TauLabs/TauLabs/files/34487/f22pretest.txt)

dustin:
I increased the battery module stack by `128` (for ease of hackery, `#define STACK_SIZE_BYTES (496 + 128)`).  I managed to poke around and get the stack remaining value to `124`.  This seems like pretty red-flaggish.

mlyle:
Fixed by #2058.

dustin:
I just flew four batteries through my test aircraft that rebooted three times in two batteries last weekend.  I was not very gentle, but saw no reboot issues.  Had alarms firing all over the place, a few crashes, etc...  If I ever could repro, I can't now.

tracernz:
Nice work @dustin!

PR #2058, opened 2015-11-14
Increase stack for battery module.

dustin:
See #2057 for more details, but after increasing it by 128, I was able
to get it to report 124 bytes free after a bit of jostling (mostly
toying with power levels from a bench supply).

I'm not 100% sure that this is the cause, but I was getting some strange
behavior resulting in watchdog reboots on this particular build.  This
experiment with the battery module indicates it's *possible* that the
battery module was the cause of these issues.

This change increases it by 80 which should be enough to keep things
contained based on current observations.

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -38,7 +38,7 @@
 
 // ****************
 // Private constants
-#define STACK_SIZE_BYTES            496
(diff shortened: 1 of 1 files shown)


mlyle:
Fixes #2057, a high criticality bug.  Merging.

PR #2060, opened 2015-11-15
picoc: add single precision math, enable colibri again

ernieift:
Adds math functions like sinf() to math.h in order to use FPU for
faster calculations.
Introduces compile switches for better integration on smaller targets
(`PICOC_NO_FP`, `PICOC_NO_DMATH` and `PICOC_NO STRINGS`).
Enables PicoC on colibri again without double precision math (in math.h).

The flash size on Colibri is now 242k and on the full featured Quanton 296k (+5). With `PICOC_NO_FP` it reduces to 225k on Colibri. So picoc needs ~24..68k depending on the configuration.


kubark42:
@mlyle, you mention in https://github.com/TauLabs/TauLabs/pull/2047#issuecomment-157798064 that you feel this is a step backwards in what you're trying to do by reducing compile flags. In a sense of not letting perfect be the enemy of good, do you think this can merge for the time being and we can see what people start doing with F3 and PicoC? It looks like the only LOC that is outside PicoC is ```#define PICOC_NO_DMATH```.

Is that an okay compromise for the time being?

mlyle:
At this point I think we should just close both PRs, and I'll work on getting llvm/clang embedded in GCS.

peabody124:
My 2c: if this PR has the same flash/ram footprint as #2047 on F3, then I'd be inclined to prefer it, if nothing else than to just defer to the original maintainer. Code equivalence shouldn't be an issue because we can use sinf on both F3 and F4. This also means if someone expected double performance (god knows why) then they won't be surprised not to get it.

If there is a performance/size benefit of 2047 I'd prefer it though.

Ultimately we don't have a ton of picoC users so changing sin->sinf won't be a big deal for anyone.

ernieift:
To avoid misunderstandings, this should not be an "attack" on #2047 . We have different understandings for what PicoC is good for. This is perfectly okay and this is only a proposed solution.
I had the impression that some discussants so long wanted to shorten the interpreter, until it finally fits in the F3 for an concentrated application. BTW PicoC needs about >15k RAM for stack, variables and script to run properly.
That would limit the functionality of the program. I have not yet had serious performance bottlenecks PicoC (looks my videos with PID control).
To get an interpreter running for F3 even F1 in a near distance racing applications, what about an external solution. In my opinion, the F1 is currently the most used controller (CC3D/NAZE32) for racing.
We have the gcs running on Android. All FCs can communicate over BT telemetry which works in the near distance. If you want to go on racing competions without CPU doping, you can do it with a referee laptop and precoded BT sticks. The script is running on the laptop an handles the UAVOs over telemetry.

mluessi:
I think this is a good measure until we have LLVM/CLANG embedded in the GPS. It let's us disable double precision support on some targets and we can encourage users to use `sinf` etc unless there is a very good reason not to do so. So only scripts that really need double won't run on the smaller targets.

mlyle:
I think the solution I favor now is to just embed clang/LLVM in GCS.  We can generate binary blobs that run on the micro without any performance limitation or resource use.  We could move a lot of the modules to being deployed this way (CameraStab?  UAVO*Bridge?), freeing up a lot of flash and resources.

mlyle:
This is a win for colibri, so if my thing isn't going to merge this should.  I prefer a unified language.

One compromise that could make it a little better is to rename the existing functions to add 'd' to the end, so people don't "accidentally" use the additional precision versions and write code that won't work on colibri.

kubark42:
Okay, thanks for the comments @mlyle. I'll merge this as is and @ernieift can open a new PR to extend as per your suggestion.

I think it makes a lot of sense in general for PicoC to implement that kind of single/double pattern. We can't be the only project that has this kind of constrained resource platform.

mlyle:
Bummed we didn't do the above suggested compromise.

kubark42:
I personally think there are good pros and cons*, but this gets the F3 people going so they can start playing with PicoC right away. Otherwise, they have to wait for us to resolve what is the best way to disambiguate between the two.

*Pro: works exactly the way @mlyle suggests. No chance of someone accidentally using the wrong function.
*Con: Changes a long-standing C pattern without improving functionality. Will lead to confusion since PicoC is supposed to run C.

mlyle:
PicoC is not supposed to run C.  It's a small, weird subset.  And it doesn't get the F3 people going; it's crammed into colibri.

I'm giving up on this crap.

dustin:
The con is that this is more confusing for users and makes picoc even less desirable.  There's less motivation to backport it because every user who picks it up and needs to do anything touching a floating point number is going to get really confused when `floor` or `pow` doesn't work except in very rare cases where someone wants a script that will only run on specific hardware.

This just looks like a "just do both" management call that leads to a harder to use product that is more complicated because of how engineering compromises are made.

The effect is that you take a small group of users (people interested in control systems), then a subset of that (people who will go against the majority and try Tau Labs), then an even smaller subset of that (people who might try to do anything with PicoC) and intersect them with people who are doing floating point math in these scripts, and then assume that that remaining fraction of a user requires double precision.  That's already narrowing us down to something less than 1, but possibly greater than 0 users, and we're already thrusting options on that user to consider based on what hardware it might need to run on.

It's a really unfortunate situation.  There's a lot of talk about making things easier, but this is a pretty clear example of pushing our own complexity to the user.

peabody124:
FWIW we actually still can't get it into F3. I just tried.

Base configuration: 63KB over flash limit
NO_DMATH configuration: 47KB over flash
NO_FP configuration: 30KB over flash
NO_FP and NO_STRINGS: 29KB over flash

This doesn't even get into memory limitaitons

mlyle:
:) Yah i guess it was technically impossible and I'm full of shit.

peabody124:
I didn't say that. Just that more work is required if we want it on F3 than using floating point instructions.

Incidentally we should update the picoC wiki https://github.com/TauLabs/TauLabs/wiki/User-Guide:-PicoC to document the float operations

mlyle:
Yah, and no one is going to do that work now because of disgust.

fujin:
:-1:

kubark42:
This is exactly why I personally feel the F3 targets aren't worth our time. There's no sense us arguing amongst ourselves about a few KB of memory in 2015.

I know we're all onboard with the value of following standards. PicoC is supposed to have a subset of C commands. So doing things in non-standard ways because a popular target this year has too little memory doesn't make sense. This PR is compliant with the existing C standard that, for example, `cosf()` uses single precision float point and `cos()` uses double.

I apologize for any controversy but the decision to merge was a result of the buy-in from all parties. The suggested additional upgrade is stylistic not functional and needs to be considered separately because its out of scope.

mlyle:
Bye.

Bye-in' from all parties.

dustin:
> This is exactly why I personally feel the F3 targets aren't worth our time. There's no sense us arguing amongst ourselves about a few KB of memory in 2015.

I want to agree with you, but you're tripling the cost of any potential user's flight controller in this statement.  The user base is small, and this keeps it small.

It's unclear what the point of Tau Labs is, but I'm thwarted by every potential entry point of adoption.

I like the codebase abstractly -- it's got a lot of potential.  F1 is basically broken.  F3 is frowned upon.  F4 targets violate core principles (unless they're popular, then they're prohibited).  Cool scripty stuff isn't generally available.  Nav needs enough work that nobody would recommend it.  There's aborted(?) OSD work, gimbal work I assume might work, but I've never heard anyone say it's great.  The most interesting and actually working aspects of the software are basically declared "not interesting" and not worth time.

What are you guys actually trying to do?  Whatever it is, you're making it very hard.

> the decision to merge was a result of the buy-in from all parties

I really wish the party were larger.  I was really interested in joining it.

This makes things Tau Labs worse and for dumb reasons.  There are effectively no users of PicoC and the places where I've been advocating it, I'm just not going willing to support any further.  This complexity benefits nobody.

mlyle:
> What are you guys actually trying to do? Whatever it is, you're making it very hard.

That's the real issue.  I want to do cool stuff, not stress out and argue.  That's not happening here.

ernieift:
Sorry my delayed answer (haven't got a stable internet account the last days). When I wrote on an interpreter, it was my goal to run it on F3 and F4. I had the decision between C or Basic. I took C for its flexibility. At this time the sparky2 wasn't born and we had only a few targets which supports PicoC. Now there are more F4 available and the interest on scripting grows.
If you want to bring scripting on F3, please try uBasic or rewrite the mechanism of PET Basic with a subset of C64 commands. This takes much less RAM but the dialect is ugly when using structures. You have the write commands and functions for every single UAVO (much more than in PicoC). But for simple things it were an alternative. IMO also very cool and suitable for @dustin ideas.
@mlyle there is no "d" in standard C at the end of the double precision math functions. I hope someone is learning a little C when playing around with PicoC scripting and I don't want to lead in a wrong direction. There is still a VM state machine in our targets. When you could create an Assembler for it, you could use it for any scripting too. It takes a lot of work and care but has the advantage of speed and memory.

kubark42:
> I want to agree with you, but you're tripling the cost of any potential user's flight controller in this statement. The user base is small, and this keeps it small.

I don't have an answer for this. Being in a position to be looking at buying LOTS of these chips, the price difference between the F3 and F4 is marginal. If you're working with the kind of margins that make VCs drool, you're looking at a MRSP difference of less than $5. So I can't explain the costs problems.

Not that that makes any difference, mind you. No one gets anywhere by tilting at windmills, so if the mass market is F3 then that will stay that way until we offer a compelling reason to go F4.

> It's unclear what the point of Tau Labs is, but I'm thwarted by every potential entry point of adoption.

That's kind of up to each person to determine? I'm not speaking for Tau Labs. I want many things that have ultimately not moved forward within the organization. It's herding cats-- and that's going to be true in any organization run completely by volunteers. The best I can hope for is to come into contact with motivated, talented developers like you guys. This might change with the proposed governance, because we can set a direction and use that as a rationale for making code compromises because we have a clear definition of the greater good.

> Nav needs enough work that nobody would recommend it.

Way back when I tried changing that, but I wasn't able to get past the drama. The way nav is done was doomed from the beginning-- KISS approaches don't work with things as rich as autonomous behaviors-- but I wasn't able to make the case convincingly enough. The upshot is we have this architectural base which is gawdawful beautiful, but we can't do anything more than manual flight with it because all the important high-level parts are missing.

All the serious professional stuff goes on at PixHawk-- which is a nowhere near as clean a codebase based on my recent perusings-- because they have a complex backend that makes the front end simple. PixHawk got so far ahead because it avoided the drama that plagued OP and then TL.

I have high hopes that slimmer governance will give the project a firmer direction that people can rally behind, or understand why it's not going in the direction they want. That will help tons with the frustration.

PR #2062, opened 2015-11-17
WIP WIP: Event callback drastic refactor

mlyle:
This is a branch I started work on today.  It is likely to be very controversial.  These are the goals:
- Eliminate the event system thread.
- Pass the actual UAVOs to callbacks to avoid multiple chained rapid traversals of locks, task switching, and associated overhead.
- Add a "context pointer" to callbacks to make it more easily possible to have multiple instances of subsystems and shared implementations of callback code between modules.

Items I still plan to implement:
- ~~A dedicated callback stack so that analysis of stack utilization is intuitive.~~
- ~~Folding the remaining event system role (periodic callbacks) into System module.~~
- ~~Prevention of auto-recursion on callbacks and improvement of callback recursion in general.~~

Note that this means that setting UAVOs can directly invoke the callbacks-- perhaps in a very high priority task.  It's my opinion this is better than before, because:
- Nearly all tasks touch UAVOs
- Nearly all of the callbacks currently trigger accesses to UAVOs, obtaining the giant UAVO lock and excluding higher priority tasks anyways
- The actual task switching / priority inheritance / obtaining of the locks are very expensive operations, so strolling through them less is a win in total execution time.
- Many tasks follow a pattern where they get inputs in the beginning and set outputs in the end; this "end time" is when it is best to be delayed.
- It is possible to obtain flight parameters synchronous to when they are generated, e.g. for autotune or logging.  The elimination of dropped and duplicated samples from race conditions in task activation is a big potential win for the system's consistency and flight performance.

Current status:  This works in the simulator.  Further cleanup and the above work is required, along with a fair degree of flight test.


lolorc:
that's impressive work !
what about creating a new branch to test that on as soon as possible ! can you name that branch betalabs pretty-please !? ;-)

kubark42:
Very interesting. I'm thinking that certain things like telemetry, HiTL, and logging might show some big improvements.

flight/Modules/Airspeed/airspeed.c
```diff
@@ -176,7 +176,7 @@ MODULE_INITCALL(AirspeedInitialize, AirspeedStart);
  */
 static void airspeedTask(void *parameters)
 {
-	AirspeedSettingsUpdatedCb(AirspeedSettingsHandle());
+	AirspeedSettingsUpdatedCb(NULL, NULL, NULL, 0);
```
kubark42: This ```NULL, NULL, NULL, 0``` paradigm is used so frequently I would suggest having a wrapper function. This will help readability and maintainability, since long lists of ```NULL```s and ```0```s can be difficult to digest and check for correctness.

flight/Modules/Airspeed/airspeed.c
```diff
@@ -359,22 +361,25 @@ void baro_airspeedGet(BaroAirspeedData *baroAirspeedData, uint32_t *lastSysTime,
...
 
-static void AirspeedSettingsUpdatedCb(UAVObjEvent * ev)
+static void AirspeedSettingsUpdatedCb(UAVObjEvent * ev, void *ctx, void *obj, int len)
 {
-	AirspeedSettingsData airspeedSettingsData;
-	AirspeedSettingsGet(&airspeedSettingsData);
-	
-	airspeedSensorType=airspeedSettingsData.AirspeedSensorType;
-	gpsSamplePeriod_ms=airspeedSettingsData.GPSSamplePeriod_ms;
+	(void) ev; (void) ctx; (void) obj; (void) len;
```
kubark42: I think we generally avoid multiple declarations/actions on a single line. To be honest, I'm a little puzzled by what's going on here. These are casts without a destination, right?

flight/Modules/Airspeed/airspeed.c
```diff
@@ -359,22 +361,25 @@ void baro_airspeedGet(BaroAirspeedData *baroAirspeedData, uint32_t *lastSysTime,
...
+
+	AirspeedSettingsData *airspeedSettingsData = obj;
```
kubark42: Isn't this ```obj = NULL``` here? Wouldn't that cause trouble below when the pointer is accessed?

flight/Modules/FixedWingPathFollower/fixedwingpathfollower.c
```diff
@@ -669,16 +668,22 @@ static uint8_t updateFixedDesiredAttitude()
...
+
 	if (ev == NULL || ev->obj == FixedWingPathFollowerSettingsHandle())
 		FixedWingPathFollowerSettingsGet(&fixedwingpathfollowerSettings);
 	if (ev == NULL || ev->obj == FixedWingAirspeedsHandle())
 		FixedWingAirspeedsGet(&fixedWingAirspeeds);
 }
 
-static void airspeedActualUpdatedCb(UAVObjEvent * ev)
+static void airspeedActualUpdatedCb(UAVObjEvent * ev, void *ctx, void *obj,
+	int len)
```
kubark42: Kinda weird to see this line wrap.

flight/Modules/FlightStats/flightstatsmodule.c
```diff
@@ -51,14 +51,13 @@
 // Private variables
 static bool module_enabled;
 static struct pios_thread *flightStatsTaskHandle;
-static FlightStatsSettingsData settings;
+static volatile FlightStatsSettingsData settings;
```
kubark42: I wonder if this won't be a little hard to maintain because it takes care to notice that this should be volatile. Any way to improve that?

flight/Modules/Airspeed/airspeed.c
mlyle: This is to mark the variables as "used"-- doing nothing with them-- so you don't get a warning about them being unused -- an alternative to __attribute__((unused)) in the declaration.  Our build flags don't currently complain about unused parameters but I wanted to avoid making them worse.

flight/Modules/Airspeed/airspeed.c
mlyle: The unfortunate thing is that it's the per-module callback function, so adding a wrapper in each module doesn't necessarily make it cleaner IMO.

An alternative would be to make the registration functions guarantee to generate at least one callback quasi-immediately so that we don't need to do it explicitly.  It would also remove race conditions-- since this could be running "at the same time" with a different version of the data if tasks switch at the wrong time.

flight/Modules/Airspeed/airspeed.c
mlyle: That's indeed bad in this one.  It's another argument for guaranteeing one immediate callback on registration.

flight/Modules/FlightStats/flightstatsmodule.c
mlyle: If you don't make it volatile you get a compile error (because the type is volatile in the registration function for the copy-callback registration).  This is at least correct, too, where the old version wasn't and could lead to undefined behavior.

flight/Modules/Airspeed/airspeed.c
kubark42: Gotcha.

Since this is boilerplate code, maybe attaching a similarly boilerplate comment might reduce confusion?

flight/Modules/FlightStats/flightstatsmodule.c
kubark42: Awesome. Sounds like it can't slip through, which is just peachy.

mluessi:
I haven't yet looked at the code in detail, but I like the premise of this PR. The ability to get the actual UAVO data in a callback without needing to copy / hold the lock would be very valuable.

flight/Modules/Airspeed/airspeed.c
```diff
@@ -359,22 +361,25 @@ void baro_airspeedGet(BaroAirspeedData *baroAirspeedData, uint32_t *lastSysTime,
...
+
+	PIOS_Assert(sizeof(*airspeedSettingsData) == len);
+
+	airspeedSensorType=airspeedSettingsData->AirspeedSensorType;
```
PTDreamer: Mixed style regarding spaces around operator.

flight/Modules/Airspeed/airspeed.c
```diff
@@ -104,10 +104,10 @@ static int8_t airspeedADCPin=-1;
 // Private functions
 static void airspeedTask(void *parameters);
 void baro_airspeedGet(BaroAirspeedData *baroAirspeedData, uint32_t *lastSysTime, uint8_t airspeedSensorType, int8_t airspeedADCPin);
-static void AirspeedSettingsUpdatedCb(UAVObjEvent * ev);
+static void AirspeedSettingsUpdatedCb(UAVObjEvent * ev, void *ctx, void *obj, int len);
```
mluessi: I'm not a big fan of very long parameter lists.. maybe we could use a "CallbackCtx" struct pointer that has all that info?

flight/Modules/Airspeed/airspeed.c
mluessi: See my comment above about using a struct

flight/Modules/Airspeed/airspeed.c
mluessi: This also wouldn't be necessary if we pass a pointer to a single struct with all the info.

flight/Modules/System/systemmod.c
```diff
@@ -67,14 +65,45 @@
...
+	UAVObjEvent ev; /** The actual event */
+	UAVObjEventCallback cb; /** The callback function, or zero if none */
+	struct pios_queue *queue; /** The queue or zero if none */
+} EventCallbackInfo;
+
+/**
+ * List of object properties that are needed for the periodic updates.
+ */
+struct PeriodicObjectListStruct {
+	EventCallbackInfo evInfo; /** Event callback information */
```
mluessi: fix indent

flight/Modules/System/systemmod.c
```diff
@@ -67,14 +65,45 @@
...
+    uint16_t updatePeriodMs; /** Update period in ms or 0 if no periodic updates are needed */
+    int32_t timeToNextUpdateMs; /** Time delay to the next update */
```
mluessi: we should be able to get aways with `uint16_t`, as the period is 16 bit as well

flight/Modules/System/systemmod.c
```diff
@@ -618,6 +693,218 @@ void vApplicationStackOverflowHook(uintptr_t pxTask, signed char * pcTaskName)
...
+	// Check that the object is not already connected
+	LL_FOREACH(objList, objEntry)
+	{
+		if (objEntry->evInfo.cb == cb &&
+			objEntry->evInfo.queue == queue &&
+			objEntry->evInfo.ev.obj == ev->obj &&
+			objEntry->evInfo.ev.instId == ev->instId &&
+			objEntry->evInfo.ev.event == ev->event)
+		{
+			// Already registered, do nothing
```
mluessi: If the entry already exists but has a different callback period, we should just update the period (or if it doesn't have a period, delete the old entry and create a new one etc.). Unless you want the same callback with multiple different periods (which doesn't make sense).

flight/Modules/System/systemmod.c
```diff
@@ -618,6 +693,218 @@ void vApplicationStackOverflowHook(uintptr_t pxTask, signed char * pcTaskName)
...
+			PIOS_Recursive_Mutex_Unlock(mutex);
+			return -1;
+		}
+	}
+    // Create handle
+	objEntry = (PeriodicObjectList*)PIOS_malloc_no_dma(sizeof(PeriodicObjectList));
```
mluessi: indentation is off

flight/Modules/System/systemmod.c
```diff
@@ -44,7 +44,9 @@
 #include "watchdogstatus.h"
```
mluessi: Whats the rationale for moving all the periodic callback stuff from the uavo manager to systemmod? To me, it seems more logical to have it in the uavo manager.

flight/Modules/Airspeed/airspeed.c
kubark42: Good idea. That would resolve my two comments regarding the ```NULL, NULL, NULL, 0``` pattern and the ```(void) ev; (void) ctx; (void) obj; (void) len;``` declaration.

flight/Modules/System/systemmod.c
mlyle: All of this is moved as originally indented for clarity.  Can be reindented later.

flight/Modules/System/systemmod.c
mlyle: This is all original periodic callback code.

flight/Modules/System/systemmod.c
mlyle: The event task used to both do the work of periodic callbacks and of pumping events from a queue to user-defined callbacks.  Since this is now direct, there is much less need for it as a separate task.

Also, periodic callbacks conform to the UAVO callback signature but really aren't related.

flight/Modules/Airspeed/airspeed.c
mlyle: I don't think 4 parameters is a long argument list.  I am against this.

1. It is an even-greater amount of error-prone manual adjustment to existing code in the same PR to put ev into that.
2. It's really important to keep the user-context pointer as a separate argument.
3. It increases marshalling and unmarashalling for no reason.  ARM ABI dedicates 4 registers for parameter lists before chewing stack, so the microarchitecture even dedicates the most core resource (registers) assuming that this many arguments are common.

flight/Modules/Airspeed/airspeed.c
mlyle: existing code

flight/Modules/Airspeed/airspeed.c
kubark42: It's not the 4 parameters that bother me so much as the default call with lots of ```NULL```s and ```0```s. That's not a pattern I'm particularly fond of, and I think we're seeing how it can be creaky with the HAL refactoring.

So if a struct pointer isn't a good approach, is there another idea for making that a little more transparent?

mlyle: It's not the "default call".  It's the existing code.  People abuse the callback by putting the entirety of configuration update handler stuff in it (very resource intensive and potentially unsafe in multiple ways).  Then they want to invoke the handler directly because they're not guaranteed that the callback ever gets called or when that first call will happen.  So they pump a "fake event" with no details (NULL, NULL, NULL, 0).

mluessi: @mlyle can you elaborate on this? 

1. Isn't a good argument IMO; yes it results in more changes but also cleaner code in the long term.
2. Why is it really important?
3. It's true that up to 4 arguments can be passed using registers, but this doesn't mean one should always use  4. It just means the ARM designers decided people rarely use more than 4 arguments, but using fewer than 4 arguments is still better.

mlyle: 1 is a great argument.  There's enough to verify and correct on this path of code change already without incurring a ton more risk.
2 is important for callback efficiency.  Most likely use of this parameter is to be assigned to a local variable struct my_state *foo = ctx;    Then do stuff with foo->whatever.  The compiler will elide the assignment and then access to foo->whatever is just an indirect load.  If it's a structure pointer, then every time you escape the optimizer's analysis by calling into another module you need to load, load, load again. 

It also lets us use things like restrict, which we can't do meaningfully with structures being passed in.  These things make a big difference on performance on giant machines like I'm used to writing code for; it's even bigger down here.

Basically, I don't see how

```
struct foo { .ctx=blah, .obj=obj, .len=len, .event=ev } bar;
cb(&bar);
```

is preferable to `cb(ev, blah, obj, len);`, and it's obviously less efficient on several axes and wordier.

mlyle:
Some further work here, still not complete.  Tested on bench on Naze32.  With a simple config (autotune module enabled, 333Hz), disarmed on bench, next uses 68% CPU and has 152 bytes of heap free.  This branch uses 63% and has 1128 bytes free, though we'll end up claiming a few hundred bytes of this to use for a callback stack.

mlyle:
P.S. No one should fly this until I'm much further along and have read this a couple times.

flight/Modules/Airspeed/airspeed.c
mluessi: I disagree with the wordier part. It's true that it is wordier for the caller, as it has to implement what you wrote above, but there is only one caller (the system module), whereas in you version, every implementation of a callback needs to add something like `(void) ev; (void) ctx; (void) obj; (void) len;` in order not to generate compiler warnings and the callback signature is wordier as well. In addition, if there are every any additions / modifications, it means that every callback function will need to be modified, whereas a struct could easily be extended.

It isn't clear to me what the use case of the "ctx" parameter is; can you give an example of it's intended purpose? What you said regarding compiler efficiency is true, but one could use `struct my_state *foo = ctx.ctx;` and after that it would be equally efficient to access  `foo->whatever`.

mluessi:
@mlyle good to see that there are memory and CPU utilization improvements :). My comments above are basically just about minor implementation details and I'm all in favor of this work.

flight/Modules/System/systemmod.c
mluessi: Well, whoever implemented the original code is clearly an idiot ;)

tracernz: Vassilis, the original author of UAVTalk, wrote it https://github.com/TauLabs/TauLabs/commit/aab47d6ed42a1728e766da6a795c3eb9c51e5225#diff-079e1e527792eda2b6a009d1a8f688f7L62 with both fields as int32_t ;-P. One of the fields was changed to uint16_t by someone else later https://github.com/TauLabs/TauLabs/commit/aab47d6ed42a1728e766da6a795c3eb9c51e5225#diff-079e1e527792eda2b6a009d1a8f688f7L62

mluessi: @tracernz I thought it came from #2030 and I wrote it based on a proof-of-concept by @mlyle. Somewhat confused now, as #2030 basically implemented the same thing (or so I thought).

flight/Modules/System/systemmod.c
mluessi: Basically what `connectObj` currently does: https://github.com/TauLabs/TauLabs/blob/next/flight/UAVObjects/uavobjectmanager.c#L1848

flight/Modules/Airspeed/airspeed.c
mlyle: The context record is so you can call the same callback multiple times for different "purposes".

This is important if we are ever going to have multiple instances of a module running.

For instance, if, say, CameraStab takes callbacks, you can (eventually) instantiate the CameraStab subsystem twice, have it register twice, and it uses the ctx variable for its state (instead of just peeking at globals which doesn't allow you to have more than one, at least not without a lot of code to handle multiple instantiations).  Or telemetry or whatever.  

Another instance of how I use it is the UAVObjCbSetFlag / UAVObjCbCopyData-- they're code that implements a very common pattern of setting a flag or copying a uav object to a given place when it changes.  The context pointer is the place where the module keeps the flag or where to copy the data to.

flight/Modules/System/systemmod.c
mlyle: @mluessi #2030 implements throttling for UAVO events.  This implements periodic callbacks that have nothing to do with uavo events (they fire at the same rate no matter what is happening in UAVOs), but use the same type signatures and have a UAVObjEvent structure for almost-no-reason.

flight/Modules/Airspeed/airspeed.c
mlyle: Well, it would be just (void)the_single_struct;

mlyle:
Verified to fly in basic case (no one else should do this :)

(On Naze32 micro quad with DSM-- attitude, acro, and ran an autotune--- but doh didn't have data gathering on this branch)

kubark42:
Impressive difference. I echo @mluessi's comments.

flight/Modules/System/systemmod.c
kubark42: :)

mlyle:
This doesn't currently work, looking into why.

mlyle:
Works on micro brushed Naze32 quad.  504 bytes free w/ autotune after allocation of a callback stack.  A few little cleanup items remaining:
- ~~Fix the airspeedactual bug~~
- ~~Fix XXX comments~~

Eh, not for this PR:
- Some reindents
- Context records also in periodic callbacks?

Also need to test on larger targets / run more sophisticated flight modes.
- ~~Test on Sparky2, basic flight / autotune~~
- Test on Sparky2, poshold
- ~~Test on Pipxtreme / radio telemetry~~

I think the idea of guaranteeing-at-least-one-callback is a good one, but should wait for a subsequent PR.

mlyle:
This current variant flies well too.  Test to be completed.  Ready for real review.

flight/targets/naze32/fw/FreeRTOSConfig.h
```diff
@@ -81,7 +81,7 @@ NVIC value of 255. */
 
 /* Enable run time stats collection */
 #if defined(DIAG_TASKS)
-#define configCHECK_FOR_STACK_OVERFLOW	2
+#define configCHECK_FOR_STACK_OVERFLOW	0
```
kubark42: Not that it bothers me, but why turn this off when ```DIAG_TASKS``` are specifically requested?

mlyle: Because if there's a context switch in the middle of using the dedicated callback stack (somewhat unlikely, but will happen eventually), it will erroneously fire the stack overflow event and lock up the system or reset.

kubark42: So is there every a time when we'd want ```configCHECK_FOR_STACK_OVERFLOW```?

flight/UAVObjects/uavobjectmanager.c
```diff
@@ -1658,24 +1681,58 @@ void UAVObjIterate(void (*iterator) (UAVObjHandle obj))
...
+static void __attribute__((used)) realInvokeCallback(struct ObjectEventEntry *event,
+		UAVObjEvent *msg, void *obj_data, int len);
 
+static void realInvokeCallback(struct ObjectEventEntry *event,
+		UAVObjEvent *msg, void *obj_data, int len) {
+	event->cb(msg, event->cbInfo.cbCtx, obj_data, len);
+}
+
+#if (!defined(SIM_POSIX)) && defined(__arm__)
+static void invokeCallback(struct ObjectEventEntry *event, UAVObjEvent *msg,
```
kubark42: Can I ask that you comment the crap out of this? I admire the assembly code and appreciate its need, but I worry that it's not easily maintainable.

I know you've already done a lot, but even with the comments I'm not sure of 1) my ability to review the code for sanity and 2) my ability to maintain the code if necessary.

mlyle: I'll try -- I've tried to put in everything relevant that I could think of, but it's worth another pass to see if anything more can be noted.

Also if there's anything particularly crazy looking let me know and I can explain / comment that in particular.

flight/Modules/PipXtreme/pipxtrememod.c
```diff
@@ -1,238 +0,0 @@
-/**
```
kubark42: Why was this entire module taken out? Is that in scope for this PR?

flight/targets/naze32/fw/FreeRTOSConfig.h
mlyle: Unfortunately this pretty much breaks it.

Mode 1, the "cheap" check (inspecting the stack pointer), is broken with this.  Mode 2, the "expensive" check (checking the end of the stack for evidence of overflow), actually still works fine.  Unfortunately, using it implies also doing mode 1 in freertos.

flight/Modules/System/systemmod.c
```diff
@@ -67,28 +65,61 @@
...
+#define STACK_SIZE_BYTES 1024
 #endif
 
-#define TASK_PRIORITY PIOS_THREAD_PRIO_LOW
+#define TASK_PRIORITY PIOS_THREAD_PRIO_NORMAL
+
+/* Generates an armed LED blink of 4.4Hz, close to the previous 5, and is
+ * a nice prime number to not generate as nasty beat frequencies as other
+ * choices */
+#define SYSTEM_UPDATE_PERIOD_MS 113
```
kubark42: Heh, nice!

flight/Modules/System/systemmod.c
```diff
@@ -67,28 +65,61 @@
 #if defined(PIOS_SYSTEM_STACK_SIZE)
 #define STACK_SIZE_BYTES PIOS_SYSTEM_STACK_SIZE
 #else
-#define STACK_SIZE_BYTES 924
+#define STACK_SIZE_BYTES 1024
 #endif
 
-#define TASK_PRIORITY PIOS_THREAD_PRIO_LOW
+#define TASK_PRIORITY PIOS_THREAD_PRIO_NORMAL
```
kubark42: How comfortable are we changing the priority?

flight/Modules/System/systemmod.c
```diff
@@ -54,10 +56,6 @@
 #define DEBUG_MSG(format, ...)
 #endif
 
-// Private constants
-#define SYSTEM_UPDATE_PERIOD_MS 1000
-#define LED_BLINK_RATE_HZ 5
-
 #ifndef IDLE_COUNTS_PER_SEC_AT_NO_LOAD
 #define IDLE_COUNTS_PER_SEC_AT_NO_LOAD 995998	// calibrated by running tests/test_cpuload.c
```
kubark42: Unrelated, but I wonder if this is still correctly calibrated...

flight/Modules/System/systemmod.c
```diff
@@ -67,28 +65,61 @@
...
-#define STACK_SIZE_BYTES 924
+#define STACK_SIZE_BYTES 1024
 #endif
 
-#define TASK_PRIORITY PIOS_THREAD_PRIO_LOW
+#define TASK_PRIORITY PIOS_THREAD_PRIO_NORMAL
+
+/* Generates an armed LED blink of 4.4Hz, close to the previous 5, and is
+ * a nice prime number to not generate as nasty beat frequencies as other
+ * choices */
```
kubark42: Heh, nice!

flight/Modules/System/systemmod.c
mlyle: Before, the event system used to run at high priority to pump periodic callbacks and to pump events.  SystemMod ran at low priority and did very little-- configuration checks and blinked an LED.

This PR changes things somewhat.

* There's no task switch involved in pumping a callback, so it runs at whatever priority the task was that generated the callback.  However it holds an important lock (the UAVO lock), so it is likely to get promoted to the priority of any pre-empting threads by priority inheritance.  IMO this is just about perfect.
* Periodic callbacks that used to run at high priority now run at normal priority in the system task.  It doesn't look like that is bad for any of the existing periodic callbacks.
* The configuration check and heartbeat LED run at normal priority.  IMO bumping the configuration check frequency is good, but one "feature" is lost.  If highest/high/normal tasks are using all the CPU, the "heartbeat" LED will stop or slow down in current code.  Bumping it to priority normal means that this will not happen as much if normal prio tasks are crowded.

There's one thing that is kinda bad with current code and this PR.  If a low priority task pushes something to a queue that goes to a high priority task, a bad little choreographed sequence of events happens.  The low priority task gets preempted, there's a (expensive) task switch to the high priority task, which attempts to get the uavo lock, blocks, increases the priority of the low priority task, which completes its actions and unlocks, and then (expensively) switches to the expensive task again.  This is bad for both throughput and latency.  Figuring out how to cure this is too tricky for now, though.

flight/Modules/System/systemmod.c
```diff
@@ -96,14 +127,20 @@ static void updateRfm22bStats();
...
+
 /**
  * Create the module task.
  * \returns 0 on success or -1 if initialization failed
  */
 int32_t SystemModStart(void)
 {
-	// Initialize vars
-	stackOverflow = false;
+        // Create periodic event that will be used to update the telemetry stats
```
kubark42: Indentation.

flight/Modules/System/systemmod.c
```diff
@@ -360,12 +447,18 @@ static void updateWDGstats()
 }
 #endif
 
+#ifdef PIPXTREME
```
kubark42: Weird to have this #ifdef in systemmod.c. Should we even continue to support the PipXtreme? It's not manufactured anymore, right? Or is this more targeting RFMB22? If so, does it make sense to have an #ifdef for a specific hardware chip?

flight/Modules/PipXtreme/pipxtrememod.c
mlyle: pipxtrememod was a duplicate of systemmod with a few changes-- turning off configuration monitoring and changing the instance number of the rfm22b stats to update.

It was a question of either performing the same changes twice in the two different systemmods or factoring them into common code.

There is a small side effect: this makes the heartbeat LED on pipxtreme blink faster when the board it is talking to is armed.

(I still need to test and verify all of this).

kubark42: Gotcha.

flight/Modules/System/systemmod.c
mlyle: How quickly it ran before was nondeterministic.  It ran faster when armed, before, to get faster LED blinks.  It also ran more if there were queue wakeups, etc.  IMO this is bad-- it could consume more resources based on unpredictable factors and the CPU use one sees on the bench unarmed would not be representative.

Now it always runs at the same speed.  It does run 9x faster when disarmed, and a little bit slower than it did before when armed.

flight/Modules/System/systemmod.c
```diff
@@ -602,20 +665,216 @@ void vApplicationIdleHook(void)
...
+
+	// Get lock
+	PIOS_Recursive_Mutex_Lock(mutex, PIOS_MUTEX_TIMEOUT_MAX);
+
+    // Iterate through each object and update its timer, if zero then transmit object.
+    // Also calculate smallest delay to next update.
+    uint32_t now = PIOS_Thread_Systime();
+    uint32_t timeToNextUpdate = PIOS_Thread_Systime() + MAX_UPDATE_PERIOD_MS;
+    LL_FOREACH(objList, objEntry)
+    {
```
kubark42: K&R style ```() {``` throughout.

flight/Modules/System/systemmod.c
```diff
@@ -602,20 +665,216 @@ void vApplicationIdleHook(void)
...
+        // If object is configured for periodic updates
+        if (objEntry->updatePeriodMs > 0)
+        {
+            // Check if time for the next update
+        	timeNow = PIOS_Thread_Systime();
```
kubark42: Indentation is combination of spaces and tabs.

flight/Modules/System/systemmod.c
```diff
@@ -602,20 +665,216 @@ void vApplicationIdleHook(void)
...
+            if (objEntry->timeToNextUpdateMs <= timeNow)
+            {
+                // Reset timer
+            	offset = ( timeNow - objEntry->timeToNextUpdateMs ) % objEntry->updatePeriodMs;
```
kubark42: spaces around ```(``` and ```)```

flight/Modules/System/systemmod.c
mlyle: It's targeting the ground-side modem-- pipxtreme/oplink mini/taulinkmodule/etc which all run on the pipxtreme target.  The ground side modem should not play with alarms as the flight board is expected to do this.

You can buy compatible hardware here: http://www.aliexpress.com/item/Free-shipping-OPLINK-MINI-CC3D-REVO-Universal-Transceiver-TX-RX-Module-Integrating-Remote-Controller-PPM-Input/32431902268.html?spm=2114.01020208.3.1.orWplJ&ws_ab_test=searchweb201556_9_79_78_77_80,searchweb201644_5,searchweb201560_9

In general, I don't like the way we handle remote vs. local objects.  I don't like that it's not possible to use sparky2 as a ground-side modem, etc.  Factoring this in common helps the ultimate cleanup to make boards more versatile, but doesn't really take any direct steps towards it.

flight/PiOS.posix/inc/pios_initcall.h
```diff
@@ -47,23 +39,33 @@ typedef struct {
...
+#define MODULE_INITSYSTEM_DECLS \
+static initmodule_t __module_initcalls[256]; \
+initmodule_t *__module_initcall_start = __module_initcalls; \
+initmodule_t *__module_initcall_end = __module_initcalls;
 
-#define MODULE_INITCALL(ifn, sfn)
+#define MODULE_INITIALISE_ALL(wdgfn)  { \
+	for (initmodule_t *fn = __module_initcall_start; fn < __module_initcall_end; fn++) { \
+		if (fn->fn_minit)                               \
+		(fn->fn_minit)();                       \
```
kubark42: spacing before ```\```

flight/Modules/System/systemmod.c
mlyle: So I had hoped to not change all of this in this PR.. This is all moved verbatim from event system.  I can uncrustify the module but I'm worried it makes the real change more obtuse.

flight/UAVObjects/uavobjectmanager.c
```diff
@@ -377,7 +397,7 @@ UAVObjHandle UAVObjRegister(uint32_t id,
  */
 UAVObjHandle UAVObjGetByID(uint32_t id)
 {
-	UAVObjHandle * found_obj = (UAVObjHandle *) NULL;
+	UAVObjHandle found_obj = NULL;
```
kubark42: Took me a second to figure out what was going on here. It's a little confusing (for me) to have a typedef be a pointer (```typedef struct UAVOBase * UAVObjHandle```).

mlyle: IMO it is preferable to what we had before, where it was a uintptr_t or void * that one always casts back and forth. 

I think we need to move to these practices:

* In public modules, always use anonymous structure pointers for handles instead of uintptr_t or void *
* Fully define these structures in private headers for the module that needs to see the information.
* Do not include private headers everywhere like we currently do.  Preferably place them somewhere so that include paths don't work out for other modules to include them.

Otherwise, there's basically no type safety (because anything can roundtrip through a void * or cast)  and there's basically no encapsulation (since everyone can see the private header informaiton).

mlyle: Also, if you've not encountered the command line tool 'cdecl' it's very helpful in explaining what declarations mean 
```cdecl> explain void *(*a)(void *)
declare a as pointer to function (pointer to void) returning pointer to void```

This is particularly helpful with things like array arguments.

kubark42:
I like this a lot and am very positive about it. Things that don't necessarily sit well with me (although none are blockers):

1. The overusage of ```NULL```, which increases maintenance burden.
1. Likewise, the varying list of variables being cast so the compiler doesn't complain: ```(void) ev; (void) ctx; (void) obj; (void) len;``` <-- Couldn't that just be a #define MACRO called at the beginning of every callback? Would be a lot more readable and maintainable.
1. Code specific to PipXtreme/RFM22B hardware in a core module. 
1. Typedef'ing a pointer ```typedef struct UAVOBase * UAVObjHandle``` just seems weird. Agree with Mike's points about why this is better than before, but still holding out hope there might be a "best" way.

mlyle:
> Likewise, the varying list of variables being cast so the compiler doesn't complain: (void) ev; (void) ctx; (void) obj; (void) len; <-- Couldn't that just be a #define MACRO called at the beginning of every callback? Would be a lot more readable and maintainable.

If you want I'll just pull this out.  We do not currently build with the flags needed to get the warning on unused args.  There's tons of functions in the code that do not use all the args (including almost all of the existing callback code).   I just was trying to make it better here.

> Code specific to PipXtreme/RFM22B hardware in a core module.

Existing system module had this-- just with different instance numbers for the UAVO to update in systemmod vs. pipxtreme mod.  This PR introduces `#ifdef PIPXTREME` to disable alarms which can ultimately be renamed.

> Typedef'ing a pointer typedef struct UAVOBase * UAVObjHandle just seems weird. Agree with Mike's points about why this is better than before, but still holding out hope there might be a "best" way.

I don't hold out much hope.  I've got 20 years of experience doing encapsulation different ways in C and this is by far the best.  There's wonky things you can do with the preprocessor that produce slightly prettier code but are too "magical".

mlyle:
Disabled stack validation in chibios-- needed for similar reasons to the freertos changes.  Also rebased on current next, untested.

mlyle:
More of systemmod had to be turned off for pipxtreme target to be happy.

flight/Modules/System/systemmod.c
kubark42: Could you do it selectively on lines that are already changing anyway? It's not a perfect solution, but reduces the number of times we'll have to review style changes.

fujin:
I ported this functionality to brain (the stack validation bits) and tested OK on bench, tested OK with power, tested radio OK with power, tested autotune OK, tested resulting tune OK (in yard, 15% throttle). branch/commit @ https://github.com/fujin/TauLabs/commit/49dd78a9ba106933e9b85a071af4afc2f515881d -- thanks @mlyle :+1:

kubark42:
@fujin, thanks! Any chance you can get us "before and after" for CPU usage as well as heap free?

mlyle:
Tested basic flight and altitude hold on sparky2 with GPS, airspeed, altitude hold, vtol path follower, path planner, and autotune modules running.  Did not catch fire.

I don't expect the larger targets to really show a benefit in CPU load or heap utilization (yet).  I have no before point, but with all those modules running CPU load = 20..21% and 91392 heapremaining.

Next step after this merges: moving autotune to get sensor points synchronously and to keep a buffer af points pending processing; move it to run synchronous with flight loop, so no longer getting a jittery, low-quality ~333Hz sample.

peabody124:
Cool. I was just testing it on BrushedSparky and it also flew fine for me with OpenLRS.

PR #2063, opened 2015-11-17
MSP support for CC3D (on PIOS_HAL)

dustin:


dustin:
Just tested this on a minimum bench setup (board, osd, cam, vtx).  Stuff on the screen seems right.

dustin:
TESTING_STATUS:  Cursory.

I've installed this on a cc3d on my desk and hooked it up to an OSD and saw it work correctly.  This change does also bring in PIOS_HAL, so a more thorough test might be a good idea.

Issue #1365, opened 2014-10-18
GCS seems to buffer events while in background


peabody124:
On OSX any recent build, I see a very odd thing. When the GCS is connected to a board and left in the background, it is like all the updates get stuck into a giant queue. When the GCS is brought the foreground it can take a while to become responsive again (with the length of time increasing the longer it has been idle for). Sometimes this lock up can take minutes to clear.

PR #2064, opened 2015-11-17
GCS/OSX: Disable application Nap

PTDreamer:
Fixes issues when application is minimized or sent to the background.

### DIFF
--- a/ground/gcs/src/app/main.cpp
+++ b/ground/gcs/src/app/main.cpp
@@ -53,6 +53,7 @@
 #include "customsplash.h"
 #include <QBitmap>
 #include "libcrashreporter-qt/libcrashreporter-handler/Handler.h"
+#include <QProcess>
 
 #include "../../../../../build/ground/gcs/gcsversioninfo.h"
 
@@ -264,6 +265,9 @@ int main(int argc, char **argv)
         // https://bugreports.qt-project.org/browse/QTBUG-32789
         QFont::insertSubstitution(".Lucida Grande UI", "Lucida Grande");
     }
+    QProcess process;
+    process.start("defaults", QStringList() << "write" << "org.taulabs.gcs" << "NSAppSleepDisabled" << "-bool" << "YES");
+    process.waitForFinished();
 #endif
 #ifdef Q_OS_LINUX
     QApplication::setAttribute(Qt::AA_X11InitThreads, true);


ground/gcs/src/app/main.cpp
```diff
@@ -264,6 +265,9 @@ int main(int argc, char **argv)
         // https://bugreports.qt-project.org/browse/QTBUG-32789
         QFont::insertSubstitution(".Lucida Grande UI", "Lucida Grande");
     }
+    QProcess process;
```
kubark42: Could you add a comment as to what this code is doing? It's relatively opaque what the problem is and where the solution is coming from. Is this inspired from a SO page which could be referred to?

PTDreamer: While working on an osx VM I noticed the "hang" behavior already described by other OSX users when the gcs is out of focus. Checking the OSX process/task list app (that OSX thing I don't recall the name) I noticed a "nap" column that turns to "yes" after a while. Googling about it turns out that it is a recent OSX feature to save energy.
According to the docs this can be disabled with a proper key on the plist file but I couldn't get that to work ( Mr Google told me I'm not alone) .
This is another approach using the "defaults" command which can write options for a given app domain

mlyle: Thanks for finding this.  Argh.  I was hoping we could just edit info.plist -- it's pretty hacky to have to run a subprocess.

mlyle:
So I was thinking about how to do this perhaps more cleanly.  It turns out QSettings can write to the org.taulabs.gcs configuration bundle.  We use this in splashscreen:

```
{
    NSNavLastRootDirectory = "~/taulabs";
    NSNavPanelExpandedSizeForOpenMode = "{1105, 697}";
    NSNavPanelExpandedSizeForSaveMode = "{712, 565}";
    "NSToolbar Configuration com.apple.NSColorPanel" =     {
        "TB Is Shown" = 1;
    };
    "splashscreen.0" = 11845;
    "splashscreen.1" = 13244;
    "splashscreen.10" = 17070;
```

Perhaps this could work to avoid executing a subprocess?

ground/gcs/src/app/main.cpp
PTDreamer: I agree this is ugly and honestly I didn't exhaustively test other approaches working on an OSX vm running on a centrino host is not very fun.
I did try adding LSAppNapIsDisabled key to the dictionary of the plist file (inside gcs/app gets copied to the bundle during build) both as bool and string and it didn't work.

PTDreamer:
Is the configuration bundle the same as the info.plist file or something else?

mlyle:
Turns out it is something else.  I had hoped otherwise.

The command I pasted output from above was ` defaults read org.taulabs.gcs`     So it seems the splashscreen writes to the configuration bundle.

peabody124:
I'm really hoping this will fix the issue where background in GCS makes it
hang when returning to foreground

kubark42:
@peabody124, do you remember which issue that was linked to?

mlyle:
@peabody124 yes that's what this is intended to fix.

It's also worth using this command to turn off nap on terminal, which is sometimes bothersome.  It's issue #1365

ground/gcs/src/app/main.cpp
kubark42: Excellent sleuthing! Could your findings be documented in the code and in the git commit message? I fear it will otherwise turn into an unmaintainable piece of code.

kubark42:
If there's no progress on the ```info.plist``` front, I'm okay to merge this as soon as the requested comment and git commit message are ready. Thanks a lot for doing this work!

mlyle:
I can write this the correct way, using QSettings, if necessary.

PTDreamer:
Superseded by #2085 
Closing

PR #2065, opened 2015-11-17
Fix linux parallel builds

PTDreamer:


kubark42:
This is related to #2053, right?

mlyle:
Yes, it fixes #2051

PR #2067, opened 2015-11-17
Update GCS_YEAR

PTDreamer:


kubark42:
No arguments here, but would this generally be better as a string which is created at compile time?

PR #2073, opened 2015-11-21
TransmitterControl: do not allow switch arming when invalid

### LINKED ISSUE #2072: User injured by incorrect behavior of arming switch
```
14:14 < droidicus> hey all, I have a report of a potentially dangerous corner 
                   case that a friend just experienced...
14:14 < droidicus> Was working on the bench and had the quad in "never arm" mode
14:16 < droidicus> we back to change that setting, and accidently chose "arming 
                   switch" instead of holding the yaw at max (cannot remember 
                   what that is called in the software)
14:17 < droidicus> however, he didn't have a arming switch set in his normal 
                   config
14:17 < droidicus> the quad immediately armed and flew at him
14:17 < droidicus> now he is in the ER getting the end of his finger sewn back on 
                   :-/
14:17 < dsal> That sounds like a bug to file in short order.
```

If an arming switch is not configured, and arming mode switch is selected, the behavior is the same as "always armed" mode.

mlyle:
fixes #2072 

needs test and validation.

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -249,6 +249,7 @@ int32_t transmitter_control_update()
 
 		if (settings.ChannelGroups[n] >= MANUALCONTROLSETTINGS_CHANNELGROUPS_NONE) {
 			cmd.Channel[n] = PIOS_RCVR_INVALID;
+			validChannel[n] = false;
 		} else {
 			cmd.Channel[n] = PIOS_RCVR_Read(pios_rcvr_group_map[settings.ChannelGroups[n]],
 							settings.ChannelNumber[n]);
@@ -306,8 +307,8 @@ int32_t transmitter_control_update()
 	    validChannel[MANUALCONTROLSETTINGS_CHANNELGROUPS_FLIGHTMODE];
 
 	// because arming is optional we must determine if it is needed before checking it is valid
-	bool arming_valid_input = settings.ChannelGroups[MANUALCONTROLSETTINGS_CHANNELGROUPS_ARMING] >= MANUALCONTROLSETTINGS_CHANNELGROUPS_NONE ||
-	    validChannel[MANUALCONTROLSETTINGS_CHANNELGROUPS_ARMING];
+	bool arming_valid_input = (settings.Arming < MANUALCONTROLSETTINGS_ARMING_SWITCH) ||
+		validChannel[MANUALCONTROLSETTINGS_CHANNELGROUPS_ARMING];
 
 	// decide if we have valid manual input or not
 	valid_input_detected &= validChannel[MANUALCONTROLSETTINGS_CHANNELGROUPS_THROTTLE] &&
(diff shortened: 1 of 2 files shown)


SystemsGuy:
Confirmed "Amputate Mode" on Endo release.  Configure PPM input, 5 channels with no arm, and select "Switch" as your arming method, and you are armed on power up.  The FC is passing throttle to the ESC as it boots if you power on with throttle on your Rx.  Only your ESC "low throttle arm" is protecting you here.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -306,8 +307,8 @@ int32_t transmitter_control_update()
 	    validChannel[MANUALCONTROLSETTINGS_CHANNELGROUPS_FLIGHTMODE];
 
 	// because arming is optional we must determine if it is needed before checking it is valid
-	bool arming_valid_input = settings.ChannelGroups[MANUALCONTROLSETTINGS_CHANNELGROUPS_ARMING] >= MANUALCONTROLSETTINGS_CHANNELGROUPS_NONE ||
-	    validChannel[MANUALCONTROLSETTINGS_CHANNELGROUPS_ARMING];
+	bool arming_valid_input = (settings.Arming >= MANUALCONTROLSETTINGS_ARMING_SWITCH) ||
```
peabody124: could we do this in a way that doesn't depend on the numerical value on the enum? that makes me nervous in case we ever reorder things.

mlyle: I put a comment in the XML file.  We have similar vulnerability to "NONE" in the channel mapping if we're concerned about that.  

I could add the 4 switch settings explicitly but it has a different problem if anyone ever adds another switch setting...

We really need to heavily refactor and clean this up in general.  Remember when we talked about it on IRC and you highlighted how mature it was in contrast to me saying the arming and failsafe logic is scary and disorganized?

mlyle: this is not correct, fixing.

mlyle:
Forcepushed ; had the conditional backwards.

flight/Modules/ManualControl/transmitter_control.c
```diff
@@ -249,6 +249,7 @@ int32_t transmitter_control_update()
 
 		if (settings.ChannelGroups[n] >= MANUALCONTROLSETTINGS_CHANNELGROUPS_NONE) {
 			cmd.Channel[n] = PIOS_RCVR_INVALID;
+			validChannel[n] = false;
```
solidgoldbomb: This should be unconditionally asserted at the top of this loop.  That way, a channel is assumed invalid until it makes it through a path that specifically sets it valid.

There is also a pre-existing bug below on (new) lines 260-266 when ```cmd.Channel[n] == PIOS_RCVR_INVALID```.  Bad idea to run ```scaleChannel()``` on an error code and then use that scaled version as input to ```validInputRange()``` to decide if the value is OK.  Line 260 should probably test for any value ```< 0```.

solidgoldbomb:
This change looks like it fixes the defined failure mode (arm via switch configuration but no arm channel defined).  I've only code inspected, haven't bench tested it.

flight/Modules/ManualControl/transmitter_control.c
mlyle: >  on an error code and then use that scaled version as input to validInputRange() 

I don't think the scaled version is input to validInputRange.

It also won't be negative because everything is coerced to a uint16.

solidgoldbomb: Ah, right.  Scaled version is not an input to ```validInputRange()```.  It's still sketchy to be passing an error code into those functions at all.  The ```if``` on (new) line 260 only tests against ```PIOS_RCVR_TIMEOUT``` and misses the other 2 potential error values.

peabody124:
I confirmed on brushedsparky and endo that I can replicate the bug on the bench, and that with this commit I cannot. Merging.

I would also like to see this merged back into master and we make a bug fix release.

peabody124:
Also confirmed arming with a switch still works appropriately

PR #2076, opened 2015-11-22
autotune: add a prediction counter

mlyle:
1. Autotune now says how many times it ran; this can be compared to the expected value (20,000) to see if points are being dropped.
2. Autotune now only updates the systemident object every 256 cycles to save a little CPU and to not spam telemetry links.

### DIFF
--- a/flight/Modules/Autotune/autotune.c
+++ b/flight/Modules/Autotune/autotune.c
@@ -108,7 +108,7 @@ int32_t AutotuneStart(void)
 MODULE_INITCALL(AutotuneInitialize, AutotuneStart)
 
 static void UpdateSystemIdent(const float *X, const float *noise,
-		float dT_s) {
+		float dT_s, uint32_t predicts) {
 	SystemIdentData relay;
 	relay.Beta[SYSTEMIDENT_BETA_ROLL]    = X[6];
 	relay.Beta[SYSTEMIDENT_BETA_PITCH]   = X[7];
@@ -123,6 +123,8 @@ static void UpdateSystemIdent(const float *X, const float *noise,
 		relay.Noise[SYSTEMIDENT_NOISE_YAW]   = noise[2];
 	}
 	relay.Period = dT_s * 1000.0f;
+
+	relay.NumAfPredicts = predicts;
 	SystemIdentSet(&relay);
 }
 
@@ -191,6 +193,8 @@ static void AutotuneTask(void *parameters)
 		const uint32_t PREPARE_TIME = 2000;
 		const uint32_t MEASURE_TIME = 60000;
 
+		static uint32_t updateCounter = 0;
+
 		bool doingIdent = false;
 
 		FlightStatusData flightStatus;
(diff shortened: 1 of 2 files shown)


flight/Modules/Autotune/autotune.c
```diff
@@ -123,6 +123,8 @@ static void UpdateSystemIdent(const float *X, const float *noise,
 		relay.Noise[SYSTEMIDENT_NOISE_YAW]   = noise[2];
 	}
 	relay.Period = dT_s * 1000.0f;
+
+	relay.NumAfPredicts = predicts;
```
mluessi: Shouldn't `Af` be `Of`, or what does `Af` stand for? Also, would prefer it to be more human readable, e.g. `NumberOfPredictions`.

mlyle: I'm not sure what it stands for; it's the number of times af_predict is called, which does one step of the EKF.  I don't know what the af prefix in the module is.

flight/Modules/Autotune/autotune.c
```diff
@@ -191,6 +193,8 @@ static void AutotuneTask(void *parameters)
 		const uint32_t PREPARE_TIME = 2000;
 		const uint32_t MEASURE_TIME = 60000;
 
+		static uint32_t updateCounter = 0;
```
tracernz, line 196: Should be avoiding adding new camel-case variables to flight code IIRC.

tracernz:
This looks fine subject to suggested edit. Any objections to merge once addressed?

flight/Modules/Autotune/autotune.c
mlyle, line 196: We should decide definitively what to do for PRs touching existing files.
1. Match existing convention in file.
2. Use a different style for new variables.
3. Re-style the whole file.

3 makes review difficult.  I'm willing to do 1 or 2-- they're each ugly in different ways.

tracernz, line 196: I think 2 in this case, especially since the file already has quite a bit of _ action. Agree that we should come up with a general rule.

flight/Modules/Autotune/autotune.c
mluessi, line 127: My bad, I thought it was a typo.

mluessi:
Looks fine to me as well, I would also favor using underscores in variables esp. given that some variables in the file already use it.

flight/Modules/Autotune/autotune.c
peabody124, line 196: Yeah historically we tried to keep camel case for UAVO names but even that I'm not consistent about. Also look at doingIdent below :(

flight/Modules/Autotune/autotune.c
peabody124, line 127: You know. I cannot for the life of me remember why it was called `af_` now

peabody124:
I'm not particularly against this. Just pointing out the down side that people will lose their preexisting SystemIdent measurements. Since this is essentially a debugging tracker there is an argument we should figure out which platforms work/don't/can't be fixed and ideally not need that.

Practically this will be useful and we can always revert it for release if we nail things down solid.

mlyle:
I think it'd be useful for GCS to respond to this ultimately and warn of short / invalid tunes.

The other half of this, the UAVO-update reduction, is kind of important too.

PR #2077, opened 2015-11-22
transmittercontrol: determine correctly whether arm switch is needed

mlyle:


peabody124:
Sorry I didn't see this when I created #2078. My bad. Do we want to collect a few bug fixes into one bump on master?

PR #1997, opened 2015-10-21
OpenLRS reliability improvements

peabody124:
This improves the timing of the OpenLRS code and makes it better stay in sync with
the channel hopping.

PR #2078, opened 2015-11-22
Endomorphism bugfixes

peabody124:
Bugfixes to endomorphis. One a safety critical bug #2073 and the other
that will cause airframes to drop while using OpenLRS #1997. Both have
been reviewed and merged to next. We can tag an endomophism bug fix
release quite quickly with this.

### DIFF
--- a/flight/Modules/ManualControl/manualcontrol.c
+++ b/flight/Modules/ManualControl/manualcontrol.c
@@ -67,13 +67,15 @@ static uint32_t lastSysTime;
 
 // Private functions
 static void manualControlTask(void *parameters);
-static bool ok_to_arm(void);
 static FlightStatusControlSourceOptions control_source_select();
 
 // Private functions for control events
 static int32_t control_event_arm();
 static int32_t control_event_arming();
 static int32_t control_event_disarm();
+
+// This is exposed to transmitter_control
+bool ok_to_arm(void);
 
 /**
  * Module starting
@@ -270,7 +272,7 @@ static FlightStatusControlSourceOptions control_source_select()
  * @brief Determine if the aircraft is safe to arm based on alarms
  * @returns True if safe to arm, false otherwise
  */
-static bool ok_to_arm(void)
+bool ok_to_arm(void)
 {
 	// read alarms
 	SystemAlarmsData alarms;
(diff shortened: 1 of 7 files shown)


mlyle:
If we're going to do a maintenance release.. can we please roll up the brushedsparky change (vtx channels) so you're not maintaining separate builds/GCSes?

Also that should get submitted to next.

mluessi:
Is brushed sparky open hardware? If so, the design files should be added. If not, the last commit should be removed.

mlyle:
@mluessi there's already brushed sparky related support in main tree.  It's just wrong.

We've also agreed so far to "look the other way" about the few lines of code needed to support Naze variants that aren't open.

mluessi:
@mlyle Does this mean I can make a target without OSD support, make the board open hardware. Then make a closed source board with an OSD as a new hardware revision? I think if we have rules, we should either follow them or change the rules if they don't work.

kubark42:
I agree with @mluessi on this.

mluessi:
Given that we already have brushed sparky support and this is safety critical, I'm in favor or merging this. But I think the open vs closed board discussion should be revisited in the near future.

mlyle:
I am going to hand cherry-pick these changes and carefully push to master.

This is necessary because for a build to be marked as a release build, it must be tagged.  Clicking "merge" will result in Jenkins being triggered to produce artifacts before the tag.  The git hash cannot be predicted and tagged beforehand.

mlyle:
created hotfix staging branch endomorphism-hotfix1

mlyle:
If artifacts spit out of this, next steps will be -- tag release, fast-forward push it onto master, wait for next set of artifacts.

mlyle:
Artifacts emerged from Jenkins.  Fast-forwarded master.

peabody124:
Sorry I was at work while this was being discussed. I'm happy to remove the BrushedSparky commit, I agree we should be consistent. There is an old commit that slippped in that relates to brushed sparky but it doesn't actually work with the production hardware.

Regarding open sourcing it, my plan is to do so unless the restructing decision is to allow closed hardware in which case I might have to think about it.

I'm more than happy to maintain my own releases until it is opened / TL policy changes. That's just fair and something I'd already done https://github.com/peabody124/TauLabs/releases

peabody124:
P.S. honestly I shouldn't have added that commit when mike asked. there was definitely no reason to couple it to safety related issues.

mlyle:
I think it cuts a very fine line to use project infrastructure to maintain releases that aren't eligible to be in the mainline project tree.

OTOH, I think it's silly for us to maintain brushedsparky separately and it causes a lot of pain to have lots of trees out there for different targets.

peabody124:
Well as I said I'm happy to do it for consistency with principles. In the mean time I pushed the schematic and assembly drawing to a PR so at least that is opened #2087

Issue #2071, opened 2015-11-21
ok_to_arm check in ManualControl unsafe


mlyle:
transmitter_control contains the arming state machine, and manualcontrol effectively has a state machine that "follows" it.

```
//! When the control system requests to arm the FC
static int32_t control_event_arm()
{
        if(ok_to_arm()) {
                FlightStatusData flightStatus;
                FlightStatusGet(&flightStatus);
                if (flightStatus.Armed != FLIGHTSTATUS_ARMED_ARMED) {
                        flightStatus.Armed = FLIGHTSTATUS_ARMED_ARMED;
                        FlightStatusSet(&flightStatus);
                }
        }
        return 0;
}
```

This means that if ok_to_arm() is false, transmitter_control will still arm.  Later, if ok_to_arm becomes true, it will instantly arm.

This is unsafe.  Say there's a transient alarm for GPS, and the user attempts to arm and then take off.  They increase throttle progressively and see nothing happen.  Then, the system gets its last satellite and zooms off into outer space.

PR #2079, opened 2015-11-23
ManualControl: ignore attempts at arming when not safe

peabody124:
Previously transmitter_control did not account for the state of
the system's safety to arm when checking the sticks. This could
result in the inner module (transmitter_control) considering
itself armed and constantly trying to fire arm events at the
outer module. As soon as the error condition cleared (e.g. GPS
lock occuring) the system would arm and process inputs.

### DIFF
--- a/flight/Modules/ManualControl/manualcontrol.c
+++ b/flight/Modules/ManualControl/manualcontrol.c
@@ -67,13 +67,15 @@ static uint32_t lastSysTime;
 
 // Private functions
 static void manualControlTask(void *parameters);
-static bool ok_to_arm(void);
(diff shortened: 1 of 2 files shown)


peabody124:
Fixes #2071

peabody124:
I can confirm with BrushedSparky I can no longer reproduce #2071 using autotune as a flight mode switch.

peabody124:
I'd also like to add this fix to #2078 once reviewed/merged.

mlyle:
Not perfectly safe / manualcontrol needs some help.  But greatly reduces the exposure to the problem per convo on IRC.

PR #2081, opened 2015-11-23
MSP Alarm support.

dustin:
This change creates a new MSP command that allows the peer (e.g. OSD) to ask for alarms from the flight controller.  The alarm message includes severity and a message.  In its current form, only one alarm is sent (fairly arbitrarily, other than severity taking precedence).  Also, as a special and very useful case, boot reason is considered a warning from the perspective of the MSP code for the first few seconds.  It might be worth generally considering that an alarm which would allow this special case code to be removed.

This has been very useful to me in showing both regular alarms (e.g. I let someone fly my sparky yesterday and saw 'MAN:NO RX' flashing on the screen because OpenTX companion turned off the radio for that profile) as well as boot reasons (saw watchdog firing in flight instead of mysteriously crashing).

STATUS:  Flown a lot.  Used to diagnose bugs.
CAVEAT:  MWOSD side is not yet accepted (waiting for @ShikOfTheRa in https://github.com/ShikOfTheRa/scarab-osd/pull/141)

### DIFF
--- a/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
+++ b/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
@@ -42,6 +42,7 @@
 #include "actuatordesired.h"
 #include "flightstatus.h"
 #include "systemstats.h"
+#include "systemalarms.h"
 #include "homelocation.h"
 #include "baroaltitude.h"
 #include "pios_thread.h"
@@ -82,6 +83,7 @@
 #define  MSP_BOXIDS     119 // get the permanent IDs associated to BOXes
 #define  MSP_NAV_STATUS 121 // Returns navigation status
 #define  MSP_CELLS      130 // FrSky SPort Telemtry
+#define  MSP_ALARMS     242 // Alarm request
 
 typedef enum {
 	MSP_BOX_ARM,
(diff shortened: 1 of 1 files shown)


flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -391,6 +397,154 @@ static void msp_send_boxids(struct msp_bridge *m) {
 	msp_send(m, MSP_BOXIDS, boxes, len);
 }
 
+static const char alarm_names[][8] = {
```
mluessi, line 400: Is it necessary to define the size of the 2nd dimension?

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -391,6 +397,154 @@ static void msp_send_boxids(struct msp_bridge *m) {
...
+	[SYSTEMALARMS_ALARM_ALTITUDEHOLD] = "A-HOLD",
+	[SYSTEMALARMS_ALARM_BOOTFAULT] = "BOOT",
+	[SYSTEMALARMS_ALARM_GEOFENCE] = "GEOFENCE",
+	[SYSTEMALARMS_ALARM_TEMPBARO] = "TEMPBARO",
+	[SYSTEMALARMS_ALARM_GYROBIAS] = "GYROBIAS",
+	[SYSTEMALARMS_ALARM_ADC] = "ADC",
+};
+
+// If someone adds a new alarm, we'd like it added to the array above.
+DONT_BUILD_IF(NELEMENTS(alarm_names) != SYSTEMALARMS_ALARM_NUMELEM, AlarmArrayMismatch);
```
mluessi, line 427: Nice :)

mluessi:
Nice work. I haven't yet had time to review the code carefully, will do so in the next few days.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
tracernz, line 400: Yes, you always need to define the size of the second and subsequent dimensions.

PR #2083, opened 2015-11-24
UAVOGen: Check that default values for enums actually exist

### LINKED ISSUE #2080: Ensure that defaultname in UAVO is present in list of options
E.g., ```<field name="FirstMotorSense" units="" type="enum" elements="1" options="CW,CCW" defaultvalue="None"/>``` passes the generator just fine. The generator should catch this and fail.

tracernz:
Checks in options list to ensure the default values are actually possible.
Fixes #2080

Apologies for the extra diff due to whitespace cleanup. I made a mess of my own diff (tabs instead of spaces) and told the editor to fix it, and it turns out the rest of the file had some problems already. I think fixing it is the right move anyway but can revert those lines if necessary.


mlyle:
Looks good!

kubark42:
Nice!

PR #2084, opened 2015-11-24
uavobjgenerator:  Ensure rebuild of flight when necessary

mlyle:
This code had one remaining weakness-- on non-windows platforms, if you
edited a file and then put it back UAVobjs might not get rebuilt due to
moving to an "old" timestamp.  Now we touch the version file to ensure
that this is OK.

Tested and confirmed to work with multiple instances of moving back and forth in tree.

### DIFF
--- a/ground/uavobjgenerator/main.cpp
+++ b/ground/uavobjgenerator/main.cpp
@@ -244,6 +244,8 @@ int main(int argc, char *argv[])
         wiresharkgen.generate(parser,templatepath,outputpath);
     }
 
+    bool changed = false;
+
     /* Symlink each of these to the current dir */
     QDir dir(outputpath);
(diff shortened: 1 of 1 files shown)


ground/uavobjgenerator/main.cpp
```diff
@@ -253,11 +255,22 @@ int main(int argc, char *argv[])
         dir.rmpath(file.fileName());
         dir.rename(file.absoluteFilePath(), file.fileName());
 #else
-        QFile::remove(file.fileName());
-        QFile::link(file.absoluteFilePath(), file.fileName());
+        if (QFile::symLinkTarget(file.fileName()) != file.absoluteFilePath()) {
```
kubark42, line 258: Could you add a comment as to how this `if()` catches a changed UAVO?

kubark42:
Looks good. There's that one suggested edit for maintainability, but aside from that I like it and am happy to merge.

PR #2085, opened 2015-11-24
gcs: OSX: set flag to prevent application nap.

mlyle:
as discovered by @PTDreamer.  Supersedes #2064.

### DIFF
--- a/ground/gcs/src/app/main.cpp
+++ b/ground/gcs/src/app/main.cpp
@@ -263,8 +263,15 @@ int main(int argc, char **argv)
         // fix Mac OS X 10.9 (mavericks) font issue
         // https://bugreports.qt-project.org/browse/QTBUG-32789
         QFont::insertSubstitution(".Lucida Grande UI", "Lucida Grande");
+
+        QSettings tmpSettings;
+
+        bool value = true;
+
+        tmpSettings.setValue(QString("NSAppSleepDisabled"), value);
     }
 #endif
+
 #ifdef Q_OS_LINUX
     QApplication::setAttribute(Qt::AA_X11InitThreads, true);
     // This should have faster performance on linux


mlyle:
Needs more verification.  Appears to do the right thing.

peabody124:
I tested this and left GCS in the background. Came back an hour later it wasn't locked up. This is a fabulous improvement.

peabody124:
@PTDreamer ok to merge with you? think works for me.

ground/gcs/src/app/main.cpp
```diff
@@ -263,8 +263,15 @@ int main(int argc, char **argv)
         // fix Mac OS X 10.9 (mavericks) font issue
         // https://bugreports.qt-project.org/browse/QTBUG-32789
         QFont::insertSubstitution(".Lucida Grande UI", "Lucida Grande");
+
+        QSettings tmpSettings;
+
+        bool value = true;
+
```
kubark42, line 270: I'd like to see documentation here. This code is opaque and hard to maintain code if there's no description about what it does.

kubark42:
> I tested this and left GCS in the background. Came back an hour later it wasn't locked up. This is a fabulous improvement.

Awesome!

ground/gcs/src/app/main.cpp
```diff
@@ -263,8 +263,15 @@ int main(int argc, char **argv)
...
+        tmpSettings.setValue(QString("NSAppSleepDisabled"), value);
```
PTDreamer, line 271: Why use 3 different lines to do this? Wouldn't:

```
QSettings().setValue(QString("NSAppSleepDisabled"), true);
```

accomplish the same thing?

PTDreamer:
> @PTDreamer ok to merge with you? think works for me.

I would prefer the more enigmatic way of doing this I proposed but hey (jk)
I made a small comment but I don't mind merging as is.

PTDreamer:
Well this is unfortunate...

peabody124:
Let's just reopen this and merge as is. It works, is tested. No need to quibble the details.

PTDreamer:
Humm that will be difficult with the source branch deleted, the same happened with other PRs. Fortunately this one is easy to reproduce

PR #2087, opened 2015-11-24
BrushedSparky: check in schematic for production version

peabody124:
This is essentially a modified Sparky2 with the addition of VTX
circuitry and replacing some of the PWM outputs with power FETs
to drive external motors.


peabody124:
Release of gerber pending TL restructing decision on open versus closed hardware, but
I want people to be able to see the schematic in case they want to play with things. I also
add the assembly drawing so people can actually map back and forth to the components
if they need to modify anything.

tracernz:
Schematic looks good, the extra inverter is a nice touch for FrSky users. Would be great if the BOM was here too so components from the assembly drawing can be identified.

tracernz:
Looked at the Jenkins log, #2083 picked up a problem :D:

```
00:20:35 - Tau Labs UAVObject Generator -
00:20:35 Error: Invalid default value for HwSparky2.VTX_Ch
```

https://github.com/TauLabs/TauLabs/pull/2087/files#diff-cb9195910640624a4525f23d47e2b317R81

shared/uavobjectdefinition/hwsparky2.xml
```diff
@@ -78,7 +78,42 @@
 		<field name="MPU9250GyroLPF" units="Hz" type="enum" elements="1" options="184,92,41,20,10,5" defaultvalue="184"/>
 		<field name="MPU9250AccelLPF" units="Hz" type="enum" elements="1" options="460,184,92,41,20,10,5" defaultvalue="184"/>
 
-		<field name="VTX_Ch" units="Hz" type="enum" elements="1" options="1,2,3,4,5,6,7,8" defaultvalue="1"/>
+                <field name="VTX_Ch" units="Hz" type="enum" elements="1" defaultvalue="BoscamACh1">
```
tracernz: The new stuff uses spaces, the rest of the file is tabs..

mluessi:
Looks good, but who is James aotton ? ;) (in the schematic signature box)

peabody124:
Thanks for catching that!

flight/targets/sparky2/fw/pios_board.c
```diff
@@ -154,30 +154,135 @@ static void panic(int32_t code) {
 void set_vtx_channel(HwSparky2VTX_ChOptions channel)
```
kubark42, line 158: Could you throw in a quick doxygen comment for this file? It's not immediately clear if you're not doing FPV that `vtx` is shorthand for "video transmission".

kubark42:
This looks like there are two things in this PR.
1. Change VTX channel changes
2. Check in a schematic for production version of BrushedSparky.

I'm happy to merge both these changesets, but can ~~they be spun into two PRs~~ the VTX changes be spun out into a separate PR first?

flight/targets/sparky2/fw/pios_board.c
peabody124, line 158: Done and commented

peabody124:
@kubark42 if you feel very strongly, I can split them apart. Personally I prefer them together since the schematic basically documents the set_vtx_channel pin selection so I see them as grouped.

kubark42:
No, they're separate commits so it's okay. I would really prefer to see this kind of thing split up in the future, though, since one part can break people's boards whereas the other is purely informational.

PR #2088, opened 2015-11-24
Gcs autoupdate

PTDreamer:
Sorry for the amount of code, this is mainly for comment, I can make separate PRs if needed.

Refactors the packaging targets across targets introducing package_{flight, ground, installer, matlab, all) and package_XXX_compress for the different platforms using similarly constructed filenames.
The ground targets produce a filelist.lst file containing all the filenames in the package to be used by the updater.
Introduces a package_info target which produces packageversioninfo.json to uploaded to the gitHub release to be used with the autoupdater.
The autoupdate searches a given configurable gitHup repo for the latest release (or prerelease requiring login credentials and privileges) at a user configurable frequency.
If a new release is found the user receives a dialog with the release info and if he agrees the package for the current platform is downloaded and decompressed to the temp directory together with an helper app from https://github.com/PTDreamer/copyApp.
The helper app is started detached and the main app quits.
The helper app deletes all the files present on the current app filelist.lst and copy the new ones from the new filelist.lst to the same destination. When done it restarts the main app.

Additionally this fixes and enables the crashreporting on osx although no dumps were tested with the generated symbols.
Striping is enabled on osx binaries to reduce package size.


m-thread:
This looks very interesting. A I understand it from briefly reading the PR this will update the whole GCS when a new version is found? That is, it does not work on separate lib level?

Correct me if I'm wrong :)

Great work!

kubark42:
@PTDreamer Started to look at this some in-depth. It would definitely be helpful to break this up into smaller PRs. It looks like there's a fair amount of changes that are generally useful outside this PR.

P.S. There's a merge conflict. Not sure if you want to resolve this before the review is further along, since it might just happen again...

kubark42:
P.S. There's a merge conflict. Not sure if you want to resolve this before the review is further along, since it might just happen again...

PR #2089, opened 2015-11-24
Package/OSX strip binaries to decrease size

PTDreamer:


peabody124:
Compiles fine and runs ok. Package size is 83M. Not much different.

PR #2091, opened 2015-11-24
Ground: Fix crash reporting on OSX

PTDreamer:
Partially mirrored at https://github.com/d-ronin/dRonin/pull/99.

### DIFF
--- a/ground/gcs/src/crashreporterapp/crashreporterapp.pro
+++ b/ground/gcs/src/crashreporterapp/crashreporterapp.pro
@@ -3,10 +3,14 @@ QT       += core gui
 
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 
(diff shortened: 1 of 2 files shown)


ground/gcs/src/crashreporterapp/crashreporterapp.pro
```diff
@@ -3,10 +3,14 @@ QT       += core gui
 
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 
+CONFIG-=app_bundle
```
peabody124, line 6: would you prefer to move that below to the macx bracket to make sure it has no unintended side effects?

PTDreamer, line 6: I can but qmake docs clearly state that this has no effect on non OS X systems.

PR #2093, opened 2015-11-24
Package refactoring

PTDreamer:
Refactors the packaging targets across targets introducing package_{flight, ground, installer, matlab, all) and package_XXX_compress for the different platforms using similarly constructed filenames.
The ground targets produce a filelist.lst file containing all the filenames in the package to be used by the updater.
Introduces a package_info target which produces packageversioninfo.json to be uploaded to the gitHub release to be used with the autoupdater.
Packages BUs together with FWs (for future use).
Adds a zip tool target for windows (requires for the package_XXX_compress targets.


PTDreamer:
Please review

PTDreamer:
Merging tomorrow if no more comments are given

peabody124:
Could you push this branch directly to jenkins, so we can see if it breaks jenkins packaging?

Makefile
```diff
@@ -218,10 +218,18 @@ help:
 	@echo "     uavobjects           - Generate source files from the UAVObject definition XML files"
 	@echo "     uavobjects_test      - parse xml-files - check for valid, duplicate ObjId's, ... "
 	@echo
-	@echo "   [Package]"
-	@echo "     package              - Executes a make all_clean and then generates a complete package build for"
-	@echo "     standalone           - Executes a make all_clean and compiles a package without packaging"
-	@echo "                            the GCS and all target board firmwares."
+	@echo "   [Packaging]"
+	@echo "     package        			- Build and package the Tau Labs distributable (DEB)"
```
peabody124: not DMG too?

peabody124:
Also @PTDreamer can you list which platforms this was tested on?

Makefile
```diff
@@ -1062,15 +1070,12 @@ endif
...
 ##############################
-
-.PHONY: package
-package:
-	$(V1) cd $@ && $(MAKE) --no-print-directory $@
-
-.PHONY: standalone
-standalone:
+PACKAGE_TARGETS = package_ground package_flight package_installer package_matlab package_all standalone package_info
+PACKAGE_TARGETS += package_ground_compress package_flight_compress package_matlab_compress package_all_compress
```
peabody124, line 1073: shouldn't `package_all_compress` cover all the previous ones?

make/templates/packageversioninfo.json
```diff
@@ -0,0 +1,5 @@
+{"package_info": {
```
peabody124, line 1: is anything done with this file or just for documentation?

package/Makefile
```diff
@@ -15,7 +15,6 @@ ROOT_DIR := $(realpath $(WHEREAMI)/../)
 include $(ROOT_DIR)/make/system-id.mk
 
 TOOLS_DIR := $(ROOT_DIR)/tools
-BUILD_DIR := $(ROOT_DIR)/build
```
peabody124, line 18: BUILD_DIR was double defined before? Where is it coming from since it is used for PACKAGE_DIR below?

Makefile
PTDreamer: This target should be removed, it was replaced by "package_installer"

Makefile
PTDreamer, line 1073: What do you mean? use it only as the .PHONY and target instead of the complete list?

make/templates/packageversioninfo.json
PTDreamer, line 1: It is used on the gcs autoupdater from a different PR

package/Makefile
PTDreamer, line 18: It is defined on line 20 which shows as original so I guess it was double defined

PTDreamer:
> Also @PTDreamer can you list which platforms this was tested on?

I have tested at least "package_ground_compress" on mint64, win7 64 and OS X 10.10

Makefile
peabody124, line 1073: Well i mean PACKAGE_TARGETS includes lots of things and then package_all_compress. doesn't package_all_compress include all those things?

package/Makefile
```diff
@@ -179,6 +206,25 @@ tlfw_resource:
 	$(V1) mkdir -p $(BUILD_DIR)/ground/$@/build
 	$(V1) echo '$(TLFW_CONTENTS)' > $(BUILD_DIR)/ground/$@/tlfw_resource.qrc
 	$(V1) $(foreach fw_targ, $(FW_TARGETS), mkdir -p $(BUILD_DIR)/ground/$@/build/$(fw_targ);)
-	$(V1)$(foreach fw_targ, $(FW_TARGETS), cp $(BUILD_DIR)/$(fw_targ)/$(fw_targ).tlfw $(BUILD_DIR)/ground/$@/build/$(fw_targ)/;)
+	$(V1) $(foreach bu_targ, $(BU_TARGETS), mkdir -p $(BUILD_DIR)/ground/$@/build/$(bu_targ);)
+	$(V1) $(foreach fw_targ, $(FW_TARGETS), cp $(BUILD_DIR)/$(fw_targ)/$(fw_targ).tlfw $(BUILD_DIR)/ground/$@/build/$(fw_targ)/;)
+	$(V1) $(foreach bu_targ, $(BU_TARGETS), cp $(BUILD_DIR)/$(bu_targ)/$(bu_targ).tlfw $(BUILD_DIR)/ground/$@/build/$(bu_targ)/;)
+
+ground_package_common: | tlfw_resource standalone
```
peabody124, line 212: would this not be more appropriate in the package/Makefile ?

package/Makefile.linux
```diff
@@ -4,98 +4,99 @@
...
-VERSION_FULL		:= $(VERNUM)-$(PACKAGE_LBL)
+VERSION_FULL		:= $(PACKAGE_LBL)-1
 
-FLIGHT_FW			:= coptercontrol sparky sparky2 sparkybgc revomini pipxtreme quanton flyingf3 naze32 aq32
+FLIGHT_FW		:= coptercontrol sparky sparky2 sparkybgc revomini pipxtreme quanton flyingf3 naze32 aq32
 
-DEB_BUILD_DIR		:= $(ROOT_DIR)/debian
+DEB_BUILD_DIR		:= $(BUILD_DIR)/debian
 
+SED_AUTHOR_NAME		= $(shell git config user.name)
```
peabody124, line 19: spacing/alignment

Makefile
peabody124, line 1073: nevermind, i see now. package_all_compress is not the same as calling all the individual commands

package/Makefile
```diff
@@ -68,23 +85,32 @@ help:
 	@echo "   Here is a summary of the available targets:"
 	@echo
 	@echo "   [Packaging]"
-	@echo "     package        - Build and package the OpenPilot distributable"
-	@echo "     package_flight - Build and package the OpenPilot flight firmware only"
+	@echo "     package        			- Build and package the Tau Labs distributable (DEB)"
```
peabody124: does this not also include DMG?

peabody124:
Code changes look reasonable. I would like to see this doesn't break jenkins, ideally also test runs off the package on other computers. I'll try and check the OSX package tonight.

package/Makefile
PTDreamer, line 212: ? this is the package/Makefile

package/Makefile
PTDreamer: Same mistake as on the top level makefile, installer is now built with "package_installer" including dmg on osx

PTDreamer:
> Code changes look reasonable. I would like to see this doesn't break jenkins, ideally also test runs off the package on other computers. I'll try and check the OSX package tonight.

I don't know what targets jenkins uses for packaging, but the idea here is also to make the packaging stuff more independent of the jenkins scripts and allow just calling "package_ground_compress" from any of the supported platforms.

peabody124:
Jenkins has it's own scripts (for better or worse), ideally we can move towards using these commands more. However, I'd like to get that fixed before merging to next rather than after. I don't think jenkins even calls make package so don't think this will break things.

The packages at http://jenkins.taulabs.org/artifacts/ look like the correct sizes. Will check them later.

peabody124:
Sorry. Finally had a chance to test compile this on OSX. This is what I get with `make package_all_compress`

```
Done: processed 119 XML files and generated 119 objects with no ID collisions. Total size of the data fields is 3849 bytes.  Id hash is aeea41edcdb80110.
generating flight code
generating gcs code
generating java code
generating matlab code
generating wireshark code
/bin/sh: /zip.exe: No such file or directory
make[1]: *** [package_all_compress] Error 127
make: *** [package_all_compress] Error 2
```

PR #2097, opened 2015-11-26
flight: remove support for original coptercontrol (not CC3D)

peabody124:
This retains support for CC3D but removes the old CC code base. CopterControl
boards aren't widely available and probably barely used (sensor drift was worse
on them than on CC3D), so this isn't likely much of a loss.


kubark42:
Sounds reasonable.

PR #2098, opened 2015-11-26
PiOS HAL: fix F3 FrSky functionality

peabody124:
On F3 targets we were passing in the same serial port configuration as the
normal one, but it requires the inversion of the lines to communicate correctly.

Ported from https://github.com/d-ronin/dronin/pull/57/


mluessi:
Has been tested in dronin, merging.

PR #2101, opened 2015-12-03
GCS-Fix racing condition which leads to a segfault on exit

PTDreamer:


tracernz:
Great work! Thanks for looking at this.

PR #2103, opened 2015-12-04
GCS: Fix compiler warnings

peabody124:
Suppress compile warnings on GCS and enable -Werror for linux


peabody124:
See https://github.com/d-ronin/dRonin/pull/59/commits and https://github.com/d-ronin/dRonin/pull/85/commits

kubark42:
LGTM, except I'm not sure we should have the default build be with error on. I like Jenkins catching that, but it's a bit of a PITA during early development.

ground/gcs/src/libs/libcrashreporter-qt/breakpad/client/linux/handler/exception_handler.cc
```diff
@@ -385,6 +385,8 @@ int ExceptionHandler::ThreadEntry(void *arg) {
 // This function runs in a compromised context: see the top of the file.
 // Runs on the crashing thread.
 bool ExceptionHandler::HandleSignal(int sig, siginfo_t* info, void* uc) {
+  (void)sig;
```
kubark42: indentation

peabody124:
@kubark42 we could have it only apply in the release configuration. I think that is what jenkins builds anyway.

kubark42:
@peabody124, yeah, I think that'd be great. I'll do that in a subsequent PR, it only affects Linux anyway.

PR #2104, opened 2015-12-08
System settings refactor

### LINKED ISSUE #418: Provide documentation for SystemSettings.GuiConfigData.
It's a UAVO field that seems to encode various vehicle frame actuator names into bitfields. It's memory efficient, since it stores the entire model setup into just 4 bytes, but it's fragile and needs to be well documented in order to prevent errors when extending models.

Alternatively, now that memory is no longer a consideration, we could consider spending a few more bytes and have category specific UAVOs. Since only one airframe UAVO would be loaded into memory at a time, the ram costs would be low.

### LINKED ISSUE #879: Output tab should have button for reloading board data
While configuring servos, oftentimes I have to refresh the settings. This would be faster and easier if there were a dedicated "Refresh board values" button.

### LINKED ISSUE #2061: Synchronize SystemSettings.AirframeCategorySpecificConfiguration union between FC and GCS
The SystemSettings UAVObject is weird. There's a AirframeCategorySpecificConfiguration which maps to a union inside GCS, but there's no mechanism to synchronize this between flight controller (FC) and GCS. The takeaway is that the FC is unaware of what its airframe's specific configuration is.

This synchronizing could be accomplished in several ways:
1. Have a common header file shared between GCS and flight.
2. Have a UAVO which encodes this information directly, instead of inside a union.

`#2` is the more Tau Labs approach, but there are a couple issues with how we could implement it.
1. If all airframe configuration options are in a single UAVO, then this UAVO will grow to be very large. That's not optimal.
2. If the configuration options were separated into several UAVOs corresponding to the vehicle class-- e.g. SystemSettingsMultiRotor, SystemSettingsFixedWing, SystemSettingsHelicopter, and SystemSettingsGroundVehicle-- then this is potentially a lot of additional memory load. We could solve this by only initializing the appropriate UAVO, but that means we'll have to reboot between changing vehicle classes. Not really the end of the world, and I can't see too many times when someone would want to frequently switch between vehicle classes.

Actually, I kind of like that last approach. It means that we can more easily commutate between vehicle types, like for a VTOL fixed wing, or a tile rotor. This would be hard right now because all the motor information is encoded into that  single AirframeCategorySpecificConfiguration field.

Any suggestions/objections before I implement this?

Related to https://github.com/TauLabs/TauLabs/issues/418

kubark42:
In the interests of saving time before full refactoring, this only targets the multirotor platform. The others will follow if this approach is found to be good.

In a nutshell, this moves away from the GuiConfigDataUnion and toward UAVOs which hold platform-specific configurations.

There are many changes which are too numerous to count. However, important changes are:
- The output widget pane finally works reliably. In the past, the pane would only show the actuator labels that were present when the board was plugged in, not the changed ones as the board was configured.
- Moving to use the UAVObjectToWidget relationship. This greatly reduces the number of explicit SIGNAL/SLOT pairs in the code.
- Better abstraction of models.
- Defining motor names in a more programatic manner.
- Fewer lines of code.
- Hopefully reduced the number of calls. I found that certain pathways were being run multiple times when they really should have only been run once.
- Singleton to move a signal between two sibling classes.

This is really a tiny firmware change with a massive GCS change. If this works for people, I'll extend it to support the other platforms.

screen shot 2015-12-10 at 4 04 00 pm

Fixes #418.
Fixes #879.
Fixes #2061.


ground/gcs/src/plugins/config/cfg_vehicletypes/configfixedwingwidget.cpp
```diff
@@ -66,19 +66,19 @@ void ConfigFixedWingWidget::setupUI(SystemSettings::AirframeTypeOptions frameTyp
 
     if (frameType == SystemSettings::AIRFRAMETYPE_FIXEDWING) {
         // Setup the UI
-        setComboCurrentIndex(m_aircraft->aircraftType, m_aircraft->aircraftType->findText("Fixed Wing"));
-        setComboCurrentIndex(m_aircraft->fixedWingType, m_aircraft->fixedWingType->findText("Elevator aileron rudder"));
-        m_aircraft->fwRudder1ChannelBox->setEnabled(true);
+//        setComboCurrentIndex(m_aircraft->aircraftType, m_aircraft->aircraftType->findText("Fixed Wing"));
```
kubark42, line 69: This kind of stuff is commented out because it's out of scope for multirotors, but definitely in scope for the next round of revisions once my direction is approved/revamped.

ground/gcs/src/plugins/config/configoutputwidget.h
```diff
@@ -40,6 +40,33 @@
 class Ui_OutputWidget;
 class OutputChannelForm;
 
+class SignalSingleton : public QObject
+{
+    Q_OBJECT
+public:
```
kubark42, line 46: Not sure if this is the best way to have accomplished sharing a signal between the sibling ConfigOutputWidget and ConfigMultirotorWidget.

kubark42:
It's unfortunate that github is barfing on `ground/gcs/src/plugins/config/cfg_vehicletypes/configmultirotorwidget.cpp`, that's where all the important changes are.

flight/Libraries/sanitycheck.c
```diff
@@ -76,23 +76,13 @@ int32_t configuration_check()
 	}
 
 	// Classify airframe type
-	bool multirotor = true;
```
kubark42, line 79: This is the kind of change I really like in the code. It makes it much cleaner to determine if something is a multirotor type, and will make it much easier to add new frame variants.

flight/Modules/System/systemmod.c
```diff
@@ -124,6 +130,29 @@ int32_t SystemModInitialize(void)
 	SystemStatsInitialize();
 	FlightStatusInitialize();
 	ObjectPersistenceInitialize();
+
+	uint8_t platformType;
+	SystemSettingsPlatformTypeGet(&platformType);
+
+	switch (platformType){
```
kubark42, line 137: This is a crucial part. It only loads the relevant UAVO. This is to save memory on constrained targets. 

The downside is that it requires rebooting the vehicle when changing airframes. Not the end of the world, but it'd be nice to be able to have all these UAVOs in memory on a normal basis.

mluessi:
Just discussed this with @kubark42:

What would be nice if the GCS would only lets a user configure flight modes that are actually supported by the vehicle type selected. One way to do this could be to move `FlightModePosition` from `ManualControlSettings` to the vehicle specific UAVO. And use a shared parent settings object to encode all flight modes. This would also let us set the correct default modes for each vehicle type (so Multirotor wouldn't have Manual as default, which can be confusing to users).

mlyle:
@mluessi / @kubark42 : I'm not certain that's a good idea.  For one, until there's better UAVO updating (which we intend to work on) it means that additions/changes to flight modes will spoil the vehicle settings.  I also think there's a lot of interest in refactoring flight modes in general to be more flexible, and it's not too great if ManualControl needs to peek in a bunch of different UAVOs depending on vehicle type.

I also fear we'd make choices that are too limiting-- for instance, is coordinatedflight allowed on a multirotor?  Some stuff I'm working on benefits from coordinatedflight on a multirotor.  Acro+ is great on planes, etc. 

I think we should probably take Manual out of the default-- the predominant use case is multirotor, and people building planes can choose it.  I think this was left to force users to look at that pane, but that's a dubious justification.

kubark42:
@mlyle good points for discussion. FWIW, additions to flight modes already spoil the vehicle settings. Worse, it spoils the ManualControlSettings UAVO, which means nuking the R/C transmitter calibration settings. (Yuck.) So I see this as a lateral step more than a step back.

It might be that there are certain models where there's a fair amount of duplication, but there are enough differences that it would help reduce complexity for users choosing the flight mode. I can't imagine that there's much usage of VBar on ground vehicles for instance, and @mluessi pointed out that Manual isn't available on multirotors.

P.S. What kind of multirotor usage do you have for coordinated flight? The way I wrote that controller is extremely limited, it just checks for a zero crossing on the y-axis accelerometer. I'm not sure that's compatible with multirotor flight.

P.S.S. Just wanted to be clear that this discussion is for a proposed feature somewhere far down the line. The refactoring hopefully doesn't introduce any functional changes from the user perspective.

mlyle:
> P.S. What kind of multirotor usage do you have for coordinated flight? The way I wrote that controller is extremely limited, it just checks for a zero crossing on the y-axis accelerometer. I'm not sure that's compatible with multirotor flight.

I have a prototype vehicle with the battery and FC in a lifting body at alpha=1deg when the multirotor is pitched forward 50 degrees.  CoordinatedFlight isn't quite right, but it ain't too wrong either. ;)

kubark42:
@peabody124 What do you think of this approach? I'd like to finish it up for the other platform types as well. FWIW, I've been using it for months and it's performing quite well.

kubark42:
@peabody124, I've been using this on the multi-rotor side for a long while now. Seems to work quite well. Do you have any thoughts about my approach before I extend it to support the other platforms as well?

peabody124:
I'll try and look at this shortly - sorry meant to do this weekend but didn't make it

kubark42:
Cool. Just rebased against latest next.

shared/uavobjectdefinition/multirotorairframesettings.xml
```diff
@@ -0,0 +1,44 @@
+<xml>
+	<object name="MultirotorAirframeSettings" singleinstance="true" settings="true" category="Platform">
+		<description>Select airframe type.  Currently used by @ref ActuatorModule to choose mixing from @ref ActuatorDesired to @ref ActuatorCommand</description>
+		
+		<field name="MultirotorType" units="" type="enum" elements="1" options="Tri,QuadP,QuadX,HexaP,HexaX,HexaCoax,Octo,OctoV,OctoCoaxP,OctoCoaxX" defaultvalue="QuadX"/>
+		
+		<field name="MotorChannel" units="" type="enum" defaultvalue="None">
```
peabody124, line 7: So just starting my comments reading here, apologies if this is out of context. Similarly to the current approach of storing the UI information, this mapping is just to cue in the system. There is nothing in code that forces this and the mixer to be consistent, right?

shared/uavobjectdefinition/fixedwingairframesettings.xml
```diff
@@ -0,0 +1,36 @@
+<xml>
+	<object name="FixedWingAirframeSettings" singleinstance="true" settings="true" category="Platform">
+		<description>Select airframe type.  Currently used by @ref ActuatorModule to choose mixing from @ref ActuatorDesired to @ref ActuatorCommand</description>
+		<field name="AirframeType" units="" type="enum" elements="1" options="Classic,Elevon,Vtail" defaultvalue="Classic"/>
+		
+		<field name="RollChannel" units="" type="enum" elements="2" defaultvalue="None">
```
peabody124, line 6: What about when you have two servos for rolling? Or elevons?

flight/Modules/System/systemmod.c
peabody124, line 137: Gotcha. Yeah on an F4 we probably could do that.

flight/Modules/VtolPathFollower/vtolpathfollower.c
```diff
@@ -142,18 +142,7 @@ static void vtolPathFollowerTask(void *parameters)
...
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_HEXA) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_HEXAX) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_HEXACOAX) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_OCTO) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_OCTOV) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_OCTOCOAXP) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_OCTOCOAXX) &&
-			(systemSettings.AirframeType != SYSTEMSETTINGS_AIRFRAMETYPE_TRI) )
-		{
+		if (systemSettings.PlatformType != SYSTEMSETTINGS_PLATFORMTYPE_MULTIROTOR) {
```
peabody124, line 145: This is where a sort of hiearchical type is definitely a code win :)

peabody124:
Ok had a chance to look although I didn't delve into it in detail. For sure, the UAVO representation is more consistent with everything else we do (e.g. is explicit about meaning). The current scope of changes also does make the fight code cleaner since we aren't comparing against a million model types.

I'm basically a bit ambivalent. It will ultimately cost more memory in firmware because I think we should keep the objects (except for maybe helicopter) initialized at all times. In principle using a union model (basically what we currently do) is more efficient but a bit odd and inconsistent so it's ok to drop it. At some point we might use more detailed information on the controller side in which case we'll be glad it's already there.

PR #2106, opened 2015-12-09
GCS Branding, and main fixes

PTDreamer:
This PR externalizes GCS artwork and strings so rebranding is easier and also make it simpler to avoid conflicts across forks.
This in practice makes the branding artwork and strings move outside the gcs src directory and to the REP_SOURCE/branding directory. The placement is so in the future it can also be used by the packaging stuff.
It also fixes main and refactors main. The command line options are now using QCommandLineParser and examples are included for both argument passing to plugins and plugin tests using QTest.
Heavy rebasing was needed so please review with caution.
Similar [PR](https://github.com/d-ronin/dRonin/pull/181) was merged on dRonin.


kubark42:
Cool! Is this basically the same as https://github.com/d-ronin/dRonin/pull/181, right?

mluessi:
For some reason this segfaults for me on Linux:

```
qlist.h L510: { Q_ASSERT_X(i >= 0 && i < p.size(), "QList<T>::at", "index out of range");
providerstrings.cpp L89:     VersionGoogleMap = decrypt("hoaGv8cd7uEX4ukZ7RwTmmudkhqWmenlbRwZK3YqyRQQ7JFyMS1uE4kUlnX266hKmYTQo74yHjO5eXl5jPvrog==").split("%@%").at(1);;
```

tracernz:
Segfaults if the organisation domain is incorrect (because %@% doesn't match if the decryption fails).

PTDreamer:
> Cool! Is this basically the same as d-ronin/dRonin#181, right?

Yes, should be the same.

PTDreamer:
> For some reason this segfaults for me on Linux:

No related to this changeset, but needs to get fixed.

kubark42:
@PTDreamer any idea why that segfault is only appearing now?

@mluessi Is that segfault happening every time you run this, and only on this branch?

mlyle:
It's because of this

```
-    QCoreApplication::setOrganizationDomain(QLatin1String("taulabs.org"));
+    QCoreApplication::setOrganizationDomain(QLatin1String(GCS_PROJECT_BRANDING_HELP));
```

Which is used as an encryption key for the TLMap URLs.

In turn, it's changed to `GCS_PROJECT_BRANDING_HELP = http://github.com/TauLabs`

kubark42:
@mlyle thanks for the insight.

@PTDreamer, is this something you can fix? It sounds like it's the only blocker.

kubark42:
@PTDreamer poke.

PR #2108, opened 2015-12-10
Actuator: Change return type to be a function of the UAVO

kubark42:
The MixerSettings object variable type can change, so this lets the actuator code reflect that.

### DIFF
--- a/flight/Modules/Actuator/actuator.c
+++ b/flight/Modules/Actuator/actuator.c
@@ -93,7 +93,7 @@ static float mix_channel(int ct, ActuatorDesiredData *desired,
 		float curve1, float curve2);
 
 static MixerSettingsMixer1TypeOptions get_mixer_type(int idx);
-static int8_t *get_mixer_vec(int idx);
+static typeof(mixerSettings.Mixer1Vector) *get_mixer_vec(int idx);
 
 /**
  * @brief Module initialization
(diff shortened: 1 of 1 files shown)


flight/Modules/Actuator/actuator.c
```diff
@@ -338,7 +338,7 @@ static void actuator_task(void* parameters)
 float process_mixer(const int index, const float curve1, const float curve2,
 		ActuatorDesiredData *desired)
 {
-	int8_t *vector = get_mixer_vec(index);
+	typeof(&(*mixerSettings.Mixer1Vector)) vector = get_mixer_vec(index);
```
mlyle: consider for these `typeof (mixerSettings.Mixer1Vector) (*vector)` .  The dereferences below would need `(*vector)[MIXERSETTINGS ...`.  The upside is that this form (pointer to array) preserves type information so smart compilers (clang, some versions of gcc) can detect accesses past the end.

note that pointer to array and pointer to first element of an array have the same value, so the `(*vector)` in access is kinda.. vestigial.

kubark42: Ugh, torn between having the compiler do smart things for us and having the code be readable. 

`(*vector)[0]` isn't a pattern we see too often. I suppose I could just put a comment in the code to make it clear what's going on. Maybe that'll be good enough so underexperienced devs don't get hung up on this.

kubark42: Actually, @mlyle, I'm not quite sure what you're suggesting. The functions can't return `typeof(mixerSettings.Mixer1Vector)` because it's an array type. So are you suggesting that only at the point of L341 that be switched to `typeof (mixerSettings.Mixer1Vector) (*vector)  = get_mixer_vec(index);`? I don't think this works, and the compiler seems to agree.

mlyle: Here's an example:

https://gist.github.com/mlyle/2fc29501dd7ad15ba2a6

During compilation it prints out the warnings you'd want:

```
jar% gcc -o example example.c -O2 -Wall
example.c: In function ‘myFunc2’:
example.c:13:2: warning: return from incompatible pointer type [enabled by default]
  return &baaz;
  ^
example.c: In function ‘main’:
example.c:27:26: warning: array subscript is above array bounds [-Warray-bounds]
  printf("%d\n", (*vector)[16]);
```

kubark42: Gotcha. Fixed as suggested.

mlyle:
LGTM.  Has it been tested?

kubark42:
Yes, flight tested and works as expected.

Works well at d-ronin, so merging here as well.

PR #2111, opened 2015-12-23
GCS performance increases through UAVObject manager

kubark42:
As discovered at https://github.com/d-ronin/dRonin/pull/254 and https://github.com/d-ronin/dRonin/pull/312.
- Uses `consts` to reduce memory usage
- Uses `QHash` instead of manually searching through list.

Used daily locally, and seems to work fine for d-ronin users as well. Will merge soon unless there are objections.

### DIFF
--- a/ground/gcs/src/plugins/uavobjects/uavobjectfield.cpp
+++ b/ground/gcs/src/plugins/uavobjects/uavobjectfield.cpp
@@ -3,6 +3,7 @@
  *
  * @file       uavobjectfield.cpp
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2015
  * @see        The GNU Public License (GPL) Version 3
  * @addtogroup GCSPlugins GCS Plugins
  * @{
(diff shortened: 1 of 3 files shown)


mlyle:
Please note if you selectively import changes from the dRonin project, which has a copyright notice attached to it and license terms (GPLv3) that require preservation of the copyright notices when dRonin code is distributed, you should include the appropriate authorship notices.

e.g.

```
 * @author     dRonin, http://dRonin.org/, Copyright (C) 2015
```

By oversight, these were omitted from these particular files originally.  But in any case, selectively pulling commits or PRs will cause these to not appropriately make their way and requires appropriate management.

kubark42:
Of course, thanks for catching this. We treat it just like we do OpenPilot and LibrePilot imports, sorry to have missed it here.

mlyle:
No problem. 

This was a good change for us.  The changes to empty-gadget-instantiation were a bigger difference in load time, though.

kubark42:
The loadtime PR is pending, I need to make some time to test it out. Load time as an idea is much lower on my list of priorities than run-time load, so I thought I'd get this one in first.

peabody124:
Awesome, thanks @kubark42 . If this has been tested and reviewed by you then I'm all for it.

mlyle:
OK, well, this is contained mostly to loadtime benefit, because the config widgets look things up by name to get the objects they're then connected to.

kubark42:
@mlyle Yeah, I saw that a lot was contained to initialize. But it seems like the better of the two to start with.

@peabody124 okay, will merge.

PR #2112, opened 2015-12-23
uavobjgenerator:  Ensure rebuild of flight when necessary

### LINKED ISSUE #2110: Firmware linking against wrong UAVO library
Could be a consequence of https://github.com/TauLabs/TauLabs/pull/1985, but I have seen a problem several times now where the firmware links against the wrong library. Not sure how repeatable this is, but it last occurred to me after building our in-house board firmware on it current branch, switching to #2108 to effect changes, and switching back to our in-house board support branch.

My current workaround is simply to delete the uavo-synthetics directory and rebuild. It can be somewhat hard to link the build error to this fix because it only occurs when an already built file is rebuilt, so it looks like things magically broke (which was my first indication the make process might be responsible).

@mlyle, have you seen this?

kubark42:
This code had one remaining weakness-- on non-windows platforms, if you edited a file and then put it back UAVobjects might not get rebuilt due to moving to an "old" timestamp.  Now we touch the version file to ensure that this is OK.

From https://github.com/d-ronin/dRonin/pull/24.

Fixes #2110

### DIFF
--- a/ground/uavobjgenerator/main.cpp
+++ b/ground/uavobjgenerator/main.cpp
@@ -3,6 +3,8 @@
  *
  * @file       main.cpp
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
+ * @author     Tau Labs, http://taulabs.org/, Copyright (C) 2013
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2015
  * @brief      UAVObjectGenerator main.
  *
  * @see        The GNU Public License (GPL) Version 3
--- a/ground/uavobjgenerator/main.cpp
+++ b/ground/uavobjgenerator/main.cpp
@@ -244,6 +244,8 @@ int main(int argc, char *argv[])
         wiresharkgen.generate(parser,templatepath,outputpath);
     }
 
+    bool changed = false;
+
     /* Symlink each of these to the current dir */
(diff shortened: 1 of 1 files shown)


kubark42:
Used this for several weeks and works fine. d-ronin has reported no problems either.

Our build process was left broken by the earlier changes to the UAVObject library blob. That led to vehicle crashes (see https://github.com/TauLabs/TauLabs/issues/2110#issuecomment-166991437) when the wrong UAVO was silently built into the firmware, so I will merge this in priority.

mlyle:
Merging your own PR?

kubark42:
Well, it's your code and it's from early in the fork, so I figure I'm more reviewer than author. Although I will take full credit for b6db52e!

mlyle:
OK.  I've gotten in trouble for similar things in the past-- making PRs from peoples' branches and then later merging them.

kubark42:
I know it's a very strong\* policy not to merge one's own PR, but I'm sorry if it escalated to getting in trouble. That's certainly not how a project should be run.

[*] But not iron-clad. It's always been acceptable to priority push bug-fixes that are safety or workflow critical.

peabody124:
Yeah pulling in other peoples changes really is that. It seems fine to act as both reviewer and committer if you leave a reasonable delay.

PR #2113, opened 2015-12-24
Sysalarms: only test the field named "Alarm"

kubark42:
Corrects a bug that I introduced years ago with the error codes, but it was only obvious earlier this year. 
The problem was that the code was testing all the fields in the UAVObject, so the solution is to only test the "Alarm" field.

See https://github.com/d-ronin/dRonin/pull/310

### DIFF
--- a/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
+++ b/ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
@@ -3,6 +3,7 @@
  *
  * @file       sysalarmsmessagingplugin.cpp
  * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2015
  * @addtogroup GCSPlugins GCS Plugins
  * @{
  * @addtogroup System Alarms Messaging Plugin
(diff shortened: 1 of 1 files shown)


ground/gcs/src/plugins/sysalarmsmessaging/sysalarmsmessagingplugin.cpp
```diff
@@ -89,42 +90,38 @@ bool SysAlarmsMessagingPlugin::initialize(const QStringList &arguments, QString
  */
 void SysAlarmsMessagingPlugin::updateAlarms(UAVObject* systemAlarm)
 {
-    Q_ASSERT(systemAlarm->getObjID() == SystemAlarms::OBJID);
```
peabody124, line 92: So this check should remain since the code below references constant definitions from the generator. 

Basically any time this error would fire, now you can get unpredictable/undefined behavior.

You can do this safely but it needs to be consistently a runtime constant, so do a string value comparison.

peabody124, line 92: Ps i wont consider this a blocker since run time errors will only happen in non clean builds. Clean builds are fine.

mlyle, line 92: The proper safe way is to use DEPENDPATH in the qt project files so that header dependencies aren't ignored like they are currently.

peabody124, line 92: And if that worked the assert would never fire :)

kubark42:
I'll go ahead and merge this this weekend as it's a useful bugfix. If we have an issue with the Q_ASSERT(), we can put it back in or, my preference, use https://github.com/TauLabs/TauLabs/pull/1974.

peabody124:
Yep nice fix. As I said the bugs will only occur on non-clean builds so no big deal for end users. Just fun surprises when jumping branches :)

PR #2114, opened 2015-12-24
naze32: add "serial" to receiver options

ernieift:
This separates the serial and ppm/pwm configuration on naze32. Now it is possible to configure the receiver port for a DMS or HoTT without PWM/PMM. This is still possible by using "PPM+SERIAL".
"SERIAL" has the advantage of 3% less CPU usage and 122 Bytes more free heap.
I was able to run an autotune procedure on the rcexplorer tricopter pcb-naze with HoTT-SUMD. Without the changes, the naze32 don't arm and there are only 4 bytes heap left.

### DIFF
--- a/flight/targets/naze32/fw/pios_board.c
+++ b/flight/targets/naze32/fw/pios_board.c
@@ -397,32 +397,6 @@ void PIOS_Board_Init(void) {
 #endif	/* PIOS_INCLUDE_PWM */
 		break;
 	case HWNAZE_RCVRPORT_PPMSERIAL:
-		{
-			uint8_t hw_rcvrserial;
-			HwNazeRcvrSerialGet(&hw_rcvrserial);
-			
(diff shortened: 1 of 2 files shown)


ernieift:
Maybe it helps on #2013 too.

mlyle:
Just FYI, this is somewhere that @jihlein made diff to do the same thing in https://github.com/d-ronin/dRonin/pull/145 and an area where we'd drift further apart.

Also, this is not sufficient to run autotune correctly on F1.  Without the event and buffering infrastructure that was introduced in dRonin roughly 1/3rd of points spill.  Some points are also duplicated, and the filter wanders randomly over a fairly large range.

ernieift:
Looks like @jihlein had the same idea before. Too bad. BTW he didn't changed the ADC part.

peabody124:
Nice work @ernieift. LGTM

peabody124:
jenkins: test this please

PR #2118, opened 2016-01-05
stack aligment fix for floating point print

droney:
For all EABI ARM implementations, an eight byte stack alignment is required for floating point printf/sprintf calls. Please see https://wiki.debian.org/ArmEabiPort#Stack_alignment.

This alignment is performed when the USE_NEWLIB define is set in flight/PiOS/Common/printf2.c. Without setting this define, luck of the draw will decide which floating point prints display the correct value and which ones do not, depending on the arguments passed to printf/sprintf.

Unfortunately, printf2 alignment code is flawed and must be addressed in addition to enabling the define. Here is the code snippet:

```
#ifdef USE_NEWLIB
char *cptr = (char *) varg ; //lint !e740 !e826 convert to double pointer
uint caddr = (uint) cptr ;
if ((caddr & 0xF) != 0) {
   cptr += 4 ;
}
double *dblptr = (double *) cptr ; //lint !e740 !e826 convert to double pointer
#else
double *dblptr = (double *) varg ; //lint !e740 !e826 convert to double pointer
#endif
```

Notice the argument pointer is adjusted if any of the lower four bits are set. The correct test to be performed should be `if ((caddr & 0x4) != 0) {`.
[printf2.zip](https://github.com/TauLabs/TauLabs/files/78466/printf2.zip)

See also https://github.com/BrainFPV/TauLabs/issues/40, https://github.com/d-ronin/dRonin/pull/394, and https://github.com/d-ronin/dRonin/pull/410


droney:
hmmm... I submitted a pull request for brainfpv but it appeared in taulabs base code. I guess pull requests not possible for brain. Closing this pull request.

PR #2119, opened 2016-01-06
openlrs: fix cast to not erroneously quantize battery data

peabody124:
This fixes a bug reported by @CrazyCoder that made the OpenLRS send
the battery voltage and current incorrectly resolved.


kubark42:
LGTM. Merging.

PR #2120, opened 2016-01-07
TauOSD support

peabody124:
Provides support TauOSD.

PR #2123, opened 2016-01-21
Math: restore sin_lookup - used by OSD

peabody124:
As requested splitting appart #2120 

This was previously in the code and used.

### DIFF
--- a/flight/Libraries/math/sin_lookup.c
+++ b/flight/Libraries/math/sin_lookup.c
@@ -1,0 +1,146 @@
+/**
+ ******************************************************************************
+ * @addtogroup TauLabsLibraries Tau Labs Libraries
+ * @{
+ * @addtogroup TauLabsMath Tau Labs math support libraries
+ * @{
+ *
+ * @file       sin_lookup.c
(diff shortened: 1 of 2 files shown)


kubark42:
Thanks! If I wrote a unit test for this, could we get in for the PR?

peabody124:
Sound like a good idea

mlyle:
Is there even any reason for this?  brainfpv/dRonin moved to just use sinf.

Sure, this is a few cycles faster, but uses a lot of flash, trades off precision (to single degree), and it's done just a few times per frame (vs. the thousands of sinf's we do per second in core flight control).

peabody124:
Flash is pretty cheap and we dont need more accuracy

mlyle:
Yes, but what is the benefit from this?  It wastes precision, adds more code to maintain, and is an undetectable optimization in CPU used.

peabody124:
@mlyle fair point. updated OSD to use sinf and no problems. I'll close this.

kubark42:
Cool, glad to see agreement. I'm fine either way as there's no clear "perfect" answer.

PR #2124, opened 2016-01-22
PIOS CAN enhancements

peabody124:
This extends PIOS_CAN support transmitting a number of other messages. It has been
demonstrated for feeding an OSD (see #2120) but could provide information for anything
and is fairly easily extensible for bidirectional data packing.

For examples of using it, please see the modules in #2120:
https://github.com/TauLabs/TauLabs/pull/2120/files#diff-12759c883388bba420062295ac381318


flight/Modules/GimbalControl/gimbalcontrol.c
```diff
@@ -72,9 +76,12 @@ int32_t GimbalControlInitialize(void)
 
 	// Create object queues
 	queue = PIOS_CAN_RegisterMessageQueue(pios_can_id, PIOS_CAN_GIMBAL);
+	queue_roll_pitch = PIOS_CAN_RegisterMessageQueue(pios_can_id, PIOS_CAN_ATTITUDE_ROLL_PITCH);
```
kubark42, line 78: Would it be reasonable to rename `queue` at the same time? When one has a shortened name and the other a long name it feels like the short name is somehow special. This isn't the intent of the code, AFAICT.

flight/PiOS/STM32F30x/pios_can.c
```diff
@@ -215,8 +215,56 @@ static void PIOS_CAN_RegisterTxCallback(uintptr_t can_id, pios_com_callback tx_o
...
+	[PIOS_CAN_BATTERY_CURR] = 0x2A1,
+	[PIOS_CAN_RSSI] = 0x2A2,
+	[PIOS_CAN_ALT] = 0x2A3,
+	[PIOS_CAN_FLIGHTSTATUS] = 0x2A4,
+	[PIOS_CAN_GPS_LATLON] = 0x2B1,
+	[PIOS_CAN_GPS_ALTSPEED] = 0x2B2,
+	[PIOS_CAN_GPS_FIX] = 0x2B3,
+	[PIOS_CAN_GPS_VEL] = 0x2B4,
+	[PIOS_CAN_POS] = 0x2B5,
+	[PIOS_CAN_ALARM] = 0x2B6
```
kubark42: If you put a `,` at the end of the final line, it will reduce (by 1) the number of changed lines in case this message is extended in the future.

flight/PiOS/STM32F4xx/pios_can.c
```diff
@@ -222,8 +222,56 @@ static void PIOS_CAN_RegisterTxCallback(uintptr_t can_id, pios_com_callback tx_o
...
+	[PIOS_CAN_BATTERY_CURR] = 0x2A1,
+	[PIOS_CAN_RSSI] = 0x2A2,
+	[PIOS_CAN_ALT] = 0x2A3,
+	[PIOS_CAN_FLIGHTSTATUS] = 0x2A4,
+	[PIOS_CAN_GPS_LATLON] = 0x2B1,
+	[PIOS_CAN_GPS_ALTSPEED] = 0x2B2,
+	[PIOS_CAN_GPS_FIX] = 0x2B3,
+	[PIOS_CAN_GPS_VEL] = 0x2B4,
+	[PIOS_CAN_POS] = 0x2B5,
+	[PIOS_CAN_ALARM] = 0x2B6,
```
kubark42: Do you think it's reasonable to put the two message structs into a private header file, rather than duplicate the code between the two?

flight/PiOS/STM32F4xx/pios_can.c
```diff
@@ -222,8 +222,56 @@ static void PIOS_CAN_RegisterTxCallback(uintptr_t can_id, pios_com_callback tx_o
...
 };
 
+static int32_t get_message_size(uint32_t msg_id) {
```
kubark42: This function could also be wrapped into a commonly shared private header as well.

flight/PiOS/inc/pios_can.h
```diff
@@ -36,8 +36,25 @@
...
+	PIOS_CAN_RSSI = 5,
+	PIOS_CAN_ALT = 6,
+	PIOS_CAN_FLIGHTSTATUS = 7,
+	PIOS_CAN_GPS_LATLON = 8,
+	PIOS_CAN_GPS_ALTSPEED = 9,
+	PIOS_CAN_GPS_FIX = 10,
+	PIOS_CAN_GPS_VEL = 11,
+	PIOS_CAN_POS = 12,
+	PIOS_CAN_ALARM = 13,
+	PIOS_CAN_LAST
```
kubark42, line 52: `,` at the end.

kubark42:
Couple style questions and the hope that code duplication can be reduced. Otherwise, LGTM.

flight/PiOS/STM32F4xx/pios_can.c
peabody124: Good idea. This is definitely meant to be common between F3/F4 definitions. Thanks for catching that.

flight/PiOS/STM32F30x/pios_can.c
peabody124: Will do

flight/PiOS/STM32F4xx/pios_can.c
peabody124: Hm.. This I'm less sure of. Really, the `implementation` should be common between F3 and F4. This method is only used internally to the driver so exposing any information about this would be going the wrong way of abstraction.

Short of trying to really factor this into a common driver between F3 and F4, I'd prefer to leave this part alone. Since the F3/F4 duplication is a large problem I'd rather not try and solve it here. Perhaps, in retrospect, I should have tried using chibios drivers but that just passes the buck.

P.S. if I had a common driver file for pios_can I'd happily move it into that, but it doesn't seem worth creating that file for this one method.

kubark42: There's a ton of code that's identical between the two drivers. I suppose refactoring to reduce function duplication would arguably happen better in a dedicated PR than here.

I'm fine to leave this one as it is for now.

flight/PiOS/STM32F4xx/pios_can.c
kubark42: Thanks! As soon as you've pushed this change, it's good to merge.

flight/PiOS/inc/pios_can.h
peabody124, line 52: Oh wait. I don't think it is needed here, because PIOS_CAN_LAST by definition better always be on the last line :)

kubark42, line 52: Point taken!

flight/Modules/GimbalControl/gimbalcontrol.c
peabody124, line 78: renamed for consistency

peabody124:
@kubark42 ok i ended up doing what you suggested and committing to a `pios_can_common.c` file. In the long run there might be more common parsing that is used by pios_can (e.g. some of what is currently in the module might be generalized) so it seemed to have long term utility.

mlyle:
I assume this one won't merge (per comments on enclosing #2120) until TauOSD schematics merge / are ready, since it just really exists to drive that flight hardware.

peabody124:
This probably ought to. It's a standalone library improvement on PIOS_CAN which is also used by SparkyBGC so it doesn't stand to bit rot. This provides a pretty nice framework for others to build on for things like CAN ESCs as well. 

It also won't drive the OSD as is, even from the Sparky side. I didn't include the OSD Module for exactly that reason since this is the library and not TauOSD support code.

mlyle:
Seems a little strange to me-- only user (and only probable user for quite some time) of these messages is TauOSD and overall they're rather OSD focused.  But OK.

kubark42:
@mlyle I understand your point that it's not optimal to have dangling tools, but it seems worth the cost in this case. I think it's a good thing to merge in an actively developed toolset, even if it isn't used in the main branch. As long as it is properly coded and supported, then it reduces the barrier to entry for others who would like to use the same tools, and it avoids the possibility of duplication of effort.

PR #2125, opened 2016-01-22
Math libraries used by TauOSD/BrainFPV

peabody124:
This contains the math libraries used for visualization on TauOSD/BrainFPV and
is a splitting up of #2120. Credit for the original commit goes to @BrainFPV

### DIFF
--- a/flight/Libraries/mgrs/mgrs.c
+++ b/flight/Libraries/mgrs/mgrs.c
@@ -1,0 +1,1366 @@
+/***************************************************************************/
+/* RSC IDENTIFIER:  MGRS
+ *
+ * ABSTRACT
+ *
+ *    This component converts between geodetic coordinates (latitude and 
+ *    Longitude) and Military Grid Reference System (MGRS) coordinates. 
+ *
+ * ERROR HANDLING
+ *
+ *    This component checks parameters for valid values.  If an invalid value
+ *    is found, the error code is combined with the current error code using 
+ *    the bitwise or.  This combining allows multiple error codes to be
+ *    returned. The possible error codes are:
+ *
+ *          MGRS_NO_ERROR          : No errors occurred in function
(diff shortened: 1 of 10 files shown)


flight/Libraries/mgrs/mgrs.c
```diff
@@ -0,0 +1,1366 @@
...
+ *    15-09-99          Reengineered upper layers
+ *    02-05-03          Corrected latitude band bug in GRID_UTM 
+ *    08-20-03          Reengineered lower layers
+ */
+
+
+/***************************************************************************/
+/*
+ *                               INCLUDES
+ */
```
kubark42, line 79: Is there any kind of ~~copyright or~~ authorship we could reference in this file?

flight/Libraries/mgrs/polarst.c
```diff
@@ -0,0 +1,523 @@
...
+ *
+ * MODIFICATIONS
+ *
+ *    Date              Description
+ *    ----              -----------
+ *    06-11-95          Original Code
+ *    03-01-97          Original Code
+ *
+ *
+ */
```
kubark42, line 87: Same question as above. It's not problem to merge in (I found the "LICENSES:  None apply to this component."), but for posterity it would be good to know who was the original author.

kubark42:
Fine to merge as is, but it would be nice to get at least a tiny bit of background on the authorship and code source. As per the code, the license is free and clear, and there's a snail-mail address for the original authors, but I wonder if we could do better.

@mluessi Do you have a URL where this code came from originally?

peabody124:
@kubark42 appears to be from https://github.com/hobu/mgrs -- I'll add a reference in the commit message

kubark42:
@peabody124 Thanks, that's perfect.

PR #2126, opened 2016-01-24
MOOSD Alarm tweaks

peabody124:
This cleans up a bit of the code for displaying alarms on MWOSD, so that
the common constants can be used in #2120 

Changes from https://github.com/d-ronin/dRonin/pull/63/commits


flight/Libraries/alarms.c
```diff
@@ -238,6 +238,157 @@ static int32_t hasSeverity(SystemAlarmsAlarmOptions severity)
...
+	[SYSTEMALARMS_ALARM_ALTITUDEHOLD] = "A-HOLD",
+	[SYSTEMALARMS_ALARM_BOOTFAULT] = "BOOT",
+	[SYSTEMALARMS_ALARM_GEOFENCE] = "GEOFENCE",
+	[SYSTEMALARMS_ALARM_TEMPBARO] = "TEMPBARO",
+	[SYSTEMALARMS_ALARM_GYROBIAS] = "GYROBIAS",
+	[SYSTEMALARMS_ALARM_ADC] = "ADC",
+};
+
+// If someone adds a new alarm, we'd like it added to the array above.
+DONT_BUILD_IF(NELEMENTS(alarm_names) != SYSTEMALARMS_ALARM_NUMELEM, AlarmArrayMismatch);
```
kubark42, line 268: Hmm... This is going to be annoying to support because it increases the amount of boilerplate to add new alarm functionality. Can we architect this so that if it doesn't see an explicitly alarm it chooses a default name?

Also, would it make sense to alphabetize this list? It's true that there's not much advantage to that, but I don't see why it would hurt, either.

kubark42:
I don't know if I'm a fan of this. My concern is that small changes to the UAVO will make a larger breakage system-wide. That's not a blocker-- indeed breakage is to expected anytime a UAVO field is changed--, but so much of it seems like boilerplate that it would be lovely to automate that. 

Maybe we can automagically generate these stringified names from the UAVO definition?

peabody124:
Let me think on this. I ultimately agree. But we dont want strings for every option in the flash so it is a bit subtle. If I cant think of a clean way to implement in the generator I might ask to defer that. Similarly in the GCS we have to add lots of stuff to make an alarm represented in a friendly way, so in comparison adding this to get errors on an OSD is very high yield...

kubark42:
@peabody124 How about creating a new tag in the UAVO definition file? Then we can explicitly turn on or off the "stringiness". Might even have a short name (for OSD) and a long name (for GCS) so that GCS alarm friendliness can increase.

peabody124:
Flag was exactly what i was thinking. I just want to get a sense of how much work that is to bundle for this feature before commiting :)  

Gcs already has strings for options exposed through the getFields method IIRC

mlyle:
Note you don't need a flag, because if the string table isn't accessed, the linker won't bring it in.. so you can always generate string table without cost.

kubark42:
@peabody124 any thoughts on this?

peabody124:
@kubark42 could we merge as is for now and create an issue for these to be automagically generated? I think it's a great idea but think getting these improvements in is probably still good without it.

kubark42:
Sounds good to me. Merging.

PR #2129, opened 2016-01-24
Fastheap check

peabody124:
This adds checking of the remaining memory in the fast heap.

From https://github.com/d-ronin/dRonin/pull/235

This adds description flags into the .xml file which he don't (currently)
support, but these should be parsed harmlessly.


kubark42:
@tracernz's description tags are net positives, whether they're used by the generator or not.

Merging.

PR #2131, opened 2016-01-24
VtolPathFollower: immediately start rising to target alt

peabody124:
From https://github.com/d-ronin/dRonin/pull/279

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_fsm.c
@@ -58,8 +58,9 @@
 
 // Various navigation constants
 const static float RTH_MIN_ALTITUDE = 15.f;  //!< Hover at least 15 m above home */
-const static float RTH_ALT_ERROR    = 1.0f;  //!< The altitude to come within for RTH */
+const static float RTH_ALT_ERROR    = 0.7f;  //!< The altitude to come within for RTH */
 const static float DT               = 0.05f; // TODO: make the self monitored
+const static float RTH_CLIMB_SPEED  = 0.75f; //!< Climb at 0.75m/s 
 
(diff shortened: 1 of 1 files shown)


kubark42:
@peabody124 did you want to respond to @pug398 or is that answered elsewhere?

pug398:
The question was aimed at the fact RTH was in a state of flux and due to events was left in an unfinished state. If at 15m altitude when triggered it would climb 1.5 m and all was good. If at extremely low altitude when triggered it would climb 1.5m and then sit there for 10s. This could lead to a condition of say sitting in the middle of a highway 3m off the ground for 10 seconds without any control. I would say set it back like it was which worked ok or adopt new RTH fsm.

peabody124:
Yeah I hadn't replied cause i need to go throught the logic and run some simulations. As far as I know there is no problem with our current code so don't want to merge this until confirmed ok.

mlyle:
Project level copyright notices.

peabody124:
@mlyle there are none from this PR that I see https://github.com/mlyle/dronin/blob/f3b3b8c6dcd8af87f1e5f98c79381918b2d21d7e/flight/Modules/VtolPathFollower/vtol_follower_fsm.c

mlyle:
It was added 5 days later.

If you want, we can just go ahead and put the no-redis notice on all files on dRonin now along with a copyright 2016 to avoid any confusion in the future.. We've avoided doing that to avoid confusion of attribution and to avoid excess conflict, but would be well within our legal rights to do so, and it'd avoid a lot of the conflicts we've had to deal with multiple PRs updating the line.  It would ensure all future changes are covered, but I'd hope you'd include the project-level copyright notices in good faith.

peabody124:
@kubark42 tested in simulation. good to merge.

PR #2132, opened 2016-01-24
ManualControl: fixed switch arming issues

peabody124:
1: SWITCHDELAY got stuck in ARM_STATE_ARMED_STILL_HOLDING,
which prevented it from auto disarming from low throttle etc.

2: SWITCHDELAY & SWITCHTHROTTLEDELAY would rearm themselves
after a low throttle disarm event. The rising edge detection
prevents that.
1. Avoid a situation that requires toggling arming switch multiple times

From https://github.com/d-ronin/dRonin/pull/434
also https://github.com/d-ronin/dRonin/pull/474

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -6,7 +6,8 @@
  * @{
  *
  * @file       transmitter_control.c
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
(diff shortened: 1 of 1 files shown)


peabody124:
I haven't fully reviewed the logic in this yet, just wanted to create PR.

mlyle:
These are good things to pick up, but be aware even with this arming is still not entirely right and we plan to rewrite the arming state machines.

PR #2133, opened 2016-01-25
Add statistics library

kubark42:
This adds a statistics library to Tau Labs. It has several functions:
1. Calculate linear mean and standard deviation
2. Calculate linear variance
3. Calculate Pearson's coefficient
4. Calculate circular mean and standard deviation
5. Calculate angular deviation
6. Calculate a Bayesian update (concerning dynamic systems)

In addition, there is a full set of unit tests. They have been tested against MATLAB (with the caveat below that linear standard deviation has an efficiency shortcut[*] ).

I wrote this library for internal analysis of sensors and have been very happy with the in-flight results. I am submitting it in hopes that others find it useful.

@mlyle You might find this useful as well. I'm especially keen to know if you see interesting compiler optimizations that I've missed.
## 

[*] The linear standard deviation is calculated using the less frequent division by N instead of N-1. This is theoretically less accurate, but for the purposes of evaluating standard deviation over 100 samples only leads to a loss of 1% accuracy, which seems worthwhile if this function will be executed frequently.
##

### DIFF
--- a/flight/Libraries/math/statistics.c
+++ b/flight/Libraries/math/statistics.c
@@ -6,10 +6,26 @@
  * @{
  *
  * @file       statistics.c
- * @author     Kenn Sebesta, Copyright (C) 2015.
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2015
  * @brief      Statistics support
+ * @see        The GNU Public License (GPL) Version 3
  *
  *****************************************************************************/
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ */
 
 #include "statistics.h"
 
--- a/flight/Libraries/math/statistics.c
+++ b/flight/Libraries/math/statistics.c
@@ -135,3 +135,124 @@ float get_linear_variance(const struct linear_mean_and_std_dev *X)
 
 	return variance;
 }
+
+
+/**
+ * @brief initialize_circular_sums Because floating point values are used,
+ * there will be some loss of precision. Thus, it is important to periodically reset the parameters.
+ *
+ * Cf. https://en.wikipedia.org/wiki/Directional_statistics
+ * @param X struct holding the stochastic parameters
+ * @param num_samples number of samples
+ * @param y vector of samples
+ */
+void initialize_circular_sums(struct circular_mean_and_std_dev *X, uint16_t window_size, uint16_t num_samples, const float y_R[])
+{
+	X->window_size = window_size;
+
+	X->T0 = num_samples;
+	X->T0_2 = num_samples * num_samples;
+
+	// Reinitialize the accumulators
+	X->S1 = 0;
+	X->C1 = 0;
+
+	for (int i=0; i<num_samples; i++) {
+		// Sum of elements
+		X->S1 += sin(/*(double)*/y_R[i]);
+		X->C1 += cos(/*(double)*/y_R[i]);
+	}
+}
+
+
+/**
+ * @brief incremental_update_circular_sums  Basically functions like a FIFO, but in the
+ * sense that it is adding and removing cumulative data. NOTE: Because this uses float point values and there
+ * will be some loss of precision, it is expected that the parameters must be periodically reset.
+ *
+ * Cf. https://en.wikipedia.org/wiki/Directional_statistics
+ * @param X struct holding the stochastic parameters
+ * @param x_oldest The chronologically oldest element in the buffer.
+ * @param x_new The chronologically newest element, to be added to the buffer.
+ */
+void incremental_update_circular_sums(struct circular_mean_and_std_dev *X, const float y_oldest, const float y_new)
+{
+	// Grow T0, number of samples, until reaching the desired size of the sample set
+	if (X->T0 < X->window_size) {
+		X->T0++; //
+		X->T0_2 = X->T0*X->T0; // Squared number of samples
+
+		// If the array isn't at full size yet, then simply add the new values to the sum
+		X->S1 += sin(y_new);
+		X->C1 += cos(y_new);
+	} else {
+		X->S1 += -sin(y_oldest) + sin(y_new);
+		X->C1 += -cos(y_oldest) + cos(y_new);
+	}
+}
+
+
+/**
+ * @brief get_circular_mean This recognizes that the mean for a dataset of angles is calculated
+ * by a rolling sum of sines and cosines of angles.
+ *
+ * Cf. https://en.wikipedia.org/wiki/Directional_statistics
+ * @param X struct holding the stochastic parameters
+ * @return the mean for the dataset represented by the stochastic parameters
+ */
+float get_circular_mean(const struct circular_mean_and_std_dev *X)
+{
+	float circular_mean = atan2f(X->S1, X->C1);
+
+	return circular_mean;
+}
+
+
+/**
+ * @brief get_circular_standard_deviation This recognizes that the standard deviation for a dataset
+ * can be calculated by a rolling sum of number of data points and sum of sines and cosines of angles.
+ *
+ * Cf. https://en.wikipedia.org/wiki/Directional_statistics
+ * @param X struct holding the stochastic parameters
+ * @return circular_std_dev the standard deviation for the dataset represented by the stochastic parameters
+ */
+float get_circular_standard_deviation(const struct circular_mean_and_std_dev *X)
+{
+	float circular_std_dev;
+
+	// Calculate R^2, the norm of the average vector
+	double R2 = (X->S1*X->S1 + X->C1*X->C1) / X->T0_2;
+
+	if (R2 <= 1) {
+		circular_std_dev = sqrtf(-log(R2));
+	} else {
+		circular_std_dev = 0;
+	}
+
+	return circular_std_dev;
+}
+
+
+/**
+ * @brief get_angular_deviation This recognizes that the angular deviation for a dataset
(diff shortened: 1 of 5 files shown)


flight/Libraries/math/statistics.c
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * Cf. https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+ * Cf. http://math.stackexchange.com/questions/102978/incremental-computation-of-standard-deviation
+ * @param X struct holding the stochastic parameters
+ * @return std_dev the variance (standard deviation squared) for the dataset represented by the stochastic parameters
+ */
+float get_linear_variance(const struct linear_mean_and_std_dev *X)
+{
+	// sigma^2 = T2/T0 - (T1/T0)^2
+	float variance = (X->T2*X->T0 - X->T1*X->T1) / (X->T0*X->T0);
```
kubark42, line 206: This can also be calculated as 

```
float mean_X = X->T1 / X->T0;
float variance = X->T2/X->T0 - mean_X*mean_X;
```

which saves two multiplications but adds a division. 

I'm unsure which is the generally more optimal approach. I'm open to arguments either way.

kubark42:
@solidgoldbomb, if you have a moment, could you briefly review my unit test framework and tell me if I'm violating any best-practices for unit tests? I'm still a little hazy on where best to draw the line on avoiding code duplication and yet still fully testing the functions.

flight/Libraries/math/statistics.c
```diff
@@ -0,0 +1,352 @@
...
+ * @return
+ */
+float pearson_correlation(const float x_autovariance, const float y_autovariance, const float variance)
+{
+	// If either autovariance is 0 (or close to), then we don't know what the correlation is.
+	// The other variable could be static or it could be varying. The result is either -Inf
+	// or +Inf, depending on from which direction the first autovariance value is
+	// approaching 0. Rather than try to calculate this, we simply return an impossible
+	// value to indicate which
+	if (x_autovariance < 1e-5) {
```
mlyle: f suffix on all of these

flight/Libraries/math/statistics.c
```diff
@@ -0,0 +1,352 @@
...
+	X->window_size = window_size;
+	X->T0 = num_samples;
+
+	// Reinitialize the accumulators
+	X->T1 = 0;
+	X->T2 = 0;
+
+	for (int i=0; i<num_samples; i++) {
+		// Sum of elements
+		X->T1 += (double)y[i];
```
mlyle, line 111: The cast is implied and has no effect.  There's a lot of these through the existing Tau code and I suggest they be avoided in general (having a cast effectively neuters type checking in the future)-- not that it matters so much here.

flight/Libraries/math/statistics.c
```diff
@@ -0,0 +1,352 @@
...
+ * @param belief_H1 The H1 hypothesis (also called the NULL HYPOTHESIS), e.g.
+ * the vehicle is still flying.
+ * @param p_sense_condition_H0 The probability of sensing the H0 condition, given the real condition
+ * @param p_sense_condition_H1 The probability of sensing the H1 condition, given the real condition
+ * @param p_H0_Ux_H0 The probability of state H0 staying H0 after the control is applied
+ * @param p_H0_Ux_H1 The probability of state H1 becoming H0 after the control is applied
+ * @param p_H1_Ux_H0 The probability of state H0 becoming H1 after the control is applied
+ * @param p_H1_Ux_H1 The probability of state H1 staying H1 after the control is applied
+ */
+void bayes_filter(double *belief_H0, double *belief_H1,
```
mlyle, line 51: Some of the stuff below, like the circular mean filters, really needs precision and it's hard to envision running it on a large data set at single precision with larger values.  

This particular one seems to be overkill at double, and using it at double would put it out of reach for most targets.  On the other hand this type of Bayesian filter is routinely run at half-float precision on GPUs for hypothesis generation in scientific computing.  Would it be possible to move this one to float?

(All targets could benefit from an are-we-flying filter for various types of safety improvement, but relatively few could spare this implementation-- requiring either two implementations for different target sizes, neglecting the flash-constrained F3 and F1 targets, or this change).

Alternatively, one could #define a type before including this file, which in turn is typedef'd and used in the prototypes.

flight/Libraries/math/statistics.c
```diff
@@ -0,0 +1,352 @@
...
+	// Reinitialize the accumulators
+	X->T1 = 0;
+	X->T2 = 0;
+
+	for (int i=0; i<num_samples; i++) {
+		// Sum of elements
+		X->T1 += (double)y[i];
+
+		// Sum of squares of elements
+		X->T2 += (double)y[i]*(double)y[i];
```
mlyle, line 114: 1.0 \* ... can be used for type promotion without the scariness of a cast on this line, though I've got less of an objection to this one.

flight/Libraries/math/statistics.c
```diff
@@ -0,0 +1,352 @@
...
+ * Cf. http://math.stackexchange.com/questions/102978/incremental-computation-of-standard-deviation
+ * @param X struct holding the stochastic parameters
+ * @return std_dev the standard deviation for the dataset represented by the stochastic parameters
+ */
+float get_linear_standard_deviation(const struct linear_mean_and_std_dev *X)
+{
+	float std_dev;
+
+	// sigma = sqrt(T0*T2 - T1^2)/T0
+	double radicand = X->T0*X->T2 - X->T1*X->T1;
```
mluessi, line 181: Shouldn't the first `T0` be `T2`?

flight/Libraries/math/statistics.c
kubark42, line 51: > really needs precision and it's hard to envision running it on a large data set at single precision with larger values.

Yeah, for sure! In fact, in my experience all the statistics functions should go to pure doubles. Single precision causes headaches because things look like they should work and then wind up not in practice because of simple design decisions like increasing the number of sample points. Bayesian hypothesis testing was one of those areas, because even five 9's of certainty is low (unless you have an excellent understanding of your PDF, which is unlikely and not particularly valuable in many scenarios).

I think it'd be cool if someone were to realize there's a neat application on the F1 and then use this code as the basis for a reduced precision version. If there were a specific, well-bounded question that we were looking to ask I bet there's a good chance we could to optimize all the math to fit. But I wrote this to be a generic solution so it's a little early for optimization. The risk of reducing precision is that people will use the lower precision in a knee-jerk fashion, not realizing the ramifications, and wind up having a miserable experience.

flight/Libraries/math/statistics.c
kubark42, line 111: ~~I think this was from back when I was experimenting with trying to make floats work. Thanks for catching it.~~

Leaving this in to be explicit, esp. in case the data format changes later.

flight/Libraries/math/statistics.c
kubark42, line 181: You mean `X->T2*X->T2 - X->T1*X->T1`? No, it needs to be `X->T0`. It's the consequence of some nice series multiplications.

flight/Libraries/math/statistics.c
kubark42, line 114: Yeah, this one is still necessary as long as the input value is float. It's arguable that the input values themselves don't need anywhere near the precision of a float, so long as the statistical analysis has double precision.

mlyle, line 114: Yes, 1.0 \* ... will promote to double though without a cast (and without the downsides of casting).  OTOH it's slightly more arcane.

kubark42, line 114: ~~Fair enough. I'll change it. It is probably less likely to be accidentally changed than a double anyway.~~

Fails to compile because of double promotion. Leaving it as is.

flight/Libraries/math/statistics.c
mlyle, line 51: Well, the example is something we definitely need for various things-- enhanced failsafe and "air mode" need to have a good estimate of whether we're in flight.  Which could be heuristic or could be based on hypothesis filtering.

And, I don't know.  For many things, 2-3 9's is great, and float is good for 6+.  Especially in many circumstances we'll have a great initial prior ("almost certainly not flying"), fairly high certainty observations along the way (high throttle command?  sustained non-unity acceleration?  varying altimeter?)  and thresholds using the test are likely to be somewhere like 90-95%.  Certainly it would be better to encourage the use of something like this-- even lacking some precision-- instead of other heuristics for efficiency.

kubark42, line 51: YMMV, but single-precision floats crapped out for me long before I had a reliable usage around five 9's. Not saying you can't get there, just that my tests had sufficient variance that important details went missing and I ultimately didn't feel I had much confidence in my "belief" due to the numerical imprecision. Things improved once I moved to doubles (much like it does with Kalman Filters).

I think the suitability for a single-precision version will have to be determined on a case-by-case scenario. At this early stage, it's probably safest to err to the side of "works in all cases, but isn't optimally efficient".

mlyle, line 51: OK.  Just note that in the things I'm thinking we'll use this for-- in dRonin the lack of the ability to specify will either force us to diverge on this (editing it to let type be specified) or not use this library.

kubark42, line 51: Understood. Because it's currently unhooked to anything, I'm expecting you won't bring this over until you have a solid use for it, at which time you'll have resolved for that usage whether singles are acceptable.

The submission is to get people's brains thinking about all the ways statistics can be powerful, and give them a unit-tested library to try stuff out. I'm excited to see what people come up with. Glad you're taking a look.

peabody124:
I'll review this and should be able to merge it this weekend. 

Just out of curiosity, the copyright line is your name. Would you be willing to change it to TL? I'm not sure if we want to start adding all individuals in that section, ideally.

kubark42:
> Just out of curiosity, the copyright line is your name. Would you be willing to change it to TL? I'm not sure if we want to start adding all individuals in that section, ideally.

Thanks for catching this. This is because until our code is released externally it's not under GPL. I can add the header in, but if we wanted to start down the path of #2155, this would be a great place to try it out. Thoughts?

peabody124:
@kubark42 thats fine with me, although might slow down this merge

kubark42:
@peabody124 I'm fine with that. This PR is pretty much guaranteed to be conflict free so I'm not worried. If that's the only remaining issue I can always go with the embedded license if time becomes critical.

peabody124:
Want to fix your header and we can merge this?

kubark42:
Yup. I'll do that as soon as the parent repo reemerges from the mandatory 10 day counter-notice period.

kubark42:
Frivolous DMCA take-down successfully refuted, so can push to repo again. Upshot: new commit relicensing code to GPLv3, merge when you're ready.

PR #2127, opened 2016-01-24
Dronin backport

peabody124:
This is a back port of some smaller fixes from https://github.com/d-ronin/dRonin/

These are all merged to their repository and in the latest release so should be fairly
stable. Each commit is also fairly atomic but I'd prefer not to create a PR for each one.
Reading the commit log should give a pretty clear idea what was done.

I specifically avoided PRs that touch on packaging / branding (duh). Also I avoided some
ones where there seemed to be instability and reversions. A few things that touch on more
critical systems or might be controversial, I created standalone PRs. Things using the new
DR event system or that were bug fixes related to the DR event system I also avoided.

Here are the links to the original PRs:
https://github.com/d-ronin/dRonin/pull/96
https://github.com/d-ronin/dRonin/pull/120
https://github.com/d-ronin/dRonin/pull/127
https://github.com/d-ronin/dRonin/pull/132
https://github.com/d-ronin/dRonin/pull/137
https://github.com/d-ronin/dRonin/pull/140
https://github.com/d-ronin/dRonin/pull/141
https://github.com/d-ronin/dRonin/pull/146
https://github.com/d-ronin/dRonin/pull/153
https://github.com/d-ronin/dRonin/pull/173
https://github.com/d-ronin/dRonin/pull/192
https://github.com/d-ronin/dRonin/pull/201
https://github.com/d-ronin/dRonin/pull/203/
https://github.com/d-ronin/dRonin/pull/250
https://github.com/d-ronin/dRonin/pull/251
https://github.com/d-ronin/dRonin/pull/259
https://github.com/d-ronin/dRonin/pull/262
https://github.com/d-ronin/dRonin/pull/267
https://github.com/d-ronin/dRonin/pull/275
https://github.com/d-ronin/dRonin/pull/285
https://github.com/d-ronin/dRonin/pull/286
https://github.com/d-ronin/dRonin/pull/311
https://github.com/d-ronin/dRonin/pull/340
https://github.com/d-ronin/dRonin/pull/341
https://github.com/d-ronin/dRonin/pull/353
https://github.com/d-ronin/dRonin/pull/365
https://github.com/d-ronin/dRonin/pull/378
https://github.com/d-ronin/dRonin/pull/385
https://github.com/d-ronin/dRonin/pull/413
https://github.com/d-ronin/dRonin/pull/414
https://github.com/d-ronin/dRonin/pull/426
https://github.com/d-ronin/dRonin/pull/441
https://github.com/d-ronin/dRonin/pull/444
[...]

PR #2136, opened 2016-01-26
GCS:ConnectionMgr: Messagebox on connect failed

kubark42:
If the attempted connection fails, we fail silently except for a message to the
debug log, not super friendly to users who will see that nothing happened when
they clicked connect. This adds a popup message box.

From https://github.com/d-ronin/dRonin/pull/96


peabody124:
Fixed title since I think that was a typo :)

kubark42:
@peabody124, thanks!

Reviewed at #2127, Jenkins says it's okay. Merging.

PR #2141, opened 2016-01-26
GCS:Config: Ignore lack of StateEstimation object

kubark42:
F1 boards don't have this object and it generates a warning when not present and the Attitude tab is saved.

From https://github.com/d-ronin/dRonin/pull/132


kubark42:

PR #2143, opened 2016-01-26
GCS:UAVTalk: Clear queues properly

kubark42:
The empty() function just returns whether the queue is empty, this was not the intent of the code. I went through the code and superficially confirmed that there are no other mistakes of this kind.

From https://github.com/d-ronin/dRonin/pull/140


kubark42:

PR #2144, opened 2016-01-26
GCS:CrashReporter: Move Objective-C sources into right place

kubark42:
Stops QMake issuing a warning.

From https://github.com/d-ronin/dRonin/pull/141


kubark42:

PR #2146, opened 2016-01-26
GCS: Remove WelcomeMode object

kubark42:
One too many lines removed in #2101, results in m_welcomeMode and not being cleaned up and a subsequent warning on exiting GCS.

From https://github.com/d-ronin/dRonin/pull/203.


kubark42:

PR #2147, opened 2016-01-26
Fix maximim yaw_scale to match the tested value

kubark42:
Slight bug in firmware. Fixes maximum yaw_scale to match the tested value.

From d-ronin/dRonin#262.


kubark42:

PR #2151, opened 2016-01-27
Attitude: increase stack size to 2504 from 2200

kubark42:
"Static stack analysis indicates function will use 2440 at peak, primarily through world magnetic model.  This is a nasty one, because most attempts to dynamically measure stack won't cause home position
update.  Hence only at the beginning of real navigation flights have we been blowing stack for quite some time!"

From https://github.com/d-ronin/dRonin/pull/275


kubark42:

PR #2153, opened 2016-01-27
UAVO: Add description element to fields, add tooltips in UAVO Browser

kubark42:
Been testing this in-house for a while, hasn't introduced any problems.

From https://github.com/d-ronin/dRonin/pull/60 (Note that this doesn't import the actual field changes)


kubark42:
P.S. @peabody124 Not sure if you caught it, but this one needs review because it wasn't in #2127.

peabody124:
Code looks sensible. Compiling to test. Looks like SystemStats has per field description

PR #2157, opened 2016-01-27
GCS: configurations: housekeeping

kubark42:
1. Reflect that `laptop` is no longer an option
2. Change the default netbook view to have the PfdQmlGadget

From https://github.com/d-ronin/dRonin/pull/365.


kubark42:

PR #2162, opened 2016-01-27
UAVOGen:GCS: Work around QML int8 bug, export UAVO classes to QML

kubark42:
- Export int16 rather than int8 to QML to avoid QT bug 37241, previously we used some javascript code but this becomes a real pain as we develop more QML code that uses enums and int8 fields.
- Export the UAVO class defs to QML, added a "Class" suffix to avoid clashing with all the existing QML code. This allows enum options to be accessed from QML among other things.

From https://github.com/d-ronin/dRonin/pull/498.

### DIFF
--- a/ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
+++ b/ground/gcs/share/taulabs/pfd/default/PfdIndicators.qml
@@ -1,5 +1,4 @@
 import QtQuick 2.0
-import "convertint8.js" as CInt8
 
 Item {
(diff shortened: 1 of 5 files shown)


kubark42:
Jenkins test this please.

peabody124:
jenkins: test this please

kubark42:
Huh, this is proving stubborn. I need to try it on my Linux VM and see if it's reproducible on other machines.

PR #2164, opened 2016-01-28
UAVOGen: Fix broken inherited enum defaults

kubark42:
When an enum type was inheriting from a parent field, and the child had a different list of options, the default value would be incorrect.

From https://github.com/d-ronin/dRonin/pull/499.

### DIFF
--- a/ground/uavobjgenerator/generators/flight/uavobjectgeneratorflight.cpp
+++ b/ground/uavobjgenerator/generators/flight/uavobjectgeneratorflight.cpp
@@ -272,9 +272,15 @@ bool UAVObjectGeneratorFlight::process_object(ObjectInfo* info)
             {
                 if ( info->fields[n]->type == FIELDTYPE_ENUM )
                 {
+                    int defaultVal;
+                    if (info->fields[n]->parent != NULL)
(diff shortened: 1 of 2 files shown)


ground/uavobjgenerator/generators/flight/uavobjectgeneratorflight.cpp
```diff
@@ -272,9 +272,15 @@ bool UAVObjectGeneratorFlight::process_object(ObjectInfo* info)
             {
                 if ( info->fields[n]->type == FIELDTYPE_ENUM )
                 {
+                    int defaultVal;
+                    if (info->fields[n]->parent != NULL)
+                        defaultVal = info->fields[n]->parent->options.indexOf( info->fields[n]->defaultValues[0] );
+                    else
+                        defaultVal = info->fields[n]->options.indexOf( info->fields[n]->defaultValues[0] );
+
```
kubark42, line 280: This code snippet is repeated 4 times in this PR. It might make sense to look at the generators and see how much code is being duplicated and if we can move functions like this into a parent class.

kubark42:
Reviewed by @peabody124 in #2127, Jenkins says it's okay. Merging.

PR #2167, opened 2016-01-29
Attitude: don't use local copies of AttitudeActual

kubark42:
We're currently using function-level copies of AttitudeActual before writing the data to the UAVObject. I think it's generally better to have a single copy which is accessible to all local functions. The downside is that if a local function were somehow to interrupt a local copy to AttitudeActual that it could get an incomplete object. However, since we don't have any mechanism for a local function interrupting another local function, this seems like an implausible scenario.

I've been using it for several months and have not had any issues.

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -166,7 +166,7 @@ static int32_t setNavigationNone();
 //! Update the complementary filter attitude estimate
 static int32_t updateAttitudeComplementary(bool first_run, bool secondary, bool raw_gps);
 //! Set the @ref AttitudeActual to the complementary filter estimate
-static int32_t setAttitudeComplementary();
+static int32_t setAttitudeComplementary(AttitudeActualData *attitudeActual);
 
 static float calc_ned_accel(float *q, float *accels);
 static void cfvert_reset(struct cfvert *cf, float baro, float time_constant);
@@ -176,7 +176,7 @@ static void cfvert_update_baro(struct cfvert *cf, float baro, float dt);
 //! Update the INSGPS attitude estimate
 static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode);
 //! Set the attitude to the current INSGPS estimate
-static int32_t setAttitudeINSGPS();
+static int32_t setAttitudeINSGPS(AttitudeActualData *attitudeActual);
 //! Set the navigation to the current INSGPS estimate
 static int32_t setNavigationINSGPS();
 static void updateNedAccel();
@@ -311,6 +311,8 @@ static void AttitudeTask(void *parameters)
 	last_algorithm = 0xfffffff;
 	last_complementary = false;
 
+	AttitudeActualData attitudeActual;
+
 	// Main task loop
 	while (1) {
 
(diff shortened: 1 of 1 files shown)


peabody124:
Just out of curiosity, why do you think it is better to have a single copy? From a general programming best practice, the more you locally scope variables, the better. This is because of what you suggest - the risk of unintended side effects if things stomp on shared variables. For sure, in embedded systems sometimes you intentionally violate this, normally to avoid large copy operations or to reduce memory usage.

In this case, the memory usage looks like it should be the same, because you aren't removing variables amongst functions that are nested. It does have less LOC which is nice :)

Anyway, I don't see a particular bug or blocker in it, but would just like to better understand since it does seem slightly worse programming style.

kubark42:
> Just out of curiosity, why do you think it is better to have a single copy?

I haven't been able to push the PR yet, but I have a batch logging function which lets me capture every single AttitudeActual data point in the logging chip. This change was necessary to make this happen cleanly.

About the best place to do the high speed write to the batch object is https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/Attitude/attitude.c#L393. There, we are guaranteed that no matter the path through the module, the AttitudeActual is now up-to-date. I could have fetched the UAVO with AttitudeActualGet, or even done a callback, but all those seem incompatible with the high efficiency needed to log all data packets. The global was a clean, lightweight way to do this.

I'm also fine for this not to merge, since it's very few LOC that are different and it's highly unlikely this cause a merge conflict for me down the road..

mlyle:
Yah, moving things from stack to globals without some kind of code generation benefit is a misfeature.

peabody124:
@kubark42 you ok not to merge this then? just for lols, you could also write your code downstream to use an additional Get operation to make it not need this diff. It would be interesting to check there were no meaningful increase in CPU usage. I think the time to grab the semaphore is normally minimal (and when it isn't, this is catching exactly what the concern is ;-).

kubark42:
Because we don't have any internal timestamp on the UAVOs it's not the best idea to add additional jitter, even if the `Get` operation is pretty quick. However, it probably doesn't add substantially either so it might be one of those "measurable but not noticeable" things.

What I find odd, though, is using the UAVObjects publish-subscribe mechanism to do interfunction communication. Maybe we could consider having a local version of AttitudeActual, scoped to `AttitudeTask`, whose pointer is then passed as an argument to the functions. Then there would be a single point where the UAVObject is set (at the end of `AttitudeTask`), instead of both in `setAttitudeINSGPS` and `setAttitudeComplementary`. This is IMO a cleaner and better way to do it, but requires more changes to the code.

If that's a suitable solution, happy to do it that way.

peabody124:
When it is any one mode (e.g. complementary or INS) there is only one set method used, right? There isn't actually intermodule communication with set/get or am I missing something?

kubark42:
You're right, there's currently no interfunction comms with the get/set. It would come if I were to put an `AttitudeActualGet()` on L393 in order to retrieve the AttitudeActual values and place them into a batch UAVObject.

As I see it, we have three choice, and I'm fine with any of them:
1. I close this PR and keep the changeset in my own branch.
2. I refactor this PR to make AttitudeActual be scoped in the main `Task` block, extend `setAttitudeComplementary()` and `setAttitudeINSGPS()` to take `AttitudeActual*` as an argument, and consolidate the two `AttitudeActualSet()` calls into one call at the end of the `while(1)`.
3. We merge this PR.

Lemme know which option you like best. I'm partial to #2.

peabody124:
2 is ok for me

kubark42:
@peabody124 Rebased to correspond to plan number 2.

PR #2168, opened 2016-01-29
Battery: Don't trigger voltage alarms when the thresholds are 0

kubark42:
This allows the user to turn off all battery alarms, without having to disable the battery module. This is useful in bench testing because it allows arming the system without having to have the battery plugged in.

Also slightly changes the battery threshold testing to match code style standards.

### DIFF
--- a/flight/Modules/Battery/battery.c
+++ b/flight/Modules/Battery/battery.c
@@ -155,13 +155,16 @@ static void batteryTask(void * parameters)
 
 			flightBatteryData.Voltage = scaled_voltage;
 
-			// generate alarms and warnings
-			if (flightBatteryData.Voltage < batterySettings.VoltageThresholds[FLIGHTBATTERYSETTINGS_VOLTAGETHRESHOLDS_ALARM])
+			// Set alarm if insufficient voltage, but only if the threshold is positive
+			if ((batterySettings.VoltageThresholds[FLIGHTBATTERYSETTINGS_VOLTAGETHRESHOLDS_ALARM]) > 0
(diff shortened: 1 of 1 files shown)


mlyle:
I don't understand.  This is vestigial / does nothing, it looks like.

tracernz:
Well, it now violates the coding standard..
https://www.kernel.org/doc/Documentation/CodingStyle

```
Also, note that this brace-placement also minimizes the number of empty
(or almost empty) lines, without any loss of readability.  Thus, as the
supply of new-lines on your screen is not a renewable resource (think
25-line terminal screens here), you have more empty lines to put
comments on.

Do not unnecessarily use braces where a single statement will do.

    if (condition)
        action();

and

    if (condition)
        do_this();
    else
        do_that();

This does not apply if only one branch of a conditional statement is a single
statement; in the latter case use braces in both branches:

    if (condition) {
        do_this();
        do_that();
    } else {
        otherwise();
    }
```

mlyle:
Yah, I just mean.. if the threshold is 0.. it will not be less than a flight battery voltage (which is clamped to nonnegative values just above)

kubark42:
In some internal code, I am relaxing the thresholding on the voltage so it can be negative. There's no reason why the voltage we're measuring (not the voltage supplied to the ADC) has to be strictly positive, even though it almost always is. This is another PR I'm happy to close if it's not desirable because it has limited immediate utility.

As for the standards, I remember sometime ago having a preference to err toward using `{}` on `if()` clauses, even though it's not kernel style.

peabody124:
Yeah we've never strictly forced the {} issue. This seems reasonable although probably won't benefit most users. Merging.

PR #2170, opened 2016-01-30
Sparky2: enable DAC as ADC port

peabody124:
This adds the DAC pin on Sparky2 as an ADC pin, which can be used
for things like RSSI monitoring. Also cleans up some comments.

Parts back ported from https://github.com/d-ronin/dRonin/pull/543


peabody124:
Closing this, it's actually not necessary given some stuff I'm doing with DAC output

PR #2171, opened 2016-01-30
Logging updates

peabody124:
Updates to the logging module to provide a default profile for logging. This
makes the defaults do something most users would find useful.

Note, that these defaults could also be pushed into the .xml files as general
defaults but would make it harder for people to change profiles. Perhaps,
later this could be added to the telemetry rate UI instead as profiles there.

This involved back porting some commits from https://github.com/d-ronin/dRonin/pull/180
that were not in #2031 with some mild adaptation.


kubark42:
I need to try this for a while. The current logging works really well for high-speed stuff. I simultaneously get every single sensor data sample from the IMU, mag, baro, _and_ GPS.

kubark42:
Does it make sense to have a "Custom" field? The UAVObjects already have the Logging field, and this would conflict with that. The system default log rate is already available in the UAVObject XML definition.

My preference would be to slightly extend the Telemetry Scheduler to handle this in addition to its other metadata tasks. Or perhaps a simple standalone widget which just chooses amongst a few profiles and saves those values to the UAVObject.

kubark42:
If we stick with the custom field, doesn't hurt to consider [#397](https://github.com/d-ronin/dRonin/pull/397) as well.

peabody124:
@kubark42 long run I prefer integrating to scheduler. Short term, I think default profile is convenient. I can add 397

kubark42:
Okay, let's do the short run, then.

kubark42:
Still haven't tried this yet, I'm sorry.

peabody124:
Rebased on next since there was a merge conflict

peabody124:
I'll merge this in a few days if there are no objections.

Issue #2155, opened 2016-01-27
Consider switching from file-scoped copyright headers


peabody124:
Maintaining file-scoped copyright headers is a fairly outdated practice for large projects. Version control provides the accurate attribution of individual changes to individuals. 

See this document for more information https://softwarefreedom.org/resources/2012/ManagingCopyrightInformation.html

It might be worth considering moving to a consolidate project wide copyright (including attributing other projects that provide code). Historically (3-4y ago) we went so far as to add individuals to files but that wasn't useful and moved to the project system we have now.

peabody124:
ping @lilvinz since you did a lot of work standardizing the copyright template in files, what do you think about this?

kubark42:
That seems reasonable. I've read in the past that header files generally\* can't be copyrighted, so reducing useless, toothless boilerplate would certainly be valuable! I'm gonna love the LOC reduction on that one. :D

[*] A header file with functional code in it could be, but a definition list of functions and `#define` constants can't be.

mlyle:
>  I think I've read that header files can't even be copyrighted

Files specifying only the calls involved in an API may or may not be copyrightable.  But headers often include a lot of other stuff, like documentation, outright code, macros, etc...

I have a couple of points of concern with this-- this would effectively lose all project-level attribution.  The "bits we've changed" that Tau has taken would no longer be apparent-- indeed who the individual copyright holder is for a block of code could be very difficult to untangle; was code that @tracernz committed from the time he was working on Tau or at dRonin?  The history is effectively stripped out.  Depending on many details of this, this may or may not be compliant.

I also think this will create a bunch of diff that will make it very difficult to move code between projects in the future; commits will not come cleanly that happen to touch the copyright block in dRonin or Tau (and a copyright block is still suggested to be present by that document), and it may trigger more complicated clerical actions in response to changes where top-level attribution information needs to be updated.

e.g. if you intend to import code from dRonin in the future, doing something like this may reduce your short-term pain but it is likely to be more problematic in general.

peabody124:
It maintains author level attribution with the granularity of what was done (i.e. commit history). It would _expand_ project level attribution to indicate that there is shared copyright with dronin. Also, file-scoped headers don't indicate what project someone was working with when they wrote a particular commit when there is simply a line that applies to a certain file. You would need something in the commit message itself to indicate that (or ick - comments in file like we used to have, ugh).

As for commit conflicts, those are really easy to fix when cherry-picking.

[...]

PR #2145, opened 2016-01-26
Sparky and Sparky2: dRonin backport 

kubark42:
Combines several Sparky and Sparky2 fixes into one PR.

In specific: 
- Updates artwork: https://github.com/d-ronin/dRonin/pull/250/commits
- Fixes EF size: https://github.com/d-ronin/dRonin/pull/251/commits
- Fixes serial telemetry: https://github.com/d-ronin/dRonin/pull/267/commits

PR #2172, opened 2016-01-31
Update copyrights

peabody124:
Include dRonin in copyrights in header where appropriate. May be outdated by
#2155 but best to do the right thing for now.

To be added: 
- [x] https://github.com/TauLabs/TauLabs/pull/2175
- [x] https://github.com/TauLabs/TauLabs/pull/2178
- [x] https://github.com/TauLabs/TauLabs/pull/2179
- [x] https://github.com/TauLabs/TauLabs/pull/2181

### DIFF
--- a/flight/targets/sparky/fw/pios_board.c
+++ b/flight/targets/sparky/fw/pios_board.c
@@ -6,7 +6,8 @@
  * @{
  *
  * @file       pios_board.c
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
+ * @author     dRonin, http://dronin.org Copyright (C) 2015
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
  * @brief      The board specific initialization routines
  * @see        The GNU Public License (GPL) Version 3
  *
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -14,7 +14,8 @@
  *
  * @file       attitude.c
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
- * @author     Tau Labs, http://taulabs.org Copyright (C) 2012-2014
+ * @author     dRonin, http://dronin.org Copyright (C) 2015
+ * @author     Tau Labs, http://taulabs.org Copyright (C) 2012-2015
  * @brief      Full attitude estimation algorithm
  *
  * @see        The GNU Public License (GPL) Version 3
--- a/flight/Modules/Stabilization/stabilization.c
+++ b/flight/Modules/Stabilization/stabilization.c
@@ -14,7 +14,8 @@
  *
  * @file       stabilization.c
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2010.
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
+ * @author     dRonin, http://dronin.org Copyright (C) 2015
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
  * @brief      Attitude stabilization.
  *
  * @see        The GNU Public License (GPL) Version 3
--- a/flight/targets/colibri/board-info/pios_board.h
+++ b/flight/targets/colibri/board-info/pios_board.h
@@ -6,7 +6,8 @@
  * @{
  *
  * @file       STM32F4xx_Colibri.c 
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
+ * @author     dRonin, http://dronin.org Copyright (C) 2015
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
  * @brief      Board specific defines for Colibri
  * @see        The GNU Public License (GPL) Version 3
  * 
--- a/flight/targets/discoveryf4/fw/pios_board.c
+++ b/flight/targets/discoveryf4/fw/pios_board.c
@@ -6,7 +6,8 @@
  * @{
  *
  * @file       pios_board.c
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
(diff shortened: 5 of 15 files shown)


mlyle:
Thank you.  Hopefully we can stay on a friendlier footing for now on.

tracernz:
Thanks @peabody124, much appreciated. Would be nice to have these discussions on IRC so we can be clear on each others intentions and reasoning in future, before things get out of hand. :+1:

kubark42:
@tracernz Seeing how emotional this was for dRonin, I don't think that would have helped. We did what we said we would, in an appropriate time frame. Why the drama, I'll never understand.

@peabody124 Thanks. When I get some bandwidth I'll make sure we caught them all and merge this.

mlyle:
@kubark42 --- I asked / raised this issue well before the first PR was open; we raised it in the initial PR; we asked before the first merge, and we brought it up calmly.  Then you guys tried to come up with a way to avoid attributing at all ;)

An appropriate timeframe is "at the same time as the merge".  The tree should not be left without the notices.

flight/targets/sparky/fw/pios_board.c
```diff
@@ -6,7 +6,8 @@
  * @{
  *
  * @file       pios_board.c
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2014
+ * @author     dRonin, http://dronin.org Copyright (C) 2015
```
kubark42, line 9: https://github.com/peabody124/TauLabs/commits/copyrights/flight/targets/sparky/fw/pios_board.c shows that this doesn't have any recent changes. Included by accident?

kubark42:
@mlyle Feel free to point to me where we tried to avoid attributing anything. The best analysis is when we didn't follow your demands hysterics ensued. Anyone is free to read GitHub discussions and judge for themselves.

@peabody124 I dunno about this PR. It really looks as if many of the changesets merged in are trivial in nature, and so don't warrant an attribution. Indeed, in addition to the spiritual violations you mentioned in another thread, it looks as if dRonin is claiming copyright of files which have had no qualifying changes. See https://github.com/d-ronin/dRonin/commits/698a1b406e6aa9cd8e91e593d0925048b7501076/flight/targets/sparky2/board-info/board_hw_defs.c, where the only dRonin change was of a non-creative nature. 

Frankly, I really don't care about dRonin falsely claiming authorship. It's not worth getting worked up over, and the git commit logs are all that really matters. So I'm proposing we set a good example and not get on dRonin's case about their attribution violations. I also propose to merge this PR without going through and rejecting authorship when only trivial changes were made. This seems to respect the open-source spirit of ensuring that attribution is spread as widely as possible, even though legally it's not required.

flight/targets/sparky/fw/pios_board.c
mlyle, line 9: https://github.com/TauLabs/TauLabs/pull/2145/files

mlyle:
>  looks as if dRonin is claiming copyright of files which have had no qualifying changes. 

Whether this is "qualifying" or not is the kind of thing that different US federal circuit courts have decided differently.  In any case, there's no harm in adding the notice-- but potential risks to anyone who decides to remove it. :)

peabody124:
@kubark42 I'd rather err on the side of caution and just do more rather than less. When we proposed going to a universal file we were basically going to give equal attribution at the top level. As you pointed out the git commit history is the only meaningful thing to see fraction of contribution. When there is a DR header in there and 99.9% of the LOC in the codebase are from TL and OP, well it is what it is :)

mlyle:
>  When there is a DR header in there and 99.9% of the LOC in the codebase are from TL and OP, well it is what it is :)

Yah, it may end up that way because I assume it's going to get more difficult to take our changesets with time.  But if you continue tracking that close, it won't; right now about 15% of flight blames to dRonin in our tree.

kubark42:
<Shrug> You're sloppy in your attribution. Consider https://github.com/d-ronin/dRonin/pull/569/files#diff-12bc03d72ddbc63727724364b733ca3a, where the only change was to rename "Throttle" to the almost synonym "Thrust". It's pretty clear that when you move files around, rename variables, add new hardware targets (which are 99% boilerplate copy pasted from original targets), and liberally sprinkle in authorship lines you'll quickly bulk up a figure. It's just hard to believe based on the PRs I've seen that in two months you've replaced 15% of the functionality built in over the past 6 years. 

These arguments thrown at us are so slippery. First it was absolute legality, then when that bubble was burst it was morality, and then now that that bubble is burst it's `git blame`. It was clear from the beginning that this discussion was unnecessary, and nothing has come out of it except a better understanding of why projects should use centralized copyright. (Shout out to @peabody124 for discovering that!)

As always, feel free to point out to us when you'd like an attribution to be added and we'll ensure it does, even when it's not legally appropriate. That just seems like the right thing to do.

mlyle:
>  You're sloppy in your attribution. 

If we want, we can add a Copyright dRonin to every file-- and perhaps we should, to avoid this type of discussion.  It might be a good idea.  The purpose of the copyright notice isn't just project-level attribution.  It's also to let people know that there (may) be a copyright interest held in that file by the project.

> As always, feel free to point out to us when you'd like an attribution to be added and we'll ensure it does, even when it's not legally appropriate. That just seems like the right thing to do.

As always, when you import from our project, ensure that you preserve the notices, as you are legally required (and now as we have very explicitly clarified.)  If you slip up a time or two, no big deal.  If you systematically ignore it (as you still are, by still not merging this PR), ... I will have to consider all remedies afforded to me by the license agreements and copyright law.

mlyle:
I mean, you say it's the right thing to do?  Go ahead, press merge, and include it directly in subsequent/new PRs.

kubark42:
The right thing to do is stick with the original plan. I'm unsure why you consciously refuse to hear this, but here it is again: once I'm done reviewing #2127/#2134 I'll merge this all in in one fell swoop. 

We're making steady progress. Hold tight.

mlyle:
In the meantime you're still distributing stuff without the copyright notices attached.

PR #2174, opened 2016-02-02
GCS: Uploader: Append board name to board ID

kubark42:
Adds the board name to the uploader. This gives normal users an easy-to-read board name.

New look: 
screen shot 2016-02-02 at 6 10 04 pm

Inspired by https://github.com/d-ronin/dRonin/pull/489.

P.S. Having two "ID" might be unnecessarily redundant.


kubark42:
@peabody124 Jenkins says this is okay, but I'd like your quick once-over to validate this different approach, since you only reviewed what @tracernz did.

peabody124:
I like the extension!

PR #2177, opened 2016-02-03
TLM: correct buffer size

peabody124:
Buffer was hardcoded as 128bytes but the comm layer was told
256. This opened the door for memory overruns.

### DIFF
--- a/flight/targets/pipxtreme/fw/pios_board.c
+++ b/flight/targets/pipxtreme/fw/pios_board.c
@@ -7,7 +7,7 @@
  *
  * @file       pios_board.c 
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2011.
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2015
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2012-2016
  * @brief      The board specific initialization routines
  * @see        The GNU Public License (GPL) Version 3
  * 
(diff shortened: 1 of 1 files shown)


peabody124:
Thanks @mlyle for pointing this out at https://github.com/d-ronin/dRonin/issues/600

kubark42:
Can't test this, but looks sane. Damned magic numbers!

What about at https://github.com/TauLabs/TauLabs/blob/8e89a6a9c5a9d8c630ec88d94e7828ec4c84da65/flight/targets/pipxtreme/fw/pios_board.c#L209 as well?

P.S. Not your responsibility, but if you're poking around there and feel like it, you could replace tabs by whitespaces in https://github.com/TauLabs/TauLabs/blob/8e89a6a9c5a9d8c630ec88d94e7828ec4c84da65/flight/targets/pipxtreme/fw/pios_board.c#L214 and https://github.com/TauLabs/TauLabs/blob/8e89a6a9c5a9d8c630ec88d94e7828ec4c84da65/flight/targets/pipxtreme/fw/pios_board.c#L246 so they're correctly aligned in all editors.

mlyle:
Just FYI, the whole rest of the code in this area is vestigial and nonfunctional.

e.g. there is nothing to separate the commands to the HM-10.  The HM-10 expects a delay of half a second between commands, which clearly ain't there.  (e.g. how does it know the end of the "+NAME" was given?  Instead, they're all buffered up in a pile and given to the RTOS which sends them with no spacing at the baud rate.  There's no way this ever worked.

(Tested the sequence on HM-10, it indeed does not init properly).

There's also misplaced/mismatched case labels and a couple other corruption bugs in this area.

peabody124:
@kubark42 I think the link you provided to line 209 is in next, because that is the line I fixed. I don't see any other constants hardcoded

peabody124:
Fixed white space per request

kubark42:
@peabody124, oh, yeah, whoops. No idea how I got to that particular commit by accident.

The other issues @mlyle raises might be correct, I haven't looked at that part of the code. However, this magic number fix is clearly sane and necessary so merging.

PR #2178, opened 2016-02-03
GCS: don't load EmptyGadgets when restoring config

kubark42:
"Previously, when we were instantiating splitter views, we'd briefly instantiate an EmptyGadget until we've gone and recursed and restored that configuration. Now, in that case, skip doing that as the appropriate gadget widget will be selected soon."

Tested, the newer version is indeed faster to load than the older version.

From https://github.com/d-ronin/dRonin/pull/311. Fixes https://github.com/d-ronin/dRonin/pull/297


kubark42:
Reviewed by @peabody124 in #2127, Jenkins says it's okay. Merging.

peabody124:
Checked and no copyright headers were updated in the upstream PR. Marking this in 2172 as complete.

PR #2180, opened 2016-02-03
Flight firmware: only look for ccache path once, use in submakes

kubark42:
Previously, building flight firmware on Windows caused repeated messages from the "which ccache" calls.

From https://github.com/d-ronin/dRonin/pull/196. Fixes https://github.com/d-ronin/dRonin/pull/195.


kubark42:
Jenkins says it's okay, reviewed in #2127. Merging.

PR #2184, opened 2016-02-08
Fix maps domain brittleness

kubark42:
The current Google API approach is very brittle to the crypto key and GCS will crash if the crypto key is wrong. This gets the base domain name from the organization name in a more robust manner, so `www.taulabs.org` returns the same results as `forums.taulabs.org` as `taulabs.org`, etc...

In the event that the key is just plain wrong, then the decrypt function will thrown an error, which is caught in the parent. The parent will send a debug message instead of GCS segfaulting.

It could be arguable, however, that we should do a `qFatal()` instead of `qDebug()`, as not having correct provider strings could lead to system instability elsewhere. Still, in limited testing on OSX 10.10 having the function return early did _not_ lead to crashes, or even any immediate decrease in functionality.

Fixes problem seen in https://github.com/TauLabs/TauLabs/pull/2106#issuecomment-163786632. Once this is in we can merge https://github.com/TauLabs/TauLabs/pull/2106.

### DIFF
--- a/ground/gcs/src/libs/tlmapcontrol/core/providerstrings.cpp
+++ b/ground/gcs/src/libs/tlmapcontrol/core/providerstrings.cpp
@@ -73,7 +73,21 @@ QString ProviderStrings::decrypt(QString str)
 ProviderStrings::ProviderStrings()
 {
     quint64 key = 0;
-    QByteArray array = QCryptographicHash::hash(QApplication::organizationDomain().toLatin1(), QCryptographicHash::Md4);
+
+    // Get the base URL and use that as the organization name registered with
+    // Google. Unfortunately, this is a hack since Qt does not provide a function
+    // to do this directly, so we have to trim it down from the full hostname.
+    QStringList domainName = QUrl(QApplication::organizationDomain()).host().split(".");
+    QString organizationBaseURL;
+    if (domainName.length() >=2) {
+        int len = domainName.length();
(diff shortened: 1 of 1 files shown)


ground/gcs/src/libs/tlmapcontrol/core/providerstrings.cpp
```diff
@@ -67,13 +68,31 @@ QString ProviderStrings::decrypt(QString str)
         ++pos;
     }
 
-    return QString::fromUtf8(qUncompress(ba));
+    QString decryptedString = QString::fromUtf8(qUncompress(ba));
+    if (decryptedString.isEmpty()) {
+        throw std::domain_error(QString("Decrypted to an empty vector!. Input vector: ").append(inputString).toStdString());
```
peabody124, line 73: What happens when this is thrown? What catches it and is it shown to the user somehow?

peabody124, line 73: Nevermind. See it later ;)

peabody124:
Code reviews well. About to test.

ground/gcs/src/libs/tlmapcontrol/core/providerstrings.cpp
kubark42, line 73: Is there a standard way to make this clear in comments? It seems unfortunate that you had to hunt through the code to find out what the catch was going to be. It sort of is like a `goto` in the sense that it can make it hard to follow the program's progression.

PR #2186, opened 2016-02-12
TlMapWidget: Initialize windspeedTxt and check for existence

kubark42:
Also fixed missing `NULL` check and made `if()` clauses more robust by explicitly placing braces.

Inspired by dRonin. Got merged (by accident?) along with https://github.com/d-ronin/dRonin/pull/523, so pulling it out separately here.

### DIFF
--- a/ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.cpp
+++ b/ground/gcs/src/libs/tlmapcontrol/mapwidget/tlmapwidget.cpp
@@ -37,7 +37,8 @@ namespace mapcontrol
     TLMapWidget::TLMapWidget(QWidget *parent, Configuration *config) : QGraphicsView(parent),
         configuration(config),UAV(0),GPS(0),Home(0),followmouse(true),
         compassRose(0),windCompass(0),showuav(false),showhome(false),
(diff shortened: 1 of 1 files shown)


mlyle:
No, it's required for #523

peabody124:
@kubark42 copyright notice

kubark42:
@peabody124 Are you asking to update the TL line? I can do that, but from what I understand the only date that really has any legal weight is the year of initial creation.

mlyle:
> but from what I understand the only date that really has any legal weight is the year of initial creation.

This is not true.  Every bit of new work that has any creative element has copyright invest in it and has its own protection and term independent from the work it is built atop.  Hence why we have new revised editions (even typesetting, etc) that are still in copyright and protected even if the original works have lapsed into the public domain.

peabody124:
@kubark42 just add DR. I don't think mlyle is remotely right legally, but honestly it isn't worth dealing their DMCA happy attitude. If you want to just add DR to all the files too and avoid the hassle that is fine.

mlyle:
He doesn't need to add dR because he reimplemented it and presumably didn't crib verbatim.  I thought you were saying bumping the tau copyright notice.

PR #2189, opened 2016-02-14
PIOS CAN: receive fix

peabody124:
This fixes a bug I found in the implementation of PIOS_CAN which prevented
receiving data on Sparky2. It also disables the currently unused PIOS_COM
layer on Sparky2 that is wrapped on PIOS_CAN.

### DIFF
--- a/flight/PiOS/STM32F30x/pios_can.c
+++ b/flight/PiOS/STM32F30x/pios_can.c
@@ -138,7 +138,8 @@ int32_t PIOS_CAN_Init(uintptr_t *can_id, const struct pios_can_cfg *cfg)
 
 	/* CAN filter init */
 	CAN_FilterInitTypeDef CAN_FilterInitStructure;
-	CAN_FilterInitStructure.CAN_FilterNumber = 0;
+	// Banks 0..13 are assigned to CAN1 and 14..28 to CAN2
+	CAN_FilterInitStructure.CAN_FilterNumber = (can_dev->cfg->regs == CAN1) ? 0 : 14;
 	CAN_FilterInitStructure.CAN_FilterMode = CAN_FilterMode_IdMask;
 	CAN_FilterInitStructure.CAN_FilterScale = CAN_FilterScale_32bit;
 	CAN_FilterInitStructure.CAN_FilterIdHigh = 0x0000;
--- a/flight/PiOS/STM32F30x/pios_can.c
+++ b/flight/PiOS/STM32F30x/pios_can.c
@@ -154,6 +154,11 @@ int32_t PIOS_CAN_Init(uintptr_t *can_id, const struct pios_can_cfg *cfg)
  	NVIC_Init((NVIC_InitTypeDef*) &can_dev->cfg->rx_irq.init);
  	NVIC_Init((NVIC_InitTypeDef*) &can_dev->cfg->tx_irq.init);
 
+	// Always enable receiving, regardless of the RxStart method as
+	// the PIOS_CAN driver does not require a PIOS_COM layer for the
+	// queue messages
+	CAN_ITConfig(can_dev->cfg->regs, CAN_IT_FMP1, ENABLE);
+
 	return(0);
 
 out_fail:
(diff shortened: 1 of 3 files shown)


flight/PiOS/STM32F30x/pios_can.c
```diff
@@ -138,7 +138,8 @@ int32_t PIOS_CAN_Init(uintptr_t *can_id, const struct pios_can_cfg *cfg)
 
 	/* CAN filter init */
 	CAN_FilterInitTypeDef CAN_FilterInitStructure;
-	CAN_FilterInitStructure.CAN_FilterNumber = 0;
+	// Banks 0..13 are assigned to CAN1 and 14..28 to CAN2
+	CAN_FilterInitStructure.CAN_FilterNumber = (can_dev->cfg->regs == CAN1) ? 0 : 14;
```
kubark42, line 142: I'm thinking this might be better as a `#define` so that the processor-specific nature of the CANbus implementation is easier to perceive.

flight/PiOS/STM32F4xx/pios_can.c
```diff
@@ -143,23 +144,36 @@ int32_t PIOS_CAN_Init(uintptr_t *can_id, const struct pios_can_cfg *cfg)
 	CAN_DeInit(can_dev->cfg->regs);
 	CAN_Init(can_dev->cfg->regs, (CAN_InitTypeDef *)&can_dev->cfg->init);
 
+	// Based on the selected RX IRQ we use either FIFO0 or FIFO1
+	if (can_dev->cfg->rx_irq.init.NVIC_IRQChannel == CAN1_RX1_IRQn ||
+		can_dev->cfg->rx_irq.init.NVIC_IRQChannel == CAN2_RX1_IRQn)
```
kubark42: Odd whitespace alignment here. At first glance, I thought there was an `if()` scoping failure. 

FWIW, this is why at work we violate the kernel style to always use braces (`{}`). We feel the odds of a bug sneaking in are too high and so for us it's not worth the tiny change in vertical spacing.

flight/PiOS/STM32F30x/pios_can.c
peabody124, line 142: I can if you really want but isn't the comment above sufficient?

flight/PiOS/STM32F4xx/pios_can.c
```diff
@@ -143,23 +144,36 @@ int32_t PIOS_CAN_Init(uintptr_t *can_id, const struct pios_can_cfg *cfg)
...
+		can_dev->rx_fifo = 1;
+	else
+		can_dev->rx_fifo = 0;
+
 	/* CAN filter init */
 	CAN_FilterInitTypeDef CAN_FilterInitStructure;
-	CAN_FilterInitStructure.CAN_FilterNumber = 0;
+	// Banks 0..13 are assigned to CAN1 and 14..28 to CAN2
+	CAN_FilterInitStructure.CAN_FilterNumber = (can_dev->cfg->regs == CAN1) ? 0 : 14;
```
kubark42, line 157: Same question about `#define`.

flight/PiOS/STM32F30x/pios_can.c
kubark42, line 142: I was thinking more about someone down the road who ports this driver to a different chip. Since this code is for the very specific implementation on STM32F3 it's not a blocker, though.

kubark42:
LGTM although I hope you can take a look at the alignment issue. It would be fine to add braces, or otherwise make it so the `if()` condition and the following line don't line up perfectly.

peabody124:
@kubark42 replaced the indentation with spaces. I'll avoid the {} for consistency with how I generally write code but totally take your point. If you are ok with it, I'll leave it with comments regarding the inbox since you could use anything in the 0...13 range for FIFO0 and I think that is made clear.

kubark42:
Fine by me. Thanks for the whitespace alignment fixes.

PR #2191, opened 2016-02-17
ActuatorSettings: default motors to off

peabody124:
Because there are so many variants of outputs now (e.g. fast ESCs, slow ESCs, OneShot, regular PWM)
we should just default to not sending out any pulses to avoid anything unsafe. I don't think there is any
actuator that will respond to a steady zero in an unsafe manner.

The wizard or importing UAV files will completely configure this and is how most users do it anyway.

### DIFF
--- a/shared/uavobjectdefinition/actuatorsettings.xml
+++ b/shared/uavobjectdefinition/actuatorsettings.xml
@@ -3,9 +3,9 @@
         <description>Settings for the @ref ActuatorModule that controls the channel assignments for the mixer based on AircraftType</description>
         <field name="TimerUpdateFreq" units="Hz" type="uint16" elements="6" defaultvalue="50"/>
         <field name="TimerPwmResolution" units="" type="enum" elements="6" options="1MHz,12MHz" defaultvalue="1MHz"/>
-        <field name="ChannelMax" units="us" type="uint16" elements="10" defaultvalue="2000"/>
-        <field name="ChannelNeutral" units="us" type="uint16" elements="10" defaultvalue="1000"/>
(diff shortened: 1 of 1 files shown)


shared/uavobjectdefinition/actuatorsettings.xml
```diff
@@ -2,10 +2,10 @@
     <object name="ActuatorSettings" singleinstance="true" settings="true">
         <description>Settings for the @ref ActuatorModule that controls the channel assignments for the mixer based on AircraftType</description>
         <field name="TimerUpdateFreq" units="Hz" type="uint16" elements="6" defaultvalue="50"/>
-        <field name="TimerPwmResolution" units="" type="enum" elements="6" options="1MHz,12MHz" defaultvalue="1MHz"/>
-        <field name="ChannelMax" units="us" type="uint16" elements="10" defaultvalue="2000"/>
-        <field name="ChannelNeutral" units="us" type="uint16" elements="10" defaultvalue="1000"/>
-        <field name="ChannelMin" units="us" type="uint16" elements="10" defaultvalue="1000"/>
+        <field name="TimerPwmResolution" units="" type="enum" elements="7" options="1MHz,12MHz" defaultvalue="1MHz"/>
```
kubark42: Is changing this to `7` intentional? It's not reflected in the git commit message or PR.

peabody124: wow no. not sure how that happened, thanks. will fix.

kubark42:
Looks good to me. I can only think of a few oddball systems where a 0 will be interpreted in an unsafe manner, and those will also arguably interpret the default 50Hz PWM unsafely as well.

PR #2193, opened 2016-02-19
VTOLPathFollower: Fix incorrect negative (-) sign

kubark42:
The assumption was made that a forward accel would have a negative value. This is confusing and unnecessary. This commit places the negative sign inside the calculation of the pitch angle, which indeed must be negative for forward acceleration.

In addition, aligned some of the symmetric code so that the differences stand out more clearly.

### DIFF
--- a/flight/Modules/VtolPathFollower/vtol_follower_control.c
+++ b/flight/Modules/VtolPathFollower/vtol_follower_control.c
@@ -462,20 +462,20 @@ int32_t vtol_follower_control_attitude(float dT, const float *att_adj)
 	// Project the north and east acceleration signals into body frame
 	float yaw;
 	AttitudeActualYawGet(&yaw);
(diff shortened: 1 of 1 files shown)


kubark42:
@lilvinz Jenkins needs a kick?

peabody124:
I don't see a problem with this because it's just shuffling math around. Is there a reason it was incorrect before? Was it actually inconsistent in our normal units?

peabody124:
BTW as long as this is tested I'm ok for it to merge. The math checks out and there is only one use of that variable.

kubark42:
It's only incorrect because the variable name is "forward", but what the variable expresses is backwards. I ran across this problem as I was extending some functionality, and couldn't figure out why everything was going the wrong direction. :D

This change is tested, and as you point out it's very easy to see the scope of the change.

Thanks!

kubark42:
PS, please pardon the above smartphone typos.

PR #2194, opened 2016-02-19
Logging: Change FileSectorNum to uint32

kubark42:
For too large files sizes (in the case of the MX25L128 more than 8MB), the logging download would fail because FileSectorNum was limited to 16 bits (65535 sectors). This increases the maximum number of sectors, although FileSectorNum is currently copied into an int32, so at most this gives 2^31 sectors. Still, this is far greater than any likely memory requirement in the short term.

Tested and works all the way up to 16MB log files.


peabody124:
LGTM - feel free to merge once Jenkins gives the green flag

kubark42:
Jenkins still seems to be down. Maybe @lilvinz can give it a kick?

PR #2195, opened 2016-02-19
Attitude: Move minimum PDOP and minimum number of satellites to UAVObject

kubark42:
This should be a configurable settings because each GPS and vehicle will have different minimum RNAV requirements. While we don't expect users to set this themselves, it should still be accessible for development and tuning purposes.

Tested and works very nicely for not having to sit around waiting for enough satellites in order to take off, gain some altitude, and get some more satellites. :D

### DIFF
--- a/flight/Modules/Attitude/attitude.c
+++ b/flight/Modules/Attitude/attitude.c
@@ -820,9 +820,9 @@ static int32_t setNavigationRaw()
 		GPSPositionData gpsPosition;
 		GPSPositionGet(&gpsPosition);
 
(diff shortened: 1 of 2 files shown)


peabody124:
It looks good to me. Has it been tested?

kubark42:
Yup. Helps tons not waiting around for a GPS lock in the Boston winter. GPS signal markedly degrades within 2m of the ground (due to ground plane reflection), so it's really helpful to get off the ground and stop freezing our fingers!

PR #2197, opened 2016-02-19
Matlab parser enhancements

kubark42:
1. The parser would fail when the UAVO had an optional timestamp.
2. The parser would fail when parsing logfiles created on the onboard flash. This is because the logfile format changed some time ago.
3. Fixes out-of-range access, as inspired by https://github.com/d-ronin/dRonin/pull/651.

NOTE: Currently, the logic to choose between logfile formats is placed at the top of the Matlab file. This can lead to a race condition if both `overo` and `onboardLogger` are set to `true`. This will be resolved in a future commit once the flash logfile identifies itself.

P.S. I realize this isn't a particularly atomic PR, but in theory it should "just work" for parsing Matlab and so that's the best litmus test for this kind of script.

### DIFF
--- a/ground/gcs/src/plugins/uavobjects/uavobjecttemplate.m
+++ b/ground/gcs/src/plugins/uavobjects/uavobjecttemplate.m
@@ -30,6 +30,7 @@ str4=[];
 str5=[];
 multipleInstanceLookup = zeros(0,2);
 overo = false;
+onboardLogger = true;
 
 fprintf('\n\n***Tau Labs log parser***\n\n');
 global crc_table;
@@ -99,8 +100,9 @@ buffer=fread(fid,Inf,'uchar=>uchar');
 
 bufferIdx=1;
 
-correctMsgByte=hex2dec('20');
 correctSyncByte=hex2dec('3C');
+correctMsgByte=hex2dec('20');
+timestampedMsgByte=hex2dec('80');
 unknownObjIDList=zeros(1,2);
 
 % Parse log file, entry by entry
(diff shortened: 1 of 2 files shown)


ground/gcs/src/plugins/uavobjects/uavobjecttemplate.m
```diff
@@ -30,6 +30,7 @@
 str5=[];
 multipleInstanceLookup = zeros(0,2);
 overo = false;
+onboardLogger = true;
```
peabody124, line 33: now won't this break it by default for GCS logs?

ground/gcs/src/plugins/uavobjects/uavobjecttemplate.m
```diff
@@ -171,38 +155,86 @@
...
 		else
 			timestamp = double(typecast(buffer(bufferIdx+12:bufferIdx+14-1),'uint16'));
 		end
-        
+		
 		% Advance buffer past header to where data is.  In the case of a
 		% multiple instance object this will be where the timstamp is but
 		% the parsing code will advance by two more.
 		bufferIdx = bufferIdx + 10;
+	elseif onboardLogger
```
peabody124, line 175: so I'm puzzled what is going on here. how does onboard logging compare to over? it looks like it has the integrated time stamp but you aren't subtracting 128 from msgType below. Does the onboard logging not set the msgType appropriately?

ground/gcs/src/plugins/uavobjects/uavobjecttemplate.m
kubark42, line 33: Whoops, totally! I knew there was a reason why I hadn't pushed this PR several months ago. I need to consider how the Matlab logger can more gracefully, or even automatically, select log file types.

ground/gcs/src/plugins/uavobjects/uavobjecttemplate.m
kubark42, line 175: Er... That might be it. I don't recall, but they're not the same format. If you post an overo logfile I can make sure this doesn't break it.

peabody124:
Here is a log file. Sorry about the delay 

https://drive.google.com/file/d/0ByPItuei3Xt6bDlFck5Cc2Jjbzg/view?usp=sharing

Some UAVOs have likely changed since then, but generally I think the format should be the same as built in logging.

kubark42:
Thanks. I'll get on it.

PR #2203, opened 2016-02-24
RFM22b COM assignment

### LINKED ISSUE #2201: RFM22b TauLink serial bridge
It would be very useful to have TauLink support a transparent
serial bridge (e.g. using VCP on the ground side) and a PIOS_COM
interface on the FC side that can be assigned like a normal
serial port.

This would be useful for things like a ROS serial link to be able
to talk through this link e.g. #1408

peabody124:
This makes the PIOS_COM assignment for the RFM22B module on
flight controllers programmable. This allows sending debug information
or other through it to a serial port. You can still use the PPM channels
for control.

Fixes #2201

### DIFF
--- a/flight/Modules/RadioComBridge/RadioComBridge.c
+++ b/flight/Modules/RadioComBridge/RadioComBridge.c
@@ -127,9 +127,6 @@ static int32_t RadioComBridgeStart(void)
 		// Check if this is the coordinator modem
 		data->isCoordinator = PIOS_RFM22B_IsCoordinator(PIOS_COM_RFM22B);
 
-		// Parse UAVTalk out of the link
-		data->parseUAVTalk = true;
-
 		// Configure our UAVObjects for updates.
 		UAVObjConnectQueue(UAVObjGetByID(RFM22BSTATUS_OBJID), data->uavtalkEventQueue,
 				   EV_UPDATED | EV_UPDATED_MANUAL | EV_UPDATE_REQ);
--- a/flight/Modules/RadioComBridge/RadioComBridge.c
+++ b/flight/Modules/RadioComBridge/RadioComBridge.c
@@ -6,8 +6,8 @@
  * @{ 
  *
  * @file       RadioComBridge.c
+ * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2016
  * @author     The OpenPilot Team, http://www.openpilot.org Copyright (C) 2012.
- * @author     Tau Labs, http://taulabs.org, Copyright (C) 2013-2014
  * @brief      Bridges from RFM22b comm channel to another PIOS_COM channel
  *             has the ability to hook and process UAVO packets for the radio
  *             board (e.g. TauLink)
--- a/flight/Modules/RadioComBridge/RadioComBridge.c
+++ b/flight/Modules/RadioComBridge/RadioComBridge.c
@@ -152,9 +152,8 @@ static int32_t RadioComBridgeStart(void)
 					   EV_UPDATED | EV_UPDATED_MANUAL | EV_UPDATE_REQ);
 		}
 
-		if (data->isCoordinator) {
-			registerObject(RadioComBridgeStatsHandle());
-		}
+		registerObject(RadioComBridgeStatsHandle());
+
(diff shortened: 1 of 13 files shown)


peabody124:
Marked as WIP. Needs a bit more work.

peabody124:
Ok. No longer WIP, this is tested and working. The only caveat is right now I hard assign the
`parseUAVTalk` flag to false when using passthrough mode. I need to get safe boot working
on HWTauLink and then this can be in the settings.

Tested this for ROS communications and it works nicely.

peabody124:
BTW my plan is to have a flag in the HwTauLink settings for parsing UAVTalk. If set to
FALSE you will need to reboot in safe mode to override it but the modem will start to
act in a dumb mode and simply pass data (and not respond to GCS queries directly).

kubark42:
> BTW my plan is to have a flag in the HwTauLink settings for parsing UAVTalk. If set to
> FALSE you will need to reboot in safe mode to override it but the modem will start to
> act in a dumb mode and simply pass data (and not respond to GCS queries directly).

How about enabling some kind of AT mode `+++` for this? I'm guessing you've tested the safe-mode operation with this, but I've found it to be kind of hit-or-miss the few times I've had to use it.

peabody124:
@kubark42 at the point I would do that, I could make the uavtalk parser continue to listen for updates to hwtaulink and object persistence so as to keep one API. That might be the best from a user friendly perspective. As long as I don't periodically send rfm22bstatus, the stream should always behave nicely.

That way safe mode functions as a fallback which sounds better now I think about it.

So then, I guess the flash is more about whether the modem injects it's UAVO unrequested into the stream or not

flight/Modules/RadioComBridge/RadioComBridge.c
```diff
@@ -385,37 +315,17 @@ static void radioRxTask( __attribute__ ((unused))
 		PIOS_WDG_UpdateFlag(PIOS_WDG_RADIORX);
 #endif
 		if (PIOS_COM_RFM22B) {
-			uint8_t serial_data[1];
+			uint8_t serial_data[64];
```
kubark42, line 319: Can the magic number be moved up to the top? Or if that's not appropriate because of the scope, refactored out of the `serial_data` declaration?

flight/PiOS/inc/pios_wdg.h
```diff
@@ -40,7 +40,6 @@
 #define PIOS_WDG_SENSORS         0x0010
 #define PIOS_WDG_AUTOTUNE        0x0020
 #define PIOS_WDG_PPMINPUT        0x0040
-#define PIOS_WDG_SERIALRX        0x0080
```
kubark42, line 43: Unfortunate that this makes a hole in the bit field.

flight/Modules/RadioComBridge/RadioComBridge.c
```diff
@@ -136,31 +127,23 @@ static int32_t RadioComBridgeStart(void)
 		// Check if this is the coordinator modem
 		data->isCoordinator = PIOS_RFM22B_IsCoordinator(PIOS_COM_RFM22B);
```
kubark42, line 128: Since you're eliminating so many uses of `data`, could you refactor it's name to `radioComBridgeData`? I know it's a little wordy, but it just makes it so much easier to find in a project because there are fewer collisions with all the strings containing just `data`.

kubark42:
> The only caveat is right now I hard assign the `parseUAVTalk` flag to false when using passthrough mode.  I need to get safe boot working on HWTauLink and then this can be in the settings.

~~It looks like `parseUAVTalk` is hardcoded to `true`. Was that a typo or am I looking in the wrong spot?~~

I see, this hasn't been implemented yet. You planned that for a subsequent PR, right?

peabody124:
It can even go in this PR since without it, this doesn't fully provide the desired functionality (it pollutes stream). I was sort of interested in any feedback about how to handle it, which I think talking it through helped clarify.

kubark42:
@peabody124 is this ready for merge?

peabody124:
No sorry, I still need to make it easier to flip back into normal mode. Will do that this weekend I hope.

PR #2204, opened 2016-02-26
Importer upgrades

kubark42:
[Add select all/none buttons](https://github.com/d-ronin/dRonin/pull/655)
[Fix the automatic application of settinsg on import](https://github.com/d-ronin/dRonin/pull/669)

Mad kudos to @jpbarraca for fixing the apply-on-import issue, that always was a weakness of the importer.

The next step in improving the uploader would be to show the actual XML so that settings can be ported over even if the correspondence is too broken, e.g. TrimAngleSettings vs. SubTrimSettings.

### DIFF
--- a/ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
+++ b/ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
@@ -65,7 +65,6 @@ ImportSummaryDialog::ImportSummaryDialog( QWidget *parent) :
 ImportSummaryDialog::~ImportSummaryDialog()
 {
     delete ui;
-    delete importedObjects;
 }
 
 /*
--- a/ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
+++ b/ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
@@ -25,6 +25,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
  */
+#include <QSignalMapper>
 #include "importsummary.h"
 
 ImportSummaryDialog::ImportSummaryDialog( QWidget *parent) :
@@ -45,6 +46,16 @@ ImportSummaryDialog::ImportSummaryDialog( QWidget *parent) :
 
    connect( ui->closeButton, SIGNAL(clicked()), this, SLOT(close()));
    connect(ui->saveToFlash, SIGNAL(clicked()), this, SLOT(doTheSaving()));
+
+   // Connect the Select All/None buttons
+   QSignalMapper* signalMapper = new QSignalMapper (this) ;
+
+   connect(ui->btnSelectAll, SIGNAL(clicked()), signalMapper, SLOT(map()));
+   connect(ui->btnSelectNone, SIGNAL(clicked()), signalMapper, SLOT(map()));
+   signalMapper->setMapping(ui->btnSelectAll, 1);
+   signalMapper->setMapping(ui->btnSelectNone, 0);
+
+   connect (signalMapper, SIGNAL(mapped(int)), this, SLOT(setCheckedState(int)));
 
    // Connect the help button
    connect(ui->helpButton, SIGNAL(clicked()), this, SLOT(openHelp()));
@@ -91,6 +102,18 @@ void ImportSummaryDialog::addLine(QString uavObjectName, QString text, bool stat
 
    this->repaint();
    this->showEvent(NULL);
+}
+
+/*
+  Sets or unsets every UAVObjet in the list
+  */
+void ImportSummaryDialog::setCheckedState(int state)
+{
+    for(int i=0; i < ui->importSummaryList->rowCount(); i++) {
+        QCheckBox *box = dynamic_cast<QCheckBox*>(ui->importSummaryList->cellWidget(i,0));
+        if(box->isEnabled())
+            box->setChecked((state == 1 ? true : false));
+    }
 }
 
 /*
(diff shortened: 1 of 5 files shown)


ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
```diff
@@ -2,6 +2,7 @@
  ******************************************************************************
  *
  * @file       importsummary.cpp
+ * @author     dRonin, http://dRonin.org/, Copyright (C) 2015
```
peabody124, line 6: can you move it after TL, also since you made some changes might as well
extend TL to 2016

ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
```diff
@@ -61,7 +107,7 @@ ImportSummaryDialog::~ImportSummaryDialog()
   */
 void ImportSummaryDialog::openHelp()
 {
-    QDesktopServices::openUrl( QUrl("https://github.com/TauLabs/TauLabs/wiki/OnlineHelp:-UAV-Settings-import-export", QUrl::StrictMode) );
+    QDesktopServices::openUrl( QUrl("https://github.com/d-ronin/dRonin/wiki/OnlineHelp:-UAV-Settings-import-export", QUrl::StrictMode) );
```
peabody124: This should be updated

ground/gcs/src/plugins/uavsettingsimportexport/uavsettingsimportexportfactory.h
```diff
@@ -36,6 +39,7 @@ class UAVSETTINGSIMPORTEXPORT_EXPORT UAVSettingsImportExportFactory : public QOb
 public:
     UAVSettingsImportExportFactory(QObject *parent = 0);
     ~UAVSettingsImportExportFactory();
+    static bool updateObject(UAVObject *obj, QDomNode * node);
```
peabody124: I don't see this new method being defined anywhere. Can you check if it works with this line (and the one
above on 33) deleted?

ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
kubark42, line 6: I didn't make any changes. git cherry-pick flubbed because the original commit history was composed of changes in commits and changes in merges. The end result was having to merge all changes into one commit. Dunno why even though I amended the commit to put Joao as the author it wound up ultimately reverting back to me.

ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
```diff
@@ -25,11 +26,27 @@
...
+// for XML object
+#include <QDomDocument>
+#include <QXmlQuery>
+
+// for Parameterized slots
+#include <QSignalMapper>
 #include "importsummary.h"
 
+#define UAV_SETIMPEXP_APPLY 1
+#define UAV_SETIMPEXP_SAVE 2
```
kubark42: Unused variables. Remove.

ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
```diff
@@ -54,6 +91,15 @@ ImportSummaryDialog::ImportSummaryDialog( QWidget *parent) :
 ImportSummaryDialog::~ImportSummaryDialog()
 {
     delete ui;
+    delete importedObjects;
```
kubark42: This doesn't look like it needs deleting. In addition, there are several `new` commands in this PR, and it looks like some could in general be removed/refactored.

kubark42: Okay, tracked it down a bit further and it looks brittle. `importedObjects` is assigned locally, but it's pointer is `importedObjectManager`, which is created in `uavsettingsimportexportfactory.cpp`. So the `delete` is happening outside the originating class, and the originating class has lost track. This doesn't feel right. Gonna investigate some more.

ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
```diff
@@ -37,14 +54,34 @@ ImportSummaryDialog::ImportSummaryDialog( QWidget *parent) :
...
+   connect(ui->btnSaveToFlash, SIGNAL(clicked()), signalMapper, SLOT(map()));
+   connect(ui->btnApply, SIGNAL(clicked()), signalMapper, SLOT(map()));
+
+   signalMapper->setMapping(ui->btnSaveToFlash, UAVSettingsAction::save);
+   signalMapper->setMapping(ui->btnApply, UAVSettingsAction::apply);
+
+   connect(signalMapper, SIGNAL(mapped(int)), this, SLOT(doTheApplySaving(int)));
+
+   // Connect the Select All/None buttons
+   signalMapper = new QSignalMapper (this);
```
kubark42: This also looks like lost memory. I don't see why not to declare these in the class.

kubark42: I've refactored all of this to be a normal signal/slot. The SignalMapping was an interesting approach to the problem of linking multiple buttons to one `SLOT`, but I prefer using `sender()` to do this.

ground/gcs/src/plugins/uavsettingsimportexport/importsummary.cpp
```diff
@@ -25,11 +26,19 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
  */
+
+// for XML object
```
kubark42: These includes also aren't used.

kubark42:
@peabody124, refactored the PR. Give it another look now.

@jpbarraca, you might give a look at the changes I added. They're another way to accomplish the same thing, but I found the structure a bit cleaner this way. What do you think?

jpbarraca:
@kubark42 You are correct regarding the cleaner approach. This PR evolved from a previous where it was possible to do Apply or Save independently.

kubark42:
Yeah, I think it's valuable to be able to "try-before-you-buy" so prefer to have both save and apply buttons. The settings importer is pretty opaque, but your upgrade is super valuable.

PR #2206, opened 2016-03-02
MS5xxx: Refactor into common driver for all members of the sensor family

kubark42:
Measurement Specialties has several sensors that are built around the same circuitry with different trim values. The upshot is that the driver is almost identical for I2C and SPI, and for all sensors in this family. At the least, this includes the:
- MS5611
- MS5637
- MS5803-01
- MS5803-02
- MS5803-05
- MS5803-07
- MS5803-14
- MS5803-30

The only differences between the above chips are the scalar multipliers and divisors for calculating the second-order compensated values.


kubark42:
Driver is now tested and works. However, as a result of a phone conversation with a Measurement Specialties application engineer, I'm going to be refactoring this into the existing MS5611 driver. He confirmed that this is identical circuitry with the only differentiator being the package.

Each sensor range has slight different math, but those are the only differences in the code.

kubark42:
@peabody124 I just pushed the refactoring. I'd appreciate help in testing any of the below.
- [ ] AQ32
- [x] Colibri
- [ ] Naze32
- [x] Quanton
- [ ] RevoMini
- [x] Sparky
- [x] Sparky 2

kubark42:
I forgot to say that this adds support for the MS5637, MS5803-01, and MS5803-02, i.e. not all the chips in the MS5xxx family.

peabody124:
I should be able to test a number of those targets by this weekend

kubark42:
Thanks. I've tested the MS5637 I2C version and MS5803-02 SPI version as well.

flight/PiOS/Common/pios_ms5xxx.c
```diff
@@ -0,0 +1,698 @@
...
+static int32_t PIOS_MS5XXX_WriteCommand(uint8_t command)
+{
+	if (PIOS_MS5XXX_Validate(dev) != 0) {
+		return -1;
+	}
+
+	int32_t rc = 0;
+
+	switch (dev->pios_bus_type) {
+	case PIOS_SPI_BUS:
```
mlyle, line 560: The issue with this structure is that it ends up including both the SPI and the I2C code irrespective of which is being used on the target.

flight/PiOS/Common/pios_ms5xxx.c
```diff
@@ -0,0 +1,698 @@
...
+	return PIOS_Semaphore_Give(dev->busy) == true ? 0 : 1;
+}
+
+/**
+ * @brief Claim the SPI bus for the baro communications and select this chip
+ * @return 0 if successful, -1 for invalid device, -2 if unable to claim bus
+ */
+static int32_t PIOS_MS5XXX_ClaimSPIBus(void)
+{
+#if defined(PIOS_INCLUDE_SPI)
```
mlyle: This particular group is wrapped in PIOS_INCLUDE_SPI but other stuff below uses SPI unconditionally.

flight/PiOS/Common/pios_ms5xxx.c
kubark42, line 560: It will increase code complexity to conditionally include I2C and SPI, whereas the existing approach seems to work fine without the conditional includes. Is there an advantage to what you're suggesting beyond minimizing impact on F1 processors?

mlyle, line 560: Sparky1 is nearly out of space, too, and is a really nice target.

flight/PiOS/Common/pios_ms5xxx.c
```diff
@@ -0,0 +1,700 @@
...
+	/* Calibration parameters */
+	for (int i = 0; i < NELEMENTS(dev->calibration); i++) {
+		PIOS_MS5XXX_Read(MS5XXX_CALIB_ADDR + i * 2, data, 2);
+		dev->calibration[i] = (data[0] << 8) | data[1];
+	}
+
+	PIOS_SENSORS_Register(PIOS_SENSOR_BARO, dev->queue);
+
+	dev->task = PIOS_Thread_Create(
+			PIOS_MS5XXX_Task, "pios_ms5xxx", MS5XXX_TASK_STACK_BYTES, NULL, MS5XXX_TASK_PRIORITY);
```
kubark42, line 171: Should this task be added to the task manager UAVObject? I don't really see any downsides to this, but it wasn't done before so I didn't add it.

flight/PiOS/Common/pios_ms5xxx.c
```diff
@@ -0,0 +1,700 @@
...
+#include "pios_ms5xxx_priv.h"
+#include "pios_semaphore.h"
+#include "pios_thread.h"
+#include "pios_queue.h"
+
+#include "physical_constants.h"
+
+/* Private constants */
+#define MS5XXX_TASK_PRIORITY	PIOS_THREAD_PRIO_HIGHEST
+#define MS5XXX_TASK_STACK_BYTES	512
```
kubark42, line 46: Missing space.

peabody124:
Tested Quanton and Sparky1

peabody124:
Tested Colibri

flight/PiOS/Common/pios_ms5xxx.c
```diff
@@ -0,0 +1,700 @@
...
+ * setup file does not define the model of chip, and -3 if the write
+ * command fails.
+ */
+int32_t PIOS_MS5XXX_Init(const struct pios_ms5xxx_cfg *cfg)
+{
+	// Set the pointer to the configuration struct. Note that it is not deep copied.
+	dev->cfg = cfg;
+
+	// Check that a model of the chip was chosen
+	if (cfg->pios_ms5xxx_model == 0) {
```
peabody124, line 150: Is there any way to confirm the device we are talking to matches the model?

flight/PiOS/Common/pios_ms5xxx.c
peabody124: Probably worth duplicating this include down around L497 or removing it here.

peabody124:
Aside from comments from @mlyle regarding the conditionals (I don't feel strongly) LGTM and good to merge. Nice improvement!

flight/PiOS/Common/pios_ms5xxx.c
kubark42, line 150: Alas, not. In Measurement Specialties's chip docs, I didn't see a WHOAMI or something that would be unique to each different model in the series. That's a real shame, because you will get the wrong absolute altitude if you use the wrong driver math, but your relative altitude changes will still be sort-of correct. (Ask me how I know!)

flight/PiOS/Common/pios_ms5xxx.c
kubark42: These #ifdefs were included because of compile-time failures when `PIOS_INCLUDE_SPI` wasn't set. I will look into why the transfer byte and transfer block sections (~L497) don't cause compiler failures (because I kinda think they should).

kubark42:
The compiler removes unused sections and so the additional #includes aren't necessary for normal use as the outputted `-Os` binaries are identical, but we still need all the conditional includes for compiling `-O0`. I added sufficient `PIOS_INCLUDE_SPI` and `PIOS_INCLUDE_I2C` to ensure this would compile correctly with `-O0` no matter the options chosen. 

As an aside, it looks like we might have breakage elsewhere in the project if we compile `-O0`. I know that's unlikely to do in practice, but it does seem we should be testing this pathway in Jenkins because I think it's pretty reasonable to expect `-O0` to compile successfully.

Issue #483, opened 2013-04-30
Move LLA2NED conversions into a library function


peabody124:
As per the title. Also a great moment to make a unit test.

PR #2208, opened 2016-03-05
LLA to NED abstraction

erika-tsutsumi:
As per issue listed below, introduces functions to generate LLA to NED conversion factors as well as a generalized 3d coordinate transformation function that can be used to do the conversion. Also contains unit tests for all functions introduced. 

Fixes https://github.com/TauLabs/TauLabs/issues/483

### DIFF
--- a/flight/Libraries/math/coordinate_conversions.c
+++ b/flight/Libraries/math/coordinate_conversions.c
@@ -366,6 +366,43 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
 }
 
 /**
+ * @brief LLA_linearization_float calculates linearized NED to LLA conversion using floats
+ * @param home_latitude is the home latitude times 10e6
+ * @param home_altitude
+ * @param T is the linearized NED to LLA conversion
+ */
+void LLA_linearization_float(int32_t home_latitude, float home_altitude, float T[])
+{
+	float lat = home_latitude / 10.0e6f * DEG2RAD;
+	float alt = home_altitude;
+
+	T[0] = alt + WGS84_A*1000.0f;
+	T[1] = cosf(lat)*(alt + WGS84_A*1000.0f);
+	T[2] = -1.0f;
+}
+
+/**
+ * @brief LLA_linearization_double calculates linearized NED to LLA conversion using doubles (used in simulation)
+ * @param home_latitude
+ * @param home_altitude
+ * @param T is the linearized NED to LLA conversion
+ */
+void LLA_linearization_double(int32_t home_latitude, double home_altitude, double T[])
+{
+	double lat = home_latitude / 10.0e6 * ((double)DEG2RAD);
+	double alt = home_altitude;
+
+	T[0] = alt + ((double)WGS84_A)*1000.0;
+	T[1] = cos(lat)*(alt + ((double)WGS84_A)*1000.0);
+	T[2] = -1.0;
+}
+
+    T[0] = alt + ((double)WGS84_A)*1000.0;
+    T[1] = cos(lat)*(alt + ((double)WGS84_A)*1000.0);
+    T[2] = -1.0;
+
+}
+/**
  * @}
  * @}
  */
--- a/flight/Libraries/math/coordinate_conversions.c
+++ b/flight/Libraries/math/coordinate_conversions.c
@@ -371,13 +371,15 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
  * @param home_altitude
  * @param T is the linearized NED to LLA conversion
  */
-void LLA_linearization_float(int32_t home_latitude, float home_altitude, float T[])
+void LLA_linearization_float(int32_t home_latitude_1e7, float home_altitude, float T[])
 {
-	float lat = home_latitude / 10.0e6f * DEG2RAD;
+    float lat_rad = home_latitude_1e7 / 10.0e6f * DEG2RAD;
 	float alt = home_altitude;
 
-	T[0] = alt + WGS84_A*1000.0f;
-	T[1] = cosf(lat)*(alt + WGS84_A*1000.0f);
+	// T is the linearization factor from LLA [rad] to NED [meters] divided by a scaling factor of
+	// 1e7*DEG2RAD. By default, input units are stored as degrees times 1e7 to avoid precision loss.
+    T[0] = ((alt + WGS84_A*1000.0f) / 10e6f) * DEG2RAD;
+    T[1] = ((cosf(lat_rad)*(alt + WGS84_A*1000.0f)) / 10e6f) * DEG2RAD ;
 	T[2] = -1.0f;
 }
 
@@ -387,20 +389,50 @@ void LLA_linearization_float(int32_t home_latitude, float home_altitude, float T
  * @param home_altitude
  * @param T is the linearized NED to LLA conversion
  */
-void LLA_linearization_double(int32_t home_latitude, double home_altitude, double T[])
+void LLA_linearization_double(int32_t home_latitude_1e7, double home_altitude, double T[])
 {
-	double lat = home_latitude / 10.0e6 * ((double)DEG2RAD);
+    double lat_rad = home_latitude_1e7 / 10.0e6 * ((double)DEG2RAD);
 	double alt = home_altitude;
 
-	T[0] = alt + ((double)WGS84_A)*1000.0;
-	T[1] = cos(lat)*(alt + ((double)WGS84_A)*1000.0);
+	// T is the linearization factor from LLA [rad] to NED [meters] divided by a scaling factor of
+	// 1e7*DEG2RAD. By default, input units are stored as degrees times 1e7 to avoid precision loss.
+    T[0] = ((alt + ((double)WGS84_A)*1000.0) / 10.0e6) * ((double)DEG2RAD);
+    T[1] = ((cos(lat_rad)*(alt + ((double)WGS84_A)*1000.0)) / 10.0e6) * ((double)DEG2RAD);
 	T[2] = -1.0;
 }
 
-    T[0] = alt + ((double)WGS84_A)*1000.0;
-    T[1] = cos(lat)*(alt + ((double)WGS84_A)*1000.0);
-    T[2] = -1.0;
+/**
+ * @brief get_linearized_3D_transformation Map position from one coordinate system to
+ * another, linearized around the origin in the original coordinate system. For instance,
+ * this can be used to map from Lattitude-Longitude-Altitude to North-East-Down.
+ *
+ * Note that the scale of the input units is not important, as that is included in
+ * the mapping coefficients. For example, if we use decimal degrees * 1e7, then the
+ * mapping coefficient includes a division by 1e7.
+ *
+ * Current position:
+ * @param [in] theta_prime (typically latitude * 1e7), as an int32_t
+ * @param [in] phi_prime   (typically longitude * 1e7), as an int32_t
+ * @param [in] r_prime     (typically altitude), as a float
+ * Home position:
+ * @param [in] theta
+ * @param [in] phi
+ * @param [in] r
+ * Scale factors and outputs:
+ * @param [in] C mapping coefficient
+ * @param [out] XYZ output coordinates
+ */
+void get_linearized_3D_transformation(const int32_t theta_prime, const int32_t phi_prime, const float r_prime,
+                                      const int32_t theta, const int32_t phi, const float r,
+                                      const float C[], float XYZ[])
+{
+	int32_t dTheta = theta_prime - theta;
+	int32_t dPhi = phi_prime - phi;
+	float dR = r_prime - r;
 
+	XYZ[0] = C[0] * dTheta;
+	XYZ[1] = C[1] * dPhi;
+	XYZ[2] = C[2] * dR;
 }
 /**
  * @}
--- a/flight/Libraries/math/coordinate_conversions.c
+++ b/flight/Libraries/math/coordinate_conversions.c
@@ -366,12 +366,12 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
 }
 
 /**
- * @brief LLA_linearization_float calculates linearized NED to LLA conversion factors using floats
+ * @brief LLA2NED_linearization_float calculates linearized LLA to NED conversion factors using floats
  * @param home_latitude_1e7 is the home latitude times 1e7
  * @param home_altitude
- * @param linearized_conversion_factor_f is the linearized NED to LLA conversion factor
+ * @param linearized_conversion_factor_f is the linearized LLA to NED conversion factor
  */
-void LLA_linearization_float(int32_t home_latitude_1e7, float home_altitude, float linearized_conversion_factor_f[])
+void LLA2NED_linearization_float(int32_t home_latitude_1e7, float home_altitude, float linearized_conversion_factor_f[])
 {
 	float lat_rad = home_latitude_1e7 / 10.0e6f * DEG2RAD;
 	float alt = home_altitude;
@@ -385,12 +385,12 @@ void LLA_linearization_float(int32_t home_latitude_1e7, float home_altitude, flo
 
 
 /**
- * @brief LLA_linearization_double calculates linearized NED to LLA conversion using doubles (used in simulation)
+ * @brief LLA2NED_linearization_double calculates linearized LLA to NED conversion using doubles (used in simulation)
  * @param home_latitude_1e7 is the home latitude times 1e7
  * @param home_altitude
- * @param linearized_conversion_factor_d is the linearized NED to LLA conversion factor
+ * @param linearized_conversion_factor_d is the linearized LLA to NED conversion factor
  */
-void LLA_linearization_double(int32_t home_latitude_1e7, double home_altitude, double linearized_conversion_factor_d[])
+void LLA2NED_linearization_double(int32_t home_latitude_1e7, double home_altitude, double linearized_conversion_factor_d[])
 {
 	double lat_rad = home_latitude_1e7 / 10.0e6 * ((double)DEG2RAD);
 	double alt = home_altitude;
--- a/flight/Libraries/math/coordinate_conversions.c
+++ b/flight/Libraries/math/coordinate_conversions.c
@@ -373,13 +373,13 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
  */
 void LLA_linearization_float(int32_t home_latitude_1e7, float home_altitude, float T[])
 {
-    float lat_rad = home_latitude_1e7 / 10.0e6f * DEG2RAD;
+	float lat_rad = home_latitude_1e7 / 10.0e6f * DEG2RAD;
 	float alt = home_altitude;
 
 	// T is the linearization factor from LLA [rad] to NED [meters] divided by a scaling factor of
 	// 1e7*DEG2RAD. By default, input units are stored as degrees times 1e7 to avoid precision loss.
-    T[0] = ((alt + WGS84_A*1000.0f) / 10e6f) * DEG2RAD;
-    T[1] = ((cosf(lat_rad)*(alt + WGS84_A*1000.0f)) / 10e6f) * DEG2RAD ;
+	T[0] = ((alt + WGS84_A*1000.0f) / 10e6f) * DEG2RAD;
+	T[1] = ((cosf(lat_rad)*(alt + WGS84_A*1000.0f)) / 10e6f) * DEG2RAD ;
 	T[2] = -1.0f;
 }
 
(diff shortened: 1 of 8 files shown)


kubark42:
Jenkins test this please.

flight/Libraries/math/coordinate_conversions.c
```diff
@@ -366,6 +366,75 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
...
 /**
+ * @brief LLA_linearization_float calculates linearized NED to LLA conversion using floats
+ * @param home_latitude is the home latitude times 10e6
+ * @param home_altitude
+ * @param T is the linearized NED to LLA conversion
+ */
+void LLA_linearization_float(int32_t home_latitude_1e7, float home_altitude, float T[])
+{
+    float lat_rad = home_latitude_1e7 / 10.0e6f * DEG2RAD;
+	float alt = home_altitude;
```
kubark42, line 377: Whitespace here and below.

flight/Modules/Attitude/attitude.c
```diff
@@ -154,6 +154,8 @@ static const float zeros[3] = {0.0f, 0.0f, 0.0f};
 static struct complementary_filter_state complementary_filter_state;
 static struct cfvert cfvert; //!< State information for vertical filter
 
+static float T[3];
```
kubark42: As long as you're moving `T` around, you might consider renaming the transformation with a more searchable, i.e. distinct, name.

flight/Modules/CameraStab/camerastab.c
```diff
@@ -58,6 +59,7 @@
 #include "positionactual.h"
 #include "tabletinfo.h"
 
+
```
kubark42: Was this whitespace intentional?

flight/Modules/Sensors/simulated/sensors.c
```diff
@@ -478,8 +476,8 @@ static void simulateModelQuadcopter()
 
 		GPSPositionData gpsPosition;
 		GPSPositionGet(&gpsPosition);
-		gpsPosition.Latitude = homeLocation.Latitude + ((pos[0] + gps_drift[0]) / T[0] * 10.0e6);
-		gpsPosition.Longitude = homeLocation.Longitude + ((pos[1] + gps_drift[1])/ T[1] * 10.0e6);
+		gpsPosition.Latitude = homeLocation.Latitude + ((pos[0] + gps_drift[0]) / T[0]);
+		gpsPosition.Longitude = homeLocation.Longitude + ((pos[1] + gps_drift[1])/ T[1]);
```
kubark42: Could you fix the whitespace around the `/`?

flight/Modules/Sensors/simulated/sensors.c
```diff
@@ -779,8 +775,8 @@ static void simulateModelAirplane()
 		
 		GPSPositionData gpsPosition;
 		GPSPositionGet(&gpsPosition);
-		gpsPosition.Latitude = homeLocation.Latitude + ((pos[0] + gps_drift[0]) / T[0] * 10.0e6);
-		gpsPosition.Longitude = homeLocation.Longitude + ((pos[1] + gps_drift[1])/ T[1] * 10.0e6);
+		gpsPosition.Latitude = homeLocation.Latitude + ((pos[0] + gps_drift[0]) / T[0]);
+		gpsPosition.Longitude = homeLocation.Longitude + ((pos[1] + gps_drift[1])/ T[1]);
```
kubark42: Likewise.

flight/Modules/Sensors/simulated/sensors.c
```diff
@@ -1036,8 +1030,8 @@ static void simulateModelCar()
...
```
kubark42: One last whitespace around `/`.

flight/tests/coordinate_conversions/unittest.cpp
```diff
@@ -85,3 +85,306 @@ TEST_F(RneFromLLATest, Equator) {
...
+  float homeLLA_D[] = {0, 0, 0};
+  float dN = 0.0111319490793274;
+  float dE = 0.0111319490793274;
+  float dD = -1;
+
+  float T[3];
+  LLA_linearization_float(homeLLA_D[0], homeLLA_D[2], T);
+
+  float eps = 0.000001f;
+
```
kubark42, line 112: Probably should add a comment that you're looking for the ratio between expected and returned values, and not absolute differences.

flight/tests/coordinate_conversions/unittest.cpp
```diff
@@ -85,3 +85,306 @@ TEST_F(RneFromLLATest, Equator) {
...
+  }
+
+  virtual void TearDown() {
+  }
+};
+
+TEST_F(LLALineraization, Equator_float) {
+  // Test location is Equator
+  float homeLLA_D[] = {0, 0, 0};
+  float dN = 0.0111319490793274;
```
kubark42, line 104: It would be helpful to embed the python script used for generating these values.

flight/tests/coordinate_conversions/unittest.cpp
```diff
@@ -85,3 +85,306 @@ TEST_F(RneFromLLATest, Equator) {
...
+
+  float T[3];
+  LLA_linearization_float(homeLLA_D[0], homeLLA_D[2], T);
+
+  float NED[3];
+  get_linearized_3D_transformation(currentLLA_D[0], currentLLA_D[1], currentLLA_D[2],
+                                   homeLLA_D[0], homeLLA_D[1], homeLLA_D[2],
+                                   T, NED);
+
+  // 0.0110574275839048
```
kubark42: Remove commented code.

flight/tests/coordinate_conversions/unittest.cpp
```diff
@@ -85,3 +85,306 @@ TEST_F(RneFromLLATest, Equator) {
...
+  LLA_linearization_float(homeLLA_D[0], homeLLA_D[2], T);
+
+  float NED[3];
+  get_linearized_3D_transformation(currentLat, currentLon, currentAlt,
+                                   homeLLA_D[0], homeLLA_D[1], homeLLA_D[2],
+                                   T, NED);
+
+  float eps = 0.003f;
+
+  double N_expected = 0.01110805;
```
kubark42, line 306: Likewise, embedding the python code would be good here for repeatability.

flight/tests/coordinate_conversions/unittest.cpp
```diff
@@ -85,3 +85,306 @@ TEST_F(RneFromLLATest, Equator) {
...
+
+  float T[3];
+  LLA_linearization_float(homeLLA_D[0], homeLLA_D[2], T);
+
+  float NED[3];
+  get_linearized_3D_transformation(currentLat, currentLon, currentAlt,
+                                   homeLLA_D[0], homeLLA_D[1], homeLLA_D[2],
+                                   T, NED);
+
+  float eps = 0.003f;
```
kubark42, line 304: So I guess this means that the linearization results in values that are only correct to 99.7%?

kubark42:
jenkins: add to whitelist

flight/Modules/Sensors/simulated/sensors.c
```diff
@@ -467,10 +467,8 @@ static void simulateModelQuadcopter()
 	if(PIOS_DELAY_DiffuS(last_gps_time) / 1.0e6 > GPS_PERIOD) {
 		// Use double precision here as simulating what GPS produces
 		double T[3];
-		T[0] = homeLocation.Altitude+6.378137E6f * DEG2RAD;
-		T[1] = cosf(homeLocation.Latitude / 10e6 * DEG2RAD)*(homeLocation.Altitude+6.378137E6) * DEG2RAD;
-		T[2] = -1.0;
-		
+		LLA_linearization_double(home_latitude, home_altitude, T);
```
kubark42: This needs to be changed to `homeLocation.Latitude, homeLocation.Altitude`.

flight/Modules/Attitude/attitude.c
```diff
@@ -1051,10 +1056,10 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 	// Be is set and a good value
 	mag_updated &= !outdoor_mode || (homeLocation.Be[0] != 0 || homeLocation.Be[1] != 0 || homeLocation.Be[2]);
 
-	// A more stringent requirement for GPS to initialize the filter
+        // A more stringent requirement for GPS to initialize the filter
```
kubark42, line 1059: whitespace

flight/Modules/Attitude/attitude.c
```diff
@@ -1068,10 +1073,10 @@ static int32_t updateAttitudeINSGPS(bool first_run, bool outdoor_mode)
 		else
 			set_state_estimation_error(SYSTEMALARMS_STATEESTIMATION_UNDEFINED);
 
-	} else if (outdoor_mode && (gpsData.Satellites < insSettings.MinRNAVSatellites || gpsData.PDOP > insSettings.MinRNAVPDOP)) {
-		if (gpsData.Satellites < insSettings.MinRNAVSatellites)
```
kubark42, line 1073: whitespace

flight/Modules/ManualControl/tablet_control.c
```diff
@@ -207,13 +208,8 @@ static int32_t tabletInfo_to_ned(TabletInfoData *tabletInfo, float *NED)
...
 	GPSPositionGet(&gpsPosition);
 
-	float lat = homeLocation.Latitude / 10.0e6f * DEG2RAD;
-	float alt = homeLocation.Altitude;
-
 	float T[3];
-	T[0] = alt+6.378137E6f;
-	T[1] = cosf(lat)*(alt+6.378137E6f);
-	T[2] = -1.0f;
+	LLA_linearization_float(homeLocation.Latitude, homeLocation.Altitude, T);
```
kubark42: This shouldn't be done every loop. It's not a blocker for this PR, but if you wanted to you could move it out.

kubark42:
I like this a lot. It's very nice how you folded the entire linearization into the matrix `T`, including the shifts by `1e7`. The resulting code is more efficient in RAM and CPU both, which is a big win.

@peabody124 I'll test fly it, but I don't have a camera rig to make sure the CameraStab module still works correctly. Is this something you can take a look at?

peabody124:
I'll review the code shortly. No guarantees on flying with my camera rig, but I'll try.

kubark42:
Cool. 

I was looking in more detail and realized that this generalized-linearization lets us move to ECEF-based navigation. In theory, this will be a a nice precision and latency win. Preparing a PR now.

flight/Libraries/math/coordinate_conversions.c
```diff
@@ -366,6 +366,75 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
 }
 
 /**
+ * @brief LLA_linearization_float calculates linearized NED to LLA conversion using floats
```
kubark42: Is this comment right? Isn't it the other way around, i.e. this is performing an LLA-->NED transformation linearization?

flight/Libraries/math/coordinate_conversions.c
```diff
@@ -366,6 +366,75 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
...
+ * @param home_latitude is the home latitude times 10e6
+ * @param home_altitude
+ * @param T is the linearized NED to LLA conversion
+ */
+void LLA_linearization_float(int32_t home_latitude_1e7, float home_altitude, float T[])
```
kubark42: Might be called `LLA2NED_linearization...`. This way it is more explicit about what the operation is, and it makes space for my `ECEF2NED_linearization...`.

flight/Libraries/math/coordinate_conversions.c
```diff
@@ -366,6 +366,75 @@ void rot_mult(float R[3][3], const float vec[3], float vec_out[3], bool transpos
...
+ * @param [in] r_prime     (typically altitude), as a float
+ * Home position:
+ * @param [in] theta
+ * @param [in] phi
+ * @param [in] r
+ * Scale factors and outputs:
+ * @param [in] C mapping coefficient
+ * @param [out] XYZ output coordinates
+ */
+void get_linearized_3D_transformation(const int32_t theta_prime, const int32_t phi_prime, const float r_prime,
```
kubark42, line 426: @peabody124 do you have any thoughts on making this more general? It bothers me that there is a mix of  `int32_t`  and `float` arguments. I see why @erika-tsutsumi did it this way, but it's not as clean as I'd like.

One thing we could do is turn `GPSPosition.Altitude` into a `int32_t`, i.e. remove the `GpsPosition->Altitude = (float)posllh->hMSL*0.001f` conversion in the GPS module. Then this function could be called completely on `int32_t`, which should be more maintainable.

Thoughts?

P.S. @erika-tsutsumi there's no need to change now, this would be for a future PR.

flight/Modules/Attitude/attitude.c
```diff
@@ -827,7 +828,11 @@ static int32_t setNavigationRaw()
 		else
 			set_state_estimation_error(SYSTEMALARMS_STATEESTIMATION_NONE);
 
-		getNED(&gpsPosition, NED);
+		// Get NED coordinates from LLA coordinates
+		get_linearized_3D_transformation(gpsPosition.Latitude,  gpsPosition.Longitude, gpsPosition.Altitude,
+		                                 homeLocation.Latitude, homeLocation.Longitude,  homeLocation.Altitude,
+		                                 T, NED);
+
```
kubark42: Extra newline.

kubark42:
@erika-tsutsumi there's also a hidden conversion at https://github.com/TauLabs/TauLabs/blob/54fff14e8b863ba7d106b8c208dfec2cc380c3b8/flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c#L259. This should be cleaned up as well.

flight/Libraries/math/coordinate_conversions.c
peabody124, line 426: I don't have a strong opinion on the units. Given we always use int32_t throughout the code this is the most efficient way to do it and probably makes sense. Casting the int32 to float directly would lose precision, and we don't want to add the burden of using doubles.

peabody124:
Code committed looks good. @erika-tsutsumi agreed it would be great to get the UAVOMSPBridge changed as well for consistency. Can you comment if you will be able to do this? Kenn has a few small comments that would be nice to fix, none major or blockers.

Has this been tested outside?

peabody124:
Tested in flight simulation without issue.

peabody124:
Simulation also shows the camera tracking working well.

flight/Libraries/math/coordinate_conversions.c
kubark42, line 426: https://github.com/TauLabs/TauLabs/blob/next/flight/Modules/GPS/UBX.c#L206 is what I was thinking of getting rid of. The native input to the system is in [mm] so I'm proposing we could remove the conversion to `float` and then the rest follows naturally. Was that clear?

On the other hand, we've been playing around with RTK GPS and sub-mm level precisioning is already something we're having fun with. So it might not make a ton of sense to optimize this for older tech GPSes when we really should be aiming at supporting an order of magnitude more precision in the near future.

erika-tsutsumi:
@peabody124 yup I'll look at the UAVOMSPBridge  when I have a chance

flight/Libraries/math/coordinate_conversions.c
erika-tsutsumi: Yup. Fixed.

flight/Libraries/math/coordinate_conversions.c
erika-tsutsumi: Also fixed.

erika-tsutsumi:
For the UAVOMSPBridge, the math wasn't quite the same so could someone double check it? I left all the original code in comments because I wasn't sure about it, I'll remove them once someone else okays it.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -256,20 +257,31 @@ static void msp_send_comp_gps(struct msp_bridge *m)
...
-			data.comp_gps.distance_to_home  = (uint16_t)(sqrtf(delta_x * delta_x + delta_y * delta_y) * 1e-4f);  // meters
+//			delta_y *= cosf(home_data.Latitude * 1e-7f * (float)DEG2RAD);  // Latitude compression correction
+
+			float linearized_conversion_factor_f[3];
+			LLA2NED_linearization_float(home_data.Latitude, home_data.Altitude, linearized_conversion_factor_f);
+
+			float deltas[3];
+			get_linearized_3D_transformation(gps_data.Latitude, gps_data.Longitude, gps_data.Altitude,
+			                                 home_data.Latitude, home_data.Longitude, home_data.Altitude,
+			                                 linearized_conversion_factor[], deltas[]);
```
peabody124: `linearized_conversion_factor` dropped the `_f` and also the [] should not be there

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -256,20 +257,31 @@ static void msp_send_comp_gps(struct msp_bridge *m)
...
 	
-			if ((delta_lon == 0) && (delta_lat == 0))
-				data.comp_gps.direction_to_home = 0;
+//			data.comp_gps.distance_to_home  = (uint16_t)(sqrtf(delta_x * delta_x + delta_y * delta_y) * 1e-4f);  // meters
+			data.comp_gps.distance_to_home  = (uint16_t)(sqrtf(deltas[0] * deltas [0] + deltas[1] * deltas[1]));  // meters
```
peabody124, line 268: I'm trying to track down why the difference of 1e-4f

kubark42, line 268: Took me a second, but I think I figured it out. The comments like `float delta_y = delta_lon * WGS84_RADIUS_EARTH_KM * DEG2RAD;  // KM * 1e7` point out that the unit is already `1e7`. So `1e-4` is appropriate for converting the final output to meters.

peabody124, line 268: Ok, great. So math is good?

peabody124, line 268: If so, looks ok to merge to me once comments removed.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
```diff
@@ -256,20 +257,31 @@ static void msp_send_comp_gps(struct msp_bridge *m)
...
+//			float delta_y = delta_lon * WGS84_RADIUS_EARTH_KM * DEG2RAD;  // KM * 1e7
+//			float delta_x = delta_lat * WGS84_RADIUS_EARTH_KM * DEG2RAD;  // KM * 1e7
 	
-			delta_y *= cosf(home_data.Latitude * 1e-7f * (float)DEG2RAD);  // Latitude compression correction
-	
-			data.comp_gps.distance_to_home  = (uint16_t)(sqrtf(delta_x * delta_x + delta_y * delta_y) * 1e-4f);  // meters
+//			delta_y *= cosf(home_data.Latitude * 1e-7f * (float)DEG2RAD);  // Latitude compression correction
+
+			float linearized_conversion_factor_f[3];
+			LLA2NED_linearization_float(home_data.Latitude, home_data.Altitude, linearized_conversion_factor_f);
```
kubark42, line 261: I'm unsure about how I feel about doing the linearization each loop. @peabody124, what's your opinion about increased code complexity vs. the CPU cost of doing this per-loop? I guess I'm kind of okay giving it a pass, since it's just a few Hz.

flight/Modules/UAVOMSPBridge/UAVOMSPBridge.c
erika-tsutsumi: yup sorry I was building flyingf3 and so didn't catch it, should be updated now

erika-tsutsumi:
Rebased commits in preparation for merge.

peabody124:
LGTM @kubark42 merge away if good with you

P.S. i'm ok with a few loops calling it extra times for code simplicity. we can change that later.

flight/Modules/CameraStab/camerastab.c
```diff
@@ -376,22 +377,18 @@ static void tablet_info_process() {
 
 		PoiLocationData poi;
 
-		float lat, alt;
-		lat = homeLocation.Latitude / 10.0e6f * DEG2RAD;
-		alt = homeLocation.Altitude;
+		float linearized_conversion_factor_f[3];
+		LLA2NED_linearization_float(homeLocation.Latitude, homeLocation.Altitude, linearized_conversion_factor_f);
```
kubark42, line 381: In a future PR, this should be moved out of the camera stab loop. No sense in recalculating it every time.

kubark42:
LGTM and tests fine with a real GPS. Merging.

@peabody124, maybe we should consider moving this linearization into a UAVObject. It seems like it's used by several modules and I foresee it being used by more in the future. Might make sense to think about sharing instead of each module having its own private instance.

peabody124:
@kubark42 that seems like a great idea and gets rid of all the duplicate linearizations!

PR #2202, opened 2016-02-23
FSK DAC: 1200bps telemetry over audio (similar to Bel202 encoding)

peabody124:
This creates a PIOS_COM interace to the DAC output that creates
a 1200bps audio modem. The stream is modulated similarly to V.23
encoding using 1 start and 1 stop bits, 1 parity bit. The two symbols
are centered at 1200 and 2200 Hz.

The data source tested with this is the light telemetry module that was
developed for this purpose of low bandwidth telemetry, previously using
a hardware serial stream to audio modem chip.

PR #2173, opened 2016-01-31
DAC Beeps

peabody124:
This is code that will generate a series of audio beeps out of the DAC pin on
Sparky2 and BrushedSparky. This is pretty nice to pump into FPV audio so 
you can hear the RSSI and voltage (RSSI is encoded into repetition rate with
higher rate being low RSSI, the pitch of the beep drops as your voltage does).

The PIOS_DAC_COMMON file is to allow the FSK DAC code (under progress)
to coexist.

PR #2209, opened 2016-03-08
RTOS compile-time selection

kubark42:
This factors out a certain amount of redundant code, and lets a compile-time flag indicate whether a board should be compiled with FreeRTOS or ChibiOS. The primary goal of this is to increase the number of target architectures which can be used with Tau Labs.

An upshot of this is that Jenkins can build for both targets in order to validate that all targets at the least compile (although of course this doesn't provide any proof that the target still works with "recessive" RTOSes). For instance, compile problems with https://github.com/TauLabs/TauLabs/pull/2173#discussion_r55444515 and https://github.com/TauLabs/TauLabs/pull/2202/files#r55444276 were found with this strategy.


peabody124:
I was trying to use this to fix those other bugs you found. 
  `make -j4 RTOS=FREERTOS fw_revomini`
and
  `make -j4 RTOS=FREERTOS fw_sparky2` 
both throw compile errors

kubark42:
Investigating. I checked all of these before each commit. Now none of them work at any commit. WTF. I must have had an environmental variable set or something like that.

kubark42:
@peabody124 This now works as it should. I realized it was the way I was backing up the `RTOS` variable which was causing problems.

Nicely enough, a command like `make all_fw_clean && make all_fw RTOS=FREERTOS` is sufficient to test all targets for successful FreeRTOS compilation. Idem for ChibiOS. Of course, the `make all_fw_clean` shouldn't be forgotten when changing RTOSes, or else some degree of hilarity might ensue.

P.S. This doesn't conflict with #2212 and can be merged separately.

flight/uavobjectlib/Makefile
```diff
@@ -66,8 +66,15 @@ LDFLAGS += -Wl,--fatal-warnings
 SRC += $(filter-out $(OPUAVSYNTHDIR)/uavobjectsinit.c,$(wildcard $(OPUAVSYNTHDIR)/*.c))
 NO_AUTO_UAVO = YES
 
+# Backup the environmental variable before entering firmware-common.mk
```
kubark42: This is the one part of this I'm uncomfortable with. It seems very hackish.

Ties in to https://github.com/TauLabs/TauLabs/pull/2209/files#diff-12b5ccf31b114976029c5c17f55b9fc6R9.

mlyle: Why do you need to do this?  The uavobjects don't need to know what RTOS it is-- why does it need to get "put back"?

kubark42: So as not to fail the check at https://github.com/TauLabs/TauLabs/pull/2209/files#diff-12b5ccf31b114976029c5c17f55b9fc6R12.

tracernz:
We didn't previously force things to use an RTOS, some out of tree targets like TL-ESC don't.

kubark42:
@tracernz There are no targets that use firmware-common.mk that do not have a RTOS, with the exception of `uavobjectlib`. 

It's generally good practice to have an argument fail explicitly instead of silently. The problem with not checking for sane values in `RTOS` is that the build can fail silently, e.g. If a user accidentally put `RTOS=FreeRTOS` then the user would not get the desired result, and without the check s/he wouldn't be aware of this. Worse, the build might fail spectacularly with little indication of the root cause.

Put another way, what's important here is to check that whatever RTOS is chosen-- even if it's no RTOS at all-- is intentionally chosen. If it would be more useful, I'm happy to change the name from `UNDEFINED_FOR_UAVOBJECT_LIB` to `THIS_PAGE_LEFT_BLANK`. ;)

flight/uavobjectlib/Makefile
mlyle: Yes, why does it need to get "put back"?

tracernz:
~~In target/<board>/fw/Makefile for targets that require it.~~

``` C
#ifndef RTOS
#error No RTOS selected
#endif
```

-e- duh, you already have that covered with `?=`. No issue I think.

tracernz:
It will fail anyway when you try to compile `pios_thread` etc without `PIOS_INCLUDE_<RTOS>` IIRC.

flight/uavobjectlib/Makefile
mlyle: Is this a problem with our understanding of environment variables?  No other makefiles will see the changed value-- environment variables are only passed to child processes and a child is unable to modify its parent's environment) You shouldn't save / restore this IMO.

kubark42: If you set an environmental variable on the command line, then it is passed to all Makefiles and cannot be overwritten without the `override` command. That means that a make command such as `make RTOS=FREERTOS quanton` will cause a compile failure without this save/restore. The earlier failure of this mechanism is why @peabody124 was having compiling issues while testing this branch.

kubark42:
> It will fail anyway when you try to compile `pios_thread` etc without PIOS_INCLUDE_<RTOS> IIRC.

@tracernz I'm not quite understanding. I don't see any failures when I compile the targets while using a command line override. The project compiles normally when not using overrides, as well. Is there coverage I'm missing?

flight/uavobjectlib/Makefile
mlyle: No, I'm saying why do you restore the original value?

This makefile doesn't need it put back, and no other makefiles see the consequence of you putting it back.  Hence, this is vestigial :/

kubark42: I see what you're getting at. I had thought that the Makefile variable was a simple global, I see now it's one of those "global unless you change it in local scope" things.

Updated.

flight/uavobjectlib/Makefile
```diff
@@ -66,6 +66,9 @@ LDFLAGS += -Wl,--fatal-warnings
 SRC += $(filter-out $(OPUAVSYNTHDIR)/uavobjectsinit.c,$(wildcard $(OPUAVSYNTHDIR)/*.c))
 NO_AUTO_UAVO = YES
 
+# Override the local version of RTOS before entering firmware-common.mk
+override RTOS = UNDEFINED_FOR_UAVOBJECT_LIB
```
kubark42, line 70: Still not terribly happy with this approach, but it's cleaner now thanks to @mlyle's input. Unless someone disagrees, I'm good to leave it as it is for a while until we have other targets that also need to unset this variable.

make/firmware-common.mk
```diff
@@ -1,3 +1,28 @@
...
+  # Do nothing here, it's only included as an escape for the
+else
+  $(error [RTOS]: $(RTOS) is not a valid RTOS. Choose between either CHIBIOS or FREERTOS)
+endif
+
+## PIOS Hardware (STM32F1/3/4xx or POSIX)
+# `ifneq ($(or $(VAR1),$(VAR2)),)` is the Makefile code for `if defined(VAR1) || defined(VAR2)`
+ifneq ($(or $(PIOSSTM32FXXX),$(PIOSPOSIX)),)
+  ifeq ($(RTOS), CHIBIOS)
+    include $(PIOS)/$(TARGET_ARCHITECTURE)/library_chibios.mk
```
peabody124, line 19: I feel like here you could have just used `PIOSSTM32FXXX` instead of defining another variable (TARGET_ARCHITECTURE)

peabody124:
So if you look here https://github.com/peabody124/TauLabs/commit/6fe6d1a5140262e003e1ae8da63d99b5ec949c29#diff-6dd112c59af36fc83d837d3a5e4f6e98 for example you can see there are a few other CFLAGS and LDFLAGS added. Are we comfortable these casual switches work? It's probably good for code rigor to have this and occasionally test it, but just wanted to confirm

make/firmware-common.mk
kubark42, line 19: I originally did it exactly that way, but the `SIM_POSIX` caused some problems there because it's non-STM32, so I refactored into this approach. This also generally opens the door to non-STM32 chips.

kubark42:
> Are we comfortable these casual switches work?

I don't see why they would fail since this sets up pre-compiler stuff on the command-line, but maybe there's some sensitivity to ordering of these `LDFLAGS` and `CDEFS` definitions? I know, for instance, that putting the ChibiOS/FreeRTOS define at the end of `firmware-common.mk` can cause compile-time failure.

peabody124:
You want to add this commit:

```
commit ae84ad03c9110bd4ec235fcd3584fb12e5b96bc6
Author: James Cotton <peabody124@gmail.com>
Date:   Sun Mar 13 17:18:12 2016 -0500

    PIOS DAC: fix to compile with FreeRTOS

diff --git a/flight/PiOS/STM32F4xx/pios_dac_common.c b/flight/PiOS/STM32F4xx/pios_dac_common.c
index 690f9b6..e352a4a 100644
--- a/flight/PiOS/STM32F4xx/pios_dac_common.c
+++ b/flight/PiOS/STM32F4xx/pios_dac_common.c
@@ -29,6 +29,11 @@
  */

 #include "pios.h"
+
+#if defined(PIOS_INCLUDE_FREERTOS)
+#include "FreeRTOS.h"
+#endif /* defined(PIOS_INCLUDE_FREERTOS) */
+
 #if defined(PIOS_INCLUDE_DAC_BEEPS) || defined(PIOS_INCLUDE_FSK)

 static void (*pios_dac_irq_cb)(void);
@@ -105,7 +110,7 @@ void DMA1_Stream5_IRQHandler(void)
        }

 #if defined(PIOS_INCLUDE_FREERTOS)
-       portEND_SWITCHING_ISR((rx_need_yield || tx_need_yield) ? pdTRUE : pdFALSE);
+       portEND_SWITCHING_ISR(pdFALSE);
 #endif /* defined(PIOS_INCLUDE_FREERTOS) */

 #if defined(PIOS_INCLUDE_CHIBIOS)
diff --git a/flight/PiOS/STM32F4xx/pios_fskdac.c b/flight/PiOS/STM32F4xx/pios_fskdac.c
index caa8161..c4339c5 100644
--- a/flight/PiOS/STM32F4xx/pios_fskdac.c
+++ b/flight/PiOS/STM32F4xx/pios_fskdac.c
@@ -339,7 +339,7 @@ static void PIOS_FSKDAC_DMA_irq_cb()
        }

 #if defined(PIOS_INCLUDE_FREERTOS)
-       portEND_SWITCHING_ISR((rx_need_yield || tx_need_yield) ? pdTRUE : pdFALSE);
+       portEND_SWITCHING_ISR(tx_need_yield ? pdTRUE : pdFALSE);
 #endif /* defined(PIOS_INCLUDE_FREERTOS) */

 }
```

to fix those other issues you raised?

kubark42:
> You want to add this commit:

Yup, will do. Thanks!

kubark42:
`make -j4 RTOS=FREERTOS fw_sparky2` tests successfully now.

PR #2210, opened 2016-03-08
Mapcontrol relative path fixes

kubark42:
Using relative directories was linked to problems on a Windows build server. There doesn't seem to be any advantage to using relative directories since the directories themselves are already along the include
path. In any case, using strings of relative paths led to obnoxiously long absolute paths full of things like `plugins/../plugins/../ground/../`, etc...

This change resolves the problem (and hopefully results in nicer code).


peabody124:
LGTM

PR #2213, opened 2016-03-10
WIP: Qt Creator automatic configuration

kubark42:
Configures flight and ground project files. For flight, the python script creates a Qt Creator build target for each board, as well as the POSIX simulator. For ground, it effectively only sets the correct build directories and sets the default application to "App" instead of randomly selecting project apps, such as  "uavobject generator" or "translator".

These templates assumes that the user does not want to use multiple cores for a build. This is an unfortunate default, since most users will be able to use multiple cores, but it seems the safest way to build on a generic platform.

screen shot 2016-03-10 at 12 20 12 pm

---

Works on:
- [x] OSX
- [ ] Windows
- [ ] Linux


kubark42:
Added support for a ground template as well. This should greatly reduce the amount of knowledge required to get a Qt environment up and running.

I'm sure that there are still additional things to be done in order to handle different Qt Kits, but that can come later.

kubark42:
P.S. I'd appreciate any reports on this. It configures automatically as a part of `make qt_sdk_install`, so it's sufficient to try this on a fresh checkout. The end result should be correctly configured build directories and a big list of targets to build from inside Qt Creator.

kubark42:
Grr... This might require some more finessing to handle cases where the Qt Kits are different. Seems to be related to 

```
   <value type="QString" key="ProjectExplorer.ProjectConfiguration.Id">qt.55.clang_64_kit</value>
```

in the `*.pro.user` files. This is probably a unique ID for a particular configured Kit. Without this, Qt Creator generates a new `*.pro.user` file. Not sure if we can programmatically determine a valid kit name.

---

UPDATE:

Looks like there's some notion in Qt Creator of shared settings. They mostly expect it to be used for editor settings, but initial tests show that it might work robustly for build directory and build targets. http://doc.qt.io/qtcreator/creator-sharing-project-settings.html

make/tools.mk
```diff
@@ -109,6 +109,10 @@ ifdef WINDOWS
 	$(V1) ./downloads/qt-opensource-windows-x86-mingw492-5.5.0.exe
 endif
 
+	# Configure Qt Creator user project files (GCS and flight firmware)
```
peabody124, line 112: Can this be a separate make target. I'd rather not conflate installing the Qt SDK with setting up an environment. It's also nice if (e.g. right now me) would like to say
`make qtcreator_environment`

peabody124:
screen shot 2016-03-12 at 11 32 18 am

This is probably my Qt Creator inexperience, but this is what I get. I'm not sure how to proceed and get it to look like yours. Is this something the python script can deal with?

make/tools.mk
kubark42, line 112: How about setting it up as a separate script that is called as part of the SDK install process?

kubark42:
I think you're running afoul of the brittleness I get with my approach. Basically, not having kits installed throws a wrench in the process. I need to refactor to use the `*.pro.shared` and see if that works.

Even if it doesn't, once the kits are in place then there won't be a problem here if we adopt your suggested approach to have this be a separate make target.

make/tools.mk
peabody124, line 112: well you might want want to clobber your configuration as part of installing the  SDK. what if you had customized it and this overwrote it? it's better to have small granular targets that do what they say

kubark42, line 112: > well you might want want to clobber your configuration as part of installing the SDK. what if you had customized it and this overwrote it?

I think this points to why I should be using the `.pro.shared` files so we don't ever clobber someone's local customizations. However, since the configuration is, in a sense, an integral part of setting up the Qt Creator IDE, I err toward keeping stuff one-stop-shopping. Introducing a second step required for people to set up the dev environment can lead to a less complete experience, as well as support issues if people have diverging environments.

I guess what I'm arguing for is that the IDE setup is an integral part of the SDK setup. I think you're arguing that many devs/users will not want the (optional) Qt Creator IDE as part of the (required) `qmake` and Qt Libraries. I can certainly see it from that side as well.

I propose I create the separate target, insulated as you suggest, and then if after a bit of testing we can show that it behaves nicely with customized configurations, consider adding it to the SDK.

P.S. I added unit testing and the ability to run `sim_posix` directly from Qt Creator as well. It's been super-useful for improving my iteration time as I run down bugs, but unless you have a preference to include them in the default configuration I feel it's a little clearer leaving them out.

peabody124:
I'll probably always have a preference of following a principle of least surprise, so not having the sdk installation start modifying local configuration files. Can always make a meta target that calls both of those for sdk_install_setup

kubark42:
Yeah, `sdk_install_setup` sounds reasonable. I think we should definitely have a minimum of commands to type to get a development environment up and running. We're considering a whole vagrant and ansible approach to setting up our build machines, where all you have to do to get an entire VM up and running is type a single command. FFMPEG is a nicely done example for this.

kubark42:
Putting this on pause this for the moment. It kind of works, and kind of doesn't. Either Qt Creator doesn't play nicely with its own shared files or  I'm doing something wrong.

I'll probably keep it around locally and see if we can find something mature that works. It might be that #2230 is the gateway to making this work robustly.

PR #2218, opened 2016-03-14
Raw hid upgrades

kubark42:
A couple cool backports from dRonin. Improves OS X usb connections and reduces console output.

https://github.com/d-ronin/dRonin/pull/795
https://github.com/d-ronin/dRonin/pull/540

Don't think we're affected by https://github.com/d-ronin/dRonin/issues/582, but should check to make sure.

P.S. Should also look at https://github.com/d-ronin/dRonin/pull/790 to see if it affects us as well.


peabody124:
So from looking at the code this appears to let you select between connecting between two TL targets when both are connected to USB. Have you verified the functionality and make sure it doesn't reduce USB stability at all?

PR #2219, opened 2016-03-14
Bootloader: Allow redefining of DFU wait period

kubark42:
Previously, it was always 6 seconds. In practice, this doesn't seem always necessary. This allows redefining the delay for individual targets.

We find that a 2 second wait isn't causing us any issues and is a lot more tolerable than 6 seconds when debugging.

### DIFF
--- a/flight/targets/bl/common/main.c
+++ b/flight/targets/bl/common/main.c
@@ -49,7 +49,13 @@ extern void PIOS_Board_Init(void);
 #define BL_DETECT_BREAK_TO_BL_TIMER MSEC_TO_USEC(500)
 #endif
 
-#define BL_WAIT_FOR_DFU_TIMER SEC_TO_USEC(6)
+// allow altering bootloader delay
+#ifndef DFU_PAUSE_DELAY_MS
+#define DFU_PAUSE_DELAY_MS 6000
+#endif
+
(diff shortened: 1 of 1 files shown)


flight/targets/bl/common/main.c
```diff
@@ -49,7 +49,11 @@ extern void PIOS_Board_Init(void);
 #define BL_DETECT_BREAK_TO_BL_TIMER MSEC_TO_USEC(500)
 #endif
 
+#if defined(DFU_PAUSE_DELAY_MS) // allow altering bootloader delay
+#define BL_WAIT_FOR_DFU_TIMER MSEC_TO_USEC(DFU_PAUSE_DELAY_MS)
```
peabody124: Just to write this more typically could you do this?

```
#ifndef DFU_PAUSE_DELAY_MS
#define DFU_PAUSE_DELAY_MS 6
#endif

#define BL_WAIT_FOR_DFU_TIMER MSEC_TO_USEC(DFU_PAUSE_DELAY_MS)
```

Less lines

mlyle:
Some old netbooks need nearly 4 seconds to enumerate and catch it, so we decided to keep it at 6 seconds in dR-- the approach we took is we have GCS kick it out of the loader / trigger boot.  If it enumerates quicker than 2 seconds, it boots even faster; it's still correct in the other case.

flight/targets/bl/common/main.c
kubark42: Yup, done.

PR #2225, opened 2016-03-15
WIP: -Wextra fixes

kubark42:
Added these flags and caught a few compiler warnings. Nothing severe.

```
CFLAGS += -Wextra
CFLAGS += -Wno-sign-compare
CFLAGS += -Wno-unused-parameter
```


flight/Libraries/timeutils.c
```diff
@@ -64,9 +64,6 @@ void date_from_timestamp(uint32_t timestamp, DateTimeT *date_time)
 	date_time->sec = rem % 60;
 	/* January 1, 1970 was a Thursday.  */
 	date_time->wday = (4 + days) % 7;
-	if (date_time->wday < 0){
```
peabody124, line 67: yeah this could never be negative. wonder why it was here.

flight/PiOS/inc/pios_can.h
```diff
@@ -35,6 +35,7 @@
 
 //! The set of CAN messages
 enum pios_can_messages {
+	PIOS_CAN_ERROR = -1,
```
peabody124, line 38: wait. why is this required?

kubark42, line 38: It's because there is a check somewhere else for something like `pios_can_messages foo; if(foo < 0)...`. Since the enum previously didn't support `<0`it would throw a warning that this condition always evaluates to false. Of course, this warning is wrong, since C does not prevent you from taking a variable `foo` and assigning any value you like to it, i.e. there's no GCC protection\* that only enum values can be assigned to enum variables. 

Upshot: by making an error flag with a negative value, GCC no longer believes that the check is false. 

To be honest, I don't think it's a great solution. It's totally non-obvious why it works, but there doesn't seem to be another way to work around this GCC bug (as suggested on stackoverflow). 

[*] clang supports checking this just fine.

flight/Libraries/timeutils.c
kubark42, line 67: IIRC, @mluessi imported the library and changed the values to be unsigned. I forget why, but I recall he had a good reason.

flight/PiOS/inc/pios_can.h
mlyle, line 38: >  Of course, this warning is wrong, since C does not prevent you from taking a variable foo and assigning any value you like to it, i.e. there's no GCC protection\* that only enum values can be assigned to enum variables.

This is actually correct; from the C99 standard:

>  4 Each enumerated type shall be compatible with char, a signed integer type, or an unsigned integer type. The choice of type is implementation-defined,110) but shall be capable of representing the values of all the members of the enumeration.

tracernz, line 38: So what I take from that is if you assigned a negative value to an enum type with only positive valued members, the behaviour is undefined. GCC should really warn you about that explicitly as well.

mlyle, line 38: In general gcc lets you assign and implicit cast unsigned and signed types without warnings-- it just gets mad about comparisons.  (Unfortunately the alternative would be way too chatty)

peabody124, line 38: @kubark42 I'd probably rather just update the code that validates the msg_id. It's actually not entirely safe anyway since it doesn't check that msg_id is too great:

```
bool pios_can_valid_msg(enum pios_can_messages msg_id)
{
    return (msg_id >= 0 && msg_id < PIOS_CAN_LAST);
}
```

Then in (for example PIOS_CAN_TxData)

```
int32_t PIOS_CAN_TxData(uintptr_t id, enum pios_can_messages msg_id, uint8_t *data)
{
    // Fetch the size of this message type or error if unknown
    int32_t bytes = get_message_size(msg_id);
    if (! pios_can_valid_msg(msg_id))
        return -1;

```

I can prepare a commit if you'd like. Since this is the only place we have to define a negative enum for this reason, I'd rather just fix it up stream.

kubark42, line 38: Yeah, totally. Or if you're happy with that blob, I can just rebase exactly that into the PR.

peabody124, line 38: Yeah if you want to do that it would be great. I think what I wrote should work, there might be one other place using that check is required. Then I'll verify CAN still works since it is a small functionality change.

peabody124:
Closing in favor of #2244

PR #2226, opened 2016-03-15
PiOS Watchdog: Validate that a flag has been registered

kubark42:
It's subtle, but an unregistered flag being updated would ultimately cause the watchdog to timeout and the board to reset. This prevents that by first checking that the watchdog flag is indeed registered before processing it.

Untested on F1 and F3 targets, works fine on F4.

### DIFF
--- a/flight/PiOS/STM32F10x/pios_wdg.c
+++ b/flight/PiOS/STM32F10x/pios_wdg.c
@@ -121,6 +121,11 @@ bool PIOS_WDG_RegisterFlag(uint16_t flag_requested)
  */
 bool PIOS_WDG_UpdateFlag(uint16_t flag) 
 {	
+	// Validate that the flag has been registered. If not, halt the program.
+	if ((wdg_configuration.used_flags & flag) == 0) {
+		PIOS_Assert(0);
+	}
+
 	// we can probably avoid using a semaphore here which will be good for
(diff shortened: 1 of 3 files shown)


mlyle:
Should this perhaps PIOS_Assert instead?  I don't like the pattern of silently-tolerate-abnormal-things; this permanently conceals that you've failed to enable a watchdog flag.

peabody124:
Could you update the headers to include TL?

peabody124:
@kubark42 I'm a bit with @mlyle -- if you just transparently drop this you are probably masking another bug, right?

kubark42:
Agreed. I was toying with that idea as well that we should PIOS_Assert() here. I didn't have time to throw the question out before leaving work, glad to see we're all on the same page here.

kubark42:
Fixed.

kubark42:
Note that the change to `PIOS_ASSERT()` causes a crash because now many tasks start before their watchdogs are enabled. The upshot is that the watchdog gets called in the task before it's been registered. This is easy to fix, as all the watchdogs must come before the task initialization.

PR landing soon. I'll merge it urgently because right now booting is broken.

P.S. We should have the simulation run for a few seconds as part of jenkins tests. This would have caught the error. Of course, so would have me testing first before pushing the newer version.

PR #2228, opened 2016-03-16
Watchdog: Change the order of task vs. watchdog initialization

kubark42:
The tasks run immediately after they are created, which means the watchdog won't have been registered yet before it the UpdateFlag reset function is called on it. The result was a crash due to `PIOS_ASSERT(0)`.

This fixes the order of initialization.

Bug introduced at https://github.com/TauLabs/TauLabs/pull/2226.

Tested on F3 and F4 targets.


kubark42:
Merging. Please let me know if there are any targets which are still broken.

peabody124:
Good call -- the upside of using Assert was catching these :)

PR #2240, opened 2016-04-10
Eigen update to version 3.2.8

peabody124:
This later version is being used for the LQR solution. Note that the
old version has been deprecated anyway.


peabody124:
I'll merge this if there are no objections.

PR #2241, opened 2016-04-27
RevoMini: extend functionality of Flex-IO port

peabody124:
This adds various UART functionality to the Flex-IO header including
- UART
- FrSky telemetry
- Serial protocols

This is a back port from https://github.com/d-ronin/dRonin/pull/617 and
credit goes to @henn1001


lolorc:
Hi,

just a quick report, I will test thoroughly with PPM, PWM at the end of day.
frsky sport and telemetry still works on flexi.
but they don't work on the rcvr port.
That the state I got stuck in last WE when I tried to backport it myself.
I took some time to check with dronin on my revomini, @henn1001 changes do works there.
have to go, will give it more time later.

Thanks !

PR #2242, opened 2016-04-27
Flight: Add external magnetometer capability to RevoMini

peabody124:
Allows connecting an external magnetometer to the RevoMini target

Back port from https://github.com/d-ronin/dRonin/pull/806
credit goes to @jihlein 


lolorc:
bench tested it with external hmc5883L and hmc5983, looks ok.
there's no error when it's set to external mag without any external mag attached tho. external_mag_fail is never used.

jihlein:
You're missing part of the change.  Look at the sensor module and search for the changes relating to external_mag_fail and the #define SUPPORTS_EXTERNAL_MAG.

lolorc:
Oh I see thanks @jihlein, in TL external_mag_fail is only defined for AQ32, dronin had 
https://github.com/d-ronin/dRonin/commit/42778abc7ba2302c76ad59ceda86df01046ed196 applied as well.

PR #2245, opened 2016-07-14
Travis

peabody124:
Support for travis-ci build verification


tracernz:
FYI: I played with Travis last year and found our build regularly took too long and timed out, not sure if they have changed the time limit since then.

PR #2257, opened 2017-01-04
RFM22B Report RSSI in ManualControlCommand

### LINKED ISSUE #2256: ManualControlSettings: support TauLink RSSI
Without this the flight controller doesn't advertise the RSSI correctly when using TauLinkModule

peabody124:
This allows getting RSSI back on TauLinkModule

fixes #2256

### DIFF
--- a/flight/Modules/ManualControl/transmitter_control.c
+++ b/flight/Modules/ManualControl/transmitter_control.c
@@ -220,6 +220,11 @@ int32_t transmitter_control_update()
 			value = PIOS_FrSkyRssi_Get();
 #endif /* PIOS_INCLUDE_FRSKY_RSSI */
 			break;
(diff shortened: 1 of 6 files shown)


peabody124:
tested on Sparky2. And "RSSI" from 0...127 reflects actually missing packets and not the RSSI, like in OpenLRS combined mode. An RSSI from 128-255 is the raw RSSI. 

	// Converting the RSSI to the same scale as with OpenLRS so add back
	// the 122 used to convert the raw register in half into dBm, and then
	// add 128 because we use (128,255) as the range for good RSSI when LQ
	// is perfect. At this point and RSSI of 150 reflects -100 dBm (very bad)
	// and 230 reflects -20 dBm (max possible register)

PR #2261, opened 2017-01-13
Python: correct GCS timestamps when jumping in log file

peabody124:
Really log files shouldn't have any corruption or require jumping
ahead, but in the case they do, handle this correctly.


peabody124:
Tested and fixes some timestamp glitches for me

